PR #639, opened 2026-04-06
Rectify: Quota Guard Cache Refresh — Kitchen Lifecycle Service Pattern

### LINKED ISSUE #638: Quota guard hook fails open 93% of the time — no cache refresh after open_kitchen
## Problem

The quota guard PreToolUse hook (`src/autoskillit/hooks/quota_check.py`) is structurally incapable of enforcing quota limits in sessions longer than 5 minutes. The quota cache file (`~/.claude/autoskillit_quota_cache.json`) is refreshed **only once** during `open_kitchen` via `_prime_quota_cache()`. The hook reads this cache with a 300-second (5-minute) `cache_max_age`. After 5 minutes, every hook invocation sees a stale cache, logs `cache_miss`, and fails open (approves unconditionally). **93.3% of all hook invocations over the last 5 days were `cache_miss` events** — the guard is effectively disabled for nearly all `run_skill` calls.

### quota_events.jsonl (2026-03-31 to 2026-04-06)

| Event Type | Count | Percentage |
|---|---|---|
| `cache_miss` | 1,944 | 93.3% |
| `approved` | 135 | 6.5% |
| `blocked` | 4 | 0.2% |
| `parse_error` | 0 | 0% |

### The Guard DOES Work When Cache Is Fresh

All 4 blocked events occurred on **2026-04-04** between 19:13-19:17 UTC at 93% utilization (threshold was 90% at the time, now lowered to 85%). This proves the blocking logic is correct — the problem is exclusively cache freshness.

| Timestamp | Utilization | Threshold | Sleep seconds |
|---|---|---|---|
| 2026-04-04T19:13:59 | 93.0% | 90.0 | 2,820s |
| 2026-04-04T19:17:27 | 93.0% | 90.0 | 2,612s |

Zero `approved` events occurred at utilization >= 85%. Highest approved utilization observed: 78.0%. When the cache is fresh, the guard makes correct decisions every time.

## The Causal Chain

```
open_kitchen
  └── _prime_quota_cache()
       └── check_and_sleep_if_needed()
            └── _fetch_quota() → writes ~/.claude/autoskillit_quota_cache.json  [ONLY WRITE POINT]

run_skill (PreToolUse hook fires BEFORE tool executes)
  └── quota_check.py::main()
       └── _read_quota_cache(cache_path, max_age=300)
            ├── cache fresh (< 300s) → read utilization → compare threshold → approve/deny
            └── cache stale (> 300s) → return None → cache_miss → sys.exit(0) → APPROVE
```

There is **no mechanism** to refresh the cache between `run_skill` calls. The cache is written once at `open_kitchen` time and then decays into staleness. Pipeline sessions routinely run for hours, but quota protection expires after 5 minutes.

## Fail-Open Paths in the Hook

The hook is designed to fail open on any error. Every one of these paths results in `sys.exit(0)` (approve):

| # | Condition | Log Event | Result |
|---|---|---|---|
| 1 | Stdin not valid JSON | (none) | Approve |
| 2 | Unexpected stdin exception | stderr only | Approve |
| 3 | Cache file does not exist | `cache_miss` | Approve |
| 4 | **Cache is stale (> 300s old)** | `cache_miss` | **Approve** |
| 5 | Cache file corrupt/unreadable | `cache_miss` | Approve |
| 6 | Cache missing `five_hour.utilization` | `parse_error` | Approve |
| 7 | Utilization < threshold | `approved` | Approve |
| 8 | Env var points to bad cache path | `cache_miss` | Approve |

**Path 4 is the one triggered 93.3% of the time.**

### Double Hook Registration

Both `~/.claude/settings.json` (global, installed package) and `.claude/settings.json` (project, dev source) register `quota_check.py` for the same matcher (`mcp__.*autoskillit.*__run_skill.*`). This causes **two hook processes per `run_skill` call**, confirmed by paired events with near-identical timestamps in the quota events log. When the hook_config file is absent, the two copies can use different `_DEFAULT_THRESHOLD` values (installed vs. dev source).

### Silent Credential Failures

If `~/.claude/.credentials.json` is missing, malformed, or has an expired token, `_prime_quota_cache()` catches the error silently (`except (OSError, ValueError, RuntimeError)`) and returns without writing the cache. No warning surfaces to the user. Every subsequent hook call fails open indefinitely.

### Silent Hook Config Write Failure

`_write_hook_config()` in `tools_kitchen.py` swallows `OSError` with only a structured logger warning. If the write fails, the hook falls back to module-level defaults with no user-visible signal in the `open_kitchen` response.

### Chicken-and-Egg Timing Problem

The hook fires as a PreToolUse hook — it runs **before** `run_skill` executes. Even if `run_skill` were to refresh the cache internally, the hook would have already made its approve/deny decision using the stale cache. This means a simple "refresh inside run_skill" doesn't solve the problem for the current invocation, only for the next one.

## Affected Components

- `src/autoskillit/hooks/quota_check.py` — The hook; correct logic but entirely dependent on cache freshness
- `src/autoskillit/execution/quota.py` — `check_and_sleep_if_needed()` / `_fetch_quota()` / `_write_cache()` — the only code that refreshes the cache; never called from `run_skill`
- `src/autoskillit/server/tools_execution.py` — `run_skill` handler; no quota refresh
- `src/autoskillit/server/helpers.py` — `_prime_quota_cache()` — called only by `open_kitchen`
- `src/autoskillit/server/tools_kitchen.py` — `open_kitchen` / `close_kitchen` — only entry point that primes the cache and writes hook config

## Constraints

- The hook script (`quota_check.py`) is **stdlib-only by design** so it can run under any Python interpreter without requiring the autoskillit package to be importable. Importing `autoskillit.execution.quota` inside the hook would break this constraint.
- The hook runs as a **subprocess** spawned by Claude Code, not inside the MCP server process. It cannot directly access the MCP server's runtime state or credentials.
- The fail-open design is intentional — the hook must never block `run_skill` due to its own bugs. Any fix must preserve this principle while reducing the stale-cache window.
- The PreToolUse hook timing means the hook fires before the tool executes. Any refresh mechanism must account for this ordering.

## Possible Fix Directions (to be explored, not prescriptive)

These are starting points for investigation, not mandated solutions. The right approach may be something not listed here:

- **Hook-level live fetch**: When the cache is stale, have the hook fetch quota directly instead of failing open. Would require either relaxing the stdlib-only constraint or shelling out to a helper script. Would eliminate the stale window entirely.
- **Background cache refresh**: `open_kitchen` starts a periodic background task (e.g., every 4 minutes) that keeps the cache file fresh for the kitchen's lifetime. Keeps the hook simple but adds background process management complexity.
- **Post-execution refresh**: `run_skill` refreshes the cache after (not before) execution, so the next hook invocation has fresh data. Reduces the stale window but doesn't eliminate it — the first call after 5 minutes still uses stale data.
- **Hybrid approach**: Combine post-execution refresh with a shorter max_age to minimize the window where stale data causes a miss.
- **Architectural rethinking**: The stale-cache-fail-open architecture may need a fundamentally different approach. The core issue is that quota enforcement degrades to zero after 5 minutes of any session.

## Test Gap

The existing test `test_approve_when_stale_cache` correctly verifies that a stale cache results in approval — the fail-open behavior is tested and intentional. The gap is architectural, not in test coverage.

Minor test issue: test T-CFG-3 (`test_quota_check_env_var_overrides_hook_config_cache_path`) writes the hook config to `tmp_path/temp/.autoskillit_hook_config.json` instead of the canonical `tmp_path/.autoskillit/temp/.autoskillit_hook_config.json` (missing the `.autoskillit` directory component). The test passes because the env var bypasses the hook config read, but it masks a potential path regression.

Trecek:
## Summary

The quota guard hook fails open 93% of the time because `open_kitchen` primes the cache once (via `_prime_quota_cache`) but no mechanism keeps it fresh. The cache TTL is 300 seconds; pipeline sessions run for hours. After 5 minutes, every hook invocation sees a stale cache, logs `cache_miss`, and approves unconditionally via `sys.exit(0)`.

The root architectural weakness is that `open_kitchen`/`close_kitchen` act as a gate toggle (open/close) but not as a **service lifecycle boundary**. There is no concept of services that start with the kitchen and stop when it closes. The fix introduces a reusable service lifecycle pattern: any background service tied to the kitchen session is started in `_open_kitchen_handler` and cancelled in `_close_kitchen_handler`, with `ToolContext` holding the task handle. The quota refresh loop is the first instance of this pattern.

A secondary structural gap is also closed: `cache_max_age` (the hook's expiry threshold) and the refresh interval previously had no enforced relationship. A new `cache_refresh_interval` config field is introduced with a structural contract `cache_refresh_interval < cache_max_age`, making it architecturally impossible for the loop to fall behind the TTL. This contract is enforced by a test that asserts it directly against `defaults.yaml`.

### Process Flow Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINALS %%
    SESSION_START([Kitchen Session Start])
    SESSION_END([Kitchen Session End])

    %% CACHE DATA NODE %%
    CACHE[("autoskillit_quota_cache.json<br/>━━━━━━━━━━<br/>fetched_at + five_hour.utilization")]

    subgraph KitchenOpen ["● open_kitchen — Kitchen Open Lifecycle"]
        direction TB
        GATE_ON["● gate.enable()<br/>━━━━━━━━━━<br/>Reveal gated MCP tools"]
        WRITE_CFG["_write_hook_config()<br/>━━━━━━━━━━<br/>Write threshold + cache_max_age<br/>to .autoskillit/temp/"]
        PRIME["_prime_quota_cache()<br/>━━━━━━━━━━<br/>check_and_sleep_if_needed()<br/>read-first → fetch if stale → write T=0"]
        START_TASK["● create_background_task(<br/>★ _quota_refresh_loop(config)<br/>)<br/>━━━━━━━━━━<br/>Stored as ctx.quota_refresh_task"]
    end

    subgraph RefreshLoop ["★ _quota_refresh_loop — Background Service (asyncio.Task)"]
        direction TB
        LOOP_SLEEP["asyncio.sleep(<br/>★ cache_refresh_interval<br/>)<br/>━━━━━━━━━━<br/>Structural contract:<br/>interval < cache_max_age"]
        REFRESH["★ _refresh_quota_cache(config)<br/>━━━━━━━━━━<br/>Always fetches unconditionally<br/>(no read-first optimization)<br/>_fetch_quota() → _write_cache()"]
        LOOP_ERR{"Exception<br/>in refresh?"}
        LOG_WARN["logger.warning(<br/>'quota_refresh_loop_error'<br/>)<br/>━━━━━━━━━━<br/>Loop continues"]
    end

    subgraph RunSkill ["● run_skill — MCP Tool Execution"]
        direction TB
        EXEC["executor.run(skill_command)<br/>━━━━━━━━━━<br/>Headless Claude session"]
        AUDIT{"success?"}
        AUDIT_OK["audit.record_success()"]
        AUDIT_FAIL["notify 'run_skill failed'"]
        POST_REFRESH["● background.submit(<br/>★ _refresh_quota_cache(config)<br/>label='quota_post_run_refresh'<br/>)<br/>━━━━━━━━━━<br/>Defense-in-depth:<br/>ensures cache fresh for next hook"]
    end

    subgraph HookProcess ["quota_check.py — PreToolUse Hook Subprocess"]
        direction TB
        READ_CACHE["_read_quota_cache(<br/>max_age=cache_max_age<br/>)<br/>━━━━━━━━━━<br/>Reads cache file from disk"]
        FRESH{"cache age<br/>≤ cache_max_age?"}
        THRESHOLD{"utilization<br/>≥ threshold?"}
        APPROVE["sys.exit(0)<br/>━━━━━━━━━━<br/>→ approve run_skill"]
        DENY["print deny JSON<br/>sys.exit(0)<br/>━━━━━━━━━━<br/>→ block run_skill"]
        FAIL_OPEN["log cache_miss<br/>sys.exit(0)<br/>━━━━━━━━━━<br/>→ fail open (approve)"]
    end

    subgraph KitchenClose ["● close_kitchen — Teardown"]
        direction TB
        CANCEL["● ctx.quota_refresh_task.cancel()<br/>ctx.quota_refresh_task = None<br/>━━━━━━━━━━<br/>Terminates _quota_refresh_loop<br/>via CancelledError"]
        GATE_OFF["gate.disable()<br/>━━━━━━━━━━<br/>Hide gated MCP tools"]
        DEL_CFG["hook_cfg_path.unlink()<br/>━━━━━━━━━━<br/>Remove hook config file"]
    end

    %% MAIN FLOW %%
    SESSION_START --> GATE_ON
    GATE_ON --> WRITE_CFG
    WRITE_CFG --> PRIME
    PRIME -->|"writes T=0"| CACHE
    PRIME --> START_TASK
    START_TASK -.->|"spawns background task"| LOOP_SLEEP

    %% BACKGROUND LOOP %%
    LOOP_SLEEP --> REFRESH
    REFRESH -->|"always writes fresh"| CACHE
    REFRESH --> LOOP_ERR
    LOOP_ERR -->|"no exception"| LOOP_SLEEP
    LOOP_ERR -->|"exception caught"| LOG_WARN
    LOG_WARN --> LOOP_SLEEP

    %% RUN_SKILL FLOW %%
    START_TASK --> EXEC
    EXEC --> AUDIT
    AUDIT -->|"yes"| AUDIT_OK
    AUDIT -->|"no"| AUDIT_FAIL
    AUDIT_OK --> POST_REFRESH
    AUDIT_FAIL --> POST_REFRESH
    POST_REFRESH -->|"fire-and-forget write"| CACHE
    POST_REFRESH --> EXEC

    %% HOOK FLOW %%
    CACHE -.->|"read by subprocess"| READ_CACHE
    READ_CACHE --> FRESH
    FRESH -->|"yes — fresh"| THRESHOLD
    FRESH -->|"no — stale"| FAIL_OPEN
    THRESHOLD -->|"below threshold"| APPROVE
    THRESHOLD -->|"at/above threshold"| DENY

    %% CLOSE FLOW %%
    EXEC -->|"session ends"| CANCEL
    CANCEL -.->|"CancelledError propagates"| LOOP_SLEEP
    CANCEL --> GATE_OFF
    GATE_OFF --> DEL_CFG
    DEL_CFG --> SESSION_END

    %% CLASS ASSIGNMENTS %%
    class SESSION_START,SESSION_END terminal;
    class CACHE stateNode;
    class GATE_ON,WRITE_CFG,PRIME,START_TASK handler;
    class LOOP_SLEEP,LOG_WARN phase;
    class REFRESH,POST_REFRESH newComponent;
    class EXEC,AUDIT_OK,AUDIT_FAIL handler;
    class CANCEL,GATE_OFF,DEL_CFG handler;
    class READ_CACHE integration;
    class FRESH,THRESHOLD,AUDIT,LOOP_ERR detector;
    class APPROVE,DENY,FAIL_OPEN output;
```

**Legend:** ★ New component · ● Modified component

### State Lifecycle Diagram

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 50, 'rankSpacing': 58, 'curve': 'basis'}}}%%
flowchart TB
    %% CLASS DEFINITIONS %%

    subgraph ConfigContract ["● QuotaGuardConfig — INIT_ONLY Fields (Set Once, Never Mutated)"]
        direction LR
        CFG_MAX_AGE["cache_max_age<br/>━━━━━━━━━━<br/>300 s (default)<br/>Hook TTL — max age<br/>before cache_miss"]
        CFG_INTERVAL["★ cache_refresh_interval<br/>━━━━━━━━━━<br/>240 s (default)<br/>Loop sleep between<br/>proactive writes"]
        STRUCT_CONTRACT["★ STRUCTURAL CONTRACT<br/>━━━━━━━━━━<br/>cache_refresh_interval<br/>MUST be < cache_max_age<br/>(enforced by test)"]
        CFG_INTERVAL -->|"enforced constraint"| STRUCT_CONTRACT
        CFG_MAX_AGE -->|"enforced constraint"| STRUCT_CONTRACT
    end

    subgraph TaskLifecycle ["★ ToolContext.quota_refresh_task — Kitchen-Scoped MUTABLE Field"]
        direction LR
        TASK_NULL_START["State: None<br/>━━━━━━━━━━<br/>Before open_kitchen<br/>No task running"]
        TASK_RUNNING["★ State: asyncio.Task<br/>━━━━━━━━━━<br/>_quota_refresh_loop coroutine<br/>Background refresh active"]
        TASK_NULL_END["State: None<br/>━━━━━━━━━━<br/>After close_kitchen<br/>Task cancelled + cleared"]

        TASK_NULL_START -->|"● open_kitchen:\ncreate_background_task()"| TASK_RUNNING
        TASK_RUNNING -->|"● close_kitchen:\ntask.cancel() + task = None"| TASK_NULL_END
    end

    subgraph CacheStateMachine ["autoskillit_quota_cache.json — Cache File State Transitions"]
        direction TB
        CACHE_ABSENT["State: ABSENT<br/>━━━━━━━━━━<br/>No cache file exists<br/>→ hook: cache_miss (fail-open)"]
        CACHE_FRESH["State: FRESH<br/>━━━━━━━━━━<br/>age ≤ cache_max_age<br/>→ hook: enforce threshold"]
        CACHE_EXPIRING["State: EXPIRING (age approaches max_age)<br/>━━━━━━━━━━<br/>★ Proactive refresh fires<br/>before expiry (interval < max_age)"]
        CACHE_EXPIRED["State: EXPIRED (age > cache_max_age)<br/>━━━━━━━━━━<br/>→ hook: cache_miss (fail-open)<br/>Only possible if loop crashes"]

        CACHE_ABSENT -->|"_prime_quota_cache() at T=0"| CACHE_FRESH
        CACHE_FRESH -->|"age increases over time"| CACHE_EXPIRING
        CACHE_EXPIRING -->|"★ _quota_refresh_loop writes\nbefore expiry (interval < max_age)"| CACHE_FRESH
        CACHE_EXPIRING -->|"★ post-run background.submit()\nafter each run_skill"| CACHE_FRESH
        CACHE_FRESH -.->|"loop crash (telemetry visible)"| CACHE_EXPIRED
        CACHE_EXPIRED -->|"loop recovery or\nnext run_skill post-refresh"| CACHE_FRESH
    end

    subgraph ValidationGates ["Hook Validation Gates (quota_check.py — PreToolUse)"]
        direction TB
        GATE_AGE["● Gate 1: Age Check<br/>━━━━━━━━━━<br/>_read_cache(max_age=cache_max_age)<br/>age ≤ cache_max_age?"]
        GATE_THRESHOLD["Gate 2: Threshold Check<br/>━━━━━━━━━━<br/>utilization < threshold?"]
        GATE_PASS["→ APPROVE<br/>━━━━━━━━━━<br/>sys.exit(0)"]
        GATE_DENY["→ DENY<br/>━━━━━━━━━━<br/>block run_skill"]
        GATE_MISS["→ FAIL-OPEN (cache_miss)<br/>━━━━━━━━━━<br/>Preserved by design\ntelemetry visible via quota_events.jsonl"]

        GATE_AGE -->|"age ≤ max_age"| GATE_THRESHOLD
        GATE_AGE -->|"age > max_age (stale)"| GATE_MISS
        GATE_THRESHOLD -->|"below threshold"| GATE_PASS
        GATE_THRESHOLD -->|"≥ threshold"| GATE_DENY
    end

    %% Cross-subgraph connections %%
    STRUCT_CONTRACT -->|"interval guarantees\ncache never expires\nduring normal operation"| CACHE_EXPIRING
    TASK_RUNNING -->|"★ loop writes every\ncache_refresh_interval"| CACHE_FRESH
    CACHE_FRESH -->|"read by subprocess"| GATE_AGE

    %% CLASS ASSIGNMENTS %%
    class CFG_MAX_AGE,CFG_INTERVAL detector;
    class STRUCT_CONTRACT newComponent;
    class TASK_NULL_START,TASK_NULL_END phase;
    class TASK_RUNNING newComponent;
    class CACHE_ABSENT detector;
    class CACHE_FRESH output;
    class CACHE_EXPIRING gap;
    class CACHE_EXPIRED detector;
    class GATE_AGE,GATE_THRESHOLD stateNode;
    class GATE_PASS output;
    class GATE_DENY handler;
    class GATE_MISS cli;
```

**Legend:** ★ New component · ● Modified component

### Concurrency Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    CACHE_FILE[("autoskillit_quota_cache.json<br/>━━━━━━━━━━<br/>Shared via atomic_write<br/>Cross-process IPC boundary")]

    subgraph EventLoop ["MCP Server — asyncio Event Loop (single-threaded cooperative)"]
        direction TB

        subgraph KitchenLifecycle ["● Kitchen Lifecycle — Coroutines"]
            direction LR
            OPEN["● _open_kitchen_handler()<br/>━━━━━━━━━━<br/>Runs in event loop as coroutine"]
            CLOSE["● _close_kitchen_handler()<br/>━━━━━━━━━━<br/>Runs in event loop as coroutine"]
        end

        subgraph BackgroundTask ["★ Kitchen-Scoped Background Task (asyncio.Task, caller-owned)"]
            direction TB
            CREATE_TASK["★ create_background_task(<br/>_quota_refresh_loop(config)<br/>)<br/>━━━━━━━━━━<br/>asyncio.create_task() — no supervision wrapper<br/>handle stored in ctx.quota_refresh_task"]
            LOOP_CORO["★ _quota_refresh_loop coroutine<br/>━━━━━━━━━━<br/>while True:<br/>  await asyncio.sleep(cache_refresh_interval)<br/>  await _refresh_quota_cache(config)<br/>CancelledError propagates from sleep → exits"]
            CANCEL["● close_kitchen:<br/>ctx.quota_refresh_task.cancel()<br/>━━━━━━━━━━<br/>Delivers CancelledError at\nnext await asyncio.sleep()\nctx.quota_refresh_task = None"]
        end

        subgraph SupervisedTasks ["● DefaultBackgroundSupervisor — Fire-and-Forget (supervised)"]
            direction TB
            SUBMIT["● background.submit(<br/>★ _refresh_quota_cache(config),<br/>label='quota_post_run_refresh'<br/>)<br/>━━━━━━━━━━<br/>Called after each run_skill completes\nReturns immediately (fire-and-forget)"]
            SUPERVISE["_supervise_task wrapper<br/>━━━━━━━━━━<br/>Catches Exception → logs + audits\nCancelledError → write_status('cancelled')"]
        end

        subgraph RunSkillHandler ["● run_skill MCP Tool Handler"]
            direction TB
            EXECUTOR["executor.run(skill_command)<br/>━━━━━━━━━━<br/>await headless session"]
            POST["● Post-completion hook:<br/>background.submit(_refresh_quota_cache)<br/>━━━━━━━━━━<br/>Fire-and-forget refresh triggered<br/>after EVERY run_skill"]
        end
    end

    subgraph HookSubprocess ["quota_check.py — OS Subprocess (separate process, stdlib-only)"]
        direction TB
        HOOK_READ["_read_quota_cache(<br/>max_age=cache_max_age<br/>)<br/>━━━━━━━━━━<br/>Read-only access to cache file\nNo Python object sharing\nNo asyncio — blocking I/O"]
        HOOK_DECIDE{"cache fresh?"}
        HOOK_APPROVE["sys.exit(0) → approve"]
        HOOK_MISS["sys.exit(0) → fail-open"]
    end

    %% FLOW %%
    OPEN --> CREATE_TASK
    CREATE_TASK --> LOOP_CORO
    LOOP_CORO -->|"every cache_refresh_interval\nawait asyncio.sleep() yields to event loop"| LOOP_CORO
    LOOP_CORO -->|"★ unconditional fetch+write"| CACHE_FILE

    EXECUTOR --> POST
    POST --> SUBMIT
    SUBMIT --> SUPERVISE
    SUPERVISE -->|"★ _refresh_quota_cache writes"| CACHE_FILE

    CLOSE --> CANCEL
    CANCEL -.->|"CancelledError at await asyncio.sleep()"| LOOP_CORO

    CACHE_FILE -->|"cross-process read\natomic_write ensures no torn reads"| HOOK_READ
    HOOK_READ --> HOOK_DECIDE
    HOOK_DECIDE -->|"fresh"| HOOK_APPROVE
    HOOK_DECIDE -->|"stale"| HOOK_MISS

    %% CLASS ASSIGNMENTS %%
    class CACHE_FILE stateNode;
    class OPEN,CLOSE handler;
    class CREATE_TASK newComponent;
    class LOOP_CORO newComponent;
    class CANCEL handler;
    class SUBMIT,POST newComponent;
    class SUPERVISE phase;
    class EXECUTOR handler;
    class HOOK_READ integration;
    class HOOK_DECIDE detector;
    class HOOK_APPROVE output;
    class HOOK_MISS phase;
```

**Legend:** ★ New component · ● Modified component

Closes #638

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260405-212016-751478/.autoskillit/temp/rectify/rectify_quota_guard_cache_refresh_2026-04-06_045000.md`

## Token Usage Summary

| investigate | 19 | 9.8k | 265.8k | 42.1k | 1 | 6m 47s |
| rectify | 21 | 29.3k | 494.6k | 55.9k | 1 | 10m 19s |
| dry_walkthrough | 2.1k | 13.9k | 1.1M | 81.2k | 1 | 4m 5s |
| implement | 81 | 29.5k | 5.6M | 89.3k | 1 | 10m 11s |
| assess | 67 | 39.4k | 4.7M | 93.7k | 1 | 16m 8s |
| open_pr | 3.0k | 31.2k | 2.2M | 101.1k | 1 | 9m 56s |
| **Total** | 5.3k | 153.2k | 14.4M | 463.2k | | 57m 30s |


src/autoskillit/execution/__init__.py
```diff
@@ -59,7 +59,11 @@
     run_managed_async,
     run_managed_sync,
 )
-from autoskillit.execution.quota import QuotaStatus, check_and_sleep_if_needed
+from autoskillit.execution.quota import (
+    QuotaStatus,
+    _refresh_quota_cache,
```
Trecek: [warning] arch: _refresh_quota_cache is a private function (underscore-prefixed) being exported from the execution package's public __init__.py and added to __all__. Exporting private implementation details in __all__ contradicts the naming convention. Either rename to a public name or do not export it.

Trecek (commented):

> **Note:** Using COMMENT event (cannot REQUEST_CHANGES on own PR). This review found blocking issues — see inline comments.

src/autoskillit/server/helpers.py
```diff
@@ -13,6 +13,7 @@
 
 from autoskillit.core import RESERVED_LOG_RECORD_KEYS, TerminationReason, get_logger
 from autoskillit.execution import (
+    _refresh_quota_cache,  # noqa: F401 — re-exported for tools_execution.py; patched by tests
```
Trecek, line 16: [warning] arch: _refresh_quota_cache is imported into helpers.py solely to be re-exported as a test-patch target (noqa: F401 confirms this). Using a module as a re-export shim to create a patchable namespace is an indirect coupling workaround rather than a proper design. The test patching target should be the canonical import site.

src/autoskillit/server/tools_execution.py
```diff
@@ -303,6 +303,13 @@ async def run_skill(
             )
         if order_id:
             skill_result.order_id = order_id
+        from autoskillit.server.helpers import _refresh_quota_cache  # noqa: PLC0415
```
Trecek, line 306: [warning] arch: Deferred inline import inside function body with noqa: PLC0415 suppression signals a circular-import workaround. The quota-post-run refresh dependency should be importable at module level, or the coupling should be resolved architecturally (e.g., pass the coroutine factory via ToolContext).

src/autoskillit/server/tools_execution.py
```diff
@@ -303,6 +303,13 @@ async def run_skill(
...
+
+        if tool_ctx.background is not None:
```
Trecek, line 308: [warning] arch: Post-run quota cache refresh scheduling is business logic embedded in the server-layer run_skill tool handler. This decision belongs in the execution or pipeline layer, not in a server-layer tool handler.

src/autoskillit/server/helpers.py
```diff
@@ -488,3 +489,23 @@ async def _prime_quota_cache() -> None:
         await check_and_sleep_if_needed(_get_ctx().config.quota_guard)
     except (OSError, ValueError, RuntimeError):
         logger.warning("quota_prime_failed", exc_info=True)
+
+
+async def _quota_refresh_loop(config: Any) -> None:
```
Trecek: [warning] defense: _quota_refresh_loop accepts `config: Any` instead of a concrete type (QuotaGuardConfig). The attribute accesses `config.cache_refresh_interval` are untyped, so a wrong config object silently produces AttributeError at runtime.

src/autoskillit/core/_type_protocols.py
```diff
@@ -459,3 +460,15 @@ def submit(
...
+@runtime_checkable
+class QuotaRefreshTask(Protocol):
+    """Protocol for a cancellable background task handle.
+
+    Satisfied by asyncio.Task — used to type the kitchen-scoped quota
+    refresh task stored in ToolContext without leaking asyncio.Task into the
+    core layer.
+    """
+
+    def cancel(self, msg: Any = None) -> bool: ...
```
Trecek, line 474: [warning] defense: QuotaRefreshTask.cancel() Protocol method is missing its return type annotation `-> bool`. Add `-> bool` to complete the typed boundary.

src/autoskillit/execution/quota.py
```diff
@@ -106,6 +106,27 @@ async def _fetch_quota(
     )
 
 
+async def _refresh_quota_cache(
+    config: Any,
```
Trecek, line 110: [warning] defense: _refresh_quota_cache accepts `config: Any` instead of QuotaGuardConfig. Attribute accesses config.credentials_path and config.cache_path are untyped — a wrong config object silently produces AttributeError at runtime.

src/autoskillit/server/tools_kitchen.py
```diff
@@ -72,6 +74,10 @@ async def _open_kitchen_handler() -> None:
     logger.info("open_kitchen", gate_state="open", kitchen_id=ctx.kitchen_id)
     _write_hook_config()
     await _prime_quota_cache()
+    ctx.quota_refresh_task = create_background_task(
```
Trecek, line 79: [warning] bugs: If open_kitchen is called while a session is already open, the existing quota_refresh_task is overwritten without cancelling it first. The old task runs indefinitely and its handle is lost, causing a task leak.

src/autoskillit/server/tools_kitchen.py
```diff
@@ -89,8 +95,12 @@ def _close_kitchen_handler() -> None:
     """Clear the tools-enabled flag. Extracted for testability."""
     from autoskillit.server import _get_ctx, logger
 
-    _get_ctx().gate.disable()
-    _get_ctx().active_recipe_packs = None
+    ctx = _get_ctx()
+    if ctx.quota_refresh_task is not None:
+        ctx.quota_refresh_task.cancel()
```
Trecek, line 102: [warning] bugs: ctx.quota_refresh_task.cancel() schedules cancellation but the reference is immediately set to None, dropping the task handle before cancellation completes. If the task is mid-flight in _refresh_quota_cache when cancelled, any subsequent exception is never retrieved, producing an 'asyncio task exception was never retrieved' warning.

tests/config/test_config.py
```diff
@@ -360,6 +360,26 @@ def test_load_quota_guard_from_yaml(self, tmp_path):
...
+        config = QuotaGuardConfig()
+        assert config.cache_refresh_interval == 240
+
+    def test_defaults_yaml_has_cache_refresh_interval(self):
+        """QG_C4: defaults.yaml defines quota_guard.cache_refresh_interval < cache_max_age."""
+        from autoskillit.core.paths import pkg_root
+
+        defaults = yaml.safe_load((pkg_root() / "config" / "defaults.yaml").read_text())
+        assert "cache_refresh_interval" in defaults["quota_guard"]
+        interval = defaults["quota_guard"]["cache_refresh_interval"]
```
Trecek, line 376: [critical] tests: Variables `interval` and `max_age` are used but never defined in test_defaults_yaml_has_cache_refresh_interval. After checking `cache_refresh_interval` key existence, the test jumps directly to `assert interval < max_age` without assigning these variables. This raises NameError at runtime.

tests/execution/test_quota.py
```diff
@@ -439,3 +441,52 @@ def test_write_cache_then_quota_check_main_reads_it(self, tmp_path, monkeypatch)
...
+        cache_path = tmp_path / "cache.json"
+
+        async def fake_fetch(credentials_path, **kwargs):
+            return QuotaStatus(utilization=0.5, resets_at=None)
+
+        monkeypatch.setattr("autoskillit.execution.quota._fetch_quota", fake_fetch)
+        config = QuotaGuardConfig(cache_path=str(cache_path))
+        await _refresh_quota_cache(config)
+        assert cache_path.exists()
+        data = json.loads(cache_path.read_text())
```
Trecek, line 491: [critical] tests: Variable `data` is used but never defined in test_refresh_quota_cache_writes_new_cache. The cache file is written by _refresh_quota_cache but the test never reads it back (missing `data = json.loads(cache_path.read_text())`). This raises NameError at runtime.

tests/server/test_quota_refresh_loop.py
```diff
@@ -0,0 +1,83 @@
...
+
+    async def fake_refresh(config):
+        nonlocal call_count
+        call_count += 1
+
+    monkeypatch.setattr("autoskillit.server.helpers.asyncio.sleep", fake_sleep)
+    monkeypatch.setattr("autoskillit.server.helpers._refresh_quota_cache", fake_refresh)
+
+    config = QuotaGuardConfig(cache_refresh_interval=240)
+    with pytest.raises(asyncio.CancelledError):
```
Trecek, line 35: [critical] tests: _quota_refresh_loop is called at lines 35, 49, and 78 but is never imported in this test file. All three tests raise NameError: name '_quota_refresh_loop' is not defined at runtime.

tests/server/test_tools_kitchen.py
```diff
@@ -497,3 +499,57 @@ def test_close_kitchen_clears_active_recipe_packs(tmp_path, monkeypatch):
...
+    """After _open_kitchen_handler(), ctx.quota_refresh_task is a running asyncio.Task."""
+    monkeypatch.chdir(tmp_path)
+    mock_ctx = _make_mock_ctx()
+
+    async def instant_loop(config):
+        await asyncio.sleep(0)
+
+    with patch("autoskillit.server._get_ctx", return_value=mock_ctx):
+        with patch("autoskillit.server.logger"):
+            with patch("autoskillit.server.tools_kitchen._prime_quota_cache", AsyncMock()):
```
Trecek, line 516: [critical] tests: Indentation inconsistency in with-block nesting: `with patch('...._prime_quota_cache', ...)` at L0516 is indented deeper than the `with patch('autoskillit.server.logger')` block at L0515 instead of being a sibling. This creates a syntax/indentation error.

tests/server/test_quota_refresh_loop.py
```diff
@@ -0,0 +1,83 @@
...
+
+@pytest.mark.anyio
+async def test_quota_refresh_loop_exits_cleanly_on_cancel(monkeypatch):
+    """CancelledError from asyncio.sleep propagates; loop does not swallow it."""
+    from autoskillit.server.helpers import _quota_refresh_loop
+
+    async def immediate_cancel(n):
+        raise asyncio.CancelledError
+
+    monkeypatch.setattr("autoskillit.server.helpers.asyncio.sleep", immediate_cancel)
```
Trecek, line 49: [warning] tests: In test_quota_refresh_loop_exits_cleanly_on_cancel, task.cancel() is called after asyncio.create_task() but the immediate_cancel stub already raises CancelledError before cancel() is reached. The test exercises internal raise rather than external cancel signal, making task.cancel() a no-op that misleads about what is being tested.

tests/server/test_quota_refresh_loop.py
```diff
@@ -0,0 +1,83 @@
...
+    monkeypatch.setattr("autoskillit.server.helpers._refresh_quota_cache", fake_refresh)
+
+    config = QuotaGuardConfig(cache_refresh_interval=240)
+    with pytest.raises(asyncio.CancelledError):
+        await _quota_refresh_loop(config)
+
+    assert call_count == 2  # one refresh per completed sleep
+
+
+@pytest.mark.anyio
```
Trecek, line 41: [warning] tests: test_quota_refresh_loop_exits_cleanly_on_cancel only asserts CancelledError propagates. It does not assert call counts on the AsyncMock for _refresh_quota_cache, so it cannot verify the loop ran at all before cancellation.

Trecek (commented):
AutoSkillit review found 15 blocking issues (4 critical, 11 warning). See inline comments.

src/autoskillit/execution/__init__.py
Trecek: Agreed — fixed in 92ffab3. Removed `"_refresh_quota_cache"` from `__all__` in `execution/__init__.py` (line 99). The import is kept for the re-export chain used by `server/helpers.py`, but the private symbol is no longer advertised as public API.

src/autoskillit/server/helpers.py
Trecek: Agreed — fixed in 4f84014. Changed `config: Any` to `config: QuotaGuardConfig` (TYPE_CHECKING import) in `_quota_refresh_loop` signature at `server/helpers.py:494`. `QuotaGuardConfig` is importable at the server layer without violating any arch constraint.

src/autoskillit/server/tools_kitchen.py
Trecek: Agreed — fixed in f628253. Added `if ctx.quota_refresh_task is not None: ctx.quota_refresh_task.cancel()` before the task assignment at `tools_kitchen.py:77`, mirroring the cancel guard already present in `_close_kitchen_handler` at line 99.

src/autoskillit/execution/quota.py
Trecek: Investigated — this is intentional. `config: Any` in `_quota_refresh_cache` at `execution/quota.py:110` is an intentional L1 isolation constraint: importing `QuotaGuardConfig` from `autoskillit.config.settings` would create a cross-subpackage L1 dependency between `execution/` and `config/`, which the architecture prohibits (L1 execution imports only L0 — contract in `tests/arch/`). The same `config: Any` pattern is used at `check_and_sleep_if_needed` line 130 for the same reason. This is not a defense gap but an enforced structural constraint.

src/autoskillit/server/tools_kitchen.py
Trecek: Investigated — this is intentional. In Python 3.8+, `CancelledError` is a subclass of `BaseException`, not `Exception`. The asyncio 'task exception was never retrieved' log is only emitted for non-`CancelledError` exceptions stored in a Task. When `.cancel()` fires and the task exits via `CancelledError` propagation from `asyncio.sleep`, Python does not treat this as an unhandled exception. Setting the handle to `None` after `.cancel()` is safe for this specific pattern (confirmed by the `_quota_refresh_loop` docstring at `helpers.py:498`: 'CancelledError from asyncio.sleep propagates uncaught, terminating the loop cleanly').

src/autoskillit/core/_type_protocols.py
Trecek: Investigated — this is intentional. `QuotaRefreshTask.cancel()` at `core/_type_protocols.py:474` already carries the `-> bool` return annotation: `def cancel(self, msg: Any = None) -> bool: ...`. The annotation was added in commit 61c7ee9. No change needed.

tests/config/test_config.py
Trecek: Investigated — this is intentional. Lines 376–377 of `tests/config/test_config.py` explicitly assign both variables before the assertion: `interval = defaults['quota_guard']['cache_refresh_interval']` (line 376) and `max_age = defaults['quota_guard']['cache_max_age']` (line 377). The `assert interval < max_age` on line 378 is therefore valid. No `NameError` is possible.

tests/execution/test_quota.py
Trecek: Investigated — this is intentional. Line 491 of `tests/execution/test_quota.py` IS the assignment: `data = json.loads(cache_path.read_text())`. The assertion on line 492 follows immediately. The reviewer appears to have confused the assignment line with the assertion line. No `NameError` is possible.

tests/server/test_quota_refresh_loop.py
Trecek: Investigated — this is intentional. Each of the three test functions in `test_quota_refresh_loop.py` opens with a local-scope import of `_quota_refresh_loop` (e.g., line 16, 44, 61: `from autoskillit.server.helpers import _quota_refresh_loop`). The reviewer only inspected the module-level import block and missed the per-function deferred imports, which are an intentional pattern used throughout this test suite to defer symbol resolution. No `NameError` at runtime.

tests/server/test_tools_kitchen.py
Trecek: Investigated — this is intentional. The indentation at lines 514–518 of `test_tools_kitchen.py` is a valid, progressively nested `with`-statement chain: 4 spaces (outer `with`), 8 spaces (first nested `with`), 12 spaces (second nested `with`), 16 spaces (third nested `with`). Python parses this without error. The reviewer misidentified correct nesting as an indentation fault.

tests/server/test_quota_refresh_loop.py
Trecek: Investigated — this is intentional. The test's stated purpose (docstring: 'CancelledError from asyncio.sleep propagates; loop does not swallow it') is narrowly scoped to error propagation — not call-count verification. Because `immediate_cancel` fires on the first `asyncio.sleep` before any refresh runs, asserting a zero call count would be redundant with the stub's behavior. The absence of a call-count assertion is an intentional scope choice.

src/autoskillit/server/helpers.py
Trecek: Valid observation — flagged for design decision. The re-export at `server/helpers.py:16` is required for REQ-IMP-003 compliance (enforced in `tests/arch/test_import_paths.py` lines 126–143), which forbids `tools_*.py` files from importing `autoskillit.execution` directly. Commit 61c7ee9 documents the routing intent explicitly. The reviewer's alternative ('canonical import site') conflicts with REQ-IMP-003. Resolving this cleanly requires a human architecture decision (e.g., passing the coroutine factory via `ToolContext`).

src/autoskillit/server/tools_execution.py
Trecek: Valid observation — flagged for design decision. The deferred import at `tools_execution.py:306` is a REQ-IMP-003 workaround (commit 61c7ee9). Moving it to module level would require importing from `autoskillit.execution` directly in a `tools_*.py` file, which violates the arch constraint. The reviewer's suggestion to pass the coroutine factory via `ToolContext` is sound but requires non-trivial changes to `ToolContext`, `_factory.py`, and `server/helpers`. Flagged for human review.

src/autoskillit/server/tools_execution.py
Trecek: Valid observation — flagged for design decision. The post-run refresh at `tools_execution.py:308` is intentional defense-in-depth (commit 32d8178: 'defense-in-depth post-run_skill background refresh'). Moving it to the execution or pipeline layer is architecturally sound but requires deciding how the background supervisor becomes accessible outside the server layer. Flagged for human review.

tests/server/test_quota_refresh_loop.py
Trecek: Valid observation — flagged for design decision. The reviewer is technically correct that `task.cancel()` on line 52 becomes a no-op: `immediate_cancel` raises `CancelledError` on the first sleep call, terminating the coroutine before the external cancel signal arrives. The test exercises the internal raise path, not the external cancellation signal path — which does not match the test name 'exits_cleanly_on_cancel'. Whether to redesign the test to use a proper async event requires a human decision.

PR #658, opened 2026-04-07
Rectify: research recipe archives wrong experiment directory; replace find heuristic with context tracking (#655)

### LINKED ISSUE #655: research recipe archives wrong experiment directory; replace find heuristic with context tracking
### What Happened

PR [TalonT-Org/spectral-init#239](https://github.com/TalonT-Org/spectral-init/pull/239) (branch `research-artifacts-20260407-004436`) was created by the research recipe. The PR body claims `artifacts.tar.gz` exists with extraction instructions, but all artifacts are committed as uncompressed plain-text files — 100+ markdown files, JSON, PNGs, .npy data files, the works.

The `commit_research_artifacts` step ran and exited 0. The `tar czf` command executed successfully — but it compressed the **wrong experiment directory**.

### Root Cause

**File:** `src/autoskillit/recipes/research.yaml`, line 403

```bash
RESEARCH_DIR=$(find "${{ context.worktree_path }}/research" -maxdepth 1 -type d -name '????-??-??-*' | sort | tail -1)
```

The worktree contained two experiment directories from the same pipeline run (the pipeline iterated/pivoted experiment design, producing both):

| Directory | `sort` Order | Selected? |
|---|---|---|
| `research/2026-04-07-kdtree-y-knn-trustworthiness/` | First (`k` < `t`) | No |
| `research/2026-04-07-tw-profiler-variant-dispatch/` | **Last** (`t` > `k`) | **Yes — wrong one** |

`sort | tail -1` picks alphabetically last, not "most recent" or "current experiment." Since both directories share the `2026-04-07` date prefix, tiebreaking falls to the slug. The tar compressed `tw-profiler-variant-dispatch`; the actual final experiment (`kdtree-y-knn-trustworthiness`) was left untouched.

This bug is inherent to any multi-iteration research run — the pipeline will always produce multiple date-prefixed directories and the heuristic will always pick whichever slug sorts last alphabetically.

### Compounding Factors

Three additional design issues amplify the impact:

1. **`create_artifact_branch` (line 606) does a blanket checkout:**
   ```bash
   git -C "${ARTIFACT_DIR}" checkout "${EXPERIMENT_BRANCH}" -- research/
   ```
   Copies the **entire** `research/` tree — including uncompressed files from the directory that should have been archived. No scoping.

2. **`open_artifact_pr` (line 633) uses a hardcoded body template** that always claims `artifacts.tar.gz` exists regardless of whether compression actually happened on the correct directory.

3. **No post-compression guard.** The step exits 0 because `tar` itself succeeds — it just targets the wrong input. Nothing verifies the intended directory was actually compressed.

### Pipeline Trace

From orchestrator session `2fa9a34e` (kitchen `2d2900f4`):

| Step | Exit | What Actually Happened |
|---|---|---|
| `commit_research_artifacts` | 0 | Compressed `tw-profiler-variant-dispatch` (8556-byte tarball), left `kdtree-y-knn-trustworthiness` untouched |
| `push_branch` | 0 | Pushed branch with mixed state (one dir compressed, one not) |
| `create_artifact_branch` | 0 | Blanket `research/` checkout pulled in all uncompressed kdtree files |
| `open_artifact_pr` | 0 | Created PR #239 with hardcoded body claiming tarball exists |
| `tag_experiment_branch` | 0 | Tagged the mixed-state branch |
| `close_experiment_pr` | 0 | Closed PR #238 with cross-references to the broken artifact PR |

Every step exited 0. The pipeline reported full success.

### Test Gap

Tests in `tests/recipe/test_research_artifact_archive.py` only do static assertions on the recipe YAML (e.g., "does the command string contain `tar czf`?"). No test covers directory-selection with multiple directories, post-compression state, or alignment between what was compressed and what lands in the artifact PR.

---

### 1. Capture research directory path into context during creation

The research directory is known at creation time (during `scope` or experiment setup). Capture it as `context.research_dir` and thread it to all downstream steps. Trace the pipeline flow to find which step first creates the `research/YYYY-MM-DD-slug/` directory and add a `capture:` clause there.

### 2. Replace find heuristic in `commit_research_artifacts`

**File:** `src/autoskillit/recipes/research.yaml`, lines 399-436

Replace:
```bash
RESEARCH_DIR=$(find "${{ context.worktree_path }}/research" -maxdepth 1 -type d -name '????-??-??-*' | sort | tail -1)
```
With:
```bash
RESEARCH_DIR="${{ context.research_dir }}"
```

### 3. Scope `create_artifact_branch` checkout

**File:** `src/autoskillit/recipes/research.yaml`, line 606

Replace the blanket `git checkout ... -- research/` with a scoped checkout that only pulls the specific experiment directory. Derive the relative path from `context.research_dir`.

### 4. Add post-compression guard

After `tar czf` and `rm -rf` in `commit_research_artifacts`, verify:
- `artifacts.tar.gz` exists in `RESEARCH_DIR`
- `artifacts/` subdirectory does NOT exist
- `README.md` exists

Exit non-zero if guard fails.

### 5. Make `open_artifact_pr` body dynamic

**File:** `src/autoskillit/recipes/research.yaml`, lines 626-648

Inspect actual files in the artifact commit instead of hardcoding "artifacts.tar.gz — Full artifact tree."

### 6. Add multi-directory test

**File:** `tests/recipe/test_research_artifact_archive.py`

Test that directory selection works correctly with two `????-??-??-*` directories sharing the same date prefix.

### Acceptance Criteria

- [ ] `context.research_dir` set during experiment creation, threaded to `commit_research_artifacts`
- [ ] `commit_research_artifacts` no longer uses `find | sort | tail -1`
- [ ] `create_artifact_branch` scoped to specific experiment directory
- [ ] Post-compression guard fails step if tarball missing or uncompressed files remain
- [ ] `open_artifact_pr` body reflects actual filesystem state
- [ ] New test covers multi-directory scenario
- [ ] Existing tests still pass

### Affected Version

`research.yaml` as of v0.7.28 (`integration`, commit `33269ae4`).

Trecek:
Closes #655

## Summary

Replaces the fragile `find | sort | tail -1` heuristic in `research.yaml` with deterministic `context.research_dir` tracking. Adds two new semantic rules (`run-cmd-emit-alignment`, `run-cmd-find-rediscovery`) to catch this class of bug at validation time.

## Changes

- **`research.yaml`**: `create_worktree` emits `research_dir` into context; `commit_research_artifacts` uses it directly; `create_artifact_branch` checkout scoped to specific experiment dir; post-compression guard added; `open_artifact_pr` body made dynamic
- **`src/autoskillit/recipe/rules_cmd.py`**: two new semantic rules registered in `__init__.py`
- **`tests/recipe/test_rules_cmd.py`**: 7 new tests for the semantic rules
- **`tests/recipe/test_research_artifact_archive.py`**: 7 new tests for context tracking and guard behavior

### DIFF
--- a/src/autoskillit/recipe/rules_cmd.py
+++ b/src/autoskillit/recipe/rules_cmd.py
@@ -1,0 +1,94 @@
+"""Semantic rules for run_cmd echo-capture alignment in recipe steps."""
+
+from __future__ import annotations
+
+import re
+
+from autoskillit.core import Severity
+from autoskillit.recipe._analysis import ValidationContext
+from autoskillit.recipe.contracts import _RESULT_CAPTURE_RE
+from autoskillit.recipe.registry import RuleFinding, semantic_rule
+
+# Raw tool output fields — these are populated directly from the tool JSON response,
+# no echo statement in the cmd is required to capture them.
+_RAW_RESULT_FIELDS = {"stdout", "stderr", "exit_code"}
+
+# Matches find ... | sort ... | (tail|head) patterns that indicate a step is
+# re-discovering a path that should have been captured by an upstream step.
+_FIND_HEURISTIC_RE = re.compile(r"\bfind\b.+\|\s*sort\b.+\|\s*(tail|head)\b", re.DOTALL)
+
+
+@semantic_rule(
+    name="run-cmd-emit-alignment",
+    description=(
+        "For every run_cmd step, each non-raw capture key K must have a matching "
+        'echo "K=..." in the cmd. A missing echo causes a silent empty-string capture.'
+    ),
+    severity=Severity.ERROR,
+)
+def _check_run_cmd_emit_alignment(ctx: ValidationContext) -> list[RuleFinding]:
+    findings: list[RuleFinding] = []
+    for name, step in ctx.recipe.steps.items():
+        if step.tool != "run_cmd":
+            continue
+        cmd = (step.with_args or {}).get("cmd", "")
+        if not isinstance(cmd, str):
+            continue
+        for cap_key, cap_val in step.capture.items():
+            m = _RESULT_CAPTURE_RE.search(cap_val)
+            if m is None:
+                # Cannot determine the result field — skip (e.g. pipe-filtered values).
+                continue
+            result_key = m.group(1)
+            if result_key in _RAW_RESULT_FIELDS:
+                continue
+            # Check that the cmd emits `echo "result_key=..."`.
+            echo_pattern = re.compile(rf'\becho\s+"?{re.escape(result_key)}=')
+            if not echo_pattern.search(cmd):
+                findings.append(
+                    RuleFinding(
+                        rule="run-cmd-emit-alignment",
+                        severity=Severity.ERROR,
+                        step_name=name,
+                        message=(
+                            f"Step '{name}' captures '{cap_key}' from result.{result_key} "
+                            f'but cmd contains no `echo "{result_key}=..."` statement. '
+                            f'Add `echo "{result_key}=${{...}}"` to the cmd or the '
+                            "captured value will always be empty."
+                        ),
+                    )
+                )
+    return findings
+
+
+@semantic_rule(
+    name="run-cmd-find-rediscovery",
+    description=(
+        "Flags run_cmd steps using find|sort|tail/head to select a path — this pattern "
+        "indicates an upstream step computed the path but did not echo it into context."
+    ),
+    severity=Severity.WARNING,
+)
+def _check_run_cmd_find_rediscovery(ctx: ValidationContext) -> list[RuleFinding]:
+    findings: list[RuleFinding] = []
+    for name, step in ctx.recipe.steps.items():
+        if step.tool != "run_cmd":
+            continue
+        cmd = (step.with_args or {}).get("cmd", "")
+        if not isinstance(cmd, str):
+            continue
+        if _FIND_HEURISTIC_RE.search(cmd):
+            findings.append(
+                RuleFinding(
+                    rule="run-cmd-find-rediscovery",
+                    severity=Severity.WARNING,
+                    step_name=name,
+                    message=(
+                        f"Step '{name}' uses a `find | sort | tail` heuristic to select "
+                        "a directory. This pattern indicates an upstream step computed "
+                        "the path but did not echo it into context. Capture the path via "
+                        "echo+capture in the originating step instead."
+                    ),
+                )
+            )
+    return findings
--- a/src/autoskillit/recipe/rules_cmd.py
+++ b/src/autoskillit/recipe/rules_cmd.py
@@ -15,7 +15,7 @@ _RAW_RESULT_FIELDS = {"stdout", "stderr", "exit_code"}
 
 # Matches find ... | sort ... | (tail|head) patterns that indicate a step is
 # re-discovering a path that should have been captured by an upstream step.
-_FIND_HEURISTIC_RE = re.compile(r"\bfind\b.+\|\s*sort\b.+\|\s*(tail|head)\b", re.DOTALL)
+_FIND_HEURISTIC_RE = re.compile(r"\bfind\b.+\|\s*sort\b.+\|\s*(tail|head)\b")
 
 
 @semantic_rule(
--- a/src/autoskillit/recipes/research.yaml
+++ b/src/autoskillit/recipes/research.yaml
@@ -642,7 +642,7 @@ steps:
         RESEARCH_SUBDIR="research/$(basename "${{ context.research_dir }}")" &&
         ARTIFACT_DIR="${{ context.worktree_path }}" &&
         if [ -f "${ARTIFACT_DIR}/${RESEARCH_SUBDIR}/artifacts.tar.gz" ]; then
-          STRUCT_DETAIL="- \`artifacts.tar.gz\` — compressed artifact tree\n\nExtract:\n\`\`\`bash\ntar xzf artifacts.tar.gz\n\`\`\`";
+          STRUCT_DETAIL=$'- `artifacts.tar.gz` \xe2\x80\x94 compressed artifact tree\n\nExtract:\n```bash\ntar xzf artifacts.tar.gz\n```';
         else
           STRUCT_DETAIL="- \`artifacts/\` — uncompressed artifact tree";
         fi &&
--- a/src/autoskillit/recipes/research.yaml
+++ b/src/autoskillit/recipes/research.yaml
@@ -640,7 +640,7 @@ steps:
         ARTIFACT="${{ context.artifact_branch }}" &&
         ORIG_PR="${{ context.pr_url }}" &&
         RESEARCH_SUBDIR="research/$(basename "${{ context.research_dir }}")" &&
-        ARTIFACT_DIR="$(git -C '${{ context.worktree_path }}' rev-parse --show-toplevel)" &&
+        ARTIFACT_DIR="${{ context.worktree_path }}" &&
         if [ -f "${ARTIFACT_DIR}/${RESEARCH_SUBDIR}/artifacts.tar.gz" ]; then
           STRUCT_DETAIL="- \`artifacts.tar.gz\` — compressed artifact tree\n\nExtract:\n\`\`\`bash\ntar xzf artifacts.tar.gz\n\`\`\`";
         else
--- a/src/autoskillit/recipes/research.yaml
+++ b/src/autoskillit/recipes/research.yaml
@@ -198,11 +198,13 @@ steps:
         for f in "${SRC_TEMP}"/plan-experiment/experiment_plan_*.md; do [ -f "$f" ] && cp "$f" "${RESEARCH_DIR}/artifacts/plan-versions/"; done;
         for f in "${SRC_TEMP}"/resolve-design-review/*.md; do [ -f "$f" ] && cp "$f" "${RESEARCH_DIR}/artifacts/review-cycles/"; done;
         cd "${RESOLVED}" && git add research/ && git commit -m "Add experiment plan and scope to research/" &&
+        echo "research_dir=${RESEARCH_DIR}" &&
         echo "worktree_path=${RESOLVED}"
       cwd: "${{ inputs.source_dir }}"
       step_name: create_worktree
     capture:
       worktree_path: "${{ result.worktree_path }}"
+      research_dir: "${{ result.research_dir }}"
     on_success: decompose_phases
     on_failure: escalate_stop
     note: >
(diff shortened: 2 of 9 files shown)


Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (COMMENT mode: cannot request changes on own PR)

src/autoskillit/recipe/rules_cmd.py
```diff
@@ -0,0 +1,94 @@
...
+from autoskillit.recipe.contracts import _RESULT_CAPTURE_RE
+from autoskillit.recipe.registry import RuleFinding, semantic_rule
+
+# Raw tool output fields — these are populated directly from the tool JSON response,
+# no echo statement in the cmd is required to capture them.
+_RAW_RESULT_FIELDS = {"stdout", "stderr", "exit_code"}
+
+# Matches find ... | sort ... | (tail|head) patterns that indicate a step is
+# re-discovering a path that should have been captured by an upstream step.
+_FIND_HEURISTIC_RE = re.compile(r"\bfind\b.+\|\s*sort\b.+\|\s*(tail|head)\b", re.DOTALL)
```
Trecek: [warning] bugs: _FIND_HEURISTIC_RE uses re.DOTALL, allowing `.` to match newlines. In a multi-line YAML cmd block, a `find` on one line and an unrelated `| sort | tail` pipeline on a later line will be incorrectly joined into a match, producing false positives. Remove re.DOTALL — the regex still matches the intended single-line pattern without it.

src/autoskillit/recipe/rules_cmd.py
```diff
@@ -0,0 +1,94 @@
...
+def _check_run_cmd_emit_alignment(ctx: ValidationContext) -> list[RuleFinding]:
+    findings: list[RuleFinding] = []
+    for name, step in ctx.recipe.steps.items():
+        if step.tool != "run_cmd":
+            continue
+        cmd = (step.with_args or {}).get("cmd", "")
+        if not isinstance(cmd, str):
+            continue
+        for cap_key, cap_val in step.capture.items():
+            m = _RESULT_CAPTURE_RE.search(cap_val)
```
Trecek, line 38: [warning] defense: If `step.capture` is None (a step with no capture block), calling `step.capture.items()` will raise AttributeError. Add a None-guard: `if not step.capture: continue` before the loop.

src/autoskillit/recipe/rules_cmd.py
```diff
@@ -0,0 +1,94 @@
...
+            if m is None:
+                # Cannot determine the result field — skip (e.g. pipe-filtered values).
+                continue
+            result_key = m.group(1)
+            if result_key in _RAW_RESULT_FIELDS:
+                continue
+            # Check that the cmd emits `echo "result_key=..."`.
```
Trecek, line 45: [warning] slop: Inline comment `# Check that the cmd emits \`echo "result_key=..."\`.` restates exactly what the next two lines do without adding any reasoning about why. Remove it.

src/autoskillit/recipe/rules_cmd.py
```diff
@@ -0,0 +1,94 @@
...
+            echo_pattern = re.compile(rf'\becho\s+"?{re.escape(result_key)}=')
```
Trecek, line 46: [warning] bugs: The echo_pattern `r'\becho\s+"?{result_key}='` allows an optional double-quote but not a single-quote. A cmd using `echo 'KEY=value'` (single-quoted) will not match, producing a false-positive alignment error. Update pattern to `r'\becho\s+[\"\\']?{result_key}='` to accept both quote styles.

src/autoskillit/recipes/research.yaml
```diff
@@ -630,7 +639,14 @@ steps:
         EXPERIMENT="${{ context.experiment_branch }}" &&
         ARTIFACT="${{ context.artifact_branch }}" &&
         ORIG_PR="${{ context.pr_url }}" &&
-        PR_BODY=$(printf '## Research Artifacts\n\nFrom experiment branch `%s`.\n\nOriginal experiment PR: %s\n\n### Structure\n\n- **README.md** — Research conclusions (browsable on GitHub)\n- **artifacts.tar.gz** — Full artifact tree (experiment plan, scope report, design evaluation, review cycles, plan versions, phase groups, phase plans)\n\nTo extract the full artifact tree:\n```\ntar xzf artifacts.tar.gz\n```\n\nContains only files under `research/` — zero production code changes.' "${EXPERIMENT}" "${ORIG_PR}") &&
+        RESEARCH_SUBDIR="research/$(basename "${{ context.research_dir }}")" &&
+        ARTIFACT_DIR="$(git -C '${{ context.worktree_path }}' rev-parse --show-toplevel)" &&
```
Trecek: [warning] bugs: `ARTIFACT_DIR` is derived via `git rev-parse --show-toplevel` on `context.worktree_path`, which just returns `context.worktree_path` itself. The variable name `ARTIFACT_DIR` is misleading — it shadows the artifact worktree path used in `create_artifact_branch`. Replace with `ARTIFACT_DIR="${{ context.worktree_path }}"` for clarity, or rename to avoid confusion with the artifact worktree path.

src/autoskillit/recipes/research.yaml
```diff
@@ -630,7 +639,14 @@ steps:
...
+        if [ -f "${ARTIFACT_DIR}/${RESEARCH_SUBDIR}/artifacts.tar.gz" ]; then
+          STRUCT_DETAIL="- \`artifacts.tar.gz\` — compressed artifact tree\n\nExtract:\n\`\`\`bash\ntar xzf artifacts.tar.gz\n\`\`\`";
```
Trecek: [warning] bugs: `STRUCT_DETAIL` is assigned as a bash double-quoted string containing `\n` escape sequences. In bash, `\n` in double-quoted strings is the literal two-char sequence `\n`, not a newline. When substituted via `printf '%s'` at L649, these remain literal, breaking the embedded code fence in the PR body. Use `$'...\n...'` (ANSI-C quoting) or a heredoc to embed real newlines.

tests/recipe/test_research_artifact_archive.py
```diff
@@ -60,3 +60,67 @@ def test_commit_research_artifacts_routing_preserved(recipe):
...
+    has_heuristic = "find" in cmd and "sort" in cmd and "tail" in cmd
+    assert not has_heuristic, (
+        "commit_research_artifacts uses find|sort|tail heuristic — "
+        "replace with ${{ context.research_dir }}"
+    )
+
+
+def test_create_artifact_branch_scopes_checkout(recipe):
+    """create_artifact_branch must scope git checkout to the specific experiment subdir."""
+    step = recipe.steps["create_artifact_branch"]
```
Trecek, line 104: [warning] tests: `assert "basename" in cmd` is a weak structural signal — any incidental use of `basename` for unrelated purposes would pass. The assertion does not verify that `basename` is applied to `context.research_dir` specifically, which is the stated intent. Strengthen to `assert 'basename' in cmd and 'context.research_dir' in cmd`.

tests/recipe/test_rules_cmd.py
```diff
@@ -0,0 +1,134 @@
+"""Tests for run_cmd echo-capture alignment rules."""
+
+from __future__ import annotations
+
+from autoskillit.recipe.validator import run_semantic_rules
+from tests.recipe.conftest import _make_workflow
```
Trecek, line 6: [warning] tests: Importing the private helper `_make_workflow` directly from `tests.recipe.conftest` creates brittle coupling — any conftest refactor silently breaks this import. Expose it as a fixture or move it to a shared test utility module.

Trecek (commented):
AutoSkillit review found 8 blocking issues. See inline comments. (verdict: changes_requested — unable to formally request changes on own PR)

src/autoskillit/recipe/rules_cmd.py
Trecek: Agreed — fixed in feb9aeb. Removed re.DOTALL from _FIND_HEURISTIC_RE; the regex confines each .+ to a single line, eliminating cross-line false positives in multi-line YAML cmd blocks while still matching the intended single-line find|sort|tail pattern.

src/autoskillit/recipe/rules_cmd.py
Trecek: Investigated — this is intentional. In schema.py, RecipeStep.capture is declared as `capture: dict[str, str] = field(default_factory=dict)`. The default_factory=dict guarantees the field is always an empty dict, never None. Calling step.capture.items() cannot raise AttributeError.

src/autoskillit/recipe/rules_cmd.py
Trecek: Flagged for design decision. This inline comment was classified below the warning threshold (cosmetic/slop). Deferring to human review — no automatic change applied.

src/autoskillit/recipe/rules_cmd.py
Trecek: Investigated — this is intentional. The rule enforces a documented double-quoted echo convention. The error message prescribes `echo "result_key=${...}"` as the required form, and the rule description states `echo "K=..."`. This validator intentionally rejects single-quoted cmds that deviate from the documented contract; widening the pattern would silently pass non-conforming recipes.

src/autoskillit/recipes/research.yaml
Trecek: Agreed — fixed in b9c32f7. Replaced the redundant `git rev-parse --show-toplevel` subshell with direct `context.worktree_path`. In a git worktree, the command returns the worktree path itself, making the subshell invocation equivalent but misleading — especially given ARTIFACT_DIR in create_artifact_branch refers to a mktemp temporary path.

src/autoskillit/recipes/research.yaml
Trecek: Agreed — fixed in 0c9cb30. Changed STRUCT_DETAIL to ANSI-C quoting ($'...') so \n sequences expand to real newlines. In bash double-quoted strings, \n is the literal two-character sequence; only ANSI-C quoting expands it. Without this fix, the embedded code fence in the PR body renders as a run-on paragraph.

tests/recipe/test_research_artifact_archive.py
Trecek: Agreed — fixed in 916c2ab. Strengthened to `assert 'basename' in cmd and 'context.research_dir' in cmd`. The create_artifact_branch cmd template uses basename specifically for context.research_dir (RESEARCH_SUBDIR line), so the combined assertion accurately captures the required relationship and closes the gap where basename used for an unrelated variable would have passed.

tests/recipe/test_rules_cmd.py
Trecek: Investigated — this is intentional. Direct import of `_make_workflow` from `tests.recipe.conftest` is the established pattern across the entire tests/recipe/ directory: test_rules_worktree.py, test_validator.py, test_rules_dataflow.py, and test_rules_structure.py all use the identical import. The underscore prefix signals it is not a pytest fixture, not that it is private. The comment misreads a deliberate, widely-replicated convention as brittleness.

PR #708, opened 2026-04-11
Rectify: Quota Dataclass Type Boundary Enforcement

Trecek:
## Summary

The `open_kitchen` tool crashed with `TypeError: float() argument must be a string or a real number, not 'NoneType'` because `QuotaStatus` and `QuotaWindowEntry` were the only L1 dataclasses lacking `__post_init__` type enforcement, allowing `None` to silently propagate through three exception-handling layers that all omitted `TypeError`.

The fix is twofold: (1) add `__post_init__` validation gates on the quota dataclasses so the bug class is structurally impossible at the type level (matching the codebase convention established by `ClaudeSessionResult`, `RecipeIngredient`, and `Recipe`), and (2) complete the exception chains so any future bypass still fails open — `_read_cache` now catches `TypeError` in its tuple, while `check_and_sleep_if_needed` and `_prime_quota_cache` are broadened to `except Exception` (with `exc_info=True` logging) to honor the explicit fail-open subsystem-boundary contract. A secondary `.get("key", default)` anti-pattern — where a present-but-`null` JSON value bypassed the default and raised `TypeError` on `float(None)` — is fixed in `pipeline/tokens.py`, `pipeline/timings.py`, and `hooks/token_summary_appender.py` by replacing the implicit default with explicit `is not None` guards.

### Error/Resilience Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% EXTERNAL SOURCES %%
    API([Anthropic Quota API<br/>━━━━━━━━━━<br/>JSON: utilization may be null])
    CACHE([Quota Cache File<br/>━━━━━━━━━━<br/>JSON: may contain nulls])
    JSONL([Session JSONL<br/>━━━━━━━━━━<br/>timing/total_seconds may be null])

    subgraph Ingestion ["● INGESTION BOUNDARY"]
        direction TB
        FETCH["● _fetch_quota<br/>━━━━━━━━━━<br/>Skip null-utilization<br/>windows pre-construction"]
        RCACHE["● _read_cache<br/>━━━━━━━━━━<br/>except tuple now includes<br/>TypeError"]
        TOKLOAD["● DefaultTokenLog.load_from_log_dir<br/>━━━━━━━━━━<br/>is-not-None guard on<br/>timing_seconds"]
        TIMLOAD["● DefaultTimingLog.load_from_log_dir<br/>━━━━━━━━━━<br/>is-not-None guard on<br/>total_seconds"]
        TSLOAD["● token_summary_appender._load_sessions<br/>━━━━━━━━━━<br/>is-not-None guard on<br/>timing_seconds"]
    end

    subgraph TypeGates ["★ TYPE BOUNDARY: __post_init__ VALIDATION"]
        direction TB
        QWE["★ QuotaWindowEntry.__post_init__<br/>━━━━━━━━━━<br/>None → raise TypeError<br/>Coerce float at construction"]
        QS["★ QuotaStatus.__post_init__<br/>━━━━━━━━━━<br/>None → raise TypeError<br/>Coerce float at construction"]
    end

    subgraph FailOpen ["● FAIL-OPEN SUBSYSTEM BOUNDARIES"]
        direction TB
        CSLEEP["● check_and_sleep_if_needed<br/>━━━━━━━━━━<br/>except Exception + exc_info<br/>Never propagates"]
        PRIME["● _prime_quota_cache<br/>━━━━━━━━━━<br/>except Exception<br/>Kitchen open never aborts"]
    end

    subgraph Callers ["CONSUMERS (protected by upstream gates)"]
        direction TB
        OPEN[open_kitchen<br/>━━━━━━━━━━<br/>calls _prime_quota_cache]
        HOOK_PRE[quota_check PreTool hook<br/>━━━━━━━━━━<br/>pre-existing TypeError catch]
        HOOK_POST[quota_post_check PostTool hook<br/>━━━━━━━━━━<br/>pre-existing TypeError catch]
        REPORT[token/timing reports<br/>━━━━━━━━━━<br/>accumulate elapsed_seconds<br/>null treated as 0.0]
    end

    %% TERMINAL STATES %%
    SAFE([FAIL-OPEN<br/>Kitchen opens; no sleep])
    CRASH([TypeError CRASH<br/>Pre-fix regression surface])

    %% DATA FLOW %%
    API --> FETCH
    CACHE --> RCACHE
    JSONL --> TOKLOAD
    JSONL --> TIMLOAD
    JSONL --> TSLOAD

    FETCH -->|"construct"| QWE
    FETCH -->|"construct"| QS
    RCACHE -->|"construct"| QS

    %% TYPE-GATE OUTCOMES %%
    QWE -->|"None utilization"| CSLEEP
    QS -->|"None utilization"| CSLEEP
    RCACHE -.->|"TypeError caught<br/>(pre-fix: propagated)"| CSLEEP

    %% FAIL-OPEN CHAIN %%
    FETCH --> CSLEEP
    CSLEEP -->|"any Exception<br/>exc_info logged"| PRIME
    PRIME -->|"any Exception<br/>logger.warning"| OPEN
    OPEN --> SAFE

    %% PRE-FIX REGRESSION EDGE (counterfactual) %%
    QS -.->|"pre-fix:<br/>float(None) TypeError<br/>uncaught"| CRASH

    %% HOOK INDEPENDENCE %%
    CACHE --> HOOK_PRE
    CACHE --> HOOK_POST

    %% LOG-LOAD FAIL-SOFT %%
    TOKLOAD -->|"null → 0.0"| REPORT
    TIMLOAD -->|"null → 0.0"| REPORT
    TSLOAD -->|"null → 0.0"| REPORT

    %% CLASS ASSIGNMENTS %%
    class API,CACHE,JSONL cli;
    class QWE,QS newComponent;
    class FETCH,RCACHE,TOKLOAD,TIMLOAD,TSLOAD handler;
    class CSLEEP,PRIME detector;
    class OPEN,HOOK_PRE,HOOK_POST,REPORT phase;
    class SAFE terminal;
    class CRASH gap;
```

### State Lifecycle Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    subgraph External ["EXTERNAL UNTYPED INPUT"]
        direction TB
        API_JSON([Anthropic API JSON<br/>━━━━━━━━━━<br/>dict with possible null])
        CACHE_JSON([Cache File JSON<br/>━━━━━━━━━━<br/>dict with possible null])
        SESS_JSON([Session JSONL<br/>━━━━━━━━━━<br/>timing_seconds may be null])
    end

    subgraph Lifecycles ["FIELD LIFECYCLE CATEGORIES"]
        direction TB
        INIT_ONLY["INIT_ONLY<br/>━━━━━━━━━━<br/>utilization, resets_at,<br/>window_name<br/>Set in __post_init__<br/>Never mutated after"]
        INIT_PRESERVE["INIT_PRESERVE<br/>━━━━━━━━━━<br/>Same three fields round-trip<br/>through cache JSON"]
        APPEND["APPEND_ONLY<br/>━━━━━━━━━━<br/>TokenEntry.elapsed_seconds<br/>TimingEntry.total_seconds"]
        DERIVED["DERIVED<br/>━━━━━━━━━━<br/>QuotaFetchResult.binding<br/>Computed from windows"]
    end

    subgraph Gates ["★ VALIDATION GATES (new __post_init__)"]
        direction TB
        GATE_QS["★ QuotaStatus.__post_init__<br/>━━━━━━━━━━<br/>utilization is None → TypeError<br/>else coerce to float"]
        GATE_QWE["★ QuotaWindowEntry.__post_init__<br/>━━━━━━━━━━<br/>utilization is None → TypeError<br/>else coerce to float"]
        GATE_SKIP["● _fetch_quota null-skip<br/>━━━━━━━━━━<br/>Pre-constructor guard<br/>raw_util is None → continue"]
        GATE_APPEND["● load_from_log_dir guards<br/>━━━━━━━━━━<br/>is-not-None guard before<br/>float() accumulation"]
    end

    subgraph Accessors ["STATE ACCESS (no mutation paths)"]
        direction TB
        FETCH["_fetch_quota<br/>━━━━━━━━━━<br/>Constructs windows dict"]
        READ_CACHE["_read_cache<br/>━━━━━━━━━━<br/>Reconstructs QuotaStatus<br/>TypeError now caught"]
        WRITE_CACHE["_write_cache<br/>━━━━━━━━━━<br/>Serializes binding<br/>to JSON"]
        COMPUTE["_compute_binding<br/>━━━━━━━━━━<br/>windows → binding<br/>(pure function)"]
    end

    subgraph Consumers ["READ-ONLY CONSUMERS"]
        direction TB
        CHECK[check_and_sleep_if_needed<br/>━━━━━━━━━━<br/>Reads binding.utilization<br/>Reads binding.resets_at]
        HOOK_PRE[quota_check hook<br/>━━━━━━━━━━<br/>Reads binding from cache]
        HOOK_POST[quota_post_check hook<br/>━━━━━━━━━━<br/>Reads binding from cache]
        REPORTS[Token / Timing reports<br/>━━━━━━━━━━<br/>Aggregate elapsed_seconds]
    end

    %% CORRUPTED TERMINAL %%
    CORRUPT([CORRUPT STATE<br/>utilization = None<br/>unreachable post-fix])

    %% CONSTRUCTION PATH %%
    API_JSON --> FETCH
    CACHE_JSON --> READ_CACHE
    SESS_JSON --> GATE_APPEND

    FETCH -->|"per window"| GATE_SKIP
    GATE_SKIP -->|"value present"| GATE_QWE
    GATE_SKIP -.->|"null<br/>(dropped)"| FETCH
    GATE_QWE -->|"valid"| INIT_ONLY

    READ_CACHE -->|"construct"| GATE_QS
    GATE_QS -->|"valid"| INIT_ONLY
    GATE_QS -.->|"TypeError<br/>caught by _read_cache"| READ_CACHE

    %% POST-CONSTRUCTION FLOW %%
    INIT_ONLY --> COMPUTE
    COMPUTE --> DERIVED
    DERIVED --> INIT_PRESERVE
    INIT_PRESERVE --> WRITE_CACHE
    WRITE_CACHE --> CACHE_JSON

    DERIVED --> CHECK
    CACHE_JSON --> HOOK_PRE
    CACHE_JSON --> HOOK_POST

    %% APPEND-ONLY FIELDS %%
    GATE_APPEND --> APPEND
    APPEND --> REPORTS

    %% CORRUPTION PATH (counterfactual) %%
    GATE_QS -.->|"pre-fix bypass<br/>(no gate)"| CORRUPT
    GATE_QWE -.->|"pre-fix bypass<br/>(no gate)"| CORRUPT

    %% CLASS ASSIGNMENTS %%
    class API_JSON,CACHE_JSON,SESS_JSON cli;
    class INIT_ONLY detector;
    class INIT_PRESERVE gap;
    class APPEND handler;
    class DERIVED phase;
    class GATE_QS,GATE_QWE newComponent;
    class GATE_SKIP,GATE_APPEND stateNode;
    class FETCH,READ_CACHE,WRITE_CACHE,COMPUTE output;
    class CHECK,HOOK_PRE,HOOK_POST,REPORTS phase;
    class CORRUPT gap;
```

## Implementation Plan

Plan file: `.autoskillit/temp/rectify/rectify_open_kitchen_float_typeerror_2026-04-10_000000.md`

### DIFF
--- a/src/autoskillit/execution/quota.py
+++ b/src/autoskillit/execution/quota.py
@@ -29,6 +29,11 @@ class QuotaStatus:
     resets_at: datetime | None  # UTC-aware; None when utilization is 0
     window_name: str = "unknown"  # which window this came from (diagnostic)
 
+    def __post_init__(self) -> None:
+        if self.utilization is None:
+            raise TypeError("QuotaStatus.utilization must not be None")
+        self.utilization = float(self.utilization)
+
 
 @dataclass
 class QuotaWindowEntry:
@@ -36,6 +41,11 @@ class QuotaWindowEntry:
 
     utilization: float
     resets_at: datetime | None
+
+    def __post_init__(self) -> None:
+        if self.utilization is None:
+            raise TypeError("QuotaWindowEntry.utilization must not be None")
+        self.utilization = float(self.utilization)
 
 
 @dataclass
@@ -109,7 +119,7 @@ def _read_cache(cache_path: str, max_age: int) -> QuotaStatus | None:
             resets_at=_parse_resets_at(b.get("resets_at")),
             window_name=str(b.get("window_name", "unknown")),
         )
-    except (FileNotFoundError, KeyError, ValueError, json.JSONDecodeError):
+    except (FileNotFoundError, KeyError, ValueError, TypeError, json.JSONDecodeError):
         return None
 
 
@@ -161,8 +171,11 @@ async def _fetch_quota(
     windows: dict[str, QuotaWindowEntry] = {}
     for name, w in data.items():
         if isinstance(w, dict) and "utilization" in w:
+            raw_util = w["utilization"]
+            if raw_util is None:
+                continue
             windows[name] = QuotaWindowEntry(
-                utilization=float(w["utilization"]),
+                utilization=float(raw_util),
                 resets_at=_parse_resets_at(w.get("resets_at")),
             )
     if not windows:
@@ -301,15 +314,12 @@ async def check_and_sleep_if_needed(
             "window_name": status.window_name,
         }
 
-    except (
-        TimeoutError,
-        OSError,
-        KeyError,
-        ValueError,
-        json.JSONDecodeError,
-        httpx.HTTPError,
-    ) as exc:
-        _log.warning("quota check failed — continuing without sleep", error=str(exc))
+    except Exception as exc:
+        _log.warning(
+            "quota check failed — continuing without sleep",
+            error=str(exc),
+            exc_info=True,
+        )
         return {
             "should_sleep": False,
             "sleep_seconds": 0,
--- a/src/autoskillit/execution/quota.py
+++ b/src/autoskillit/execution/quota.py
@@ -328,13 +328,20 @@ async def check_and_sleep_if_needed(
             json.JSONDecodeError,
             httpx.HTTPError,
         )
-        _log_fn = _log.warning if isinstance(exc, _operational_types) else _log.error
-        _log_fn(
-            "quota check failed — continuing without sleep",
-            error=str(exc),
-            error_type=type(exc).__name__,
-            exc_info=True,
-        )
+        if isinstance(exc, _operational_types):
+            _log.warning(
+                "quota check failed — continuing without sleep",
+                error=str(exc),
+                error_type=type(exc).__name__,
+                exc_info=True,
+            )
+        else:
+            _log.error(
+                "quota check failed (unexpected error) — continuing without sleep",
+                error=str(exc),
+                error_type=type(exc).__name__,
+                exc_info=True,
+            )
         return {
             "should_sleep": False,
             "sleep_seconds": 0,
--- a/src/autoskillit/execution/quota.py
+++ b/src/autoskillit/execution/quota.py
@@ -315,9 +315,24 @@ async def check_and_sleep_if_needed(
         }
 
     except Exception as exc:
-        _log.warning(
+        # Fail-open subsystem-boundary contract: never raise on quota errors.
+        # Split severity so operational failures stay at WARNING while programming
+        # bugs (AttributeError, NameError, ImportError, ...) surface at ERROR in
+        # dashboards instead of being masked as routine transient errors.
+        _operational_types = (
+            TimeoutError,
+            OSError,
+            KeyError,
+            ValueError,
+            TypeError,
+            json.JSONDecodeError,
+            httpx.HTTPError,
+        )
+        _log_fn = _log.warning if isinstance(exc, _operational_types) else _log.error
+        _log_fn(
             "quota check failed — continuing without sleep",
             error=str(exc),
+            error_type=type(exc).__name__,
             exc_info=True,
         )
         return {
--- a/src/autoskillit/hooks/token_summary_appender.py
+++ b/src/autoskillit/hooks/token_summary_appender.py
@@ -231,7 +231,8 @@ def _load_sessions(
         entry["output_tokens"] += data.get("output_tokens", 0)
         entry["cache_creation_input_tokens"] += data.get("cache_creation_input_tokens", 0)
         entry["cache_read_input_tokens"] += data.get("cache_read_input_tokens", 0)
-        entry["elapsed_seconds"] += float(data.get("timing_seconds", 0.0))
+        _raw_timing = data.get("timing_seconds")
+        entry["elapsed_seconds"] += float(_raw_timing) if _raw_timing is not None else 0.0
         entry["invocation_count"] += 1
 
     return aggregated
--- a/src/autoskillit/pipeline/timings.py
+++ b/src/autoskillit/pipeline/timings.py
@@ -140,7 +140,8 @@ class DefaultTimingLog:
             if key not in self._entries:
                 self._entries[key] = TimingEntry(step_name=step_name)
             e = self._entries[key]
-            e.total_seconds += max(0.0, float(data.get("total_seconds", 0.0)))
+            _raw_total = data.get("total_seconds")
+            e.total_seconds += max(0.0, float(_raw_total) if _raw_total is not None else 0.0)
             e.invocation_count += 1
             count += 1
 
--- a/src/autoskillit/pipeline/tokens.py
+++ b/src/autoskillit/pipeline/tokens.py
@@ -216,7 +216,8 @@ class DefaultTokenLog:
             e.cache_read_input_tokens += data.get("cache_read_input_tokens", 0)
             # timing_seconds is the on-disk key name written by session_log.py;
             # elapsed_seconds is the in-memory field name on TokenEntry.
-            e.elapsed_seconds += float(data.get("timing_seconds", 0.0))
+            _raw_timing = data.get("timing_seconds")
+            e.elapsed_seconds += float(_raw_timing) if _raw_timing is not None else 0.0
             # Each token_usage.json file represents a single run_skill invocation
             # (one file = one invocation). Incrementing here reconstructs the
             # invocation count that was accumulated live via record().
--- a/src/autoskillit/server/helpers.py
+++ b/src/autoskillit/server/helpers.py
@@ -488,7 +488,7 @@ async def _prime_quota_cache() -> None:
 
     try:
         await check_and_sleep_if_needed(_get_ctx().config.quota_guard)
-    except (OSError, ValueError, RuntimeError):
+    except Exception:
         logger.warning("quota_prime_failed", exc_info=True)
 
 
--- a/tests/execution/test_quota.py
+++ b/tests/execution/test_quota.py
@@ -99,6 +99,62 @@ class TestReadCache:
         cache_path.write_text("not valid json{{")
         assert _read_cache(str(cache_path), max_age=120) is None
 
+    def test_read_cache_null_utilization_returns_none(self, tmp_path):
+        """Cache with utilization: null in binding must return None (fail-open)."""
+        from autoskillit.execution.quota import _read_cache
+
+        cache_path = tmp_path / "usage_cache.json"
+        payload = {
+            "fetched_at": datetime.now(UTC).isoformat(),
+            "binding": {"utilization": None, "resets_at": None, "window_name": "five_hour"},
+        }
+        cache_path.write_text(json.dumps(payload))
+        assert _read_cache(str(cache_path), max_age=300) is None
+
+    def test_read_cache_typeerror_returns_none(self, tmp_path):
+        """_read_cache must catch TypeError (e.g., float(None)) and return None."""
+        from autoskillit.execution.quota import _read_cache
+
+        cache_path = tmp_path / "usage_cache.json"
+        payload = {
+            "fetched_at": datetime.now(UTC).isoformat(),
+            "binding": {"utilization": None, "resets_at": None},
+        }
+        cache_path.write_text(json.dumps(payload))
+        # With __post_init__, constructing QuotaStatus with None raises TypeError.
+        # _read_cache must catch it and return None.
+        assert _read_cache(str(cache_path), max_age=300) is None
+
+
+class TestQuotaDataclassPostInit:
+    """QuotaStatus and QuotaWindowEntry must enforce types at construction."""
+
+    def test_quota_window_entry_rejects_none_utilization(self):
+        from autoskillit.execution.quota import QuotaWindowEntry
+
+        with pytest.raises(TypeError, match="utilization"):
+            QuotaWindowEntry(utilization=None, resets_at=None)
+
+    def test_quota_status_rejects_none_utilization(self):
+        from autoskillit.execution.quota import QuotaStatus
+
+        with pytest.raises(TypeError, match="utilization"):
+            QuotaStatus(utilization=None, resets_at=None)
+
+    def test_quota_window_entry_coerces_string_utilization(self):
+        from autoskillit.execution.quota import QuotaWindowEntry
+
+        entry = QuotaWindowEntry(utilization="85.5", resets_at=None)
+        assert entry.utilization == 85.5
+        assert isinstance(entry.utilization, float)
+
+    def test_quota_status_coerces_string_utilization(self):
+        from autoskillit.execution.quota import QuotaStatus
+
+        status = QuotaStatus(utilization="42.0", resets_at=None)
+        assert status.utilization == 42.0
+        assert isinstance(status.utilization, float)
+
 
 class TestWriteCache:
     def test_writes_readable_cache(self, tmp_path):
--- a/tests/server/test_tools_kitchen.py
+++ b/tests/server/test_tools_kitchen.py
@@ -648,6 +648,7 @@ async def test_prime_quota_cache_catches_typeerror(monkeypatch):
     mock_ctx.config.quota_guard = MagicMock()
 
     with patch("autoskillit.server._get_ctx", return_value=mock_ctx):
-        with patch("autoskillit.server.helpers.logger"):
+        with patch("autoskillit.server.helpers.logger") as mock_logger:
             # Must not raise — fails open
             await _prime_quota_cache()
+            mock_logger.warning.assert_called_once_with("quota_prime_failed", exc_info=True)
--- a/tests/server/test_tools_kitchen.py
+++ b/tests/server/test_tools_kitchen.py
@@ -631,3 +631,23 @@ async def test_open_kitchen_warns_on_missing_hook_scripts(tmp_path, monkeypatch)
     assert "Hook scripts not found" in result, (
         "open_kitchen() must include the exact _build_hook_diagnostic_warning phrase"
     )
+
+
+@pytest.mark.anyio
+async def test_prime_quota_cache_catches_typeerror(monkeypatch):
+    """_prime_quota_cache must catch TypeError and not propagate — 'fails open' contract."""
+    import autoskillit.execution as _exec_mod
+    from autoskillit.server.helpers import _prime_quota_cache
+
+    async def raise_type_error(*a, **kw):
+        raise TypeError("float() argument must be a string or a real number, not 'NoneType'")
+
+    monkeypatch.setattr(_exec_mod, "check_and_sleep_if_needed", raise_type_error)
+
+    mock_ctx = MagicMock()
+    mock_ctx.config.quota_guard = MagicMock()
+
+    with patch("autoskillit.server._get_ctx", return_value=mock_ctx):
+        with patch("autoskillit.server.helpers.logger"):
+            # Must not raise — fails open
+            await _prime_quota_cache()
--- a/tests/execution/test_quota_http.py
+++ b/tests/execution/test_quota_http.py
@@ -195,6 +195,39 @@ async def test_multi_window_worst_case_governs(mock_http_server, quota_config):
     assert result["window_name"] == "one_hour"
 
 
+async def test_null_utilization_window_skipped(mock_http_server, quota_config):
+    """Windows with utilization: null must be skipped, not crash."""
+    mock_http_server.register(
+        "GET",
+        QUOTA_ENDPOINT,
+        PyResponseSpec(
+            body={
+                "five_hour": {"utilization": None, "resets_at": "2026-04-10T12:00:00Z"},
+                "daily": {"utilization": 50.0, "resets_at": "2026-04-10T18:00:00Z"},
+            }
+        ),
+    )
+    result = await check_and_sleep_if_needed(quota_config, base_url=mock_http_server.url)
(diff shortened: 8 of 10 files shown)


src/autoskillit/execution/quota.py
```diff
@@ -29,6 +29,11 @@ class QuotaStatus:
     resets_at: datetime | None  # UTC-aware; None when utilization is 0
     window_name: str = "unknown"  # which window this came from (diagnostic)
 
+    def __post_init__(self) -> None:
+        if self.utilization is None:
```
Trecek, line 33: [warning] defense: The explicit None-guard is redundant — `float(None)` on L35 already raises TypeError. More importantly, the field is annotated `float`, so None should never reach construction. Consider either (a) removing the None-check and relying on the TypeError from `float()`, or (b) re-typing the field as `float | None` to make the runtime trap explicit at the type boundary.

Trecek (commented):

Self-review (reviewer is PR author); posted as COMMENT event per GitHub constraint. Found 8 actionable warnings (clear fixes), 3 decision findings, and 2 info findings. See inline comments.

src/autoskillit/execution/quota.py
```diff
@@ -37,6 +42,11 @@ class QuotaWindowEntry:
     utilization: float
     resets_at: datetime | None
 
+    def __post_init__(self) -> None:
```
Trecek, line 45: [warning] defense: Same pattern as QuotaStatus.__post_init__: the None-guard duplicates what `float(None)` would raise on L48, and the field is annotated `float`. If None is actually a legitimate input from callers, the annotation should be `float | None`; otherwise the guard is a hidden runtime trap for a case the type system already forbids.

src/autoskillit/execution/quota.py
```diff
@@ -109,7 +119,7 @@ def _read_cache(cache_path: str, max_age: int) -> QuotaStatus | None:
             resets_at=_parse_resets_at(b.get("resets_at")),
             window_name=str(b.get("window_name", "unknown")),
         )
-    except (FileNotFoundError, KeyError, ValueError, json.JSONDecodeError):
+    except (FileNotFoundError, KeyError, ValueError, TypeError, json.JSONDecodeError):
```
Trecek, line 122: [info] defense: The except tuple catches `TypeError` but not `PermissionError` / other `OSError` subclasses. If the cache file exists but is unreadable, the exception propagates instead of returning None, breaking the fail-open contract for that edge case. Consider adding `OSError` (which covers `FileNotFoundError` and `PermissionError` under a single base).

src/autoskillit/execution/quota.py
```diff
@@ -161,8 +171,11 @@ async def _fetch_quota(
     windows: dict[str, QuotaWindowEntry] = {}
     for name, w in data.items():
         if isinstance(w, dict) and "utilization" in w:
+            raw_util = w["utilization"]
+            if raw_util is None:
```
Trecek, line 175: [info] bugs: Silent `continue` here is asymmetric with `QuotaWindowEntry.__post_init__`, which raises TypeError on None utilization. A caller that constructs `QuotaWindowEntry` directly (not via `_fetch_quota`) will see a hard failure instead of a silent skip. Prefer a single policy — either drop the pre-filter and let __post_init__ decide, or document the asymmetry.

src/autoskillit/execution/quota.py
```diff
@@ -301,15 +314,12 @@ async def check_and_sleep_if_needed(
...
-    except (
-        TimeoutError,
-        OSError,
-        KeyError,
-        ValueError,
-        json.JSONDecodeError,
-        httpx.HTTPError,
-    ) as exc:
-        _log.warning("quota check failed — continuing without sleep", error=str(exc))
+    except Exception as exc:
```
Trecek, line 317: [warning] bugs: Broadening from a narrow tuple to `except Exception` will silently swallow coding bugs (AttributeError, NameError, ImportError) as mere warnings. The fail-open contract is valid, but mixing transient network errors with programming bugs at the same log level (`warning`) makes regressions invisible. Consider re-raising on programming-error types (AttributeError, NameError, TypeError outside the dataclass construction path), or logging these at ERROR severity so they surface in dashboards.

src/autoskillit/server/helpers.py
```diff
@@ -488,7 +488,7 @@ async def _prime_quota_cache() -> None:
 
     try:
         await check_and_sleep_if_needed(_get_ctx().config.quota_guard)
-    except (OSError, ValueError, RuntimeError):
+    except Exception:
```
Trecek, line 491: [warning] defense: This is a second `except Exception` layered around `check_and_sleep_if_needed`, which already has its own `except Exception` on L317 and never raises on quota errors. The only exceptions that can reach here are from `_get_ctx()` or `config.quota_guard` attribute access — i.e. configuration/bootstrap bugs, not quota domain failures. Swallowing these silently under a `quota_prime_failed` label hides misconfiguration. Narrow to the exceptions those calls can actually raise, or at minimum log the exception type in the message so maintainers can tell quota-domain from structural failures.

src/autoskillit/hooks/token_summary_appender.py
```diff
@@ -231,7 +231,8 @@ def _load_sessions(
         entry["output_tokens"] += data.get("output_tokens", 0)
         entry["cache_creation_input_tokens"] += data.get("cache_creation_input_tokens", 0)
         entry["cache_read_input_tokens"] += data.get("cache_read_input_tokens", 0)
-        entry["elapsed_seconds"] += float(data.get("timing_seconds", 0.0))
+        _raw_timing = data.get("timing_seconds")
+        entry["elapsed_seconds"] += float(_raw_timing) if _raw_timing is not None else 0.0
```
Trecek, line 235: [warning] bugs: The None-guard protects against `null` but `float()` still raises `ValueError` on non-numeric strings. If upstream ever emits a sentinel like `"N/A"` or `""`, this line will propagate an unhandled ValueError. Also: this idiom is duplicated in `pipeline/timings.py:144` and `pipeline/tokens.py:220` with subtle differences (timings.py wraps in `max(0.0, ...)`). Extract a shared `_coerce_seconds(val) -> float` helper so the policy lives in one place.

src/autoskillit/pipeline/timings.py
```diff
@@ -140,7 +140,8 @@ def load_from_log_dir(
             if key not in self._entries:
                 self._entries[key] = TimingEntry(step_name=step_name)
             e = self._entries[key]
-            e.total_seconds += max(0.0, float(data.get("total_seconds", 0.0)))
+            _raw_total = data.get("total_seconds")
+            e.total_seconds += max(0.0, float(_raw_total) if _raw_total is not None else 0.0)
```
Trecek, line 144: [warning] bugs + cohesion: Two issues at one line. (1) `float(_raw_total)` still raises `ValueError` on non-numeric strings — the None-guard only covers nulls. (2) This site uses `max(0.0, ...)` to clamp negatives, but the identical idiom in `pipeline/tokens.py:220` and `hooks/token_summary_appender.py:235` does NOT clamp. Either negative values are invalid for seconds everywhere (and the clamp should be added to the other sites) or they are valid (and the clamp here is gratuitous). Extract a single helper and converge the three sites.

src/autoskillit/pipeline/tokens.py
```diff
@@ -216,7 +216,8 @@ def load_from_log_dir(
             e.cache_read_input_tokens += data.get("cache_read_input_tokens", 0)
             # timing_seconds is the on-disk key name written by session_log.py;
             # elapsed_seconds is the in-memory field name on TokenEntry.
-            e.elapsed_seconds += float(data.get("timing_seconds", 0.0))
+            _raw_timing = data.get("timing_seconds")
+            e.elapsed_seconds += float(_raw_timing) if _raw_timing is not None else 0.0
```
Trecek, line 220: [warning] bugs: Same as `token_summary_appender.py:235` — the None-guard does not protect against non-numeric string values, which will raise `ValueError`. Also asymmetric with `pipeline/timings.py:144` which clamps via `max(0.0, ...)`. Should be converged through a shared coercion helper.

tests/server/test_tools_kitchen.py
```diff
@@ -631,3 +631,23 @@ async def test_open_kitchen_warns_on_missing_hook_scripts(tmp_path, monkeypatch)
     assert "Hook scripts not found" in result, (
         "open_kitchen() must include the exact _build_hook_diagnostic_warning phrase"
     )
+
+
+@pytest.mark.anyio
+async def test_prime_quota_cache_catches_typeerror(monkeypatch):
```
Trecek, line 637: [warning] tests: This test asserts nothing. It only verifies that `_prime_quota_cache()` does not raise — but a silently-swallowed TypeError is indistinguishable from a correct fail-open log call. To verify the 'fails open' contract, capture the patched logger and assert that `logger.warning` was called with the expected key. As written, a regression that drops the warning entirely would still pass this test.

tests/server/test_tools_kitchen.py
```diff
@@ -631,3 +631,23 @@ async def test_open_kitchen_warns_on_missing_hook_scripts(tmp_path, monkeypatch)
...
+    async def raise_type_error(*a, **kw):
+        raise TypeError("float() argument must be a string or a real number, not 'NoneType'")
+
+    monkeypatch.setattr(_exec_mod, "check_and_sleep_if_needed", raise_type_error)
+
+    mock_ctx = MagicMock()
+    mock_ctx.config.quota_guard = MagicMock()
+
+    with patch("autoskillit.server._get_ctx", return_value=mock_ctx):
+        with patch("autoskillit.server.helpers.logger"):
```
Trecek: [warning] tests: `patch('autoskillit.server.helpers.logger')` suppresses logger output but the mock is never bound (`with patch(...):` without `as mock_logger`) and no assertion is made on it. Bind the mock and assert `mock_logger.warning.assert_called_once_with('quota_prime_failed', exc_info=True)` (or equivalent) to actually verify the log contract the test claims to cover.

tests/execution/test_quota.py
```diff
@@ -99,6 +99,62 @@ def test_corrupted_cache_returns_none(self, tmp_path):
...
+
+        cache_path = tmp_path / "usage_cache.json"
+        payload = {
+            "fetched_at": datetime.now(UTC).isoformat(),
+            "binding": {"utilization": None, "resets_at": None, "window_name": "five_hour"},
+        }
+        cache_path.write_text(json.dumps(payload))
+        assert _read_cache(str(cache_path), max_age=300) is None
+
+    def test_read_cache_typeerror_returns_none(self, tmp_path):
```
Trecek, line 114: [info] tests (requires_decision): `test_read_cache_typeerror_returns_none` exercises nearly the same path as `test_read_cache_null_utilization_returns_none` above — both write a cache payload with `utilization: null` and assert `_read_cache` returns None. The only structural difference is the presence/absence of `window_name` in the payload. Consider parametrizing over the payload shapes, or explicitly documenting which distinct code path each exercises (the current docstrings do not make the distinction clear).

tests/execution/test_quota.py
```diff
@@ -99,6 +99,62 @@ def test_corrupted_cache_returns_none(self, tmp_path):
...
+        """_read_cache must catch TypeError (e.g., float(None)) and return None."""
+        from autoskillit.execution.quota import _read_cache
+
+        cache_path = tmp_path / "usage_cache.json"
+        payload = {
+            "fetched_at": datetime.now(UTC).isoformat(),
+            "binding": {"utilization": None, "resets_at": None},
+        }
+        cache_path.write_text(json.dumps(payload))
+        # With __post_init__, constructing QuotaStatus with None raises TypeError.
```
Trecek, line 124: [info] slop: The inline comments on L124-L125 restate what the docstring on L115 already says. The docstring covers the intent (`_read_cache must catch TypeError and return None`); the two comment lines add no new information. Remove them.

src/autoskillit/execution/quota.py
Trecek: Investigated — this is intentional. The explicit None-guard is a deliberate boundary assertion added in commit b030eeaa ("Rectify: Quota Dataclass Type Boundary Enforcement"), which introduced `__post_init__` type validation specifically because the Anthropic API was returning `utilization: null` at runtime. The named `TypeError` with a contract-specific message ("QuotaStatus.utilization must not be None") is more diagnostic than the generic `float(None)` error. The `float` annotation is aspirational — this PR is precisely the fix for the gap between annotation and runtime reality, so retyping to `float | None` would undo the contract the guard is meant to enforce.

src/autoskillit/execution/quota.py
Trecek: Investigated — this is intentional. Same design as `QuotaStatus` (see reply on line 33): `_fetch_quota` already skips None-utilization windows before construction, and `__post_init__` is the defense-in-depth second barrier added in commit b030eeaa. The named error ("QuotaWindowEntry.utilization must not be None") documents the boundary invariant explicitly.

src/autoskillit/execution/quota.py
Trecek: Agreed — fixed in daa93ba4 (with the initial attempt in 18ae0514, adjusted for ARCH-003 AST compliance). The broad `except Exception` now routes by `isinstance`: operational types (TimeoutError, OSError, KeyError, ValueError, TypeError, json.JSONDecodeError, httpx.HTTPError) continue to log at WARNING, while unexpected types (AttributeError, NameError, ImportError, ...) log at ERROR. Both branches preserve the fail-open contract and add `error_type` to the log record so quota-domain failures are distinguishable from programming regressions in dashboards.

src/autoskillit/server/helpers.py
Trecek: Valid observation — flagged for design decision. You are correct that the outer `except Exception` catches bootstrap/structural bugs (`RuntimeError` from `_get_ctx()`, `AttributeError` from `config.quota_guard`) under the same `quota_prime_failed` label as quota-domain failures; `exc_info=True` does preserve the exception type in log output but the label is uniform. The broadening was intentional (honor the fail-open subsystem-boundary contract so `open_kitchen` cannot abort on quota issues), so narrowing vs. adding a discriminating severity is a policy tradeoff worth human review — not auto-fixed here.

src/autoskillit/hooks/token_summary_appender.py
Trecek: Investigated — this is intentional. The `timing_seconds` value in `token_usage.json` is written exclusively by `session_log.py:216` as a typed `float | None` Python value, so it serializes to a JSON number (or is replaced with 0.0) — never a string sentinel like "N/A" or "". The None-guard added in b030eeaa correctly handles the only real path (JSON `null` from schema drift), and `float()` cannot raise `ValueError` from this closed write path. A shared helper across the three sites would conflate semantically distinct on-disk contracts; see the reply on `timings.py:144` for the clamp asymmetry rationale.

src/autoskillit/pipeline/timings.py
Trecek: Investigated — this is intentional. The `max(0.0, ...)` asymmetry mirrors the distinct write-path contracts in `session_log.py`: `step_timing.json.total_seconds` is pre-clamped at the source (L227 writes `max(0.0, timing_seconds)`), while `token_usage.json.timing_seconds` is not (L216 writes the raw float). So the clamp here is doubly defensive (harmless), and its absence in `tokens.py` / `token_summary_appender.py` is correct for that source. The `ValueError` concern also does not apply — both files are written by typed internal Python code, never external/user input, so non-numeric strings cannot flow in.

src/autoskillit/pipeline/tokens.py
Trecek: Investigated — this is intentional. Same rationale as the `timings.py:144` reply: the value is a typed Python float at write time in `session_log.py:216`, never a string, so `float()` has no `ValueError` risk. Omitting the `max(0.0, ...)` clamp matches the source contract (`token_usage.json` is not pre-clamped, unlike `step_timing.json`). A shared helper would conflate two distinct on-disk contracts.

tests/server/test_tools_kitchen.py
Trecek: Agreed — fixed in ff34432b. Bound the logger mock as `mock_logger` and added `mock_logger.warning.assert_called_once_with("quota_prime_failed", exc_info=True)` — matching the exact fail-open contract at `helpers.py:492`. A regression that drops the warning will now fail the test.

tests/server/test_tools_kitchen.py
Trecek: Agreed — fixed in ff34432b. Changed `with patch(...)` to `with patch(...) as mock_logger` and added `mock_logger.warning.assert_called_once_with("quota_prime_failed", exc_info=True)`. The patch target already matched the SUT's module-level `logger` at `helpers.py:38`.

PR #731, opened 2026-04-11
Rectify: IDE Env Leak Across All Claude Subprocess Launch Sites

Trecek:
## Summary

The stray `LSP(workspaceSymbol)` call observed in `autoskillit order` is one symptom of a **systemic pattern weakness**: every claude-launching subprocess in autoskillit inherits the parent process's full environment without sanitization. When autoskillit runs under a shell that has `CLAUDE_CODE_SSE_PORT` set (VS Code integrated terminal, any descendant process), that var is propagated into the cook subprocess, which then attaches to `~/.claude/ide/$PORT.lock` over a websocket and merges IDE-provided LSP tools into the model's manifest.

This PR introduces a canonical L0 env builder (`build_claude_env()`) that applies a principled IDE-denylist and injects `CLAUDE_CODE_AUTO_CONNECT_IDE=0` to suppress the lock-file-scan fallback. All four claude-launching sites are wired through it, and an AST rule enforces compliance structurally.

### Changes

**New components (★):**
- `src/autoskillit/core/_claude_env.py` — IDE-scrubbing env builder with denylist, prefix denylist, and always-extras
- `tests/core/test_claude_env.py` — 14 unit tests for the builder
- `tests/cli/test_cook_env_scrub.py` — launch-site env-capture tests for cook/order
- `tests/cli/test_cook_ide_isolation.py` — end-to-end IDE-lock simulation smoke canary
- `tests/execution/test_headless_env_scrub.py` — headless runner env-capture test

**Modified components (●):**
- `execution/commands.py` — `ClaudeInteractiveCmd`/`ClaudeHeadlessCmd.env` now hold fully resolved scrubbed env; `build_full_headless_cmd` returns `ClaudeHeadlessCmd` (was `list[str]`); `["env", ...]` argv prefix eliminated
- `cli/app.py`, `cli/_cook.py` — drop inline `{**os.environ}`, use `spec.env` directly
- `execution/headless.py` — pass `env=spec.env` to runner
- `_llm_triage.py` — pass `env=build_claude_env()` to `run_managed_async`
- `execution/recording.py` — read `SCENARIO_STEP_NAME` from `env` kwarg (not argv prefix); delete obsolete `_extract_env_and_args`
- `execution/process.py`, `core/_type_subprocess.py` — widen `env` param type to `Mapping[str, str] | None`
- `tests/cli/test_subprocess_env_contracts.py` — new `test_no_raw_claude_env` AST rule with allowlist
- Test migrations: `test_commands.py`, `test_recording.py`, `test_headless_env_injection.py`, `test_tools_execution.py`, `test_llm_triage.py`

### Immunity claim

- The specific bug is impossible to reintroduce without (a) bypassing the builder (blocked by AST rule), (b) removing a var from the denylist, or (c) removing the `CLAUDE_CODE_AUTO_CONNECT_IDE=0` injection — all deliberate, reviewable single-file changes
- Two-layer immunity: env scrub closes the direct-signal path; auto-connect disable closes the lock-file-scan fallback
- Test discipline is now symmetric with argv discipline: both layers have a builder, a dataclass field, and an AST rule

## Implementation Plan

Plan file: `.autoskillit/temp/rectify/rectify_ide_env_leak_2026-04-11_120000.md`

### DIFF
--- a/src/autoskillit/cli/app.py
+++ b/src/autoskillit/cli/app.py
@@ -428,7 +428,9 @@ def _launch_cook_session(
         print("ERROR: 'claude' not found. Install: https://docs.anthropic.com/en/docs/claude-code")
         sys.exit(1)
     spec = build_interactive_cmd(
-        initial_prompt=initial_message, resume_session_id=resume_session_id
+        initial_prompt=initial_message,
+        resume_session_id=resume_session_id,
+        env_extras=extra_env,
     )
     cmd = spec.cmd + [
         ClaudeFlags.PLUGIN_DIR,
@@ -438,11 +440,8 @@ def _launch_cook_session(
         ClaudeFlags.APPEND_SYSTEM_PROMPT,
         system_prompt,
     ]
-    env = {**os.environ, **spec.env}
-    if extra_env:
-        env.update(extra_env)
     with terminal_guard():
-        result = subprocess.run(cmd, env=env)
+        result = subprocess.run(cmd, env=spec.env)
     if result.returncode != 0:
         sys.exit(result.returncode)
 
--- a/src/autoskillit/core/_claude_env.py
+++ b/src/autoskillit/core/_claude_env.py
@@ -1,0 +1,89 @@
+"""Canonical env builder for claude-launching subprocesses.
+
+Every subprocess that invokes the ``claude`` CLI must route its environment
+through :func:`build_claude_env` so that host-process IDE state (VS Code,
+Cursor, Zed, JetBrains, Neovim bridges) cannot leak across the trust
+boundary and silently widen the child's tool surface.
+
+Two layers of immunity are applied:
+
+1. **Denylist scrub** — IDE discovery variables such as
+   ``CLAUDE_CODE_SSE_PORT`` and the ``CLAUDE_CODE_IDE_*`` family are
+   stripped from ``base``. Removing the port env closes the direct-signal
+   attach path.
+2. **Implicit auto-connect disable** — ``CLAUDE_CODE_AUTO_CONNECT_IDE=0``
+   is always injected. This suppresses the ``~/.claude/ide/*.lock`` scan
+   fallback that the Claude CLI follows at startup even when no IDE env
+   vars are set; without it, third-party IDE bridges (e.g.
+   ``claudecode.nvim``) can still attach via the lock-file mechanism.
+"""
+
+from __future__ import annotations
+
+import os
+from collections.abc import Mapping
+from types import MappingProxyType
+
+# Exact-match IDE discovery variable names stripped from the child env.
+IDE_ENV_DENYLIST: frozenset[str] = frozenset(
+    {
+        "CLAUDE_CODE_SSE_PORT",
+        "ENABLE_IDE_INTEGRATION",
+        "CLAUDE_CODE_WEBSOCKET_AUTH_FILE_DESCRIPTOR",
+        "VSCODE_GIT_ASKPASS_MAIN",
+        "CURSOR_TRACE_ID",
+        "ZED_TERM",
+    }
+)
+
+# Prefix-match IDE variable patterns stripped from the child env.
+IDE_ENV_PREFIX_DENYLIST: tuple[str, ...] = (
+    "CLAUDE_CODE_IDE_",
+    "CLAUDE_CODE_SSE",
+)
+
+# Variables injected into every built env regardless of caller. These close
+# discovery paths that cannot be closed by scrubbing alone — notably
+# CLAUDE_CODE_AUTO_CONNECT_IDE=0, which suppresses the ~/.claude/ide/*.lock
+# scan path that fires even when SSE_PORT is absent.
+IDE_ENV_ALWAYS_EXTRAS: Mapping[str, str] = MappingProxyType(
+    {
+        "CLAUDE_CODE_AUTO_CONNECT_IDE": "0",
+    }
+)
+
+
+def build_claude_env(
+    base: Mapping[str, str] | None = None,
+    *,
+    extras: Mapping[str, str] | None = None,
+) -> Mapping[str, str]:
+    """Return a scrubbed, sealed env dict suitable for a claude subprocess.
+
+    Parameters
+    ----------
+    base
+        Starting environment. Defaults to ``os.environ``.
+    extras
+        Caller-supplied overrides merged last. Used to carry
+        ``AUTOSKILLIT_HEADLESS=1``, ``SCENARIO_STEP_NAME=...`` and similar
+        into the child.
+
+    Returns
+    -------
+    Mapping[str, str]
+        A ``MappingProxyType`` over the resolved env. Both
+        ``subprocess.run(env=...)`` and ``anyio.open_process(env=...)``
+        accept any ``Mapping``, so this is a drop-in for the underlying
+        runners. The read-only view prevents post-build mutation.
+    """
+    src = os.environ if base is None else base
+    out: dict[str, str] = {
+        k: v
+        for k, v in src.items()
+        if k not in IDE_ENV_DENYLIST and not any(k.startswith(p) for p in IDE_ENV_PREFIX_DENYLIST)
+    }
+    out.update(IDE_ENV_ALWAYS_EXTRAS)
+    if extras:
+        out.update(extras)
+    return MappingProxyType(out)
--- a/src/autoskillit/execution/commands.py
+++ b/src/autoskillit/execution/commands.py
@@ -3,23 +3,43 @@
 from __future__ import annotations
 
 import os
-from collections.abc import Sequence
+from collections.abc import Mapping, Sequence
 from dataclasses import dataclass, field
 from pathlib import Path
 
-from autoskillit.core import ClaudeFlags, ValidatedAddDir, temp_dir_display_str
+from autoskillit.core import (
+    ClaudeFlags,
+    ValidatedAddDir,
+    build_claude_env,
+    temp_dir_display_str,
+)
 
 
 @dataclass(frozen=True)
 class ClaudeInteractiveCmd:
+    """Resolved argv + env for a claude interactive subprocess.
+
+    ``env`` is the fully resolved environment returned by
+    :func:`build_claude_env` — pass directly to ``subprocess.run(env=...)``.
+    Callers must NOT merge in ``os.environ`` again; the sanitization layer
+    has already applied the denylist and the auto-connect suppressor.
+    """
+
     cmd: list[str]
-    env: dict[str, str]
+    env: Mapping[str, str] = field(default_factory=dict)
 
 
 @dataclass(frozen=True)
 class ClaudeHeadlessCmd:
+    """Resolved argv + env for a claude headless subprocess.
+
+    ``env`` is the fully resolved environment returned by
+    :func:`build_claude_env`, including any headless-only extras such as
+    ``AUTOSKILLIT_HEADLESS=1``. Pass directly to the subprocess runner.
+    """
+
     cmd: list[str]
-    env: dict[str, str] = field(default_factory=dict)  # always {}
+    env: Mapping[str, str] = field(default_factory=dict)
 
 
 def build_interactive_cmd(
@@ -29,6 +49,7 @@ def build_interactive_cmd(
     plugin_dir: Path | None = None,
     add_dirs: Sequence[Path | str | ValidatedAddDir] = (),
     resume_session_id: str | None = None,
+    env_extras: Mapping[str, str] | None = None,
 ) -> ClaudeInteractiveCmd:
     """Build a Claude interactive session command.
 
@@ -46,6 +67,8 @@ def build_interactive_cmd(
         Each entry is appended as ``--add-dir <path>``.
     resume_session_id
         When provided, appended as ``--resume <id>`` before any positional prompt.
+    env_extras
+        Optional caller overrides merged into the resolved env after IDE scrubbing.
     """
     cmd = ["claude", ClaudeFlags.DANGEROUSLY_SKIP_PERMISSIONS]
     if resume_session_id is not None:
@@ -58,15 +81,20 @@ def build_interactive_cmd(
         cmd += [ClaudeFlags.ADD_DIR, str(d)]
     if initial_prompt is not None:
         cmd.append(initial_prompt)
-    return ClaudeInteractiveCmd(cmd=cmd, env={})
+    return ClaudeInteractiveCmd(cmd=cmd, env=build_claude_env(extras=env_extras))
 
 
-def build_headless_cmd(prompt: str, *, model: str | None = None) -> ClaudeHeadlessCmd:
+def build_headless_cmd(
+    prompt: str,
+    *,
+    model: str | None = None,
+    env_extras: Mapping[str, str] | None = None,
+) -> ClaudeHeadlessCmd:
     """Build a Claude headless session command for skill execution."""
     cmd = ["claude", ClaudeFlags.PRINT, prompt, ClaudeFlags.DANGEROUSLY_SKIP_PERMISSIONS]
     if model:
         cmd += [ClaudeFlags.MODEL, model]
-    return ClaudeHeadlessCmd(cmd=cmd, env={})
+    return ClaudeHeadlessCmd(cmd=cmd, env=build_claude_env(extras=env_extras))
 
 
 def _ensure_skill_prefix(skill_command: str) -> str:
@@ -121,12 +149,13 @@ def build_full_headless_cmd(
     exit_after_stop_delay_ms: int = 0,
     scenario_step_name: str = "",
     temp_dir_relpath: str | None = None,
-) -> list[str]:
-    """Build the complete headless command list ready for subprocess invocation.
+) -> ClaudeHeadlessCmd:
+    """Build the complete headless command spec ready for subprocess invocation.
 
     Applies prompt transformations (skill prefix, completion directive, cwd anchor),
     then constructs the full CLI command including plugin-dir, output-format,
-    add-dir entries, and the ``env AUTOSKILLIT_HEADLESS=1`` prefix.
+    and add-dir entries. The environment carries ``AUTOSKILLIT_HEADLESS=1`` and any
+    scenario / exit-delay extras on ``.env`` — it is NOT serialized as an argv prefix.
 
     Parameters
     ----------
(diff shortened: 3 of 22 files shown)


tests/core/test_claude_env.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def test_build_claude_env_returns_mappingproxy() -> None:
+    result = build_claude_env(base={"HOME": "/tmp"})
+    assert isinstance(result, MappingProxyType)
+    with pytest.raises(TypeError):
+        result["X"] = "Y"  # type: ignore[index]
+
+
+def test_ide_env_denylist_contains_expected_names() -> None:
```
Trecek, line 105: [warning] tests: test_ide_env_denylist_contains_expected_names, test_ide_env_prefix_denylist_covers_ide_and_sse, and test_ide_env_always_extras_includes_auto_connect_off are tautological constant-inspection tests: they import the exact constants being tested and assert their own contents. If a key is removed from IDE_ENV_DENYLIST the parametrized scrubbing tests will catch it; these three tests add no independent protection.

tests/core/test_claude_env.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@pytest.mark.parametrize(
+    "key",
+    [
+        "CLAUDE_CODE_SSE_PORT",
+        "ENABLE_IDE_INTEGRATION",
+        "CLAUDE_CODE_WEBSOCKET_AUTH_FILE_DESCRIPTOR",
+        "VSCODE_GIT_ASKPASS_MAIN",
+        "CURSOR_TRACE_ID",
```
Trecek, line 48: [info] tests: test_build_claude_env_strips_expanded_denylist overlaps with test_build_claude_env_strips_sse_port and prefix-denylist parametrize. CLAUDE_CODE_SSE_PORT appears in both, creating a redundant case.

tests/execution/test_commands.py
```diff
@@ -35,9 +37,18 @@ def test_starts_with_claude(self) -> None:
...
-    def test_env_is_empty(self) -> None:
+    def test_env_is_populated_and_scrubbed(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        monkeypatch.setenv("CLAUDE_CODE_SSE_PORT", "23270")
+        monkeypatch.setenv("HOME", "/tmp/home")
+        result = build_interactive_cmd()
+        assert "CLAUDE_CODE_SSE_PORT" not in result.env
+        assert result.env.get("HOME") == "/tmp/home"
+        assert result.env["CLAUDE_CODE_AUTO_CONNECT_IDE"] == "0"
+
+    def test_env_strips_sse_port(self, monkeypatch: pytest.MonkeyPatch) -> None:
```
Trecek: [warning] tests: test_env_strips_sse_port (for both build_interactive_cmd and build_headless_cmd) is a strict subset of the immediately preceding test_env_is_populated_and_scrubbed: both set the same monkeypatch env and assert the same SSE port exclusion. The strip-only tests add no new assertions.

tests/cli/test_cook_ide_isolation.py
```diff
@@ -0,0 +1,90 @@
...
+    monkeypatch.setenv("VSCODE_GIT_ASKPASS_MAIN", "/fake/vscode")
+    monkeypatch.setenv("CLAUDE_CODE_IDE_HOST_OVERRIDE", "localhost")
+
+    # Wrap Path.open so any parent-process touch of the IDE lock path is
+    # detectable. We count reads against the exact lock_file path object.
+    lock_touches = {"count": 0}
+    real_path_open = Path.open
+    resolved_lock = lock_file.resolve()
+
+    def tracking_open(self: Path, *args: object, **kwargs: object) -> object:
```
Trecek: [warning] tests: test_cook_session_ignores_ide_lock_file assertion (3) that lock_touches['count'] == 0 is vacuously true: the production code path never calls Path.open on a lock file — there is no lock-file scan in the Python source, only in the C/Go claude binary which is mocked away via subprocess.run patch. The tracking_open wrapper monitors the parent Python process only, so assertion (3) can never fail.

Trecek (commented):

tests/cli/test_cook_ide_isolation.py
```diff
@@ -0,0 +1,90 @@
...
+VS Code (or any IDE) has ``CLAUDE_CODE_SSE_PORT`` set and an active
+``~/.claude/ide/$PORT.lock`` file. When autoskillit launches the cook session,
+the child must NOT attach to the IDE channel via either discovery path:
+
+1. **Env scrub** — ``CLAUDE_CODE_SSE_PORT`` and the expanded IDE denylist are
+   stripped from the child env.
+2. **Auto-connect disable** — ``CLAUDE_CODE_AUTO_CONNECT_IDE=0`` is injected,
+   which suppresses the ``~/.claude/ide/*.lock`` scan fallback that fires even
+   when no IDE env vars are set.
+3. **Lock file is not opened** by the parent process between the builder call
```
Trecek: [info] tests: test_cook_ide_isolation.py and test_cook_env_scrub.py cover almost identical launch-site scrub assertions for _launch_cook_session. The isolation canary file could be folded into test_cook_env_scrub.py.

tests/execution/test_headless_env_injection.py
```diff
@@ -46,8 +46,11 @@ async def test_headless_command_includes_headless_env_var(tmp_path: Path) -> Non
     await run_headless_core("/investigate foo", str(tmp_path), ctx)
 
     assert mock_runner.call_args_list, "runner was never called"
-    cmd, *_ = mock_runner.call_args_list[0]
-    assert "AUTOSKILLIT_HEADLESS=1" in cmd, (
-        "run_headless_core must inject AUTOSKILLIT_HEADLESS=1 into the subprocess command "
+    cmd, _cwd, _timeout, kwargs = mock_runner.call_args_list[0]
```
Trecek, line 49: [warning] tests: Unpacking mock_runner.call_args_list[0] as `cmd, _cwd, _timeout, kwargs` assumes positional arg layout of the runner. If MockSubprocessRunner changes to keyword-only, this destructuring silently breaks. Prefer call_args.args / call_args.kwargs accessors.

tests/execution/test_recording.py
```diff
@@ -74,18 +73,13 @@ async def test_session_call_routes_to_record_step(tmp_path):
...
-        "do stuff",
-    ]
+    cmd = ["claude", "--model", "sonnet", "--print", "do stuff"]
+    env = {
+        "AUTOSKILLIT_HEADLESS": "1",
+        "SCENARIO_STEP_NAME": "investigate",
+    }
 
-    result = await runner(cmd, cwd=Path("/tmp"), timeout=300, pty_mode=True)
+    result = await runner(cmd, cwd=Path("/tmp"), timeout=300, env=env, pty_mode=True)
```
Trecek, line 82: [info] tests: Multiple test functions unpack mock call_args as positional tuple — fragile positional destructuring. Prefer mock.call_args.args and mock.call_args.kwargs.

src/autoskillit/execution/commands.py
```diff
@@ -3,23 +3,43 @@
...
+
+    ``env`` is the fully resolved environment returned by
+    :func:`build_claude_env` — pass directly to ``subprocess.run(env=...)``.
+    Callers must NOT merge in ``os.environ`` again; the sanitization layer
+    has already applied the denylist and the auto-connect suppressor.
+    """
+
     cmd: list[str]
-    env: dict[str, str]
+    env: Mapping[str, str] = field(default_factory=dict)
```
Trecek, line 29: [warning] defense: ClaudeInteractiveCmd.env field default is `field(default_factory=dict)`, which produces an empty plain dict — not a scrubbed MappingProxyType. Any code path that constructs ClaudeInteractiveCmd directly without calling build_claude_env() will silently inherit an unscrubbed empty env. Consider replacing the default with `field(default_factory=build_claude_env)` or removing the default to force callers to always supply a scrubbed env.

src/autoskillit/execution/commands.py
```diff
@@ -3,23 +3,43 @@
...
+    """Resolved argv + env for a claude headless subprocess.
+
+    ``env`` is the fully resolved environment returned by
+    :func:`build_claude_env`, including any headless-only extras such as
+    ``AUTOSKILLIT_HEADLESS=1``. Pass directly to the subprocess runner.
+    """
+
     cmd: list[str]
-    env: dict[str, str] = field(default_factory=dict)  # always {}
+    env: Mapping[str, str] = field(default_factory=dict)
```
Trecek, line 42: [warning] defense: ClaudeHeadlessCmd.env field default is also `field(default_factory=dict)`, sharing the same problem: direct construction without build_claude_env() produces an empty unscrubbed env. This contradicts the docstring's claim that env is the fully resolved environment.

src/autoskillit/core/_claude_env.py
```diff
@@ -0,0 +1,89 @@
...
+# scan path that fires even when SSE_PORT is absent.
+IDE_ENV_ALWAYS_EXTRAS: Mapping[str, str] = MappingProxyType(
+    {
+        "CLAUDE_CODE_AUTO_CONNECT_IDE": "0",
+    }
+)
+
+
+def build_claude_env(
+    base: Mapping[str, str] | None = None,
```
Trecek, line 57: [info] defense: build_claude_env() has no runtime type validation that base and extras values are actually strings. Given the trust-boundary purpose, a runtime isinstance check would provide defense-in-depth.

src/autoskillit/execution/recording.py
```diff
@@ -102,15 +86,13 @@ async def __call__(
         completion_drain_timeout: float = 5.0,
         linux_tracing_config: Any | None = None,
     ) -> SubprocessResult:
-        env_dict, clean_args = _extract_env_and_args(cmd)
-        step_name = env_dict.get(SCENARIO_STEP_NAME_ENV, "")
+        step_name = (env or {}).get(SCENARIO_STEP_NAME_ENV, "")
```
Trecek, line 89: [info] defense: The expression (env or {}).get(SCENARIO_STEP_NAME_ENV, '') silently treats None env as an empty dict. For RecordingSubprocessRunner this is the intended behavior, but env=None silently skips the recording path without any diagnostic.

src/autoskillit/cli/app.py
```diff
@@ -438,11 +440,8 @@ def _launch_cook_session(
         ClaudeFlags.APPEND_SYSTEM_PROMPT,
         system_prompt,
     ]
-    env = {**os.environ, **spec.env}
-    if extra_env:
-        env.update(extra_env)
     with terminal_guard():
-        result = subprocess.run(cmd, env=env)
+        result = subprocess.run(cmd, env=spec.env)
```
Trecek, line 444: [warning] defense: In _launch_cook_session, extra_env is passed via env_extras without a public type annotation visible in the diff for the extra_env parameter, making it an unvalidated boundary.

tests/cli/test_subprocess_env_contracts.py
```diff
@@ -196,3 +212,221 @@ def test_source_drift_all_subprocess_calls_have_env_kwarg() -> None:
...
+            ):
+                return False, "env=os.environ is not allowed"
+            # env=spec.env / env=foo.env
+            if env_val.attr == "env":
+                return True, ""
+        if isinstance(env_val, ast.Call):
+            fn = _call_func_name(env_val.func)
+            if fn == "build_claude_env":
+                return True, ""
+        if isinstance(env_val, ast.Name):
```
Trecek, line 365: [warning] bugs: _env_shape_ok accepts any ast.Name as valid without resolving the binding. A pattern like `my_env = os.environ; subprocess.run(cmd, env=my_env)` would pass the check because only the call-site Name node is inspected. This creates a false negative in the enforcement rule: it can be bypassed by assigning os.environ to a local variable. Fix: resolve the Name through _bindings and re-apply the shape check.

tests/cli/test_cook_env_scrub.py
```diff
@@ -0,0 +1,102 @@
+"""Launch-site env-scrub contract tests for _launch_cook_session and cook().
```
Trecek, line 1: [info] cohesion: test_cook_env_scrub.py and test_cook_ide_isolation.py both test _launch_cook_session IDE isolation with overlapping assertions. The naming does not consistently signal the layering hierarchy.

tests/cli/test_subprocess_env_contracts.py
```diff
@@ -196,3 +212,221 @@ def test_source_drift_all_subprocess_calls_have_env_kwarg() -> None:
...
+
+# Allowlist: administrative ``claude plugin ...`` subprocess calls that are
+# plugin marketplace / registration operations, NOT skill sessions. They do
+# not present a tool surface to the model, so IDE-channel attach is not a
+# concern. Each entry is (filename, enclosing_function_name).
+#
+# Adding a new entry requires a comment justifying why the call cannot use
+# build_claude_env — matching the convention in
+# ``tests/arch/test_ast_rules.py::test_no_raw_claude_list_construction``.
+_CLAUDE_ENV_RULE_ALLOWED: frozenset[tuple[str, str]] = frozenset(
```
Trecek, line 399: [info] cohesion: _CLAUDE_ENV_RULE_ALLOWED allowlist and _ClaudeLaunchWalker AST enforcement rule are in test_subprocess_env_contracts.py (cli/ test dir) but reference tests/arch/test_ast_rules.py as convention source, suggesting this enforcement rule may belong in arch/ test directory.

tests/execution/test_commands.py
```diff
@@ -143,9 +154,18 @@ def test_does_not_include_allow_dangerously_skip_permissions(self) -> None:
...
+    def test_env_is_populated_and_scrubbed(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        monkeypatch.setenv("CLAUDE_CODE_SSE_PORT", "23270")
+        monkeypatch.setenv("HOME", "/tmp/home")
         result = build_headless_cmd("some prompt")
-        assert result.env == {}
+        assert "CLAUDE_CODE_SSE_PORT" not in result.env
+        assert result.env.get("HOME") == "/tmp/home"
+        assert result.env["CLAUDE_CODE_AUTO_CONNECT_IDE"] == "0"
+
+    def test_env_strips_sse_port(self, monkeypatch: pytest.MonkeyPatch) -> None:
```
Trecek: [warning] slop: test_env_strips_sse_port in TestBuildHeadlessCmd is a strict subset of test_env_is_populated_and_scrubbed directly above it — redundant, zero additional coverage.

tests/execution/test_headless_env_injection.py
```diff
@@ -46,8 +46,11 @@ async def test_headless_command_includes_headless_env_var(tmp_path: Path) -> Non
...
+    env = kwargs.get("env")
+    assert env is not None
+    assert env["AUTOSKILLIT_HEADLESS"] == "1", (
+        "run_headless_core must inject AUTOSKILLIT_HEADLESS=1 via the env kwarg "
         "so PreToolUse hooks can identify headless sessions."
     )
+    assert cmd[0] != "env", "argv must no longer carry a leading ['env', ...] prefix"
```
Trecek, line 56: [info] slop: assert cmd[0] != 'env' is a weak negative assertion. The meaningful check would be assert cmd[0] == 'claude'.

tests/execution/test_recording.py
```diff
@@ -221,8 +202,11 @@ async def test_run_headless_core_injects_scenario_step_name(tmp_path):
 
     await run_headless_core("/investigate foo", str(tmp_path), ctx, step_name="investigate")
 
-    cmd = mock_runner.call_args_list[0][0]
-    assert "SCENARIO_STEP_NAME=investigate" in cmd
+    cmd, _cwd, _timeout, kwargs = mock_runner.call_args_list[0]
+    assert cmd[0] != "env"
```
Trecek, line 206: [info] slop: assert cmd[0] != 'env' is redundant — the env-prefix removal is enforced by construction. A positive assertion like assert cmd[0] == 'claude' would be more meaningful.

tests/execution/test_commands.py
Trecek: Agreed — fixed in 121473ed. test_env_strips_sse_port was a strict subset of test_env_is_populated_and_scrubbed (lines 40-46); every assertion was already covered by the superset test. Removed from both TestBuildInteractiveCmd and TestBuildHeadlessCmd.

tests/cli/test_cook_ide_isolation.py
Trecek: Agreed — fixed in d2fe8853. The lock-file scan executes in the claude binary (C/Go), not in the Python _launch_cook_session path. The Path.open wrapper was monkeypatching Python-level I/O that the production code never exercises during this call chain, making the assertion structurally guaranteed to pass. Removed the lock_touches tracking and assertion; env scrub + auto-connect assertions remain.

tests/cli/test_subprocess_env_contracts.py
Trecek: Agreed — fixed in 85a907d2. _env_shape_ok now calls self._resolve_name(env_val) for ast.Name nodes and recursively checks the bound value. A pattern like `my_env = os.environ; subprocess.run(cmd, env=my_env)` now correctly resolves through the alias and rejects the os.environ binding.

tests/execution/test_commands.py
Trecek: Agreed — fixed in 121473ed. Same redundancy as 3068605292 — the headless variant of test_env_strips_sse_port was also a strict subset of its sibling test_env_is_populated_and_scrubbed. Removed.

tests/core/test_claude_env.py
Trecek: Investigated — this is intentional. The three flagged tests use hardcoded string literals as expected values (e.g. `assert "CURSOR_TRACE_ID" in IDE_ENV_DENYLIST` at line 108), not references to the constants under test. If a key were removed from IDE_ENV_DENYLIST in _claude_env.py, the assertion would raise AssertionError — the expected set does not shrink with the constant. These are deliberate spot-check membership tests documenting security-critical contract keys.

tests/execution/test_headless_env_injection.py
Trecek: Investigated — this is intentional. MockSubprocessRunner is a bespoke test double (conftest.py lines 72-80), not unittest.mock.Mock. Its `__call__` manually appends `(cmd, cwd, timeout, kwargs)` tuples to call_args_list, documented in the docstring. There is no `.call_args.args`/`.call_args.kwargs` API on this class — the positional destructuring directly matches the documented, stable tuple contract.

src/autoskillit/execution/commands.py
Trecek: Investigated — this is intentional. All construction sites of ClaudeInteractiveCmd pass env= explicitly: commands.py:84 calls `ClaudeInteractiveCmd(cmd=cmd, env=build_claude_env(extras=env_extras))`. All callers (cli/app.py:430, cli/_cook.py:154) go through build_interactive_cmd(), never the constructor directly. The default_factory=dict is a dataclass field sentinel never exercised at runtime. The AST enforcement test test_no_raw_claude_env statically verifies all claude-launching sites route env through build_claude_env().

src/autoskillit/execution/commands.py
Trecek: Investigated — this is intentional. Same pattern as ClaudeInteractiveCmd: all construction sites of ClaudeHeadlessCmd pass env= explicitly (commands.py:97, commands.py:207 via spec.env). No caller constructs ClaudeHeadlessCmd directly — grep across the entire source tree finds zero direct constructor invocations outside commands.py. The AST enforcement test test_no_raw_claude_env guards against regression.

src/autoskillit/cli/app.py
Trecek: Investigated — this is intentional. The extra_env parameter IS annotated: app.py line 423 declares `extra_env: dict[str, str] | None = None` in the _launch_cook_session signature. Additionally, build_interactive_cmd in execution/commands.py line 52 annotates `env_extras: Mapping[str, str] | None = None`. Both boundaries are fully typed and visible in the diff.

PR #749, opened 2026-04-12
Rectify: SIGTERM Bypasses atexit — ScenarioRecorder.finalize() Never Runs

### LINKED ISSUE #745: Scenario recording: finalize() never runs under SIGTERM — scenario.json and MCP middleware buffer are lost on /exit
## Summary

After running `task test-smoke-record` end-to-end and cleanly `/exit`ing the Claude Code session, the scenario output directory (`tests/fixtures/scenarios/smoke-happy/`) contains **only** the PTY cassette for the one headless session step (`001_implement_task`). The top-level `scenario.json` manifest is missing, the `mcp_calls` buffer from `McpRecordingMiddleware` (wired in PR #727) is lost, and any `record_non_session_step()` entries are lost. The recorded directory is not replay-ready.

**Root cause in one sentence:** `ScenarioRecorder.finalize()` is the only writer of `scenario.json` and the only code path that flushes the in-memory `mcp_calls` / non-session-step buffers to disk, but it is registered *only* via `atexit.register(recorder.finalize)` and Claude Code terminates the MCP server subprocess with **SIGTERM** on `/exit`, which Python's default SIGTERM handler kills without running `atexit` hooks.

## Symptoms

Running `task test-smoke-record` with the smoke recipe completing all steps successfully at the MCP level (the recipe's own routing logic classified `implement_task` as `empty_output` — **unrelated to this bug**, see note at bottom), then `/exit`ing the interactive Claude Code session, produces this directory layout:

```
tests/fixtures/scenarios/smoke-happy/
└── sessions/
    └── 001_implement_task/
        ├── stdout.jsonl       46 KB   (healthy stream-json)
        ├── metadata.json     124 B    (exit_code=0, model=sonnet, duration_ms=32195, termination_reason=end_turn)
        ├── input.json        1.8 KB   (claude argv)
        └── session_log.jsonl   0 B    (empty)
```

What's missing:
1. **No top-level `scenario.json`** at the root of `smoke-happy/`. `ScenarioPlayer.scenario()` in `build_replay_runner()` expects one — this directory is not replay-ready.
2. **No `mcp_calls` record** from `McpRecordingMiddleware`. The middleware successfully registered and intercepted every MCP tool call (`open_kitchen`, `clone_repo`, `create_unique_branch`, `run_cmd`, `run_skill`, `register_clone_status`), but all of that is buffered in memory only.
3. **No `record_non_session_step()` entries.** Even if they had been captured (see secondary issue below), they too are memory-only until `finalize()`.

The `implement_task` PTY cassette is fully healthy: 46 KB of well-formed stream-json, exit code 0, clean `end_turn` termination, 32-second duration. This isolates the failure entirely to the finalize/shutdown path, not to the capture path.

### The finalize contract lives in the api-simulator Rust extension

`ScenarioRecorder.finalize()` (in `crates/api-simulator/src/claude/scenario.rs`, lines 758–782) is the **only** writer of `scenario.json`. It:
1. Computes `total_duration_ms` from wall-clock start time.
2. Assembles a `Scenario` struct from accumulated in-memory `steps`, `mcp_calls`, and `recipe_name`, setting `terminal_step` to the last step's name.
3. Calls `scenario.save(&self.output_dir)`, which writes `<output_dir>/scenario.json` (pretty-printed JSON).

Everything else in the recording path is either incremental per-step PTY capture (`record_step()`, which writes `sessions/NNN_<step>/*` files incrementally during capture — this is why we see the one cassette) or pure in-memory buffering:

- `ScenarioRecorder.record_step()` appends a `StepRecord` entry to `self.steps` and writes cassette files to disk incrementally. **Does not update `scenario.json`.**
- `ScenarioRecorder.record_non_session_step()` (lines 713–729) **writes nothing to disk at all.** It only appends a `StepRecord` (with `session_dir: None`, `result_summary: ...`) to `self.steps`. The data reaches disk only via `finalize()`.
- `McpRecordingMiddleware.__call__` (from `api_simulator.mcp.middleware`, lines 57–81) wraps each MCP `tools/call`, serializes the result, and calls `self._recorder.record_mcp_call(...)` — which is an in-memory append to `ScenarioRecorder.mcp_calls`. **No disk write.**

So `scenario.json` + the entire `mcp_calls` array + every non-session step record is only written when `finalize()` runs.

### The autoskillit shutdown path never runs finalize

Finalize is wired in exactly one place:

```python
# src/autoskillit/execution/recording.py:67
class RecordingSubprocessRunner:
    def __init__(self, recorder: ScenarioRecorder, inner: SubprocessRunner) -> None:
        self._inner = inner
        self.recorder = recorder
        atexit.register(recorder.finalize)
```

Neither `src/autoskillit/server/__init__.py` nor `src/autoskillit/server/_state.py` contains:
- Any FastMCP `lifespan` async context manager with shutdown logic.
- Any `on_shutdown` / `on_exit` callback.
- Any signal handler for SIGTERM.
- Any `try/finally` wrapper around `mcp.run()` that calls `recorder.finalize()`.

The only flush is `atexit`. Python's default SIGTERM disposition is to terminate the process immediately, **without running `atexit` hooks**. Claude Code shuts down stdio MCP servers by sending SIGTERM to the child process on `/exit`. So `finalize()` is never called in practice.

### Evidence from this run

- `sessions/001_implement_task/metadata.json` shows `exit_code=0, duration_ms=32195, termination_reason=end_turn` — the PTY capture layer works.
- `stdout.jsonl` is 46 KB of well-formed stream-json — capture works.
- No `scenario.json` at the root → `finalize()` did not run.
- No Python exception traceback printed during `/exit` → consistent with SIGTERM bypassing the interpreter's normal shutdown.
- The autoskillit session diagnostic log at `~/.local/share/autoskillit/logs/sessions/no_session_2026-04-12T04-16-40.*` corresponds to the *inner* `run_skill` headless subprocess (32 s, matching `duration_ms`), not to the outer MCP server process. There is no diagnostic log for the outer MCP server process itself.

## Affected Components

| File | Lines | Role |
|---|---|---|
| `src/autoskillit/execution/recording.py` | 67 | Sole `atexit.register(recorder.finalize)` call. No SIGTERM handler, no explicit finalize in any teardown path. |
| `src/autoskillit/execution/recording.py` | 89, 99, 113–121 | `RecordingSubprocessRunner.__call__` — non-session branch. Dead code in practice (see secondary issue). |
| `src/autoskillit/server/__init__.py` | — | No `lifespan` hook, no shutdown handler, no `try/finally` around `mcp.run()`. |
| `src/autoskillit/server/_state.py` | ~60–88 | `_initialize()` registers `McpRecordingMiddleware(ctx.runner.recorder)` (PR #727). Correct, but irrelevant when `finalize()` never runs. |
| `src/autoskillit/server/_factory.py` | 155–165 | `make_context()` creates `recorder = make_scenario_recorder(output_dir, recipe_name)` and wraps `RecordingSubprocessRunner` around the inner runner. Correct. |
| `src/autoskillit/pipeline/helpers.py` | 359–361 | `_run_subprocess` never passes `env` to `runner(...)`. Makes the non-session branch of `RecordingSubprocessRunner` effectively dead (see secondary issue). |
| `api_simulator.claude.ScenarioRecorder.finalize` | Rust, `claude/scenario.rs:758–782` | The actual writer of `scenario.json`. Only runs when explicitly called. |
| `api_simulator.mcp.McpRecordingMiddleware` | Python, `mcp/middleware.py:57–81` | Buffers via `recorder.record_mcp_call()`. No direct disk writes. |

## Data Flow Trace

```
User: task test-smoke-record
  → autoskillit CLI sets RECORD_SCENARIO=1, RECORD_SCENARIO_DIR=..., RECORD_SCENARIO_RECIPE=smoke-test
  → autoskillit order smoke-test → interactive Claude Code session
  → Claude Code spawns autoskillit MCP server as stdio subprocess
    → MCP server process:
       1. make_context() runs in _factory.py:155-165
       2. recorder = make_scenario_recorder(output_dir, "smoke-test")
       3. RecordingSubprocessRunner wraps runner
       4. atexit.register(recorder.finalize)   ← registered, will NOT run under SIGTERM
       5. _initialize() in _state.py registers McpRecordingMiddleware(recorder) on the FastMCP app
       6. mcp.run() enters stdio loop

  → Interactive session runs the smoke-test recipe:
    - open_kitchen           → middleware appends to recorder.mcp_calls (MEMORY)
    - clone_repo             → middleware appends; tool handler uses workspace/clone.py raw subprocess.run (BYPASSES runner entirely)
    - create_unique_branch   → middleware appends; _run_subprocess → runner() with no env → falls through, unrecorded at subprocess level
    - run_cmd (setup)        → same as create_unique_branch
    - run_skill              → middleware appends; _run_subprocess → runner() with SCENARIO_STEP_NAME=\"implement_task\" and pty_mode=True → recorder.record_step() → Rust spawns PTY, writes sessions/001_implement_task/ INCREMENTALLY ← THIS IS THE ONLY THING THAT SURVIVES
    - register_clone_status  → middleware appends; pure file I/O

  → User: /exit
  → Claude Code sends SIGTERM to MCP server subprocess
  → Python default SIGTERM disposition: terminate immediately
  → atexit hooks DO NOT RUN
  → recorder.finalize() NEVER CALLED
  → scenario.json, mcp_calls, non-session steps → LOST
```

## Secondary Architectural Issue (deferred, not a blocker)

The non-session branch of `RecordingSubprocessRunner.__call__` (`recording.py:113–121`) is effectively **dead code** for every MCP tool call:

1. `RecordingSubprocessRunner.__call__` reads `step_name = (env or {}).get(SCENARIO_STEP_NAME_ENV, \"\")` at line 89.
2. `_run_subprocess` in `src/autoskillit/pipeline/helpers.py:359–361` never passes an `env` kwarg:
   ```python
   runner = _get_ctx().runner
   result = await runner(cmd, cwd=Path(cwd), timeout=timeout)
   ```
3. Therefore `step_name` is always `\"\"` when MCP tool handlers invoke the runner, and `record_non_session_step()` never fires.
4. `SCENARIO_STEP_NAME` is set in exactly one place: `src/autoskillit/execution/commands.py:192`, inside the headless session command builder — used **only** when `run_skill` launches a `claude` PTY subprocess.

On top of that, `clone_repo` bypasses `ctx.runner` entirely. Its implementation (`src/autoskillit/workspace/clone.py`, lines 51, 69, 87, 103, 123, 133, 293, 317, 335, 352, 362, 367) calls `subprocess.run(...)` directly. The `workspace/` package is L1 and cannot import from `pipeline/`, so it has no access to the DI container.

**Consequence:** the only recording layer that observes `clone_repo`, `run_cmd`, `create_unique_branch`, `register_clone_status` is `McpRecordingMiddleware` at the MCP protocol layer — and that's buffered in memory until `finalize()`.

**Options once the primary fix is in:**
- (a) Remove the non-session branch entirely — `McpRecordingMiddleware` covers this territory at the protocol layer.
- (b) Plumb `SCENARIO_STEP_NAME` through from MCP tool `step_name` parameters into `_run_subprocess`'s env kwarg, so subprocess-level recording actually fires.

Defer this decision until the primary fix ships.

## Test Gap Analysis

`tests/execution/test_recording.py` has 22+ tests but none of them:

1. Assert that a complete scenario directory contains `scenario.json` or any specific file layout. Every test mocks `ScenarioPlayer` and `ScenarioRecorder` entirely — `recorder = Mock()`, `finalize` is never called in any test.
2. Cover `McpRecordingMiddleware` registration. The class name does not appear anywhere in the test file.
3. Cover the full end-to-end flow of recording a scenario to disk and replaying it from that same disk artifact. The replay tests construct `session_map` dicts manually in memory.
4. Test the interaction between `atexit`, `finalize()`, and SIGTERM delivery. `atexit.register` itself is often mocked out via `monkeypatch.setattr(\"atexit.register\", Mock())`.

`tests/recipe/test_smoke_pipeline.py` never references `tests/fixtures/scenarios/smoke-happy/` at all. It is described as a \"documentation anchor\"; it checks only YAML structure of `smoke-test.yaml` and skips `TestSmokePipelineExecution` unless `SMOKE_TEST=1`.

**No existing test would have caught this.** The finalize-on-exit contract is untested end-to-end. The actual shutdown signal delivery has no test coverage. This is an integration gap between autoskillit's process lifecycle and api-simulator's in-memory buffering strategy.

### 1. [Primary] Wire `finalize()` through FastMCP `lifespan`

Add a `lifespan` async context manager to the FastMCP instance in `src/autoskillit/server/__init__.py`. In its `__aexit__`, do:

```python
if isinstance(ctx.runner, RecordingSubprocessRunner):
    ctx.runner.recorder.finalize()
```

FastMCP guarantees the lifespan shutdown runs before the stdio loop exits, so SIGTERM delivery is no longer a problem — the event loop is torn down cleanly. Deterministic, visible in the call graph, easily testable.

### 2. [Belt-and-suspenders] SIGTERM → SystemExit handler

Install `signal.signal(signal.SIGTERM, lambda *_: sys.exit(0))` in `server/__init__.py` so that even if future code paths forget the lifespan wiring, SIGTERM converts to a clean `SystemExit` and `atexit` hooks run. One line. Optional safety net.

### 3. Delete the `atexit.register(recorder.finalize)` line once (1) is in place

`recording.py:67` is misleading: it suggests a working shutdown contract that doesn't actually work for the signal this project actually receives. Keeping it obscures where the real shutdown contract lives.

### 4. Add an end-to-end record-replay integration test

Spawn the MCP server as a subprocess under `RECORD_SCENARIO=1`, issue a few MCP tool calls over stdio, send SIGTERM, then assert:
- `scenario.json` exists at the output path
- `scenario.json` contains the expected `mcp_calls` entries
- At least one per-step cassette directory exists under `sessions/`

This is the only test that would have caught this bug and would guard any future regression.

### 5. [Deferred] Decide the fate of the non-session subprocess recording branch

See \"Secondary Architectural Issue\" above. Defer until after (1) ships.

## Unrelated — explicitly out of scope

The `run_skill` session was classified as `needs_retry: true, retry_reason: empty_output` by the autoskillit orchestrator despite producing 46 KB of well-formed stream-json, exit code 0, and a clean `end_turn` termination. That is a **session-adjudication** problem (probably in `execution/session_adjudication.py` or the `_run_skill` retry-reason classifier), not a recording problem. The PTY cassette itself is healthy. **Do not conflate these two issues.** A separate investigation should cover the `empty_output` false positive.

## Bottom Line

One fix unblocks the entire recording feature: **call `recorder.finalize()` from a deterministic shutdown hook on the MCP server process** — FastMCP lifespan is the cleanest choice. Every other piece of the recording machinery (PTY capture, MCP middleware buffering, non-session record buffering) is already working correctly; they're all just waiting for a flush that never happens.

Trecek:
## Summary

The MCP server process (`autoskillit serve`) terminates under SIGTERM (delivered by Claude Code on `/exit`), which bypasses Python's `atexit` hooks entirely. `ScenarioRecorder.finalize()` — the sole writer of `scenario.json` and the only flush of the in-memory MCP call buffer — is registered only via `atexit.register()` at `recording.py:67`. It never runs. Every recording session loses all scenario data except PTY cassettes already written to disk incrementally by `run_skill`.

The architectural weakness is that **server resource teardown is scattered into the resource's constructor** (`recording.py:67`) rather than owned by the server's lifecycle system. Worse: the server has no lifecycle system at all — no FastMCP `lifespan`, no signal handler, no `try/finally` around `mcp.run()`. Any future resource that needs cleanup on server exit faces the identical failure mode.

The immunity comes from two structural changes:

1. **Introduce a FastMCP `lifespan` async context manager** — the single authoritative teardown point for all server resources. No other shutdown mechanism is valid; all resources register their teardown here.
2. **Install a SIGTERM → `sys.exit(0)` signal handler** — converts signal-based termination into a clean Python exit so the lifespan's `__aexit__` always fires.

Together, these make the class of "resource registered for cleanup but bypassed by signal" bugs impossible: teardown is now a framework contract enforced at the server boundary, not a voluntary constructor-level `atexit` registration.

### State Lifecycle Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% ── STARTUP PHASE ── %%
    subgraph Startup ["SERVER STARTUP — make_context() + _initialize()"]
        direction LR

        CtxFields["ToolContext fields<br/>━━━━━━━━━━<br/>INIT_ONLY: config, gate, audit,<br/>executor, runner, recorder<br/>Set once by make_context()"]

        MWFlag["● _mcp_middleware_registered<br/>━━━━━━━━━━<br/>MUTABLE (write-once flag)<br/>Guard: prevents double middleware<br/>Set by _initialize()"]

        GateClosed["ctx.gate.enabled = False<br/>━━━━━━━━━━<br/>MUTABLE: starts CLOSED<br/>Toggled by open_kitchen /<br/>close_kitchen at runtime"]

        AuditLoad["ctx.audit entries<br/>━━━━━━━━━━<br/>APPEND_ONLY<br/>Disk load in _initialize()<br/>Tools append at runtime"]
    end

    %% ── FASTMCP WIRING ── %%
    subgraph MCPLayer ["● server/__init__.py — FastMCP App"]
        direction TB

        MCPApp["● mcp = FastMCP('autoskillit',...)<br/>━━━━━━━━━━<br/>INIT_ONLY at module import<br/>lifespan=_autoskillit_lifespan<br/>mcp.disable(tags={'kitchen'}) on import"]

        SigInstall["● signal.signal(SIGTERM, handler)<br/>━━━━━━━━━━<br/>INIT_ONLY: installed before mcp.run()<br/>Converts SIGTERM → KeyboardInterrupt<br/>Registered in serve()"]
    end

    %% ── SHUTDOWN TRIGGERS ── %%
    subgraph ShutdownTriggers ["SHUTDOWN TRIGGERS"]
        direction LR

        SIGTERM(["OS: SIGTERM<br/>━━━━━━━━━━<br/>e.g. Claude Code /exit"])

        SigHandler["● _sigterm_handler(*_)<br/>━━━━━━━━━━<br/>raise KeyboardInterrupt<br/>No state read/written<br/>Converts signal → exception"]

        NormalExit(["User EOF / Ctrl-C<br/>━━━━━━━━━━<br/>Direct KeyboardInterrupt"])
    end

    KBI(["KeyboardInterrupt<br/>━━━━━━━━━━<br/>Both paths converge here"])

    MCPRun["mcp.run() → shutdown<br/>━━━━━━━━━━<br/>Catches KeyboardInterrupt<br/>Triggers FastMCP lifespan __aexit__"]

    %% ── LIFESPAN TEARDOWN ── %%
    subgraph LifespanTeardown ["★ server/_lifespan.py — _autoskillit_lifespan.__aexit__"]
        direction TB

        LifespanCtx["★ _autoskillit_lifespan<br/>━━━━━━━━━━<br/>@asynccontextmanager<br/>Startup: no-op (yield only)<br/>Teardown: ALL work in __aexit__"]

        Gate1{"GATE 1<br/>ctx is not None<br/>━━━━━━━━━━<br/>_get_ctx_or_none()"}

        Gate2{"GATE 2<br/>isinstance(runner,<br/>RecordingSubprocessRunner)<br/>━━━━━━━━━━<br/>Type guard"}
    end

    %% ── RECORDING STATE ── %%
    subgraph RecordingState ["● execution/recording.py — RecordingSubprocessRunner"]
        direction TB

        RunnerFields["● RecordingSubprocessRunner<br/>━━━━━━━━━━<br/>recorder: INIT_ONLY (set in __init__)<br/>_inner: INIT_ONLY (delegation target)<br/>_mcp_middleware_registered: MUTABLE"]

        FinalizeCall["recorder.finalize()<br/>━━━━━━━━━━<br/>Called exactly once by lifespan<br/>No idempotency guard in this repo<br/>(api-simulator owns finalize logic)"]
    end

    ScenarioOut[/"scenario.json<br/>━━━━━━━━━━<br/>Written to RECORD_SCENARIO_DIR<br/>Contains recipe name + steps<br/>Issue #745: now guaranteed on SIGTERM"/]

    NoOp(["no-op<br/>━━━━━━━━━━<br/>Safe skip — not a recording run"])

    %% ── CONNECTIONS ── %%
    Startup --> MCPLayer
    CtxFields -->|"runner field"| RunnerFields
    MWFlag -->|"guards"| RunnerFields

    SIGTERM --> SigHandler
    SigHandler -->|"raises"| KBI
    NormalExit --> KBI
    KBI -->|"propagates into"| MCPRun
    SigInstall -->|"installed before"| MCPRun
    MCPApp -->|"lifespan= wired"| MCPRun
    MCPRun -->|"FastMCP __aexit__"| LifespanCtx
    LifespanCtx --> Gate1

    Gate1 -->|"ctx not None"| Gate2
    Gate1 -->|"ctx is None"| NoOp

    Gate2 -->|"isinstance pass"| FinalizeCall
    Gate2 -->|"type mismatch"| NoOp

    RunnerFields -->|"recorder field"| FinalizeCall
    FinalizeCall --> ScenarioOut

    %% ── CLASS ASSIGNMENTS ── %%
    class CtxFields,AuditLoad stateNode;
    class MWFlag,GateClosed gap;
    class MCPApp phase;
    class SigInstall,SigHandler cli;
    class SIGTERM,NormalExit,KBI,NoOp terminal;
    class MCPRun phase;
    class LifespanCtx newComponent;
    class Gate1,Gate2 detector;
    class RunnerFields handler;
    class FinalizeCall gap;
    class ScenarioOut output;
```

### Error/Resilience Diagram

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 40, 'rankSpacing': 52, 'curve': 'basis'}}}%%
flowchart TB
    %% CLASS DEFINITIONS %%

    %% ENTRY POINTS %%
    SIGTERM(["SIGTERM<br/>External Signal"])
    KEYBOARD_INT(["KeyboardInterrupt<br/>Ctrl-C"])

    subgraph SignalGate ["● SIGNAL HANDLING — cli/app.py serve()"]
        direction TB
        SIGTERM_H["● _sigterm_handler<br/>━━━━━━━━━━<br/>raise KeyboardInterrupt<br/>(converts OS signal → Python exc)"]
        MCP_RUN["● mcp.run()<br/>━━━━━━━━━━<br/>try: mcp.run()<br/>except KeyboardInterrupt: pass"]
    end

    subgraph LifespanTeardown ["★ LIFESPAN TEARDOWN — server/_lifespan.py"]
        direction TB
        LIFESPAN["★ _autoskillit_lifespan<br/>━━━━━━━━━━<br/>@asynccontextmanager<br/>__aexit__ side fires on shutdown"]
        CTX_GUARD{"_get_ctx_or_none()<br/>returns None?"}
        TYPE_GUARD{"isinstance(runner,<br/>RecordingSubprocessRunner)?"}
        FINALIZE["★ recorder.finalize()<br/>━━━━━━━━━━<br/>Writes scenario.json<br/>NO try/except here"]
    end

    subgraph RecordingPath ["● RECORDING SESSION — execution/recording.py"]
        direction TB
        REC_CALL["● RecordingSubprocessRunner.__call__<br/>━━━━━━━━━━<br/>Routes: pty_mode + step_name → _record_session<br/>else → inner runner (optionally records summary)"]
        RECORD_SESSION["● _record_session()<br/>━━━━━━━━━━<br/>asyncio.to_thread(recorder.record_step)"]
        REC_TRY{"try:<br/>record_step()"}
        REC_EXCEPT["logger.exception(<br/>'record_step failed')<br/>then re-raise"]
    end

    subgraph ReplayValidation ["● REPLAY VALIDATION GATES — execution/recording.py"]
        direction TB
        REPLAY_STEP{"SCENARIO_STEP_NAME<br/>present in env?"}
        REPLAY_MAP{"step_name in<br/>sessions or non_session?"}
        REPLAY_RUN["Consume from deque /<br/>return non_session result"]
    end

    subgraph InitRecovery ["WARN-AND-CONTINUE RECOVERY — server/_state.py _initialize()"]
        direction TB
        SUBSET_GUARD["subset mcp.disable()<br/>━━━━━━━━━━<br/>ImportError → logger.error, continue"]
        MIDDLEWARE_GUARD["McpRecordingMiddleware add<br/>━━━━━━━━━━<br/>ImportError / Exception → logger.warning, continue"]
        CRASH_GUARD["recover_crashed_sessions()<br/>━━━━━━━━━━<br/>Any Exception → logger.debug, continue"]
        TELEMETRY_GUARD["telemetry recovery<br/>━━━━━━━━━━<br/>Any Exception → logger.warning, continue"]
        SKILL_GUARD["session_skill_manager.cleanup_stale()<br/>━━━━━━━━━━<br/>Any Exception → logger.warning, continue"]
    end

    %% TERMINAL STATES %%
    T_SHUTDOWN(["CLEAN_SHUTDOWN<br/>process exits 0"])
    T_SCENARIO_OK(["★ scenario.json WRITTEN<br/>Recording preserved"])
    T_SKIP_NONE(["SKIP: ctx=None<br/>Silent, safe"])
    T_SKIP_PLAIN(["SKIP: plain runner<br/>No recording to save"])
    T_RECORD_FAIL(["RECORD_STEP FAILED<br/>Exception propagates<br/>to tool handler"])
    T_VAL_MISSING(["ValueError:<br/>SCENARIO_STEP_NAME absent"])
    T_REPLAY_MISS(["ScenarioReplayError:<br/>step not found<br/>+ diagnostic names"])
    T_FINALIZE_UNCAUGHT(["UNHANDLED finalize()<br/>Exception bubbles<br/>through FastMCP"])
    T_CTX_UNINIT(["RuntimeError:<br/>serve() not called before<br/>_get_ctx() on gated tool"])
    T_INIT_OK(["_initialize() complete<br/>non-fatal errors logged,<br/>server continues"])

    %% SIGNAL FLOW %%
    SIGTERM --> SIGTERM_H
    KEYBOARD_INT --> MCP_RUN
    SIGTERM_H -->|"raises KeyboardInterrupt"| MCP_RUN
    MCP_RUN -->|"except KeyboardInterrupt: pass"| LIFESPAN
    MCP_RUN -->|"clean exit"| T_SHUTDOWN

    %% LIFESPAN TEARDOWN FLOW %%
    LIFESPAN --> CTX_GUARD
    CTX_GUARD -->|"None"| T_SKIP_NONE
    CTX_GUARD -->|"ctx present"| TYPE_GUARD
    TYPE_GUARD -->|"False"| T_SKIP_PLAIN
    TYPE_GUARD -->|"True"| FINALIZE
    FINALIZE -->|"success"| T_SCENARIO_OK
    FINALIZE -->|"raises"| T_FINALIZE_UNCAUGHT

    %% RECORDING PATH %%
    REC_CALL -->|"pty_mode=True + step_name"| RECORD_SESSION
    RECORD_SESSION --> REC_TRY
    REC_TRY -->|"success"| REC_CALL
    REC_TRY -->|"Exception"| REC_EXCEPT
    REC_EXCEPT --> T_RECORD_FAIL

    %% REPLAY VALIDATION GATES %%
    REPLAY_STEP -->|"absent"| T_VAL_MISSING
    REPLAY_STEP -->|"present"| REPLAY_MAP
    REPLAY_MAP -->|"found"| REPLAY_RUN
    REPLAY_MAP -->|"not found"| T_REPLAY_MISS

    %% INIT RECOVERY %%
    SUBSET_GUARD --> MIDDLEWARE_GUARD
    MIDDLEWARE_GUARD --> CRASH_GUARD
    CRASH_GUARD --> TELEMETRY_GUARD
    TELEMETRY_GUARD --> SKILL_GUARD
    SKILL_GUARD --> T_INIT_OK

    %% CLASS ASSIGNMENTS %%
    class SIGTERM,KEYBOARD_INT integration;
    class SIGTERM_H,MCP_RUN handler;
    class LIFESPAN,FINALIZE newComponent;
    class CTX_GUARD,TYPE_GUARD,REPLAY_STEP,REPLAY_MAP detector;
    class REC_CALL,RECORD_SESSION handler;
    class REC_TRY stateNode;
    class REC_EXCEPT gap;
    class REPLAY_RUN output;
    class SUBSET_GUARD,MIDDLEWARE_GUARD,CRASH_GUARD,TELEMETRY_GUARD,SKILL_GUARD output;
    class T_SHUTDOWN,T_SCENARIO_OK,T_SKIP_NONE,T_SKIP_PLAIN,T_RECORD_FAIL,T_VAL_MISSING,T_REPLAY_MISS,T_FINALIZE_UNCAUGHT,T_CTX_UNINIT,T_INIT_OK terminal;
```

### Process Flow Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINALS %%
    START([START — autoskillit serve])
    CLEAN_EXIT([CLEAN EXIT])
    FINALIZED([FINALIZED — scenario.json written])
    ERROR([ERROR — unexpected exception])

    subgraph Startup ["● Phase 1 — Server Startup (cli/app.py:serve)"]
        direction TB
        ConfigLog["● Configure Logging<br/>━━━━━━━━━━<br/>INFO or DEBUG to stderr<br/>before server import"]
        LoadConfig["● Load Config<br/>━━━━━━━━━━<br/>load_config(project_dir)<br/>resolve config path"]
        ImportServer["● Import autoskillit.server<br/>━━━━━━━━━━<br/>Triggers FastMCP construction<br/>tool module registration<br/>mcp.disable(tags={'kitchen'})"]
        HeadlessGate{"AUTOSKILLIT_HEADLESS<br/>== '1'?"}
        EnableHeadless["Enable headless tag<br/>━━━━━━━━━━<br/>mcp.enable(tags={'headless'})<br/>exposes test_check only"]
        SigtermHandler["● Install SIGTERM Handler<br/>━━━━━━━━━━<br/>signal.signal(SIGTERM,<br/>_sigterm_handler)<br/>converts SIGTERM → KeyboardInterrupt"]
        McpRun["● mcp.run()<br/>━━━━━━━━━━<br/>blocking event loop<br/>wrapped in try/except KeyboardInterrupt"]
    end

    subgraph Lifespan ["★ Phase 2 — FastMCP Lifespan (server/_lifespan.py)"]
        direction TB
        LifespanYield["★ _autoskillit_lifespan: yield<br/>━━━━━━━━━━<br/>startup: no-op<br/>server active, serving MCP requests"]
    end

    subgraph Recording ["● Phase 3 — Recording Dispatch (execution/recording.py)"]
        direction TB
        RecCall["● RecordingSubprocessRunner.__call__<br/>━━━━━━━━━━<br/>invoked per tool subprocess request<br/>no atexit registered (lifespan owns teardown)"]
        StepNameGate{"SCENARIO_STEP_NAME<br/>in env?"}
        PtyGate{"pty_mode<br/>== True?"}
        RecordSession["● _record_session()<br/>━━━━━━━━━━<br/>asyncio.to_thread(recorder.record_step)<br/>returns synthetic SubprocessResult<br/>termination=NATURAL_EXIT, pid=0"]
        PassThrough["Pass-through<br/>━━━━━━━━━━<br/>delegate to self._inner<br/>DefaultSubprocessRunner"]
        RecordNonSession["● record_non_session_step()<br/>━━━━━━━━━━<br/>captures exit_code + stdout[:500]<br/>after inner runner completes"]
    end

    subgraph Shutdown ["★ Phase 4 — SIGTERM Teardown (_lifespan.py __aexit__)"]
        direction TB
        SigtermReceived["● SIGTERM received<br/>━━━━━━━━━━<br/>_sigterm_handler raises KeyboardInterrupt<br/>asyncio/anyio unwinds event loop"]
        LifespanExit["★ _autoskillit_lifespan: __aexit__<br/>━━━━━━━━━━<br/>lifespan context manager exits<br/>FastMCP-driven teardown path"]
        GetCtx["★ _get_ctx_or_none()<br/>━━━━━━━━━━<br/>safe retrieval: returns None<br/>if no session context active"]
        CtxGate{"ctx is<br/>not None?"}
        RunnerGate{"isinstance(ctx.runner,<br/>RecordingSubprocessRunner)?"}
        Finalize["★ recorder.finalize()<br/>━━━━━━━━━━<br/>writes scenario.json to disk<br/>persists all captured steps"]
        NoOpExit["No-op teardown<br/>━━━━━━━━━━<br/>non-recording mode<br/>or no context initialized"]
    end

    %% STARTUP FLOW %%
    START --> ConfigLog
    ConfigLog --> LoadConfig
    LoadConfig --> ImportServer
    ImportServer --> HeadlessGate
    HeadlessGate -->|"yes"| EnableHeadless
    HeadlessGate -->|"no"| SigtermHandler
    EnableHeadless --> SigtermHandler
    SigtermHandler --> McpRun
    McpRun --> LifespanYield

    %% RECORDING DISPATCH (repeated per tool call) %%
    LifespanYield -->|"tool subprocess requested"| RecCall
    RecCall --> StepNameGate
    StepNameGate -->|"no step name"| PassThrough
    StepNameGate -->|"step name present"| PtyGate
    PtyGate -->|"pty_mode=True"| RecordSession
    PtyGate -->|"pty_mode=False"| PassThrough
    PassThrough -->|"non-pty + step name"| RecordNonSession
    RecordSession -->|"returns to lifespan"| LifespanYield
    RecordNonSession -->|"returns to lifespan"| LifespanYield
    PassThrough -->|"no step name: returns to lifespan"| LifespanYield

    %% SHUTDOWN FLOW %%
    McpRun -->|"KeyboardInterrupt caught"| SigtermReceived
    SigtermReceived --> LifespanExit
    LifespanExit --> GetCtx
    GetCtx --> CtxGate
    CtxGate -->|"None (no context)"| NoOpExit
    CtxGate -->|"ctx exists"| RunnerGate
    RunnerGate -->|"not RecordingSubprocessRunner"| NoOpExit
    RunnerGate -->|"is RecordingSubprocessRunner"| Finalize
    Finalize --> FINALIZED
    NoOpExit --> CLEAN_EXIT

    %% ERROR PATH %%
    RecordSession -->|"exception from record_step"| ERROR

    %% CLASS ASSIGNMENTS %%
    class START,CLEAN_EXIT,FINALIZED,ERROR terminal;
    class ConfigLog,LoadConfig,ImportServer,EnableHeadless phase;
    class SigtermHandler,McpRun,RecCall,PassThrough handler;
    class HeadlessGate,StepNameGate,PtyGate,CtxGate,RunnerGate stateNode;
    class SigtermReceived,Finalize detector;
    class LifespanYield,LifespanExit,GetCtx,RecordSession,RecordNonSession,NoOpExit newComponent;
```

Closes #745

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-745-20260411-213724-303410/.autoskillit/temp/rectify/rectify_sigterm_finalize_2026-04-11_120000.md`

## Token Usage Summary

| investigate | 106 | 9.3k | 408.0k | 39.9k | 1 | 4m 53s |
| rectify | 121 | 25.0k | 579.5k | 57.2k | 1 | 9m 46s |
| review | 5.0k | 10.1k | 809.8k | 70.9k | 1 | 5m 18s |
| dry_walkthrough | 57 | 19.3k | 1.3M | 81.1k | 1 | 7m 25s |
| implement | 91 | 17.4k | 3.0M | 70.0k | 1 | 6m 24s |
| assess | 161 | 32.7k | 7.3M | 133.2k | 2 | 20m 22s |
| prepare_pr | 34 | 5.5k | 214.7k | 30.7k | 1 | 1m 55s |
| run_arch_lenses | 1.9k | 23.6k | 569.8k | 92.6k | 3 | 9m 59s |
| compose_pr | 75 | 10.5k | 260.3k | 34.6k | 1 | 2m 30s |
| **Total** | 7.6k | 153.2k | 14.5M | 610.2k | | 1h 8m |

### DIFF
--- a/src/autoskillit/cli/app.py
+++ b/src/autoskillit/cli/app.py
@@ -115,8 +115,16 @@ def serve(*, verbose: Annotated[bool, Parameter(name=["--verbose", "-v"])] = Fal
     plugin_dir = str(pkg_root())
     ctx = make_context(cfg, plugin_dir=plugin_dir)
     _initialize(ctx)
-    signal.signal(signal.SIGTERM, lambda *_: sys.exit(0))
-    mcp.run()
+
+    def _sigterm_handler(*_):
+        """Convert SIGTERM into KeyboardInterrupt so asyncio shuts down cleanly."""
+        raise KeyboardInterrupt
+
+    signal.signal(signal.SIGTERM, _sigterm_handler)
+    try:
+        mcp.run()
+    except KeyboardInterrupt:
+        pass
 
 
 @app.command
--- a/src/autoskillit/cli/app.py
+++ b/src/autoskillit/cli/app.py
@@ -121,6 +121,7 @@ def serve(*, verbose: Annotated[bool, Parameter(name=["--verbose", "-v"])] = Fal
         raise KeyboardInterrupt
 
     signal.signal(signal.SIGTERM, _sigterm_handler)
+    get_logger(__name__).info("sigterm_handler_ready")
     try:
         mcp.run()
     except KeyboardInterrupt:
--- a/src/autoskillit/cli/app.py
+++ b/src/autoskillit/cli/app.py
@@ -7,6 +7,7 @@ import json
 import os
 import random
 import shutil
+import signal
 import subprocess
 import sys
 from datetime import UTC
@@ -114,6 +115,7 @@ def serve(*, verbose: Annotated[bool, Parameter(name=["--verbose", "-v"])] = Fal
     plugin_dir = str(pkg_root())
     ctx = make_context(cfg, plugin_dir=plugin_dir)
     _initialize(ctx)
+    signal.signal(signal.SIGTERM, lambda *_: sys.exit(0))
     mcp.run()
 
 
--- a/src/autoskillit/execution/recording.py
+++ b/src/autoskillit/execution/recording.py
@@ -3,8 +3,6 @@
 from __future__ import annotations
 
 import asyncio
-import atexit
-import shutil
 import tempfile
 from collections import deque
 from collections.abc import Mapping
@@ -64,7 +62,6 @@ class RecordingSubprocessRunner(SubprocessRunner):
         inner: SubprocessRunner | None = None,
     ) -> None:
         self.recorder = recorder
-        atexit.register(recorder.finalize)
         if inner is None:
             from autoskillit.execution.process import DefaultSubprocessRunner
 
@@ -169,6 +166,8 @@ class ReplayingSubprocessRunner(SubprocessRunner):
     dispatches to the matching step queue.
     """
 
+    _tmp_replay_dir: tempfile.TemporaryDirectory[str] | None
+
     def __init__(
         self,
         session_map: dict[str, deque[tuple[Any, Any]]],
@@ -180,6 +179,7 @@ class ReplayingSubprocessRunner(SubprocessRunner):
         self._non_session = non_session_results
         self.player: ScenarioPlayer | None = player
         self.call_log: list[tuple[str, list[str]]] = []
+        self._tmp_replay_dir = None
 
     async def __call__(
         self,
@@ -236,8 +236,8 @@ class ReplayingSubprocessRunner(SubprocessRunner):
 def build_replay_runner(replay_dir: str) -> ReplayingSubprocessRunner:
     """Build a ReplayingSubprocessRunner from a scenario directory.
 
-    Creates a temporary output directory for the player, registers an atexit
-    handler to clean it up, then parses the scenario manifest and constructs
+    Creates a temporary output directory for the player, then parses the
+    scenario manifest and constructs
     the deque-based session map.  All domain logic for replay setup lives here
     (L1) rather than in the L3 composition root.
 
@@ -262,8 +262,8 @@ def build_replay_runner(replay_dir: str) -> ReplayingSubprocessRunner:
             "Install it to enable scenario replay."
         ) from exc
 
-    tmp_replay = tempfile.mkdtemp(prefix="autoskillit-replay-")
-    atexit.register(shutil.rmtree, tmp_replay, True)
+    _tmp_replay_dir = tempfile.TemporaryDirectory(prefix="autoskillit-replay-")
+    tmp_replay = _tmp_replay_dir.name
 
     player = make_scenario_player(
         scenario_dir=replay_dir,
@@ -284,4 +284,6 @@ def build_replay_runner(replay_dir: str) -> ReplayingSubprocessRunner:
         raise
 
     session_map = {k: deque(v) for k, v in raw_map.items()}
-    return ReplayingSubprocessRunner(session_map, non_session, player=player)
+    runner = ReplayingSubprocessRunner(session_map, non_session, player=player)
+    runner._tmp_replay_dir = _tmp_replay_dir
+    return runner
--- a/src/autoskillit/server/__init__.py
+++ b/src/autoskillit/server/__init__.py
@@ -14,9 +14,12 @@ Transport: stdio (default for FastMCP).
 
 from __future__ import annotations
 
+from contextlib import asynccontextmanager
+
 from fastmcp import FastMCP
 
 from autoskillit.core import get_logger
+from autoskillit.execution.recording import RecordingSubprocessRunner
 from autoskillit.pipeline import (  # noqa: F401
     GATED_TOOLS,
     UNGATED_TOOLS,
@@ -28,12 +31,23 @@ from autoskillit.server._state import (  # noqa: E402, F401
     _ctx,
     _get_config,
     _get_ctx,
+    _get_ctx_or_none,
     _get_plugin_dir,
     _initialize,
     version_info,
 )
 
-mcp: FastMCP = FastMCP("autoskillit")
+
+@asynccontextmanager
+async def _autoskillit_lifespan(server):
+    """Server lifecycle: teardown recording on shutdown."""
+    yield
+    ctx = _get_ctx_or_none()
+    if ctx is not None and isinstance(ctx.runner, RecordingSubprocessRunner):
+        ctx.runner.recorder.finalize()
+
+
+mcp: FastMCP = FastMCP("autoskillit", lifespan=_autoskillit_lifespan)
 
 logger = get_logger(__name__)
 
--- a/src/autoskillit/server/__init__.py
+++ b/src/autoskillit/server/__init__.py
@@ -24,12 +24,11 @@ from autoskillit.pipeline import (  # noqa: F401
     ToolContext,
     gate_error_result,
 )
-from autoskillit.server._lifespan import _autoskillit_lifespan  # noqa: F401
+from autoskillit.server._lifespan import _autoskillit_lifespan
 from autoskillit.server._state import (  # noqa: E402, F401
     _ctx,
     _get_config,
     _get_ctx,
-    _get_ctx_or_none,
     _get_plugin_dir,
     _initialize,
     version_info,
--- a/src/autoskillit/server/__init__.py
+++ b/src/autoskillit/server/__init__.py
@@ -14,12 +14,9 @@ Transport: stdio (default for FastMCP).
 
 from __future__ import annotations
 
-from contextlib import asynccontextmanager
-
 from fastmcp import FastMCP
 
 from autoskillit.core import get_logger
-from autoskillit.execution.recording import RecordingSubprocessRunner
 from autoskillit.pipeline import (  # noqa: F401
     GATED_TOOLS,
     UNGATED_TOOLS,
@@ -27,6 +24,7 @@ from autoskillit.pipeline import (  # noqa: F401
     ToolContext,
     gate_error_result,
 )
+from autoskillit.server._lifespan import _autoskillit_lifespan  # noqa: F401
 from autoskillit.server._state import (  # noqa: E402, F401
     _ctx,
     _get_config,
@@ -36,16 +34,6 @@ from autoskillit.server._state import (  # noqa: E402, F401
     _initialize,
     version_info,
 )
-
-
-@asynccontextmanager
-async def _autoskillit_lifespan(server):
-    """Server lifecycle: teardown recording on shutdown."""
-    yield
-    ctx = _get_ctx_or_none()
-    if ctx is not None and isinstance(ctx.runner, RecordingSubprocessRunner):
-        ctx.runner.recorder.finalize()
-
 
 mcp: FastMCP = FastMCP("autoskillit", lifespan=_autoskillit_lifespan)
 
--- a/src/autoskillit/server/_lifespan.py
+++ b/src/autoskillit/server/_lifespan.py
@@ -20,6 +20,7 @@ logger = get_logger(__name__)
 @asynccontextmanager
 async def _autoskillit_lifespan(server: Any) -> Any:
     """Server lifecycle: teardown recording on shutdown."""
+    logger.info("lifespan_started")
     yield
     ctx = _get_ctx_or_none()
     if ctx is not None and isinstance(ctx.runner, RecordingSubprocessRunner):
--- a/src/autoskillit/server/_lifespan.py
+++ b/src/autoskillit/server/_lifespan.py
@@ -1,0 +1,22 @@
+"""FastMCP lifespan for server resource teardown.
+
+Provides the async context manager wired into FastMCP via ``lifespan=``.
+The ``__aexit__`` side calls ``recorder.finalize()`` so scenario data survives
+SIGTERM (issue #745).
+"""
+
+from __future__ import annotations
+
+from contextlib import asynccontextmanager
+
+from autoskillit.execution import RecordingSubprocessRunner
+from autoskillit.server._state import _get_ctx_or_none
+
+
+@asynccontextmanager
+async def _autoskillit_lifespan(server):
+    """Server lifecycle: teardown recording on shutdown."""
+    yield
+    ctx = _get_ctx_or_none()
+    if ctx is not None and isinstance(ctx.runner, RecordingSubprocessRunner):
+        ctx.runner.recorder.finalize()
--- a/src/autoskillit/server/_lifespan.py
+++ b/src/autoskillit/server/_lifespan.py
@@ -8,15 +8,22 @@ SIGTERM (issue #745).
 from __future__ import annotations
 
 from contextlib import asynccontextmanager
+from typing import Any
 
+from autoskillit.core import get_logger
 from autoskillit.execution import RecordingSubprocessRunner
 from autoskillit.server._state import _get_ctx_or_none
 
+logger = get_logger(__name__)
+
 
 @asynccontextmanager
-async def _autoskillit_lifespan(server):
+async def _autoskillit_lifespan(server: Any) -> Any:
     """Server lifecycle: teardown recording on shutdown."""
     yield
     ctx = _get_ctx_or_none()
     if ctx is not None and isinstance(ctx.runner, RecordingSubprocessRunner):
-        ctx.runner.recorder.finalize()
+        try:
+            ctx.runner.recorder.finalize()
+        except Exception:
+            logger.exception("recorder.finalize() failed during lifespan teardown")
--- a/tests/cli/test_serve_sigterm.py
+++ b/tests/cli/test_serve_sigterm.py
@@ -7,7 +7,7 @@ import pytest
 
 
 def test_serve_installs_sigterm_handler(monkeypatch, tmp_path):
-    """serve() installs signal.SIGTERM → sys.exit(0) before mcp.run()."""
+    """serve() installs signal.SIGTERM handler before mcp.run()."""
     installed_handlers = {}
 
     original_signal = signal_mod.signal
@@ -38,6 +38,5 @@ def test_serve_installs_sigterm_handler(monkeypatch, tmp_path):
     assert signal_mod.SIGTERM in installed_handlers, "SIGTERM handler not installed by serve()"
 
     handler = installed_handlers[signal_mod.SIGTERM]
-    with pytest.raises(SystemExit) as exc_info:
+    with pytest.raises(KeyboardInterrupt):
         handler(signal_mod.SIGTERM, None)
-    assert exc_info.value.code == 0
--- a/tests/cli/test_serve_sigterm.py
+++ b/tests/cli/test_serve_sigterm.py
@@ -1,0 +1,43 @@
+"""Test that serve() installs a SIGTERM handler. Regression guard for issue #745."""
+
+import signal as signal_mod
+from unittest.mock import MagicMock
+
+import pytest
+
+
+def test_serve_installs_sigterm_handler(monkeypatch, tmp_path):
+    """serve() installs signal.SIGTERM → sys.exit(0) before mcp.run()."""
+    installed_handlers = {}
+
+    original_signal = signal_mod.signal
+
+    def capture(sig, handler):
+        installed_handlers[sig] = handler
+        return original_signal(sig, handler)
+
+    monkeypatch.chdir(tmp_path)
+
+    mock_cfg = MagicMock()
+    mock_cfg.logging.level = "INFO"
+    mock_cfg.logging.json_output = None
+    mock_cfg.safety.protected_branches = []
+
+    # Patch the source modules that serve() imports from at call time.
+    monkeypatch.setattr("autoskillit.config.load_config", lambda _: mock_cfg)
+    monkeypatch.setattr("autoskillit.core.configure_logging", lambda **kw: None)
+    monkeypatch.setattr("autoskillit.server.make_context", lambda *a, **kw: MagicMock())
+    monkeypatch.setattr("autoskillit.server._initialize", lambda ctx: None)
+    monkeypatch.setattr("autoskillit.server.mcp.run", lambda: None)
+    monkeypatch.setattr(signal_mod, "signal", capture)
+
+    from autoskillit.cli.app import serve
+
+    serve()
+
+    assert signal_mod.SIGTERM in installed_handlers, "SIGTERM handler not installed by serve()"
+
+    handler = installed_handlers[signal_mod.SIGTERM]
+    with pytest.raises(SystemExit) as exc_info:
+        handler(signal_mod.SIGTERM, None)
+    assert exc_info.value.code == 0
--- a/tests/execution/test_recording.py
+++ b/tests/execution/test_recording.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from collections import deque
 from dataclasses import dataclass
 from pathlib import Path
-from unittest.mock import Mock
+from unittest.mock import Mock, patch
 
 import pytest
 
@@ -46,6 +46,21 @@ class FakeMeta:
     exit_code: int
     model: str = "test"
     duration_ms: int = 1000
+
+
+# --- T0: RecordingSubprocessRunner must NOT call atexit.register ---
+
+
+def test_recording_runner_does_not_register_atexit():
+    """
+    RecordingSubprocessRunner must not register atexit hooks.
+    Teardown is owned by the FastMCP server lifespan, not the constructor.
+    Regression guard for issue #745.
+    """
+    mock_recorder = Mock()
+    with patch("atexit.register") as mock_atexit:
+        RecordingSubprocessRunner(recorder=mock_recorder, inner=Mock())
+    mock_atexit.assert_not_called()
 
 
 # --- T1: Protocol compliance ---
@@ -224,13 +239,15 @@ def test_make_context_wraps_runner_when_record_scenario(monkeypatch, tmp_path):
     monkeypatch.setattr(
         _api_sim_claude, "make_scenario_recorder", Mock(return_value=mock_recorder), raising=False
     )
-    monkeypatch.setattr("atexit.register", Mock())
+    mock_atexit = Mock()
+    monkeypatch.setattr("atexit.register", mock_atexit)
 
     from autoskillit.config import AutomationConfig
     from autoskillit.server._factory import make_context
 
     ctx = make_context(AutomationConfig(), plugin_dir=str(tmp_path))
     assert isinstance(ctx.runner, RecordingSubprocessRunner)
+    mock_atexit.assert_not_called()
 
 
 # --- T11: make_context default runner unchanged without env var ---
@@ -459,7 +476,8 @@ def test_replay_takes_precedence_over_record(monkeypatch, tmp_path):
     monkeypatch.setattr(
         _api_sim_claude, "make_scenario_recorder", mock_make_recorder, raising=False
     )
-    monkeypatch.setattr("atexit.register", Mock())
+    mock_atexit = Mock()
+    monkeypatch.setattr("atexit.register", mock_atexit)
 
     from autoskillit.config import AutomationConfig
     from autoskillit.server._factory import make_context
@@ -467,6 +485,7 @@ def test_replay_takes_precedence_over_record(monkeypatch, tmp_path):
     ctx = make_context(AutomationConfig(), plugin_dir=str(tmp_path))
     assert isinstance(ctx.runner, ReplayingSubprocessRunner)
     mock_make_recorder.assert_not_called()  # REPLAY takes precedence over RECORD
+    mock_atexit.assert_not_called()
 
 
 # --- T22: Cross-scenario session override (integration, requires api-simulator) ---
@@ -497,10 +516,12 @@ async def test_cross_scenario_override(tmp_path):
 
 def test_recording_runner_recorder_is_public(monkeypatch):
     """RecordingSubprocessRunner exposes recorder as a public attribute."""
-    monkeypatch.setattr("atexit.register", Mock())
+    mock_atexit = Mock()
+    monkeypatch.setattr("atexit.register", mock_atexit)
     mock_recorder = Mock()
     runner = RecordingSubprocessRunner(recorder=mock_recorder)
     assert runner.recorder is mock_recorder
+    mock_atexit.assert_not_called()
 
 
 # --- T-REPLAY-PLAYER: player attribute stored ---
@@ -540,7 +561,9 @@ def test_build_replay_runner_stores_player_on_runner(tmp_path, monkeypatch):
     monkeypatch.setattr(
         _api_sim_claude, "make_scenario_player", Mock(return_value=mock_player), raising=False
     )
-    monkeypatch.setattr("atexit.register", Mock())
+    mock_atexit = Mock()
+    monkeypatch.setattr("atexit.register", mock_atexit)
 
     result = build_replay_runner(str(tmp_path))
     assert result.player is mock_player
+    mock_atexit.assert_not_called()
--- a/tests/execution/test_recording_sigterm.py
+++ b/tests/execution/test_recording_sigterm.py
@@ -50,6 +50,7 @@ def test_sigterm_writes_scenario_json(tmp_path):
     # teardown which writes scenario.json. Close stdin so the stdio transport
     # detects EOF and the event loop can fully unwind.
     proc.stdin.close()
+    proc.stdin = None  # prevent communicate() from flushing the closed pipe
     proc.send_signal(signal.SIGTERM)
     try:
         stdout_bytes, stderr_bytes = proc.communicate(timeout=10)
--- a/tests/execution/test_recording_sigterm.py
+++ b/tests/execution/test_recording_sigterm.py
@@ -13,6 +13,8 @@ import time
 
 import pytest
 
+_READY_TOKEN = "lifespan_started"
+
 
 @pytest.mark.integration
 def test_sigterm_writes_scenario_json(tmp_path):
@@ -36,13 +38,25 @@ def test_sigterm_writes_scenario_json(tmp_path):
         env=env,
     )
 
-    # Poll stderr for any output (server startup log) to detect readiness,
-    # rather than using a fixed sleep. Falls back to a 5-second ceiling so
-    # the test is both responsive on fast hosts and resilient on slow CI.
+    # Poll stderr line-by-line for the "sigterm_handler_ready" token which
+    # serve() emits immediately after installing the SIGTERM handler. This
+    # guarantees the handler is active before we send SIGTERM, while still
+    # being responsive (no fixed sleep). Falls back after 5 s on slow CI.
+    stderr_lines: list[str] = []
     deadline = time.monotonic() + 5.0
     while time.monotonic() < deadline:
-        ready, _, _ = select.select([proc.stderr], [], [], 0.1)
-        if ready or proc.poll() is not None:
+        remaining = deadline - time.monotonic()
+        readable, _, _ = select.select([proc.stderr], [], [], min(remaining, 0.2))
+        if not readable:
+            if proc.poll() is not None:
+                break
+            continue
+        line = proc.stderr.readline()
+        if not line:
+            break  # EOF — process died
+        decoded = line.decode(errors="replace")
+        stderr_lines.append(decoded)
+        if _READY_TOKEN in decoded:
             break
 
     # SIGTERM is the exact signal Claude Code sends on /exit.
@@ -53,13 +67,13 @@ def test_sigterm_writes_scenario_json(tmp_path):
     proc.stdin = None  # prevent communicate() from flushing the closed pipe
     proc.send_signal(signal.SIGTERM)
     try:
-        stdout_bytes, stderr_bytes = proc.communicate(timeout=10)
+        stdout_bytes, remaining_stderr_bytes = proc.communicate(timeout=10)
     except subprocess.TimeoutExpired:
         proc.kill()
-        stdout_bytes, stderr_bytes = proc.communicate()
+        stdout_bytes, remaining_stderr_bytes = proc.communicate()
 
     stdout = stdout_bytes.decode(errors="replace")
-    stderr = stderr_bytes.decode(errors="replace")
+    stderr = "".join(stderr_lines) + remaining_stderr_bytes.decode(errors="replace")
 
     scenario_json = output_dir / "scenario.json"
     assert scenario_json.exists(), (
--- a/tests/execution/test_recording_sigterm.py
+++ b/tests/execution/test_recording_sigterm.py
@@ -7,8 +7,8 @@ import json
 import os
 import signal
 import subprocess
+import sys
 import time
-from pathlib import Path
 
 import pytest
 
@@ -25,8 +25,10 @@ def test_sigterm_writes_scenario_json(tmp_path):
         "RECORD_SCENARIO_DIR": str(output_dir),
         "RECORD_SCENARIO_RECIPE": "test-recipe",
     }
+    # Use sys.executable -m to ensure we run the worktree-installed version,
+    # not a system-wide `autoskillit` binary that may lack the lifespan fix.
     proc = subprocess.Popen(
-        ["autoskillit", "serve"],
+        [sys.executable, "-m", "autoskillit"],
         stdin=subprocess.PIPE,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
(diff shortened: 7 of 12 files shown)


tests/execution/test_recording.py
```diff
@@ -48,6 +48,21 @@ class FakeMeta:
...
+
+def test_recording_runner_does_not_register_atexit():
+    """
+    RecordingSubprocessRunner must not register atexit hooks.
+    Teardown is owned by the FastMCP server lifespan, not the constructor.
+    Regression guard for issue #745.
+    """
+    mock_recorder = Mock()
+    with patch("atexit.register") as mock_atexit:
+        RecordingSubprocessRunner(recorder=mock_recorder, inner=Mock())
```
Trecek, line 62: [critical] bugs: RecordingSubprocessRunner(...) is outside the `with patch('atexit.register') as mock_atexit:` block due to missing indentation. The atexit mock is not active when the runner is constructed, so mock_atexit.assert_not_called() on L63 is vacuously true — the test proves nothing. Indent the constructor call to sit inside the `with` block.

src/autoskillit/server/_lifespan.py
```diff
@@ -0,0 +1,22 @@
...
+from autoskillit.server._state import _get_ctx_or_none
+
+
+@asynccontextmanager
+async def _autoskillit_lifespan(server):
+    """Server lifecycle: teardown recording on shutdown."""
+    yield
+    ctx = _get_ctx_or_none()
+    if ctx is not None and isinstance(ctx.runner, RecordingSubprocessRunner):
+        ctx.runner.recorder.finalize()
```
Trecek: [critical] defense: recorder.finalize() is called with no exception handling in __aexit__. If finalize() raises, the exception propagates out of the lifespan context manager and can crash server shutdown. Wrap in try/except/log. Additionally, if finalize() is ever made async, the missing await would silently create an unawaited coroutine — verify it is sync.

src/autoskillit/execution/recording.py
```diff
@@ -284,4 +284,6 @@ def build_replay_runner(replay_dir: str) -> ReplayingSubprocessRunner:
         raise
 
     session_map = {k: deque(v) for k, v in raw_map.items()}
-    return ReplayingSubprocessRunner(session_map, non_session, player=player)
+    runner = ReplayingSubprocessRunner(session_map, non_session, player=player)
+    runner._tmp_replay_dir = _tmp_replay_dir
```
Trecek, line 288: [warning] cohesion: runner._tmp_replay_dir is assigned via post-construction attribute mutation, leaving the object in inconsistent state between ReplayingSubprocessRunner() and the assignment. This bypasses encapsulation. Pass the TemporaryDirectory into __init__ or make build_replay_runner a classmethod/factory to keep initialization atomic.

tests/execution/test_recording_sigterm.py
```diff
@@ -0,0 +1,58 @@
...
+    proc = subprocess.Popen(
+        [sys.executable, "-m", "autoskillit"],
+        stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        env=env,
+    )
+
+    # Allow server to start and install its signal handler
+    time.sleep(1.0)
```
Trecek: [warning] tests: time.sleep(1.0) is a fixed wait for server startup before SIGTERM. This is inherently flaky: too short on slow CI, wasteful otherwise. Use a readiness signal (e.g., poll stdout for a ready line or probe for EOF on a startup pipe) to make the test deterministic.

src/autoskillit/server/_lifespan.py
```diff
@@ -0,0 +1,22 @@
...
+from __future__ import annotations
+
+from contextlib import asynccontextmanager
+
+from autoskillit.execution import RecordingSubprocessRunner
+from autoskillit.server._state import _get_ctx_or_none
+
+
+@asynccontextmanager
+async def _autoskillit_lifespan(server):
```
Trecek: [warning] defense: _autoskillit_lifespan has no type annotation on the `server` parameter. Add `server: Any` (or the appropriate FastMCP server type) to satisfy typed boundaries at L3 public-facing functions.

src/autoskillit/cli/app.py
```diff
@@ -114,7 +115,16 @@ def serve(*, verbose: Annotated[bool, Parameter(name=["--verbose", "-v"])] = Fal
     plugin_dir = str(pkg_root())
     ctx = make_context(cfg, plugin_dir=plugin_dir)
     _initialize(ctx)
-    mcp.run()
+
+    def _sigterm_handler(*_):
+        """Convert SIGTERM into KeyboardInterrupt so asyncio shuts down cleanly."""
+        raise KeyboardInterrupt
```
Trecek, line 121: [warning] defense: Raising KeyboardInterrupt directly from a SIGTERM signal handler is unsafe when an asyncio event loop is running. Signal handlers execute between bytecodes and can interrupt asyncio internals at arbitrary points, potentially corrupting loop state. The conventional safe approach is to schedule shutdown via the event loop: `loop.call_soon_threadsafe(loop.stop)` or `asyncio.get_event_loop().call_soon_threadsafe(raise_ki)` rather than raising synchronously from the handler.

tests/execution/test_recording_sigterm.py
```diff
@@ -0,0 +1,58 @@
...
+    proc.stdin.close()
+    proc.send_signal(signal.SIGTERM)
+    try:
+        proc.wait(timeout=10)
+    except subprocess.TimeoutExpired:
+        proc.kill()
+        proc.wait()
+
+    scenario_json = output_dir / "scenario.json"
+    assert scenario_json.exists(), (
```
Trecek, line 79: [warning] tests: When scenario_json.exists() fails (e.g., because proc.wait() timed out and the process was killed), the error message only says 'finalize() likely bypassed'. Include captured proc.stderr output in the assertion message to improve debuggability of CI failures.

tests/cli/test_serve_sigterm.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # Patch the source modules that serve() imports from at call time.
+    monkeypatch.setattr("autoskillit.config.load_config", lambda _: mock_cfg)
+    monkeypatch.setattr("autoskillit.core.configure_logging", lambda **kw: None)
+    monkeypatch.setattr("autoskillit.server.make_context", lambda *a, **kw: MagicMock())
+    monkeypatch.setattr("autoskillit.server._initialize", lambda ctx: None)
+    monkeypatch.setattr("autoskillit.server.mcp.run", lambda: None)
+    monkeypatch.setattr(signal_mod, "signal", capture)
+
+    from autoskillit.cli.app import serve
```
Trecek, line 34: [warning] tests: `from autoskillit.cli.app import serve` is imported inside the test body after monkeypatching. If autoskillit.cli.app was already imported by another test in the same worker (e.g., via autoskillit.__init__), the import is a no-op and the cached module object is used — meaning the monkeypatches on autoskillit.server.mcp.run etc. may not apply. Move the import to module level or use importlib.reload() after patching.

src/autoskillit/server/__init__.py
```diff
@@ -24,16 +24,18 @@
     ToolContext,
     gate_error_result,
 )
+from autoskillit.server._lifespan import _autoskillit_lifespan  # noqa: F401
```
Trecek: [warning] slop: `# noqa: F401` on the `_autoskillit_lifespan` import is misleading — the symbol is used immediately on L38 (`lifespan=_autoskillit_lifespan`), so no F401 warning would fire. Remove the noqa comment.

Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (2 critical, 9 warning findings; see inline comments)

src/autoskillit/execution/recording.py
```diff
@@ -169,6 +166,8 @@ class ReplayingSubprocessRunner(SubprocessRunner):
     dispatches to the matching step queue.
     """
 
+    _tmp_replay_dir: tempfile.TemporaryDirectory[str] | None
```
Trecek, line 169: [warning] cohesion: _tmp_replay_dir is declared as a class-level annotation (L169), set to None in __init__ (L182), then assigned its real value externally in build_replay_runner (L288). This two-phase initialization creates a window where the object is in an inconsistent state and breaks encapsulation. Accept the TemporaryDirectory as a constructor parameter instead.

src/autoskillit/server/__init__.py
```diff
@@ -24,16 +24,18 @@
...
 from autoskillit.server._state import (  # noqa: E402, F401
     _ctx,
     _get_config,
     _get_ctx,
+    _get_ctx_or_none,
```
Trecek: [warning] slop: _get_ctx_or_none is re-exported from server/__init__.py but is only used in _lifespan.py, which already imports it directly from autoskillit.server._state. This re-export is dead and adds noise to the public package surface. Remove it.

tests/execution/test_recording_sigterm.py
```diff
@@ -0,0 +1,58 @@
...
+import signal
+import subprocess
+import sys
+import time
+
+import pytest
+
+
+@pytest.mark.integration
+def test_sigterm_writes_scenario_json(tmp_path):
```
Trecek, line 20: [warning] tests: Integration test spawns a real MCP server subprocess. With xdist -n workers, multiple instances could compete for shared system resources (SQLite session log, file locks). The @pytest.mark.integration marker is present but there is no @pytest.mark.xdist_group serialization guard. Evaluate whether concurrent runs of this test across workers could cause interference.

Trecek (commented):
AutoSkillit review found 11 blocking issues (2 critical, 9 warning). See inline comments on the review. verdict=changes_requested

tests/execution/test_recording.py
Trecek: Investigated — this is intentional. L62 has correct 8-space indentation, placing `RecordingSubprocessRunner(...)` inside the `with patch('atexit.register') as mock_atexit:` block (confirmed via bytecode). `RecordingSubprocessRunner.__init__` (recording.py:59–69) contains no `atexit.register` call, so the test correctly guards against future regressions that would add one. No change needed.

src/autoskillit/server/_lifespan.py
Trecek: Agreed — fixed in 80fa6f7. Wrapped `ctx.runner.recorder.finalize()` in `try/except Exception` with `logger.exception()` so a native extension error during SIGTERM teardown cannot propagate out of the asynccontextmanager and abort cleanup.

src/autoskillit/execution/recording.py
Trecek: Investigated — this is intentional. The post-construction assignment in `build_replay_runner` is a deliberate factory pattern introduced in c7912c6. `build_replay_runner` is the sole production factory (_factory.py:142). All 8+ direct `ReplayingSubprocessRunner({}, {})` constructions in tests intentionally omit `tmp_dir`. Making it a constructor parameter would either break all tests (if required) or preserve None-initialization (if optional) — identical semantics with worse test ergonomics.

tests/execution/test_recording_sigterm.py
Trecek: Agreed — fixed in bad688c. Replaced `time.sleep(1.0)` with `select.select()` line-by-line polling on `proc.stderr` for the `lifespan_started` token — emitted inside the asyncio event loop immediately before `yield` in `_autoskillit_lifespan`. This guarantees the lifespan is active (not just the signal handler installed) before SIGTERM is sent. Falls back to a 5-second ceiling for CI resilience.

src/autoskillit/server/_lifespan.py
Trecek: Agreed — fixed in 80fa6f7. Added `from typing import Any` and annotated `server: Any` to satisfy typed L3 boundaries. mypy pre-commit hook passes with the annotation.

src/autoskillit/cli/app.py
Trecek: Investigated — this is intentional. `mcp.run()` uses `anyio.run()` → `asyncio.Runner` → `loop.run_until_complete()`. Python delivers signals between bytecodes in the main thread at the C-level eval breaker checkpoint, so `KeyboardInterrupt` surfaces cleanly at the `run_until_complete()` boundary — not inside asyncio internals. `loop.call_soon_threadsafe()` is only the correct remedy when the signal handler runs in a *different thread*. Here both run in the main thread. The pattern was deliberately chosen in a69221d for clean FastMCP lifespan teardown, and `tests/cli/test_serve_sigterm.py:41-42` explicitly validates this contract. No change needed.

tests/execution/test_recording_sigterm.py
Trecek: Agreed — fixed in a486bbd. Switched from `proc.wait()` + manual reads to `proc.communicate()` to safely drain both stdout and stderr without pipe-buffer deadlock. Added captured stdout and stderr to the assertion failure message so CI failures expose subprocess output directly.

tests/cli/test_serve_sigterm.py
Trecek: Investigated — this is intentional. All monkeypatches target live object attributes (`autoskillit.server.mcp.run`, `autoskillit.config.load_config`, etc.), not module-level name bindings. `serve()` does deferred imports from `autoskillit.server` inside the function body, so it always gets the live module with the patches applied. Cached module state is irrelevant here. The same import-inside-test pattern is used consistently across `test_cook.py` and `test_cook_interactive.py`. No change needed.

src/autoskillit/server/__init__.py
Trecek: Agreed — fixed in e6ca883. Removed `# noqa: F401` from the `_autoskillit_lifespan` import line. Confirmed via ruff that no F401 fires without the comment since the symbol is actively used on line 38 (`lifespan=_autoskillit_lifespan`).

src/autoskillit/execution/recording.py
Trecek: Investigated — this is intentional. Same factory pattern as the related comment on line 288. The class-level annotation and `self._tmp_replay_dir = None` in `__init__` were added in c7912c6 as the typed lifecycle contract for a test-constructible lightweight class with a separate production factory. No change needed.

src/autoskillit/server/__init__.py
Trecek: Agreed — fixed in e6ca883. Verified via grep that nothing imports `_get_ctx_or_none` from `autoskillit.server` (package level) — all callers import directly from `_state`, `helpers`, or `_lifespan`. Removed from the re-export block in `__init__.py`.

tests/execution/test_recording_sigterm.py
Trecek: Valid observation — flagged for design decision. The test uses `tmp_path` for filesystem isolation (unique per xdist worker). The main shared-state risk is the XDG session log (`~/.local/share/autoskillit/logs/sessions.jsonl`). Test assertions don't read that file, so interference is unlikely to cause test failures directly. Recommend evaluating `@pytest.mark.xdist_group('integration')` defensively if session-log WAL contention becomes a concern in CI.

Issue #756, opened 2026-04-12
clone_registry.batch_delete never prunes the registry after successful deletion


Trecek:
## Problem

`clone_registry.batch_delete` at `src/autoskillit/workspace/clone_registry.py:88-107` reads registered clones, deletes them on disk, and returns a summary — but never writes the registry file back. Entries accumulate indefinitely.

Live evidence: `.autoskillit/temp/clone-cleanup-registry.json` currently holds 67 entries, oldest from 2026-04-05. Every `batch_cleanup_clones` invocation re-scans all of them.

## Symptoms

1. **Noisy output**: subsequent runs report 60+ `delete_failures` with `reason: not_found` for clones deleted days/weeks ago. Looks like failures; is actually cosmetic noise.
2. **Misleading `deleted: []`**: when called after a prior cleanup has already emptied the disk, the tool reports `deleted: []` with a huge `delete_failures` list. Easy to misread as "cleanup did nothing." This actually happened during an investigation where a user thought three clones had been deleted without permission because they were looking at a second cleanup call's `deleted: []` output instead of the first call's (which had already done the work 5.5 hours earlier in a parallel orchestrator session).
3. **Latent cross-session hazard**: the registry is shared across all orchestrator sessions via `_resolve_registry_path` (`clone_registry.py:27-31`). A session calling `batch_cleanup_clones` will attempt to delete entries registered by other in-flight sessions. Not observed to cause loss yet, but dangerous if parallel orchestrators become common. (Session-scoped ownership tagging is a larger change and is being filed as a separate issue.)

## Evidence

Session logs from 2026-04-12 show two parallel `batch_cleanup_clones` calls against the same registry:

- Session `927d5724-877f-492e-b780-990643537619` at `2026-04-12T09:29:56Z`: `deleted: [8 items]`, `delete_failures: [57 not_found entries from 2026-04-05 onward]`
- Session `2ae5a3dc-dd06-43b3-95c1-9be9e697083d` at `2026-04-12T14:53:47Z`: `deleted: []`, `delete_failures: [60+ not_found entries including the 8 the first call just deleted]`

The second call's `delete_failures` list contained the entries the first call had just deleted, proving `batch_delete` never rewrote the registry file after the first call.

## Root Cause

`src/autoskillit/workspace/clone_registry.py:88-107`:

```python
def batch_delete(
    registry_path: str,
    remove_fn: Callable[[str, str], dict[str, str]],
    temp_dir: Path | None = None,
[...]

PR #761, opened 2026-04-12
Implementation Plan: Clone Cleanup Registry — Session-Scoped Ownership Tagging

### LINKED ISSUE #757: clone cleanup registry needs session-scoped ownership tagging for safe parallel orchestration
Related to #756 (`batch_delete` pruning bug). Filing separately because this is a larger architectural change and should be planned independently.

## Problem

The clone cleanup registry at `src/autoskillit/workspace/clone_registry.py` is a single shared file (`<temp_dir>/clone-cleanup-registry.json`) that every orchestrator session reads and writes, regardless of which session owns which clone. The registry entry schema today is:

```json
{"clones": [{"path": "/abs/path", "status": "success|error"}]}
```

There is no notion of which session registered which entry. When any session calls `batch_cleanup_clones`, it iterates every `success`-status entry in the registry and calls `remove_fn` on every clone path — including clones that belong to a different, still-running orchestrator session.

## Why It Matters

Today, `batch_cleanup_clones` is only called at the end of a single orchestrator's pipeline, and the registry is effectively used for intra-pipeline (parallel pipeline-within-session) cleanup coordination. PR #493 ("Defer Clone Cleanup Until All Parallel Pipelines Complete") introduced the pattern for this use case.

But as soon as two orchestrator sessions run in parallel against the same project temp dir:

- Session A registers clone `/runs/impl-foo-123` as `success` and starts working on it.
- Session B (a different orchestrator, started later) finishes its own work and calls `batch_cleanup_clones`.
- Session B's cleanup call sees session A's `/runs/impl-foo-123` entry, sees status `success`, and deletes it out from under session A.

On 2026-04-12 two parallel orchestrator sessions were confirmed via session logs:

- Session `927d5724-877f-492e-b780-990643537619` (remediation orchestrator, started 04:34Z)
- Session `2ae5a3dc-dd06-43b3-95c1-9be9e697083d` (impl orchestrator, started 05:31Z)

Both wrote into the same `clone-cleanup-registry.json`. Both called `batch_cleanup_clones` against the same pool of entries. No loss was observed on this run because their timings did not cross — session 927d5724 cleaned up at 09:29Z, session 2ae5a3dc cleaned up at 14:53Z, and the former had already finished all its clone work before the latter's cleanup fired. But the race window is real and will eventually be hit.

## Design Options (to be decided during planning)

1. **Owner = session UUID from Claude Code session ID**. Requires the clone registry to have access to the session ID. Session ID is available to MCP tools via `ctx` (Context) — the tool already uses `_get_ctx()`, and the session ID is stored on the ToolContext. This is the cleanest path.

2. **Owner = kitchen_id (from open_kitchen)**. Also unique per session. Already propagated through `ToolContext`. Slightly coarser than session UUID but guaranteed unique per orchestrator instance.

3. **Owner = caller-provided opaque string**. `register_clone` and `batch_cleanup_clones` accept an `owner` argument; the recipe passes the same constant throughout. Most flexible but relies on recipe discipline.

Recommendation: (2) kitchen_id. It's already unique per orchestrator run, already flows through every MCP tool call, and has a natural "all work under this kitchen" scope semantics.

## Test Plan

- Unit: simulate two owners registering into the same registry; call `batch_cleanup_clones` as owner A; assert only A's entries are deleted and B's remain untouched.
- Unit: call `batch_cleanup_clones` with explicit "all owners" flag; assert all eligible entries across owners are considered.
- Integration: spin up two parallel orchestrator sessions (kitchen A and kitchen B) sharing the same project temp dir; verify that A's cleanup leaves B's registered clones intact.
- Backward-compat: feed a pre-migration registry (no owner field) into `batch_cleanup_clones`; assert the decided legacy behavior (treat as current session owned OR skip — depending on REQ-REG-006 decision).

## Out of Scope

- Pruning `batch_delete` to rewrite the registry after successful deletion — filed as #756.
- Cross-host / remote cleanup coordination — this proposal is local-filesystem only.

## Affected Files

- `src/autoskillit/workspace/clone_registry.py` — schema, `register_clone`, `cleanup_candidates`, `batch_delete`
- `src/autoskillit/server/tools_clone.py` — `register_clone_status`, `batch_cleanup_clones` — plumb owner from `ToolContext`
- `src/autoskillit/pipeline/context.py` — ensure `kitchen_id` / session UUID is reachable
- `tests/workspace/test_clone_registry.py` — new owner-scoping tests
- `tests/server/test_tools_clone.py` — new owner-scoping tests

## Requirements

**REQ-REG-001:** The registry entry schema must include an owner field identifying which orchestrator session registered the entry.

**REQ-REG-002:** The owner identifier must uniquely distinguish concurrent orchestrator sessions on the same host.

**REQ-REG-003:** `register_clone` must populate the owner field at registration time, sourced from the current session context.

**REQ-REG-004:** `cleanup_candidates` must accept an owner filter and return only entries matching that owner, defaulting to the current session's owner.

**REQ-REG-005:** `batch_cleanup_clones` must scope deletion to the caller's owner by default.

**REQ-REG-006:** The schema change must handle existing registry files whose entries lack an owner field without corrupting or losing data.

**REQ-REG-007:** An explicit "clean up orphaned entries from all owners" escape hatch must exist for operator recovery.

Trecek:
## Summary

The clone cleanup registry (`src/autoskillit/workspace/clone_registry.py`) is a single shared JSON file that every orchestrator session reads and writes. Today, entries carry only `{path, status}`. When any session calls `batch_cleanup_clones`, it deletes every `status="success"` entry — including clones owned by other running sessions. Two parallel orchestrator sessions on 2026-04-12 exposed the race window (sessions `927d5724…` and `2ae5a3dc…` both wrote into the same `clone-cleanup-registry.json`).

This plan extends the registry entry schema with an `owner` field (populated from `ToolContext.kitchen_id` at registration time), threads an owner filter through `register_clone`, `cleanup_candidates`, and `batch_delete`, updates the two server tool handlers (`register_clone_status`, `batch_cleanup_clones`) to source the owner from `tool_ctx.kitchen_id`, and adds an explicit `all_owners="true"` escape hatch on `batch_cleanup_clones` for operator recovery. Legacy entries without an `owner` field are classified as orphans — invisible to owner-scoped calls, deletable only via the escape hatch. Registry pruning after successful deletion is explicitly out of scope (tracked in #756).

## Requirements

**REQ-REG-001:** The registry entry schema must include an owner field identifying which orchestrator session registered the entry.

**REQ-REG-002:** The owner identifier must uniquely distinguish concurrent orchestrator sessions on the same host.

**REQ-REG-003:** `register_clone` must populate the owner field at registration time, sourced from the current session context.

**REQ-REG-004:** `cleanup_candidates` must accept an owner filter and return only entries matching that owner, defaulting to the current session's owner.

**REQ-REG-005:** `batch_cleanup_clones` must scope deletion to the caller's owner by default.

**REQ-REG-006:** The schema change must handle existing registry files whose entries lack an owner field without corrupting or losing data.

**REQ-REG-007:** An explicit "clean up orphaned entries from all owners" escape hatch must exist for operator recovery.

### State Lifecycle Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINALS %%
    REG_IN([● register_clone_status])
    CLEANUP_IN([● batch_cleanup_clones])

    subgraph ValidationGates ["VALIDATION GATES (Registration Path)"]
        direction TB
        GATE_KITCHEN["● kitchen_id gate<br/>━━━━━━━━━━<br/>kitchen_id == '' → error<br/>No active kitchen"]
        GATE_STATUS["status gate<br/>━━━━━━━━━━<br/>must be 'success'|'error'<br/>FAIL-FAST"]
        GATE_OWNER["● owner validation<br/>━━━━━━━━━━<br/>owner == '' → ValueError<br/>register_clone rejects empty"]
    end

    subgraph LockWrite ["CONCURRENT WRITE SAFETY"]
        direction TB
        FLOCK["fcntl.LOCK_EX<br/>━━━━━━━━━━<br/>Exclusive advisory lock<br/>Read-modify-write atomic"]
    end

    subgraph RegistryEntry ["REGISTRY ENTRY (INIT_ONLY fields)"]
        direction LR
        FIELD_PATH["path<br/>━━━━━━━━━━<br/>INIT_ONLY<br/>Clone dir — never changes"]
        FIELD_STATUS["status<br/>━━━━━━━━━━<br/>'success' | 'error'<br/>INIT_ONLY — terminal outcome"]
        FIELD_OWNER["● owner<br/>━━━━━━━━━━<br/>kitchen_id<br/>INIT_ONLY — session tag"]
    end

    subgraph LegacyPath ["LEGACY ENTRY (no owner field)"]
        LEGACY["orphan entry<br/>━━━━━━━━━━<br/>No 'owner' key<br/>Pre-owner schema"]
    end

    subgraph CleanupScope ["● OWNER SCOPE RESOLUTION"]
        direction TB
        SCOPE_CHECK{"all_owners<br/>== 'true'?"}
        SCOPE_KITCHEN["owner_filter = kitchen_id<br/>━━━━━━━━━━<br/>Default: scoped to session<br/>Other kitchen entries invisible"]
        SCOPE_ALL["owner_filter = None<br/>━━━━━━━━━━<br/>Escape hatch<br/>All entries including legacy"]
        GATE_EMPTY_K["● kitchen_id empty gate<br/>━━━━━━━━━━<br/>kitchen_id == '' and<br/>not escape hatch → error"]
    end

    subgraph CleanupFilter ["● _entry_matches_owner FILTER"]
        direction TB
        FILTER_FN["_entry_matches_owner(entry, owner)<br/>━━━━━━━━━━<br/>owner=None → True (pass all)<br/>owner=X → entry.get('owner')==X"]
        PARTITION["cleanup_candidates partition<br/>━━━━━━━━━━<br/>to_delete: status==success ∧ match<br/>to_preserve: status==error ∧ match"]
    end

    subgraph RecipeValidation ["● RECIPE VALIDATOR (rules_tools.py)"]
        RULE_PARAMS["dead-with-param rule<br/>━━━━━━━━━━<br/>batch_cleanup_clones params:<br/>{registry_path, all_owners, step_name}"]
    end

    RESULT_DELETE(["remove_clone() × N<br/>MUTABLE: deleted[]<br/>DERIVED: to_delete list"])
    RESULT_PRESERVE(["preserved on disk<br/>error-status clones<br/>manual investigation"])

    %% REGISTRATION FLOW %%
    REG_IN --> GATE_KITCHEN
    GATE_KITCHEN -->|kitchen_id valid| GATE_STATUS
    GATE_KITCHEN -->|kitchen_id empty| GATE_KITCHEN_ERR["registered=false<br/>no active kitchen"]
    GATE_STATUS -->|valid| GATE_OWNER
    GATE_OWNER -->|owner non-empty| FLOCK
    FLOCK --> FIELD_PATH
    FLOCK --> FIELD_STATUS
    FLOCK --> FIELD_OWNER

    %% CLEANUP FLOW %%
    CLEANUP_IN --> SCOPE_CHECK
    SCOPE_CHECK -->|"false (default)"| GATE_EMPTY_K
    SCOPE_CHECK -->|true| SCOPE_ALL
    GATE_EMPTY_K -->|kitchen_id present| SCOPE_KITCHEN
    GATE_EMPTY_K -->|kitchen_id empty| CLEANUP_ERR["deleted=[], error:<br/>pass all_owners='true'<br/>to force-cleanup"]
    SCOPE_KITCHEN --> FILTER_FN
    SCOPE_ALL --> FILTER_FN
    LEGACY -.->|visible only via owner=None| FILTER_FN
    FIELD_OWNER -.->|read back at cleanup| FILTER_FN
    FILTER_FN --> PARTITION
    PARTITION --> RESULT_DELETE
    PARTITION --> RESULT_PRESERVE

    %% RECIPE VALIDATOR READS PARAM TABLE %%
    CLEANUP_IN -.->|recipe YAML validation| RULE_PARAMS

    %% CLASS ASSIGNMENTS %%
    class REG_IN,CLEANUP_IN terminal;
    class GATE_KITCHEN,GATE_OWNER detector;
    class GATE_STATUS detector;
    class GATE_EMPTY_K detector;
    class FLOCK phase;
    class FIELD_PATH,FIELD_STATUS stateNode;
    class FIELD_OWNER handler;
    class LEGACY gap;
    class SCOPE_CHECK phase;
    class SCOPE_KITCHEN,SCOPE_ALL output;
    class FILTER_FN,PARTITION phase;
    class RULE_PARAMS stateNode;
    class RESULT_DELETE,RESULT_PRESERVE output;
    class GATE_KITCHEN_ERR,CLEANUP_ERR gap;
```

### Concurrency Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% ── ENTRY ── %%
    START([FastMCP Event Loop<br/>receives tool call])

    %% ── ASYNC LAYER ── %%
    subgraph AsyncLayer ["ASYNC EVENT LOOP  (single-threaded, per kitchen process)"]
        direction TB
        CTX_RESET["● clear_contextvars()<br/>━━━━━━━━━━<br/>thread-local log reset<br/>per invocation"]
        KITCHEN_GUARD{"kitchen_id<br/>empty?"}
        ERROR_NO_KIT["return error<br/>━━━━━━━━━━<br/>no active kitchen"]
        OWNERS_CHECK{"all_owners<br/>== 'true'?"}
        DISPATCH_REG["await asyncio.to_thread(<br/>━━━━━━━━━━<br/>register_clone, owner=kitchen_id)"]
        DISPATCH_CLEAN["await asyncio.to_thread(<br/>━━━━━━━━━━<br/>batch_delete, owner_filter)"]
        BARRIER["await result<br/>━━━━━━━━━━<br/>event loop resumes here"]
    end

    %% ── THREAD POOL ── %%
    subgraph ThreadPool ["DEFAULT ThreadPoolExecutor  (asyncio.to_thread workers)"]
        direction LR
        REG_WORKER["● register_clone()<br/>━━━━━━━━━━<br/>owner=kitchen_id<br/>validates non-empty"]
        CLEAN_WORKER["● batch_delete()<br/>━━━━━━━━━━<br/>owner_filter scoped<br/>or None (all-owners)"]
    end

    %% ── FILE LOCK ── %%
    subgraph FileLock ["CROSS-PROCESS MUTEX  (fcntl.LOCK_EX on .lock sidecar)"]
        direction TB
        LOCK_ACQ["● fcntl.LOCK_EX<br/>━━━━━━━━━━<br/>.lock sidecar file<br/>blocks concurrent writers"]
        READ_REG["read registry JSON<br/>━━━━━━━━━━<br/>full snapshot while locked"]
        APPEND["append entry<br/>━━━━━━━━━━<br/>{path, status, owner}"]
        ATOMIC["● atomic_write<br/>━━━━━━━━━━<br/>tmp → rename<br/>crash-safe"]
        LOCK_REL["fcntl.LOCK_UN<br/>━━━━━━━━━━<br/>next writer unblocked"]
    end

    %% ── OWNER SCOPING ── %%
    subgraph OwnerScope ["OWNER SCOPING  (session isolation layer)"]
        direction TB
        FILTER["● _entry_matches_owner()<br/>━━━━━━━━━━<br/>entry.owner == kitchen_id<br/>legacy entries hidden"]
        ESCAPE["● all_owners='true'<br/>━━━━━━━━━━<br/>operator escape hatch<br/>bypasses owner filter"]
        SPLIT["partition to_delete<br/>━━━━━━━━━━<br/>status==success → delete<br/>status==error → preserve"]
    end

    %% ── PARALLEL KITCHENS (cross-process) ── %%
    subgraph Parallel ["PARALLEL KITCHEN PROCESSES  (each runs identical flow above)"]
        direction LR
        KIT_A["Kitchen A<br/>━━━━━━━━━━<br/>kitchen_id=kit-A<br/>own registry entries only"]
        KIT_B["Kitchen B<br/>━━━━━━━━━━<br/>kitchen_id=kit-B<br/>own registry entries only"]
    end

    %% ── REGISTRY ── %%
    REGISTRY[("● clone-cleanup-registry.json<br/>━━━━━━━━━━<br/>shared across all kitchens<br/>entries tagged by owner")]

    %% ── END ── %%
    END([return result to MCP caller])

    %% ── FLOWS ── %%
    START --> CTX_RESET
    CTX_RESET --> KITCHEN_GUARD
    KITCHEN_GUARD -->|"empty"| ERROR_NO_KIT
    KITCHEN_GUARD -->|"register path"| DISPATCH_REG
    KITCHEN_GUARD -->|"cleanup path"| OWNERS_CHECK
    OWNERS_CHECK -->|"no (default)"| DISPATCH_CLEAN
    OWNERS_CHECK -->|"yes"| ESCAPE

    DISPATCH_REG --> REG_WORKER
    DISPATCH_CLEAN --> CLEAN_WORKER
    ESCAPE --> CLEAN_WORKER

    REG_WORKER --> LOCK_ACQ
    LOCK_ACQ --> READ_REG --> APPEND --> ATOMIC --> LOCK_REL
    LOCK_REL --> BARRIER

    CLEAN_WORKER --> FILTER
    FILTER --> SPLIT
    SPLIT --> BARRIER

    BARRIER --> END

    %% cross-process parallel writers serialize on lock
    KIT_A -.->|"concurrent register"| LOCK_ACQ
    KIT_B -.->|"blocked until lock free"| LOCK_ACQ

    %% registry is shared state
    ATOMIC -->|"write"| REGISTRY
    FILTER -->|"read snapshot"| REGISTRY

    %% class assignments
    class START,END terminal;
    class CTX_RESET,KITCHEN_GUARD,OWNERS_CHECK stateNode;
    class ERROR_NO_KIT gap;
    class DISPATCH_REG,DISPATCH_CLEAN,BARRIER phase;
    class REG_WORKER,CLEAN_WORKER handler;
    class LOCK_ACQ,LOCK_REL detector;
    class READ_REG,APPEND phase;
    class ATOMIC output;
    class FILTER,ESCAPE,SPLIT newComponent;
    class REGISTRY stateNode;
    class KIT_A,KIT_B terminal;
```

### Module Dependency Diagram

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 55, 'rankSpacing': 80, 'curve': 'basis'}}}%%
graph TB
    %% CLASS DEFINITIONS %%

    subgraph L3 ["L3 — SERVER LAYER"]
        direction LR
        TOOLS_CLONE["● server/tools_clone.py<br/>━━━━━━━━━━<br/>MCP tools: clone_repo,<br/>remove_clone, register_clone_status,<br/>batch_cleanup_clones<br/>(plumbs kitchen_id as owner)"]
        HELPERS["server/helpers.py<br/>━━━━━━━━━━<br/>Facade + re-export bridge<br/>clone_registry re-exported<br/>for tools_clone.py"]
    end

    subgraph L2 ["L2 — RECIPE VALIDATION LAYER"]
        direction LR
        RULES_TOOLS["● recipe/rules_tools.py<br/>━━━━━━━━━━<br/>Semantic rules: unknown-tool,<br/>subset-disabled-tool,<br/>dead-with-param<br/>(registers all_owners param)"]
        RECIPE_REGISTRY["recipe/registry.py<br/>━━━━━━━━━━<br/>RuleFinding, semantic_rule<br/>_RULE_REGISTRY, run_semantic_rules"]
        RECIPE_ANALYSIS["recipe/_analysis.py<br/>━━━━━━━━━━<br/>ValidationContext<br/>make_validation_context"]
    end

    subgraph L1 ["L1 — WORKSPACE LAYER"]
        direction LR
        CLONE_REGISTRY["● workspace/clone_registry.py<br/>━━━━━━━━━━<br/>register_clone(owner=...)<br/>cleanup_candidates(owner=...)<br/>batch_delete(owner=...)<br/>File-locked JSON registry"]
    end

    subgraph L0 ["L0 — CORE FOUNDATION"]
        direction LR
        CORE_CONSTANTS["core/_type_constants.py<br/>━━━━━━━━━━<br/>GATED_TOOLS, HEADLESS_TOOLS<br/>UNGATED_TOOLS, TOOL_SUBSET_TAGS<br/>Severity"]
        CORE_IO["core/io.py + logging.py<br/>━━━━━━━━━━<br/>atomic_write, resolve_temp_dir<br/>get_logger"]
    end

    %% L3 → L3 (same-layer: tools_clone → helpers) %%
    TOOLS_CLONE -->|"imports: _notify, _require_enabled,<br/>clone_registry, track_response_size"| HELPERS

    %% L3 → L0 %%
    TOOLS_CLONE -->|"get_logger"| CORE_IO

    %% L3 → L1 (via helpers re-export bridge) %%
    HELPERS -->|"re-exports clone_registry<br/>(noqa: F401)"| CLONE_REGISTRY

    %% L3 helpers → L0 %%
    HELPERS -->|"atomic_write, get_logger<br/>resolve_temp_dir"| CORE_IO

    %% L2 → L2 (same-layer) %%
    RULES_TOOLS -->|"RuleFinding, semantic_rule"| RECIPE_REGISTRY
    RULES_TOOLS -->|"ValidationContext"| RECIPE_ANALYSIS
    RECIPE_REGISTRY -->|"ValidationContext<br/>make_validation_context"| RECIPE_ANALYSIS

    %% L2 → L0 %%
    RULES_TOOLS -->|"GATED_TOOLS, HEADLESS_TOOLS<br/>UNGATED_TOOLS, TOOL_SUBSET_TAGS<br/>Severity"| CORE_CONSTANTS
    RECIPE_REGISTRY -->|"Severity"| CORE_CONSTANTS

    %% L1 → L0 %%
    CLONE_REGISTRY -->|"atomic_write, get_logger<br/>resolve_temp_dir"| CORE_IO

    %% CLASS ASSIGNMENTS %%
    class TOOLS_CLONE cli;
    class HELPERS handler;
    class RULES_TOOLS,RECIPE_REGISTRY,RECIPE_ANALYSIS phase;
    class CLONE_REGISTRY stateNode;
    class CORE_CONSTANTS,CORE_IO integration;
```

Closes #757

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260412-092010-886732/.autoskillit/temp/make-plan/clone_registry_owner_scoping_plan_2026-04-12_092354.md`

## Token Usage Summary

| plan | 215 | 23.3k | 1.3M | 144.1k | 1 | 5m 40s |
| verify | 150 | 16.7k | 909.8k | 76.7k | 1 | 4m 31s |
| implement | 2.5k | 32.9k | 3.7M | 93.0k | 1 | 8m 45s |
| fix | 174 | 5.7k | 838.5k | 41.4k | 1 | 5m 46s |
| prepare_pr | 86 | 5.0k | 246.1k | 34.7k | 1 | 1m 32s |
| run_arch_lenses | 3.6k | 22.0k | 735.1k | 101.4k | 3 | 7m 48s |
| compose_pr | 59 | 7.7k | 198.4k | 29.7k | 1 | 1m 57s |
| **Total** | 6.8k | 113.2k | 7.9M | 521.0k | | 36m 2s |

### DIFF
--- a/src/autoskillit/server/tools_clone.py
+++ b/src/autoskillit/server/tools_clone.py
@@ -298,7 +298,12 @@ async def register_clone_status(
     logger.info("register_clone_status", clone_path=clone_path, status=status)
 
     if status not in ("success", "error"):
-        return json.dumps({"error": f"Invalid status '{status}'. Must be 'success' or 'error'."})
+        return json.dumps(
+            {
+                "registered": "false",
+                "reason": f"Invalid status '{status}'. Must be 'success' or 'error'.",
+            }
+        )
 
     from autoskillit.server import _get_ctx
 
--- a/src/autoskillit/server/tools_clone.py
+++ b/src/autoskillit/server/tools_clone.py
@@ -303,6 +303,14 @@ async def register_clone_status(
     from autoskillit.server import _get_ctx
 
     tool_ctx = _get_ctx()
+    owner = tool_ctx.kitchen_id
+    if owner == "":
+        return json.dumps(
+            {
+                "registered": "false",
+                "reason": "no active kitchen (kitchen_id empty) — cannot register clone",
+            }
+        )
     _start = time.monotonic()
     typed_status: Literal["success", "error"] = "success" if status == "success" else "error"
     try:
@@ -310,6 +318,7 @@ async def register_clone_status(
             clone_registry.register_clone,
             clone_path,
             typed_status,
+            owner,
             registry_path,
             tool_ctx.temp_dir,
         )
@@ -326,6 +335,7 @@ async def register_clone_status(
 @track_response_size("batch_cleanup_clones")
 async def batch_cleanup_clones(
     registry_path: str = "",
+    all_owners: str = "false",
     step_name: str = "",
     ctx: Context = CurrentContext(),
 ) -> str:
@@ -334,11 +344,18 @@ async def batch_cleanup_clones(
     Called ONCE after all parallel pipelines have completed (after batch_confirm_cleanup
     action: confirm). Never raises — failures are reported in the result dict.
 
+    By default, deletion is scoped to the current kitchen's entries (owner = kitchen_id).
+    Pass all_owners="true" as an operator escape hatch to delete all success-status entries
+    regardless of owner, including legacy orphan entries from pre-owner-schema registries.
+
     Returns {"deleted": [...], "delete_failures": [...], "preserved": [...]}.
 
     Args:
         registry_path: Absolute path to the shared registry file. Defaults to
                        .autoskillit/temp/clone-cleanup-registry.json in cwd.
+        all_owners: "false" (default) scopes deletion to the current kitchen's entries;
+                    "true" is the operator escape hatch that ignores owner and deletes
+                    all success-status entries including legacy orphans.
         step_name: Optional YAML step key for wall-clock timing accumulation.
     """
     try:
@@ -346,7 +363,7 @@ async def batch_cleanup_clones(
             return gate
         structlog.contextvars.clear_contextvars()
         structlog.contextvars.bind_contextvars(tool="batch_cleanup_clones")
-        logger.info("batch_cleanup_clones", registry_path=registry_path)
+        logger.info("batch_cleanup_clones", registry_path=registry_path, all_owners=all_owners)
 
         from autoskillit.server import _get_ctx
 
@@ -361,6 +378,23 @@ async def batch_cleanup_clones(
                 }
             )
 
+        is_escape_hatch = all_owners.strip().lower() == "true"
+        owner_filter: str | None
+        if is_escape_hatch:
+            owner_filter = None
+        else:
+            owner_filter = tool_ctx.kitchen_id
+            if owner_filter == "":
+                return json.dumps(
+                    {
+                        "deleted": [],
+                        "delete_failures": [],
+                        "preserved": [],
+                        "error": "no active kitchen (kitchen_id empty) — "
+                        "pass all_owners='true' to force-cleanup legacy entries",
+                    }
+                )
+
         _start = time.monotonic()
         try:
             result = await asyncio.to_thread(
@@ -368,6 +402,7 @@ async def batch_cleanup_clones(
                 registry_path,
                 tool_ctx.clone_mgr.remove_clone,
                 tool_ctx.temp_dir,
+                owner_filter,
             )
             return json.dumps(result)
         finally:
--- a/src/autoskillit/workspace/clone_registry.py
+++ b/src/autoskillit/workspace/clone_registry.py
@@ -5,7 +5,7 @@ After all pipelines complete, batch_cleanup_clones reads the registry and
 removes only success-status clones.
 
 Registry file format:
-    {"clones": [{"path": "/abs/path/to/clone", "status": "success|error"}]}
+    {"clones": [{"path": "/abs/path/to/clone", "status": "success|error", "owner": "kitchen-uuid"}]}
 """
 
 from __future__ import annotations
@@ -31,14 +31,23 @@ def _resolve_registry_path(registry_path: str, temp_dir: Path | None = None) -> 
     return base / _DEFAULT_REGISTRY_NAME
 
 
+def _entry_matches_owner(entry: dict[str, str], owner: str | None) -> bool:
+    if owner is None:
+        return True
+    return entry.get("owner") == owner
+
+
 def register_clone(
     clone_path: str,
     status: CloneStatus,
+    owner: str,
     registry_path: str = "",
     temp_dir: Path | None = None,
 ) -> dict[str, str]:
     """Append a clone entry to the registry. Safe for parallel callers — holds an
     exclusive advisory lock across the entire read-modify-write sequence."""
+    if owner == "":
+        raise ValueError("owner is required")
     path = _resolve_registry_path(registry_path, temp_dir)
     lock_path = path.with_suffix(".lock")
     lock_path.parent.mkdir(parents=True, exist_ok=True)
@@ -50,7 +59,7 @@ def register_clone(
                 existing = json.loads(path.read_text()).get("clones", [])
             except (json.JSONDecodeError, OSError) as exc:
                 _log.warning("clone_registry: could not read %s: %s", path, exc)
-        existing.append({"path": clone_path, "status": status})
+        existing.append({"path": clone_path, "status": status, "owner": owner})
         atomic_write(path, json.dumps({"clones": existing}, indent=2))
     return {"registered": "true", "registry_path": str(path)}
 
@@ -73,15 +82,28 @@ def read_registry(
 def cleanup_candidates(
     registry_path: str = "",
     temp_dir: Path | None = None,
+    owner: str | None = None,
 ) -> tuple[list[str], list[str]]:
     """Return (to_delete, to_preserve) path lists from the registry.
 
     to_delete  — clones with status='success' (safe to remove)
     to_preserve — clones with status='error'  (preserve for investigation)
+
+    When owner is provided, only entries belonging to that owner are considered.
+    Entries without an 'owner' field (legacy orphans) are only visible when
+    owner is None (all-owners mode).
     """
     entries = read_registry(registry_path, temp_dir)
-    to_delete = [e["path"] for e in entries if e.get("status") == "success" and "path" in e]
-    to_preserve = [e["path"] for e in entries if e.get("status") == "error" and "path" in e]
+    to_delete = [
+        e["path"]
+        for e in entries
+        if e.get("status") == "success" and "path" in e and _entry_matches_owner(e, owner)
+    ]
+    to_preserve = [
+        e["path"]
+        for e in entries
+        if e.get("status") == "error" and "path" in e and _entry_matches_owner(e, owner)
+    ]
     return to_delete, to_preserve
 
 
@@ -89,13 +111,17 @@ def batch_delete(
     registry_path: str,
     remove_fn: Callable[[str, str], dict[str, str]],
     temp_dir: Path | None = None,
+    owner: str | None = None,
 ) -> dict[str, Any]:
     """Read registry and delete success-status clones via remove_fn.
 
     Calls remove_fn(path, "false") for each success clone. Error clones are
     preserved. Returns {"deleted": [...], "delete_failures": [...], "preserved": [...]}.
+
+    When owner is provided, only entries belonging to that owner are considered.
+    Pass owner=None to operate on all entries (escape hatch mode).
     """
-    to_delete, to_preserve = cleanup_candidates(registry_path, temp_dir)
+    to_delete, to_preserve = cleanup_candidates(registry_path, temp_dir, owner=owner)
     deleted: list[str] = []
     delete_failures: list[dict[str, str]] = []
     for path in to_delete:
--- a/src/autoskillit/workspace/clone_registry.py
+++ b/src/autoskillit/workspace/clone_registry.py
@@ -5,7 +5,7 @@ After all pipelines complete, batch_cleanup_clones reads the registry and
 removes only success-status clones.
 
 Registry file format:
-    {"clones": [{"path": "/abs/path/to/clone", "status": "success|error", "owner": "kitchen-uuid"}]}
+    {"clones": [{"path": "/abs/path", "status": "success|error", "owner": "kitchen-uuid"}]}
 """
 
 from __future__ import annotations
--- a/tests/server/test_tools_clone.py
+++ b/tests/server/test_tools_clone.py
@@ -273,6 +273,7 @@ class TestRegisterCloneStatusTool:
     @pytest.mark.anyio
     async def test_register_clone_status_success(self, tool_ctx, tmp_path):
         """register_clone_status status='success' writes registry and returns registered=true."""
+        tool_ctx.kitchen_id = "kit-test"
         registry_path = str(tmp_path / "registry.json")
         result = json.loads(
             await register_clone_status(
@@ -287,6 +288,7 @@ class TestRegisterCloneStatusTool:
     @pytest.mark.anyio
     async def test_register_clone_status_error(self, tool_ctx, tmp_path):
         """register_clone_status status='error' writes registry and returns registered=true."""
+        tool_ctx.kitchen_id = "kit-test"
         registry_path = str(tmp_path / "registry.json")
         result = json.loads(
             await register_clone_status(
@@ -318,13 +320,14 @@ class TestBatchCleanupClonesTool:
     @pytest.mark.anyio
     async def test_batch_cleanup_clones_deletes_success_preserves_error(self, tool_ctx, tmp_path):
         """batch_cleanup_clones removes success clones, skips error clones."""
+        tool_ctx.kitchen_id = "kit-test"
         registry_path = str(tmp_path / "registry.json")
         success_path = str(tmp_path / "success_clone")
         error_path = str(tmp_path / "error_clone")
 
         # Pre-populate registry with one success and one error entry
-        clone_registry.register_clone(success_path, "success", registry_path)
-        clone_registry.register_clone(error_path, "error", registry_path)
+        clone_registry.register_clone(success_path, "success", "kit-test", registry_path)
+        clone_registry.register_clone(error_path, "error", "kit-test", registry_path)
 
         # Mock clone_mgr so remove_clone reports success for the success clone
         mock_mgr = MagicMock()
@@ -341,6 +344,7 @@ class TestBatchCleanupClonesTool:
     @pytest.mark.anyio
     async def test_batch_cleanup_clones_empty_registry(self, tool_ctx, tmp_path):
         """batch_cleanup_clones with missing registry returns deleted=[], preserved=[]."""
+        tool_ctx.kitchen_id = "kit-test"
         registry_path = str(tmp_path / "nonexistent.json")
         result = json.loads(await batch_cleanup_clones(registry_path=registry_path))
         assert result == {"deleted": [], "delete_failures": [], "preserved": []}
@@ -348,11 +352,12 @@ class TestBatchCleanupClonesTool:
     @pytest.mark.anyio
     async def test_batch_cleanup_clones_nonexistent_path_does_not_raise(self, tool_ctx, tmp_path):
         """batch_cleanup_clones reports failure gracefully when a success clone path is gone."""
+        tool_ctx.kitchen_id = "kit-test"
         registry_path = str(tmp_path / "registry.json")
         missing_path = str(tmp_path / "gone_clone")
 
         # Register a success clone whose directory does not exist on disk
-        clone_registry.register_clone(missing_path, "success", registry_path)
+        clone_registry.register_clone(missing_path, "success", "kit-test", registry_path)
 
         # Mock clone_mgr to report removal failure (path not found)
         mock_mgr = MagicMock()
@@ -365,3 +370,191 @@ class TestBatchCleanupClonesTool:
         assert result["deleted"] == []
         # Confirm no exception was raised (we received a well-formed result dict)
         assert "error" not in result
+
+
+# ---------------------------------------------------------------------------
+# New tests: T12–T19 (owner-scoping feature for server tools)
+# ---------------------------------------------------------------------------
+
+
+class TestRegisterCloneStatusOwner:
+    """T12–T13: register_clone_status propagates kitchen_id as owner."""
+
+    @pytest.mark.anyio
+    async def test_register_clone_status_propagates_kitchen_id_as_owner(
+        self, tool_ctx, tmp_path
+    ):
+        """T12 — register_clone_status writes entry with owner == kitchen_id."""
+        tool_ctx.kitchen_id = "kit-xyz"
+        reg = str(tmp_path / "registry.json")
+        result = json.loads(
+            await register_clone_status(clone_path="/c", status="success", registry_path=reg)
+        )
+        assert result["registered"] == "true"
+
+        import json as _json
+        from pathlib import Path as _Path
+
+        data = _json.loads(_Path(reg).read_text())
+        assert len(data["clones"]) == 1
+        assert data["clones"][0]["owner"] == "kit-xyz"
+
+    @pytest.mark.anyio
+    async def test_register_clone_status_rejects_when_kitchen_id_empty(
+        self, tool_ctx, tmp_path
+    ):
+        """T13 — register_clone_status returns registered=false when kitchen_id is empty."""
+        tool_ctx.kitchen_id = ""
+        reg = str(tmp_path / "registry.json")
+        result = json.loads(
+            await register_clone_status(clone_path="/c", status="success", registry_path=reg)
+        )
+        assert result["registered"] == "false"
+        assert "kitchen_id" in result["reason"] or "kitchen" in result["reason"]
+        from pathlib import Path as _Path
+
+        assert not _Path(reg).exists()
+
+
+class TestBatchCleanupClonesOwner:
+    """T14–T19: batch_cleanup_clones owner-scoping and escape hatch."""
+
+    @pytest.mark.anyio
+    async def test_batch_cleanup_clones_default_scopes_to_current_kitchen_id(
+        self, tool_ctx, tmp_path
+    ):
+        """T14 — default call only removes current kitchen's clones."""
+        reg = str(tmp_path / "registry.json")
+        clone_registry.register_clone("/clone-A", "success", "kit-A", reg)
+        clone_registry.register_clone("/clone-B", "success", "kit-B", reg)
+
+        tool_ctx.kitchen_id = "kit-A"
+        mock_mgr = MagicMock()
+        mock_mgr.remove_clone.return_value = {"removed": "true"}
+        tool_ctx.clone_mgr = mock_mgr
+
+        result = json.loads(await batch_cleanup_clones(registry_path=reg))
+
+        assert "/clone-A" in result["deleted"]
+        assert "/clone-B" not in result["deleted"]
+        mock_mgr.remove_clone.assert_called_once_with("/clone-A", "false")
+
+    @pytest.mark.anyio
+    async def test_batch_cleanup_clones_all_owners_true_removes_every_success(
+        self, tool_ctx, tmp_path
+    ):
+        """T15 — all_owners='true' escape hatch removes all success entries."""
+        reg = str(tmp_path / "registry.json")
+        clone_registry.register_clone("/clone-A", "success", "kit-A", reg)
+        clone_registry.register_clone("/clone-B", "success", "kit-B", reg)
+
+        tool_ctx.kitchen_id = "kit-A"
+        mock_mgr = MagicMock()
+        mock_mgr.remove_clone.return_value = {"removed": "true"}
+        tool_ctx.clone_mgr = mock_mgr
+
+        result = json.loads(
+            await batch_cleanup_clones(registry_path=reg, all_owners="true")
+        )
+
+        deleted = result["deleted"]
+        assert "/clone-A" in deleted
+        assert "/clone-B" in deleted
+
+    @pytest.mark.anyio
+    async def test_batch_cleanup_clones_empty_kitchen_id_and_all_owners_false_returns_error(
+        self, tool_ctx, tmp_path
+    ):
+        """T16 — empty kitchen_id with default all_owners='false' returns error."""
+        tool_ctx.kitchen_id = ""
+        reg = str(tmp_path / "registry.json")
+        mock_mgr = MagicMock()
+        tool_ctx.clone_mgr = mock_mgr
+
+        result = json.loads(await batch_cleanup_clones(registry_path=reg))
+
+        assert "error" in result
+        assert "kitchen_id" in result["error"] or "kitchen" in result["error"]
+        mock_mgr.remove_clone.assert_not_called()
+
+    @pytest.mark.anyio
+    async def test_batch_cleanup_clones_empty_kitchen_id_with_all_owners_true_succeeds(
+        self, tool_ctx, tmp_path
+    ):
+        """T17 — escape hatch works even when kitchen_id is empty (legacy recovery)."""
+        from pathlib import Path as _Path
+
+        reg_path = tmp_path / "registry.json"
+        reg_path.write_text(
+            __import__("json").dumps(
+                {"clones": [{"path": "/legacy", "status": "success"}]}
+            )
+        )
+
+        tool_ctx.kitchen_id = ""
+        mock_mgr = MagicMock()
+        mock_mgr.remove_clone.return_value = {"removed": "true"}
+        tool_ctx.clone_mgr = mock_mgr
+
+        result = json.loads(
+            await batch_cleanup_clones(registry_path=str(reg_path), all_owners="true")
+        )
+
+        assert "/legacy" in result["deleted"]
+        mock_mgr.remove_clone.assert_called_once_with("/legacy", "false")
+
+    @pytest.mark.anyio
+    async def test_batch_cleanup_clones_invalid_all_owners_literal_treated_as_false(
+        self, tool_ctx, tmp_path
+    ):
+        """T18 — all_owners='yes' is not the escape hatch; scoped behaviour applies."""
+        reg = str(tmp_path / "registry.json")
+        clone_registry.register_clone("/clone-A", "success", "kit-A", reg)
+        clone_registry.register_clone("/clone-B", "success", "kit-B", reg)
+
+        tool_ctx.kitchen_id = "kit-A"
+        mock_mgr = MagicMock()
+        mock_mgr.remove_clone.return_value = {"removed": "true"}
+        tool_ctx.clone_mgr = mock_mgr
+
+        result = json.loads(
+            await batch_cleanup_clones(registry_path=reg, all_owners="yes")
+        )
+
+        assert "/clone-A" in result["deleted"]
+        assert "/clone-B" not in result["deleted"]
+
+    @pytest.mark.anyio
+    async def test_two_kitchens_register_and_cleanup_isolated(self, tool_ctx, tmp_path):
+        """T19 — kitchen A's cleanup does not touch kitchen B's registry entry."""
+        import json as _json
+        from pathlib import Path as _Path
+
+        reg = str(tmp_path / "registry.json")
+
+        # Session 1: kit-1 registers
+        tool_ctx.kitchen_id = "kit-1"
+        await register_clone_status(clone_path="/clone-1", status="success", registry_path=reg)
+
+        # Session 2: kit-2 registers
+        tool_ctx.kitchen_id = "kit-2"
+        await register_clone_status(clone_path="/clone-2", status="success", registry_path=reg)
+
+        # Session 1 cleans up
+        tool_ctx.kitchen_id = "kit-1"
+        mock_mgr = MagicMock()
+        mock_mgr.remove_clone.return_value = {"removed": "true"}
+        tool_ctx.clone_mgr = mock_mgr
+
+        result = json.loads(await batch_cleanup_clones(registry_path=reg))
+
+        # Only kit-1's clone is removed
+        assert "/clone-1" in result["deleted"]
+        assert "/clone-2" not in result["deleted"]
+        mock_mgr.remove_clone.assert_called_once_with("/clone-1", "false")
+
+        # kit-2's entry is still on disk
+        data = _json.loads(_Path(reg).read_text())
+        remaining = {e["path"]: e.get("owner") for e in data["clones"]}
+        assert "/clone-2" in remaining
+        assert remaining["/clone-2"] == "kit-2"
--- a/tests/server/test_tools_clone.py
+++ b/tests/server/test_tools_clone.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+from pathlib import Path
 from unittest.mock import MagicMock
 
 import pytest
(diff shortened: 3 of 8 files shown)


src/autoskillit/server/tools_clone.py
```diff
@@ -303,13 +303,22 @@ async def register_clone_status(
     from autoskillit.server import _get_ctx
 
     tool_ctx = _get_ctx()
+    owner = tool_ctx.kitchen_id
+    if owner == "":
```
Trecek, line 312: [warning] arch: L3 (tools_clone.py) short-circuits with a JSON `registered=false` response when `kitchen_id` is empty, before calling L1's `register_clone`. L1 independently raises `ValueError('owner is required')` for empty owner (clone_registry.py:49). This creates a dual-validation pattern where L3 duplicates an ownership guard L1 already enforces, rather than delegating and translating the resulting exception into a response. The layers can diverge silently.

src/autoskillit/server/tools_clone.py
```diff
@@ -303,13 +303,22 @@ async def register_clone_status(
...
+        return json.dumps(
+            {
+                "registered": "false",
+                "reason": "no active kitchen (kitchen_id empty) — cannot register clone",
```
Trecek, line 316: [warning] cohesion: `register_clone_status` early-return uses key `reason` for the error message, but `batch_cleanup_clones` empty-kitchen-id early-return uses key `error`. Both represent tool-level guard failures — the error key name should be consistent (`reason` or `error`) across both tools so callers can inspect the failure field by name without branching.

src/autoskillit/server/tools_clone.py
```diff
@@ -361,13 +378,31 @@ async def batch_cleanup_clones(
                 }
             )
 
+        is_escape_hatch = all_owners.strip().lower() == "true"
```
Trecek, line 386: [warning] arch: The escape-hatch parsing logic (`all_owners.strip().lower() == 'true'`) in L3 defines *which string value* triggers all-owners mode. This is domain logic — the semantic mapping from MCP string param to `owner_filter=None`. Placing it in L1 (`batch_delete` accepting an explicit sentinel or bool) would keep L3 responsible only for parameter extraction, not for defining the escape-hatch contract. As-is, the rule that `'true'` maps to `owner=None` exists only in L3; L1 is unaware of the escape hatch's activation condition.

src/autoskillit/server/tools_clone.py
```diff
@@ -361,13 +378,31 @@ async def batch_cleanup_clones(
...
+        owner_filter: str | None
+        if is_escape_hatch:
+            owner_filter = None
+        else:
+            owner_filter = tool_ctx.kitchen_id
```
Trecek, line 391: [warning] bugs: `owner_filter` is assigned `tool_ctx.kitchen_id` then only guarded with `if owner_filter == ""`. If `kitchen_id` is `None` (not empty string), the guard does not fire and `owner_filter=None` is passed to `batch_delete`, silently activating all-owners mode without the caller passing `all_owners="true"`. Change the guard to `if not owner_filter:` to catch both empty string and `None`.

src/autoskillit/workspace/clone_registry.py
```diff
@@ -73,29 +82,46 @@ def read_registry(
 def cleanup_candidates(
     registry_path: str = "",
     temp_dir: Path | None = None,
+    owner: str | None = None,
```
Trecek, line 85: [warning] defense: `cleanup_candidates` accepts `owner: str | None = None` but an empty string `""` silently becomes a valid filter predicate — `_entry_matches_owner` would match entries whose `owner` field is also `""` (i.e., no entries in practice, but distinct from `None` all-owners mode). A caller who derives `owner` from an unvalidated source and passes `""` gets silent all-miss results rather than an error. Add an explicit guard: `if owner is not None and owner == "": raise ValueError("owner must be non-empty or None for all-owners mode")`.

Trecek (commented):

**10 actionable findings** (clear fix exists) and **2 decision findings** (design trade-offs) across 5 audit dimensions. See inline comments. [Note: REQUEST_CHANGES replaced with COMMENT — GitHub disallows requesting changes on own PRs]

src/autoskillit/workspace/clone_registry.py
```diff
@@ -73,29 +82,46 @@ def read_registry(
...
 ) -> tuple[list[str], list[str]]:
     """Return (to_delete, to_preserve) path lists from the registry.
```
Trecek, line 87: [warning] defense: The `cleanup_candidates` docstring documents `owner=None` (all-owners) and mentions legacy orphan behavior, but does not state what `owner=""` (empty string) does. The empty-string behavior is undocumented, creating a silent footgun — it would match only entries whose `owner` field is explicitly `""`, which is an impossible value given `register_clone` rejects empty owner. Document or add a guard.

tests/server/test_tools_clone.py
```diff
@@ -365,3 +370,180 @@ async def test_batch_cleanup_clones_nonexistent_path_does_not_raise(self, tool_c
...
+    async def test_register_clone_status_propagates_kitchen_id_as_owner(self, tool_ctx, tmp_path):
+        """T12 — register_clone_status writes entry with owner == kitchen_id."""
+        tool_ctx.kitchen_id = "kit-xyz"
+        reg = str(tmp_path / "registry.json")
+        result = json.loads(
+            await register_clone_status(clone_path="/c", status="success", registry_path=reg)
+        )
+        assert result["registered"] == "true"
+
+        import json as _json
```
Trecek: [warning] slop: Inline `import json as _json` and `from pathlib import Path as _Path` inside the test method body are unnecessary — `json` and `Path` should be top-level imports. Using aliased inline imports to avoid collisions signals missing top-level imports rather than warranting per-method imports. Move to top-level and use `json`/`Path` directly.

tests/server/test_tools_clone.py
```diff
@@ -365,3 +370,180 @@ async def test_batch_cleanup_clones_nonexistent_path_does_not_raise(self, tool_c
...
+    async def test_register_clone_status_rejects_when_kitchen_id_empty(self, tool_ctx, tmp_path):
+        """T13 — register_clone_status returns registered=false when kitchen_id is empty."""
+        tool_ctx.kitchen_id = ""
+        reg = str(tmp_path / "registry.json")
+        result = json.loads(
+            await register_clone_status(clone_path="/c", status="success", registry_path=reg)
+        )
+        assert result["registered"] == "false"
+        assert "kitchen_id" in result["reason"] or "kitchen" in result["reason"]
+        from pathlib import Path as _Path
```
Trecek: [warning] slop: Inline `from pathlib import Path as _Path` inside test method body. Same pattern as L393 — should be a top-level import. Move to the module-level import block.

tests/server/test_tools_clone.py
```diff
@@ -365,3 +370,180 @@ async def test_batch_cleanup_clones_nonexistent_path_does_not_raise(self, tool_c
...
+
+    @pytest.mark.anyio
+    async def test_batch_cleanup_clones_empty_kitchen_id_with_all_owners_true_succeeds(
+        self, tool_ctx, tmp_path
+    ):
+        """T17 — escape hatch works even when kitchen_id is empty (legacy recovery)."""
+
+        reg_path = tmp_path / "registry.json"
+        reg_path.write_text(
+            __import__("json").dumps({"clones": [{"path": "/legacy", "status": "success"}]})
```
Trecek: [warning] slop: `__import__('json').dumps(...)` is an AI-generated workaround for a missing top-level import. Replace with a normal top-level `import json` and `json.dumps(...)` directly.

tests/server/test_tools_clone.py
```diff
@@ -365,3 +370,180 @@ async def test_batch_cleanup_clones_nonexistent_path_does_not_raise(self, tool_c
...
+        mock_mgr.remove_clone.return_value = {"removed": "true"}
+        tool_ctx.clone_mgr = mock_mgr
+
+        result = json.loads(await batch_cleanup_clones(registry_path=reg, all_owners="yes"))
+
+        assert "/clone-A" in result["deleted"]
+        assert "/clone-B" not in result["deleted"]
+
+    @pytest.mark.anyio
+    async def test_two_kitchens_register_and_cleanup_isolated(self, tool_ctx, tmp_path):
```
Trecek, line 512: [warning] tests: T19 simulates two sessions by mutating `tool_ctx.kitchen_id` sequentially on the same shared object. This works because `register_clone_status` reads `kitchen_id` at call time, but it conflates shared-state mutation with session isolation. If the implementation ever reads `kitchen_id` earlier in the call path, silent cross-contamination could occur. Document this dependency explicitly or use two distinct `tool_ctx` instances to model two real sessions.

tests/server/test_tools_clone.py
```diff
@@ -365,3 +370,180 @@ async def test_batch_cleanup_clones_nonexistent_path_does_not_raise(self, tool_c
...
+        """T19 — kitchen A's cleanup does not touch kitchen B's registry entry."""
+        import json as _json
```
Trecek: [warning] slop: Inline `import json as _json` and `from pathlib import Path as _Path` inside T19 test method body. Repeated AI-generated pattern of avoiding top-level imports by aliasing inline. Move to module-level imports.

tests/workspace/test_clone_registry.py
```diff
@@ -91,3 +105,161 @@ def test_cleanup_candidates_empty_registry(self, tmp_path: Path) -> None:
...
+        reg = str(tmp_path / "registry.json")
+        register_clone("/tmp/a", "success", "kit-1", registry_path=reg)
+        register_clone("/tmp/b", "success", "kit-2", registry_path=reg)
+        register_clone("/tmp/c", "success", "kit-3", registry_path=reg)
+
+        to_delete, _ = cleanup_candidates(registry_path=reg, owner=None)
+
+        assert sorted(to_delete) == ["/tmp/a", "/tmp/b", "/tmp/c"]
+
+    def test_cleanup_candidates_with_owner_filters_to_matching_only(self, tmp_path: Path) -> None:
```
Trecek, line 155: [warning] tests: T4 uses two negative membership checks (`not in to_delete`, `not in to_preserve`) but does not assert `to_preserve == []` explicitly. Since kit-2's entry should be completely invisible to owner='kit-1' scoping (absent from both lists), an explicit `assert to_preserve == []` makes the invisibility contract unambiguous rather than relying on two negative checks.

src/autoskillit/server/tools_clone.py
Trecek, line 312: [warning] arch: L3 (tools_clone.py) short-circuits with a JSON `registered=false` when `kitchen_id` is empty, before calling L1's `register_clone`. L1 independently raises `ValueError('owner is required')` for empty owner (clone_registry.py:49). Dual-validation — L3 duplicates an ownership guard L1 already enforces. The layers can diverge silently.

src/autoskillit/server/tools_clone.py
Trecek, line 316: [warning] cohesion: `register_clone_status` early-return uses key `reason` for the error message, but `batch_cleanup_clones` empty-kitchen-id early-return (L393) uses key `error`. Both are tool-level guard failures — the error key name should be consistent across both tools so callers don't need to branch on key name.

src/autoskillit/server/tools_clone.py
Trecek, line 386: [warning] arch: The escape-hatch parsing logic (`all_owners.strip().lower() == 'true'`) in L3 defines *which string value* triggers all-owners mode — that's domain logic. Moving to L1 (e.g., `batch_delete` accepting an explicit sentinel) would keep L3 responsible only for parameter extraction, not for defining the escape-hatch contract.

src/autoskillit/server/tools_clone.py
Trecek, line 391: [warning] bugs: `owner_filter` is assigned `tool_ctx.kitchen_id` and then only guarded with `if owner_filter == ""`. If `kitchen_id` is `None` (not empty string), the guard doesn't fire and `owner_filter=None` is passed to `batch_delete`, silently activating all-owners mode without `all_owners="true"`. Fix: `if not owner_filter:`.

src/autoskillit/workspace/clone_registry.py
Trecek, line 85: [warning] defense: `cleanup_candidates(owner: str | None = None)` accepts an empty string `""` as a valid owner filter — `_entry_matches_owner` would match entries with `owner == ""` (impossible in practice but undocumented). A caller passing `""` gets silent all-miss results rather than an error. Add guard: `if owner is not None and owner == "": raise ValueError(...)`.

src/autoskillit/workspace/clone_registry.py
Trecek, line 87: [warning] defense: The docstring documents `owner=None` and legacy orphan behavior but omits what `owner=""` does. Silent footgun — document or add a guard at the signature level.

tests/server/test_tools_clone.py
Trecek: [warning] slop: Inline `import json as _json` and `from pathlib import Path as _Path` inside the test method body — these should be top-level imports. Using aliased inline imports to avoid collisions signals missing top-level imports rather than warranting per-method imports.

tests/server/test_tools_clone.py
Trecek: [warning] slop: Inline `from pathlib import Path as _Path` inside test method body. Same pattern as L393 — move to module-level import block.

tests/server/test_tools_clone.py
Trecek: [warning] slop: `__import__('json').dumps(...)` is an AI-generated workaround for a missing top-level import. Use a normal `import json` at module level and call `json.dumps()` directly.

tests/server/test_tools_clone.py
Trecek, line 512: [warning] tests: T19 simulates two sessions by mutating `tool_ctx.kitchen_id` sequentially on the same object. Works today because `register_clone_status` reads `kitchen_id` at call time, but conflates shared-state mutation with session isolation. Document this dependency or use two distinct `tool_ctx` instances.

tests/server/test_tools_clone.py
Trecek: [warning] slop: Inline `import json as _json` / `from pathlib import Path as _Path` in T19. Repeated AI-generated pattern — move to module-level imports.

tests/workspace/test_clone_registry.py
Trecek, line 155: [warning] tests: T4 uses two negative membership checks but doesn't assert `to_preserve == []` explicitly. Owner scoping should make kit-2's entry completely invisible (absent from both lists). An explicit empty-list assertion makes the invisibility contract unambiguous.

Trecek (commented):
AutoSkillit review found 10 blocking issues across 5 audit dimensions (arch, bugs, defense, tests, slop). See inline comments. [Note: REQUEST_CHANGES submitted as COMMENT — PR is authored by the authenticated user]

src/autoskillit/server/tools_clone.py
Trecek: Investigated — this is intentional. L1 (clone_registry.register_clone line 49) independently raises ValueError when owner is empty, so the L3 early-exit is defense-in-depth, not a layering violation. The fast-fail avoids asyncio.to_thread overhead and produces the same JSON shape the except-block would produce. Introduced intentionally in 1f8cef8.

Trecek: Investigated — this is intentional. L1 (clone_registry.register_clone line 49) independently raises ValueError when owner is empty, so the L3 early-exit is defense-in-depth, not a layering violation. The fast-fail avoids asyncio.to_thread overhead and produces the same JSON shape the except-block would produce. Introduced intentionally in 1f8cef8.

src/autoskillit/server/tools_clone.py
Trecek: Agreed — fixed in 6ddacbd. Line 301 (status-validation guard) returned {"error":...} while the documented shape and all other code paths use {"registered":"false","reason":...}. Normalized to the documented shape throughout, including updating the existing test assertion.

Trecek: Agreed — fixed in 6ddacbd. Line 301 (status-validation guard) returned {"error":...} while the documented shape and all other code paths use {"registered":"false","reason":...}. Normalized to the documented shape throughout, including updating the existing test assertion.

src/autoskillit/server/tools_clone.py
Trecek: Valid observation — flagged for design decision. The string-to-bool parsing is a single expression used exactly once; L1 (clone_registry.batch_delete) takes owner: str | None and has no concept of the string 'true'. Centralizing at L1 would misplace it; a shared parse_bool_param() utility at the server layer would be the right abstraction if this pattern grows. Leaving as-is for this PR scope.

Trecek: Valid observation — flagged for design decision. The string-to-bool parsing is a single expression used exactly once; L1 (clone_registry.batch_delete) takes owner: str | None and has no concept of the string 'true'. Centralizing at L1 would misplace it; a shared parse_bool_param() utility at the server layer would be the right abstraction if this pattern grows. Leaving as-is for this PR scope.

src/autoskillit/server/tools_clone.py
Trecek: Investigated — this is intentional. ToolContext.kitchen_id is declared `kitchen_id: str = field(default="")` in pipeline/context.py — hard str type, never None. Python dataclasses enforce this at construction. The `== ""` guard is correct and sufficient; using `if not owner_filter` would conflate the empty-string sentinel with None (the all-owners escape-hatch value), which would be a logic error.

Trecek: Investigated — this is intentional. ToolContext.kitchen_id is declared `kitchen_id: str = field(default="")` in pipeline/context.py — hard str type, never None. Python dataclasses enforce this at construction. The `== ""` guard is correct and sufficient; using `if not owner_filter` would conflate the empty-string sentinel with None (the all-owners escape-hatch value), which would be a logic error.

src/autoskillit/workspace/clone_registry.py
Trecek: Investigated — this is intentional. The empty-string path is structurally blocked before reaching cleanup_candidates: tools_clone.py lines 387-396 check `if owner_filter == ""` and return an error JSON without calling batch_delete or cleanup_candidates. Additionally register_clone (line 49-50) raises ValueError on empty owner, so no registry entry can carry owner="". The concern is valid mechanically but the code path is doubly unreachable.

Trecek: Investigated — this is intentional. The empty-string path is structurally blocked before reaching cleanup_candidates: tools_clone.py lines 387-396 check `if owner_filter == ""` and return an error JSON without calling batch_delete or cleanup_candidates. Additionally register_clone (line 49-50) raises ValueError on empty owner, so no registry entry can carry owner="". The concern is valid mechanically but the code path is doubly unreachable.

src/autoskillit/workspace/clone_registry.py
Trecek: Investigated — this is intentional. The docstring documents the two supported calling modes (owner=None for all-owners, owner=<non-empty string> for scoped). Documenting owner="" would implicitly suggest it is a supported calling convention when it is structurally prevented from reaching this function (blocked by the kitchen_id guard in tools_clone.py lines 387-396 before any call here).

Trecek: Investigated — this is intentional. The docstring documents the two supported calling modes (owner=None for all-owners, owner=<non-empty string> for scoped). Documenting owner="" would implicitly suggest it is a supported calling convention when it is structurally prevented from reaching this function (blocked by the kitchen_id guard in tools_clone.py lines 387-396 before any call here).

tests/server/test_tools_clone.py
Trecek: Agreed — fixed in acf791e. Added module-level `from pathlib import Path` import; removed inline `import json as _json` and `from pathlib import Path as _Path` from T12 method body.

Trecek: Agreed — fixed in acf791e. Added module-level `from pathlib import Path` import; removed inline `import json as _json` and `from pathlib import Path as _Path` from T12 method body.

tests/server/test_tools_clone.py
Trecek: Agreed — fixed in acf791e. Removed inline `from pathlib import Path as _Path` from T13 method body; now uses module-level Path.

Trecek: Agreed — fixed in acf791e. Removed inline `from pathlib import Path as _Path` from T13 method body; now uses module-level Path.

tests/server/test_tools_clone.py
Trecek: Agreed — fixed in acf791e. Replaced `__import__("json").dumps(...)` with `json.dumps(...)` using the existing module-level import.

Trecek: Agreed — fixed in acf791e. Replaced `__import__("json").dumps(...)` with `json.dumps(...)` using the existing module-level import.

tests/server/test_tools_clone.py
Trecek: Agreed — fixed in acf791e. Refactored T19 to use clone_registry.register_clone directly for both session registrations (consistent with T14-T18 pattern) instead of mutating tool_ctx.kitchen_id twice on a shared context. The MCP tool is now only used for the cleanup phase.

Trecek: Agreed — fixed in acf791e. Refactored T19 to use clone_registry.register_clone directly for both session registrations (consistent with T14-T18 pattern) instead of mutating tool_ctx.kitchen_id twice on a shared context. The MCP tool is now only used for the cleanup phase.

tests/server/test_tools_clone.py
Trecek: Agreed — fixed in acf791e. Removed inline `import json as _json` and `from pathlib import Path as _Path` from T19; these are now module-level imports used directly.

Trecek: Agreed — fixed in acf791e. Removed inline `import json as _json` and `from pathlib import Path as _Path` from T19; these are now module-level imports used directly.

tests/workspace/test_clone_registry.py
Trecek: Agreed — fixed in f458c18. Replaced two negative membership checks with `assert to_preserve == []` directly, matching the stronger assertion pattern already used in T5.

Trecek: Agreed — fixed in f458c18. Replaced two negative membership checks with `assert to_preserve == []` directly, matching the stronger assertion pattern already used in T5.

PR #767, opened 2026-04-12
Implementation Plan: Detect sustained D-state and high-CPU anomalies in headless sessions

### LINKED ISSUE #760: Detect sustained D-state and high-CPU anomalies in headless sessions
## Summary

Extend `execution/anomaly_detection.py` to flag two new anomaly kinds for hung and infinite-loop headless sessions. The underlying data is already collected by `execution/linux_tracing.py` via psutil — we just don't alert on it.

## Motivation

Today, when a headless Claude subprocess stalls on an uninterruptible disk/NFS/IPC wait, or burns into a tight CPU loop, the pipeline sees nothing until the outer timeout kills it. The symptoms look like "session still running" in the CLI. We already capture the Linux process-state code (`R`/`S`/`D`/`T`/`Z`) and `wchan` (the kernel function the process is sleeping inside) in every `ProcSnapshot`, plus CPU%. Two small additions to `detect_anomalies()` would convert that raw data into actionable alerts.

Related: #631 is the complementary suppression side (don't *kill* sessions whose children are legitimately busy). This issue is the detection side: flag parent sessions that are silently stuck or looping.

## Current state

`execution/anomaly_detection.py:13-21` defines:

```python
class AnomalyKind(StrEnum):
    OOM_SPIKE = "oom_spike"
    OOM_CRITICAL = "oom_critical"
    ZOMBIE_DETECTED = "zombie_detected"
    ZOMBIE_PERSISTENT = "zombie_persistent"
    SIGNALS_PENDING = "signals_pending"
    RSS_GROWTH = "rss_growth"
    FD_HIGH = "fd_high"
```

The only state-based check is `consecutive_zombie` for `ZOMBIE_PERSISTENT`. `state` beyond `Z` and `wchan` are collected but never consumed.

## Proposed changes

1. **Add two new anomaly kinds:**
   - `D_STATE_SUSTAINED` — fires when ≥2 consecutive snapshots show `state == "D"` (uninterruptible sleep, typically disk or NFS I/O hang).
   - `HIGH_CPU_SUSTAINED` — fires when ≥2 consecutive snapshots show CPU% ≥ 90 (likely infinite loop).

2. **Implementation in `detect_anomalies()`:** add two counters alongside the existing `consecutive_zombie`, following the same pattern (increment on match, reset on miss, emit anomaly when counter crosses threshold).

3. **Include `wchan` in the D-state anomaly detail** so operators can see *which kernel function* the process is stuck on. Examples:
   - `wchan=ext4_file_write_iter` → ext4 disk write
   - `wchan=unix_stream_recvmsg` → unix-socket receive
   - `wchan=nfs_wait_on_request` → NFS stall
   - `wchan=do_nanosleep` → actually sleeping on purpose (should NOT fire; add to the ignore list)

4. **Optional: add a check to `cli/_doctor.py`** that runs `ps -axo pid,state,pcpu,comm | grep claude` and prints the state breakdown of any running claude processes — useful for interactive "is my pipeline stuck" debugging.

## Acceptance criteria

- `AnomalyKind` enum has `D_STATE_SUSTAINED` and `HIGH_CPU_SUSTAINED` values.
- `detect_anomalies()` fires `D_STATE_SUSTAINED` on ≥2 consecutive D-state snapshots, with `wchan` in the anomaly detail.
- `detect_anomalies()` fires `HIGH_CPU_SUSTAINED` on ≥2 consecutive snapshots with CPU% ≥ 90.
- D-state anomaly skips false positives where `wchan` indicates a legitimate sleep (e.g. `do_nanosleep`, `do_epoll_wait`, `schedule_hrtimeout_range`).
- Unit tests in `tests/execution/test_anomaly_detection.py` cover both new kinds, including:
  - Fires on 2 consecutive matching snapshots
  - Does NOT fire on a single matching snapshot
  - Resets counter when a non-matching snapshot intervenes
  - D-state with benign `wchan` does not fire
  - Anomaly detail includes `wchan` for D-state
- Session diagnostics JSONL entries (via `execution/session_log.py`) surface the new anomaly kinds.

## Out of scope

- Process killing / recovery actions. This ticket is diagnostic only — fire the anomaly, write to the log, continue running. Kill-on-hang is a follow-up design decision.
- Changes to `linux_tracing.py`. The data is already collected there.
- Cross-platform support (macOS psutil exposes `status` but not `wchan`; this detection is Linux-specific and the existing `linux_tracing.py` already gates on Linux).

## Files likely touched

- `src/autoskillit/execution/anomaly_detection.py` — enum additions + detection logic
- `src/autoskillit/execution/session_log.py` — if anomaly kinds are enumerated anywhere for filtering/routing
- `tests/execution/test_anomaly_detection.py` — new test cases
- `src/autoskillit/cli/_doctor.py` — optional `ps` check

### ENUM — Anomaly Enum Extensions

- **REQ-ENUM-001:** `AnomalyKind` must include a `D_STATE_SUSTAINED` member.
- **REQ-ENUM-002:** `AnomalyKind` must include a `HIGH_CPU_SUSTAINED` member.

### DETECT — Detection Logic

- **REQ-DETECT-001:** `detect_anomalies()` must fire `D_STATE_SUSTAINED` when two or more consecutive snapshots show `state == "D"`.
- **REQ-DETECT-002:** `detect_anomalies()` must fire `HIGH_CPU_SUSTAINED` when two or more consecutive snapshots show CPU percent ≥ 90.
- **REQ-DETECT-003:** The consecutive-match counter for each new anomaly kind must reset when a non-matching snapshot intervenes.
- **REQ-DETECT-004:** Neither new anomaly kind must fire on a single matching snapshot.

### WCHAN — wchan Handling

- **REQ-WCHAN-001:** The `D_STATE_SUSTAINED` anomaly detail must include the `wchan` value from the matching snapshot.
- **REQ-WCHAN-002:** `D_STATE_SUSTAINED` must not fire when `wchan` indicates a legitimate sleep (e.g., `do_nanosleep`, `do_epoll_wait`, `schedule_hrtimeout_range`).
- **REQ-WCHAN-003:** The benign-`wchan` ignore list must be centrally defined so it can be extended without touching the detection loop.

### TEST — Test Coverage

- **REQ-TEST-001:** Tests must verify `D_STATE_SUSTAINED` fires on two consecutive D-state snapshots.
- **REQ-TEST-002:** Tests must verify `HIGH_CPU_SUSTAINED` fires on two consecutive snapshots where CPU percent is ≥ 90.
- **REQ-TEST-003:** Tests must verify neither new anomaly fires on a single matching snapshot.
- **REQ-TEST-004:** Tests must verify the consecutive-match counter resets when a non-matching snapshot intervenes for each new kind.
- **REQ-TEST-005:** Tests must verify D-state snapshots with benign `wchan` values do not fire `D_STATE_SUSTAINED`.
- **REQ-TEST-006:** Tests must verify the `D_STATE_SUSTAINED` anomaly detail includes the `wchan` field.

### DIAG — Session Diagnostics Surface

- **REQ-DIAG-001:** Session diagnostics JSONL entries written by `execution/session_log.py` must surface `D_STATE_SUSTAINED` and `HIGH_CPU_SUSTAINED` anomaly kinds alongside the existing kinds.

### DOCTOR — Diagnostic CLI (Optional)

- **REQ-DOCTOR-001:** If the `cli/_doctor.py` check is implemented, it must run `ps -axo pid,state,pcpu,comm`, filter to `claude` processes, and print the state breakdown for interactive hang debugging.

Trecek:
## Summary

Extend `execution/anomaly_detection.py` with two new `AnomalyKind` members —
`D_STATE_SUSTAINED` and `HIGH_CPU_SUSTAINED` — that fire when ≥2 consecutive
`ProcSnapshot`s show the process stuck in uninterruptible sleep (I/O hang) or
burning CPU ≥ 90% (suspected infinite loop). Wire the detection into
`detect_anomalies()` following the existing `consecutive_zombie` counter
pattern, filter out benign `wchan` values via a centrally-defined
`BENIGN_WCHANS` frozenset, and include `wchan` in the D-state anomaly detail
so operators can see which kernel function the process is stuck in.

A prerequisite correction is required: the issue asserts CPU% is already
captured in `ProcSnapshot`, but the current `ProcSnapshot` dataclass has no
`cpu_percent` field. The plan therefore adds `cpu_percent` to `ProcSnapshot`
and teaches `proc_monitor` to persist a single `psutil.Process` instance so
that `cpu_percent(interval=0)` returns meaningful values across snapshots.
Without this, `HIGH_CPU_SUSTAINED` has no input signal.

Session diagnostics surfacing (REQ-DIAG-001) is automatic once detection
fires: `flush_session_log()` already serialises every anomaly record to
`anomalies.jsonl` via `json.dumps(a, sort_keys=True)` without enumerating
kinds. An integration test locks that behaviour in.

The optional doctor check (REQ-DOCTOR-001) is included: a new numbered
Check 15 runs `ps -axo pid,state,pcpu,comm`, filters to `claude`
processes, groups by state, and prints the breakdown, warning when any
claude process is currently in D state.

### ENUM — Anomaly Enum Extensions

- **REQ-ENUM-001:** `AnomalyKind` must include a `D_STATE_SUSTAINED` member.
- **REQ-ENUM-002:** `AnomalyKind` must include a `HIGH_CPU_SUSTAINED` member.

### DETECT — Detection Logic

- **REQ-DETECT-001:** `detect_anomalies()` must fire `D_STATE_SUSTAINED` when two or more consecutive snapshots show `state == "D"`.
- **REQ-DETECT-002:** `detect_anomalies()` must fire `HIGH_CPU_SUSTAINED` when two or more consecutive snapshots show CPU percent ≥ 90.
- **REQ-DETECT-003:** The consecutive-match counter for each new anomaly kind must reset when a non-matching snapshot intervenes.
- **REQ-DETECT-004:** Neither new anomaly kind must fire on a single matching snapshot.

### WCHAN — wchan Handling

- **REQ-WCHAN-001:** The `D_STATE_SUSTAINED` anomaly detail must include the `wchan` value from the matching snapshot.
- **REQ-WCHAN-002:** `D_STATE_SUSTAINED` must not fire when `wchan` indicates a legitimate sleep (e.g., `do_nanosleep`, `do_epoll_wait`, `schedule_hrtimeout_range`).
- **REQ-WCHAN-003:** The benign-`wchan` ignore list must be centrally defined so it can be extended without touching the detection loop.

### TEST — Test Coverage

- **REQ-TEST-001:** Tests must verify `D_STATE_SUSTAINED` fires on two consecutive D-state snapshots.
- **REQ-TEST-002:** Tests must verify `HIGH_CPU_SUSTAINED` fires on two consecutive snapshots where CPU percent is ≥ 90.
- **REQ-TEST-003:** Tests must verify neither new anomaly fires on a single matching snapshot.
- **REQ-TEST-004:** Tests must verify the consecutive-match counter resets when a non-matching snapshot intervenes for each new kind.
- **REQ-TEST-005:** Tests must verify D-state snapshots with benign `wchan` values do not fire `D_STATE_SUSTAINED`.
- **REQ-TEST-006:** Tests must verify the `D_STATE_SUSTAINED` anomaly detail includes the `wchan` field.

### DIAG — Session Diagnostics Surface

- **REQ-DIAG-001:** Session diagnostics JSONL entries written by `execution/session_log.py` must surface `D_STATE_SUSTAINED` and `HIGH_CPU_SUSTAINED` anomaly kinds alongside the existing kinds.

### DOCTOR — Diagnostic CLI (Optional)

- **REQ-DOCTOR-001:** If the `cli/_doctor.py` check is implemented, it must run `ps -axo pid,state,pcpu,comm`, filter to `claude` processes, and print the state breakdown for interactive hang debugging.

### Process Flow Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% ── TERMINALS ── %%
    START([START: run_headless_core])
    START2([START: autoskillit doctor])
    SESSION_DONE([SESSION COMPLETE])
    DOCTOR_OK([DOCTOR OK])
    DOCTOR_WARN([DOCTOR WARNING])

    %% ─────────────────────────────────── %%
    %% FLOW A: SESSION MONITORING          %%
    %% ─────────────────────────────────── %%

    subgraph Monitor ["● Phase 1 — Process Monitor Loop  (linux_tracing.py)"]
        direction TB
        PersistProc["● Persist psutil.Process once<br/>━━━━━━━━━━<br/>proc_monitor() constructs Process(pid)<br/>calls cpu_percent(interval=0) to prime baseline<br/>reuses same object every iteration"]
        SnapTick{"next interval tick<br/>━━━━━━━━━━<br/>await anyio.sleep(proc_interval)"}
        ReadSnap["● read_proc_snapshot(pid, process=_proc)<br/>━━━━━━━━━━<br/>psutil.oneshot(): state, vm_rss_kb, threads,<br/>  fd_count, fd_soft_limit, ctx_switches<br/>● cpu_percent(interval=0) — delta since last call<br/>/proc reads: oom_score, wchan, SigPnd/Blk/Cgt"]
        ProcDied{"snap is None?<br/>━━━━━━━━━━<br/>NoSuchProcess /<br/>AccessDenied"}
        AccSnap["accumulate + stream snapshot<br/>━━━━━━━━━━<br/>handle._snapshots.append(snap)<br/>writes JSON line → /dev/shm/autoskillit_trace_{pid}.jsonl"]
    end

    subgraph SessionEnd ["Phase 2 — Session Termination"]
        direction TB
        TrigFire["race trigger fires<br/>━━━━━━━━━━<br/>Channel A (stdout JSONL result)<br/>Channel B (Claude session JSONL)<br/>process exit / idle timeout"]
        StopTrace["tracing_handle.stop()<br/>━━━━━━━━━━<br/>cancel CancelScope → stops monitor task<br/>flush+close tmpfs file<br/>returns list[ProcSnapshot]"]
        FinalSnap["read final snapshot<br/>━━━━━━━━━━<br/>read_proc_snapshot(pid) — no process=<br/>cpu_percent = 0.0 (post-death baseline)"]
    end

    subgraph AnomalyDetect ["● Phase 3 — Post-Hoc Anomaly Detection  (anomaly_detection.py)"]
        direction TB
        DetectEntry["detect_anomalies(snapshots, pid)<br/>━━━━━━━━━━<br/>called once from flush_session_log()<br/>initialises counters: consecutive_zombie,<br/>consecutive_d_state, consecutive_high_cpu"]
        SnapIter{"for seq, snap in enumerate(snapshots)<br/>━━━━━━━━━━<br/>linear pass — no early exit"}
        DStateGate{"● state == disk-sleep?<br/>━━━━━━━━━━<br/>AND wchan not in BENIGN_WCHANS<br/>(do_nanosleep, do_epoll_wait,<br/>schedule_hrtimeout_range, '', '0')"}
        DStateCount{"● consecutive_d_state >= 2?"}
        EmitDState["● emit D_STATE_SUSTAINED<br/>━━━━━━━━━━<br/>severity: WARNING<br/>detail: {state, wchan, consecutive_count}"]
        CPUGate{"● cpu_percent >= 90.0?"}
        CPUCount{"● consecutive_high_cpu >= 2?"}
        EmitHighCPU["● emit HIGH_CPU_SUSTAINED<br/>━━━━━━━━━━<br/>severity: WARNING<br/>detail: {cpu_percent, consecutive_count}"]
        OtherDet["other detectors (unchanged)<br/>━━━━━━━━━━<br/>OOM_SPIKE (delta > 200) · OOM_CRITICAL (≥ 800)<br/>ZOMBIE_DETECTED · ZOMBIE_PERSISTENT (≥ 3)<br/>SIGNALS_PENDING (sig_pnd rising edge)<br/>FD_HIGH (fd ratio > 0.80)"]
        PostLoop["post-loop: RSS_GROWTH<br/>━━━━━━━━━━<br/>final_rss > 2x initial_rss<br/>AND len(snapshots) >= 5"]
    end

    subgraph LogFlush ["Phase 4 — Session Log Flush  (session_log.py)"]
        direction TB
        WriteTrace["write proc_trace.jsonl<br/>━━━━━━━━━━<br/>one line per snapshot<br/>tracks peak_rss_kb, peak_oom_score, peak_fd_ratio"]
        HasAnom{"anomalies list non-empty?"}
        WriteAnom["write anomalies.jsonl<br/>━━━━━━━━━━<br/>{ts, seq, event, kind, severity,<br/>pid, detail, snapshot} per line"]
        WriteSummary["write summary.json (atomic)<br/>━━━━━━━━━━<br/>anomaly_count · peak fields<br/>session metadata · termination_reason"]
        AppendIdx["append sessions.jsonl<br/>━━━━━━━━━━<br/>global index with anomaly_count,<br/>peak_rss_kb, peak_oom_score"]
        Retention["_enforce_retention()<br/>━━━━━━━━━━<br/>trim to 500 sessions, rewrite index"]
    end

    %% ─────────────────────────────────── %%
    %% FLOW B: DOCTOR CHECK 15             %%
    %% ─────────────────────────────────── %%

    subgraph DoctorFlow ["● Phase 5 — Doctor Check 15  (cli/_doctor.py)"]
        direction TB
        RunPS["● _check_claude_process_state_breakdown()<br/>━━━━━━━━━━<br/>run: ps -axo pid,state,pcpu,comm<br/>timeout=5s · returns OK on any ps failure"]
        FilterClaude{"comm == 'claude'<br/>or ends with '/claude'?"}
        AnyProcs{"claude processes<br/>found?"}
        BuildBreakdown["build state breakdown<br/>━━━━━━━━━━<br/>count by ps state letter<br/>format: 'D=1, S=3'"]
        CheckDStatePS{"any row<br/>state == 'D'?"}
        DoctorOKMsg["OK: 'claude process state breakdown: S=2'<br/>━━━━━━━━━━<br/>no D-state processes"]
        DoctorWarnMsg["WARNING: lists stuck PIDs<br/>━━━━━━━━━━<br/>'pid=X pcpu=Y' for each D-state claude proc<br/>includes full breakdown"]
    end

    %% ─────────────────────────────────── %%
    %% CONNECTIONS — FLOW A                %%
    %% ─────────────────────────────────── %%
    START --> PersistProc
    PersistProc --> SnapTick
    SnapTick --> ReadSnap
    ReadSnap --> ProcDied
    ProcDied -->|"yes — generator returns"| TrigFire
    ProcDied -->|"no"| AccSnap
    AccSnap -->|"sleep interval"| SnapTick

    TrigFire --> StopTrace
    StopTrace --> FinalSnap
    FinalSnap --> DetectEntry

    DetectEntry --> SnapIter
    SnapIter -->|"each snapshot"| DStateGate
    DStateGate -->|"yes — non-benign D-state"| DStateCount
    DStateGate -->|"no — reset counter"| CPUGate
    DStateCount -->|">= 2"| EmitDState
    DStateCount -->|"< 2"| CPUGate
    EmitDState --> CPUGate
    CPUGate -->|"yes"| CPUCount
    CPUGate -->|"no — reset counter"| OtherDet
    CPUCount -->|">= 2"| EmitHighCPU
    CPUCount -->|"< 2"| OtherDet
    EmitHighCPU --> OtherDet
    OtherDet -->|"next snapshot"| SnapIter
    SnapIter -->|"loop done"| PostLoop

    PostLoop --> WriteTrace
    WriteTrace --> HasAnom
    HasAnom -->|"yes"| WriteAnom
    HasAnom -->|"no"| WriteSummary
    WriteAnom --> WriteSummary
    WriteSummary --> AppendIdx
    AppendIdx --> Retention
    Retention --> SESSION_DONE

    %% ─────────────────────────────────── %%
    %% CONNECTIONS — FLOW B                %%
    %% ─────────────────────────────────── %%
    START2 --> RunPS
    RunPS --> FilterClaude
    FilterClaude -->|"match"| AnyProcs
    FilterClaude -->|"skip row"| FilterClaude
    AnyProcs -->|"none found"| DoctorOKMsg
    AnyProcs -->|"found"| BuildBreakdown
    BuildBreakdown --> CheckDStatePS
    CheckDStatePS -->|"no D-state"| DoctorOKMsg
    CheckDStatePS -->|"D-state found"| DoctorWarnMsg
    DoctorOKMsg --> DOCTOR_OK
    DoctorWarnMsg --> DOCTOR_WARN

    %% ─────────────────────────────────── %%
    %% CLASS ASSIGNMENTS                   %%
    %% ─────────────────────────────────── %%
    class START,START2,SESSION_DONE,DOCTOR_OK,DOCTOR_WARN terminal;
    class PersistProc,ReadSnap,AccSnap,FinalSnap,TrigFire,StopTrace handler;
    class DStateGate,DStateCount,CPUGate,CPUCount,SnapTick,ProcDied stateNode;
    class DetectEntry,SnapIter phase;
    class EmitDState,EmitHighCPU,OtherDet,PostLoop detector;
    class WriteTrace,WriteAnom,WriteSummary,AppendIdx,Retention output;
    class RunPS,FilterClaude,AnyProcs,BuildBreakdown,CheckDStatePS,DoctorOKMsg,DoctorWarnMsg handler;
```

### Operational Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% ── COLUMN A: autoskillit doctor ── %%
    subgraph Doctor ["AUTOSKILLIT DOCTOR  (autoskillit doctor)"]
        direction TB
        DOCTOR_CMD["● autoskillit doctor<br/>━━━━━━━━━━<br/>15 project-health checks<br/>--json flag available<br/>cli/_doctor.py"]

        subgraph DoctorChecks ["Checks 1–14 (unchanged)"]
            direction TB
            CHECKS_PREV["stale_mcp_servers · mcp_server_registered<br/>autoskillit_on_path · project_config<br/>hook_health · hook_registration<br/>version_consistency · script_version_health<br/>gitignore · secret_scanning · quota_cache_schema<br/>source_version_drift · editable_install · stale_entry_points"]
        end

        CHECK15["● Check 15: claude_process_state<br/>━━━━━━━━━━<br/>ps -axo pid,state,pcpu,comm<br/>Detects D-state + high-CPU<br/>Reports per-process breakdown<br/>_check_claude_process_state_breakdown()"]
    end

    %% ── COLUMN B: runtime process monitoring ── %%
    subgraph Runtime ["HEADLESS SESSION RUNTIME  (execution/linux_tracing.py)"]
        direction TB
        PROC_MON["● proc_monitor(pid)<br/>━━━━━━━━━━<br/>Async generator, 5 s interval<br/>Persists psutil.Process handle<br/>Primes cpu_percent baseline<br/>Monotonic captured_at guarantee"]

        PROC_SNAP["● ProcSnapshot (dataclass)<br/>━━━━━━━━━━<br/>state · vm_rss_kb · threads<br/>fd_count · fd_soft_limit<br/>sig_pnd · sig_blk · sig_cgt<br/>oom_score · wchan<br/>★ cpu_percent (new field)"]

        TMPFS["/dev/shm/autoskillit_trace_{pid}.jsonl<br/>━━━━━━━━━━<br/>Crash-resilient tmpfs stream<br/>Line-buffered JSONL<br/>Recovered on next startup"]
    end

    %% ── COLUMN C: anomaly detection ── %%
    subgraph AnomalyDetect ["ANOMALY DETECTION  (execution/anomaly_detection.py)"]
        direction TB
        DETECT["● detect_anomalies(snapshots, pid)<br/>━━━━━━━━━━<br/>Post-hoc, runs at flush time<br/>Multi-snapshot pattern matching"]

        subgraph ExistingAnomalies ["Existing Anomaly Kinds"]
            direction TB
            EXISTING["OOM_SPIKE · OOM_CRITICAL<br/>ZOMBIE_DETECTED · ZOMBIE_PERSISTENT<br/>SIGNALS_PENDING · RSS_GROWTH · FD_HIGH"]
        end

        D_STATE["● D_STATE_SUSTAINED<br/>━━━━━━━━━━<br/>state == disk-sleep<br/>wchan ∉ BENIGN_WCHANS<br/>≥ 2 consecutive snapshots<br/>severity: WARNING"]

        HIGH_CPU["● HIGH_CPU_SUSTAINED<br/>━━━━━━━━━━<br/>cpu_percent ≥ 90 %<br/>≥ 2 consecutive snapshots<br/>severity: WARNING"]

        BENIGN["● BENIGN_WCHANS filter<br/>━━━━━━━━━━<br/>do_nanosleep · do_epoll_wait<br/>schedule_hrtimeout_range<br/>empty · 0  (benign D-states)"]
    end

    %% ── COLUMN D: session diagnostics logs ── %%
    subgraph SessionLogs ["SESSION DIAGNOSTICS LOGS  (execution/session_log.py)"]
        direction TB
        FLUSH["flush_session_log()<br/>━━━━━━━━━━<br/>Called post-session<br/>Runs detect_anomalies()<br/>Writes structured files"]

        ANOMALIES_OUT["anomalies.jsonl<br/>━━━━━━━━━━<br/>Written only if anomalies exist<br/>Includes D_STATE_SUSTAINED<br/>and HIGH_CPU_SUSTAINED records<br/>~/.local/share/autoskillit/logs/"]

        SUMMARY["summary.json<br/>━━━━━━━━━━<br/>anomaly_count field<br/>peak_rss_kb · peak_oom_score<br/>peak_fd_ratio"]

        INDEX["sessions.jsonl (append-only index)<br/>━━━━━━━━━━<br/>anomaly_count per entry<br/>jq-queryable for failed sessions<br/>500-session retention"]
    end

    %% ── FLOWS ── %%
    DOCTOR_CMD --> CHECKS_PREV
    DOCTOR_CMD --> CHECK15

    PROC_MON -->|"yields every 5 s"| PROC_SNAP
    PROC_SNAP -->|"streamed (line-buffered)"| TMPFS
    PROC_SNAP -->|"accumulated in memory"| FLUSH

    FLUSH --> DETECT
    DETECT --> EXISTING
    DETECT --> D_STATE
    DETECT --> HIGH_CPU
    D_STATE -.->|"wchan filtered by"| BENIGN
    HIGH_CPU -.->|"threshold: cpu_percent"| PROC_SNAP

    DETECT -->|"anomalies list"| ANOMALIES_OUT
    FLUSH --> SUMMARY
    FLUSH --> INDEX

    %% ── CLASS ASSIGNMENTS ── %%
    class DOCTOR_CMD cli;
    class CHECKS_PREV,CHECK15 phase;
    class PROC_MON handler;
    class PROC_SNAP stateNode;
    class TMPFS,ANOMALIES_OUT,SUMMARY,INDEX output;
    class DETECT handler;
    class EXISTING,D_STATE,HIGH_CPU,BENIGN detector;
    class FLUSH handler;
```

Closes #760

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260412-120704-347557/.autoskillit/temp/make-plan/detect_sustained_d_state_and_high_cpu_plan_2026-04-12_121132.md`

## Token Usage Summary

| plan | 244 | 32.4k | 1.5M | 137.0k | 1 | 8m 5s |
| verify | 2.0k | 11.0k | 439.5k | 42.8k | 1 | 6m 23s |
| implement | 554 | 34.9k | 5.2M | 104.3k | 1 | 9m 40s |
| prepare_pr | 52 | 6.8k | 167.0k | 29.8k | 1 | 1m 50s |
| run_arch_lenses | 202 | 29.0k | 636.0k | 164.6k | 3 | 10m 50s |
| compose_pr | 59 | 8.6k | 208.3k | 29.9k | 1 | 2m 34s |
| **Total** | 3.1k | 122.7k | 8.2M | 508.4k | | 39m 25s |

### DIFF
--- a/src/autoskillit/cli/_doctor.py
+++ b/src/autoskillit/cli/_doctor.py
@@ -462,7 +462,7 @@ def _check_claude_process_state_breakdown() -> DoctorResult:
         if len(parts) < 4:
             continue
         comm = parts[3]
-        if comm != "claude" and not comm.endswith("/claude"):
+        if comm != "claude":
             continue
         try:
             claude_rows.append((int(parts[0]), parts[1], float(parts[2])))
--- a/src/autoskillit/cli/_doctor.py
+++ b/src/autoskillit/cli/_doctor.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 import shutil
+import subprocess
 import urllib.parse
 from dataclasses import dataclass
 from pathlib import Path
@@ -32,8 +33,6 @@ class DoctorResult:
 
 def _check_mcp_server_registered(claude_json_path: Path | None = None) -> DoctorResult:
     """Check that autoskillit MCP server is registered (via mcpServers or plugin)."""
-    import subprocess
-
     if claude_json_path is None:
         claude_json_path = Path.home() / ".claude.json"
 
@@ -268,8 +267,6 @@ def _check_editable_install_source_exists() -> DoctorResult:
 
 def _check_stale_entry_points() -> DoctorResult:
     """Detect autoskillit binaries on PATH outside ~/.local/bin (stale/poisoned installs)."""
-    import subprocess
-
     check_name = "stale_entry_points"
     primary = shutil.which("autoskillit")
     if not primary:
@@ -432,6 +429,67 @@ def _check_quota_cache_schema(cache_path: Path | None = None) -> DoctorResult:
         check_name,
         f"Quota cache schema drift at {path}: observed={observed!r}, "
         f"expected={QUOTA_CACHE_SCHEMA_VERSION}.",
+    )
+
+
+def _check_claude_process_state_breakdown() -> DoctorResult:
+    """Check current D-state and CPU usage of claude processes via ps."""
+    check_name = "claude_process_state"
+    try:
+        result = subprocess.run(
+            ["ps", "-axo", "pid,state,pcpu,comm"],
+            capture_output=True,
+            text=True,
+            timeout=5,
+        )
+    except (FileNotFoundError, subprocess.TimeoutExpired, OSError) as exc:
+        return DoctorResult(
+            Severity.OK,
+            check_name,
+            f"ps unavailable ({type(exc).__name__}); skipping claude process check",
+        )
+
+    if result.returncode != 0:
+        return DoctorResult(
+            Severity.OK,
+            check_name,
+            f"ps exited {result.returncode}; skipping claude process check",
+        )
+
+    claude_rows: list[tuple[int, str, float]] = []
+    for line in result.stdout.splitlines()[1:]:
+        parts = line.split(maxsplit=3)
+        if len(parts) < 4:
+            continue
+        comm = parts[3]
+        if comm != "claude" and not comm.endswith("/claude"):
+            continue
+        try:
+            claude_rows.append((int(parts[0]), parts[1], float(parts[2])))
+        except ValueError:
+            continue
+
+    if not claude_rows:
+        return DoctorResult(Severity.OK, check_name, "No claude processes running")
+
+    breakdown: dict[str, int] = {}
+    for _, state, _ in claude_rows:
+        breakdown[state] = breakdown.get(state, 0) + 1
+
+    summary = ", ".join(f"{s}={c}" for s, c in sorted(breakdown.items()))
+
+    d_rows = [f"pid={pid} pcpu={pcpu}" for pid, state, pcpu in claude_rows if state == "D"]
+    if d_rows:
+        return DoctorResult(
+            Severity.WARNING,
+            check_name,
+            f"claude processes in D state: {', '.join(d_rows)} (breakdown: {summary})",
+        )
+
+    return DoctorResult(
+        Severity.OK,
+        check_name,
+        f"claude process state breakdown: {summary}",
     )
 
 
@@ -635,6 +693,9 @@ def run_doctor(*, output_json: bool = False) -> None:
 
     # Check 14: Quota cache schema version
     results.append(_check_quota_cache_schema())
+
+    # Check 15: claude process state breakdown
+    results.append(_check_claude_process_state_breakdown())
 
     # Output
     if output_json:
--- a/src/autoskillit/execution/anomaly_detection.py
+++ b/src/autoskillit/execution/anomaly_detection.py
@@ -18,12 +18,29 @@ class AnomalyKind(StrEnum):
     SIGNALS_PENDING = "signals_pending"
     RSS_GROWTH = "rss_growth"
     FD_HIGH = "fd_high"
+    D_STATE_SUSTAINED = "d_state_sustained"
+    HIGH_CPU_SUSTAINED = "high_cpu_sustained"
 
 
 class AnomalySeverity(StrEnum):
     INFO = "info"
     WARNING = "warning"
     CRITICAL = "critical"
+
+
+# Kernel wait-channel values that are normal for a healthy process in
+# uninterruptible sleep.  D-state snapshots whose wchan matches any of these
+# are NOT counted toward the D_STATE_SUSTAINED threshold.
+# "" and "0" guard against missing-data snapshots (unreadable /proc/wchan).
+BENIGN_WCHANS: frozenset[str] = frozenset(
+    {
+        "do_nanosleep",
+        "do_epoll_wait",
+        "schedule_hrtimeout_range",
+        "",  # /proc returns empty when unreadable — treat as benign
+        "0",  # kernel reports literal "0" when thread is runnable
+    }
+)
 
 
 def _anomaly(
@@ -60,6 +77,8 @@ def detect_anomalies(
         return anomalies
 
     consecutive_zombie = 0
+    consecutive_d_state = 0
+    consecutive_high_cpu = 0
     prev_sig_pnd: str | None = None
     initial_rss: int | None = None
 
@@ -128,6 +147,49 @@ def detect_anomalies(
                 )
         else:
             consecutive_zombie = 0
+
+        # D-state sustained: process stuck in uninterruptible sleep
+        wchan = snap.get("wchan", "")
+        if state == "disk-sleep" and isinstance(wchan, str) and wchan not in BENIGN_WCHANS:
+            consecutive_d_state += 1
+            if consecutive_d_state >= 2:
+                anomalies.append(
+                    _anomaly(
+                        AnomalyKind.D_STATE_SUSTAINED,
+                        AnomalySeverity.WARNING,
+                        {
+                            "state": state,
+                            "wchan": wchan,
+                            "consecutive_count": consecutive_d_state,
+                        },
+                        snap,
+                        seq,
+                        pid,
+                    )
+                )
+        else:
+            consecutive_d_state = 0
+
+        # High-CPU sustained: process burning CPU >= 90% (suspected infinite loop)
+        cpu_percent = snap.get("cpu_percent", 0.0)
+        if isinstance(cpu_percent, (int, float)) and cpu_percent >= 90.0:
+            consecutive_high_cpu += 1
+            if consecutive_high_cpu >= 2:
+                anomalies.append(
+                    _anomaly(
+                        AnomalyKind.HIGH_CPU_SUSTAINED,
+                        AnomalySeverity.WARNING,
+                        {
+                            "cpu_percent": float(cpu_percent),
+                            "consecutive_count": consecutive_high_cpu,
+                        },
+                        snap,
+                        seq,
+                        pid,
+                    )
+                )
+        else:
+            consecutive_high_cpu = 0
 
         # Signals pending: transition from all-zeros to non-zero
         _all_zeros = "0000000000000000"
--- a/src/autoskillit/execution/linux_tracing.py
+++ b/src/autoskillit/execution/linux_tracing.py
@@ -78,6 +78,8 @@ class ProcSnapshot:
     sig_cgt: str
     oom_score: int
     wchan: str
+    # CPU utilisation (0.0 when process arg is not supplied to read_proc_snapshot)
+    cpu_percent: float
 
 
 def _parse_proc_status(content: str) -> dict[str, str]:
@@ -101,13 +103,19 @@ def _parse_proc_status(content: str) -> dict[str, str]:
     return fields
 
 
-def read_proc_snapshot(pid: int) -> ProcSnapshot | None:
-    """Read a complete snapshot for pid. Returns None if process gone."""
+def read_proc_snapshot(pid: int, *, process: psutil.Process | None = None) -> ProcSnapshot | None:
+    """Read a complete snapshot for pid. Returns None if process gone.
+
+    When *process* is provided, it is reused instead of constructing a fresh
+    psutil.Process(pid); cpu_percent(interval=0) then returns a meaningful
+    delta against the baseline primed by the caller.  When *process* is None,
+    cpu_percent defaults to 0.0.
+    """
     if not LINUX_TRACING_AVAILABLE:
         return None
     captured_at = datetime.now(UTC).isoformat()
     try:
-        p = psutil.Process(pid)
+        p = process if process is not None else psutil.Process(pid)
         with p.oneshot():
             state = p.status()
             mem = p.memory_info()
@@ -115,6 +123,7 @@ def read_proc_snapshot(pid: int) -> ProcSnapshot | None:
             num_fds = p.num_fds()
             fd_soft_limit = p.rlimit(psutil.RLIMIT_NOFILE)[0]
             ctx = p.num_ctx_switches()
+            cpu_pct = p.cpu_percent(interval=0) if process is not None else 0.0
     except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
         return None
 
@@ -149,6 +158,7 @@ def read_proc_snapshot(pid: int) -> ProcSnapshot | None:
         sig_cgt=sig_fields.get("sig_cgt", ""),
         oom_score=oom,
         wchan=wchan,
+        cpu_percent=cpu_pct,
     )
 
 
@@ -161,8 +171,13 @@ async def proc_monitor(pid: int, interval: float = 5.0) -> AsyncIterator[ProcSna
     maintain the monotonic ordering invariant at the production site.
     """
     _last_captured_at: str = ""
+    try:
+        _proc = psutil.Process(pid)
+        _proc.cpu_percent(interval=0)  # prime the psutil-internal baseline
+    except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
+        return
     while True:
-        snap = read_proc_snapshot(pid)
+        snap = read_proc_snapshot(pid, process=_proc)
         if snap is None:
             return
         captured_at = snap.captured_at
--- a/tests/cli/test_doctor.py
+++ b/tests/cli/test_doctor.py
@@ -1466,3 +1466,83 @@ def test_doctor_reports_drift_in_project_scope(
     assert result.severity == Severity.ERROR
     assert "[project]" in result.message
     assert "pretty_output.py" in result.message
+
+
+# ---------------------------------------------------------------------------
+# REQ-DOCTOR-001 — _check_claude_process_state_breakdown
+# ---------------------------------------------------------------------------
+
+
+class TestCheckClaudeProcessStateBreakdown:
+    """Tests for the claude_process_state doctor check (Check 15)."""
+
+    def _ps_result(self, stdout: str, returncode: int = 0):
+        return type(
+            "CompletedProcess",
+            (),
+            {"returncode": returncode, "stdout": stdout},
+        )()
+
+    def test_ok_when_only_sleeping(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        """Single sleeping claude process → Severity.OK with state breakdown."""
+        import subprocess
+
+        from autoskillit.cli._doctor import Severity, _check_claude_process_state_breakdown
+
+        header = "PID STAT %CPU COMMAND\n"
+        monkeypatch.setattr(
+            subprocess,
+            "run",
+            lambda *a, **kw: self._ps_result(header + "1234 S 0.5 claude"),
+        )
+        result = _check_claude_process_state_breakdown()
+        assert result.severity == Severity.OK
+        assert "S=1" in result.message
+
+    def test_warns_on_d_state(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        """claude process in D state → Severity.WARNING with pid and pcpu in message."""
+        import subprocess
+
+        from autoskillit.cli._doctor import Severity, _check_claude_process_state_breakdown
+
+        header = "PID STAT %CPU COMMAND\n"
+        monkeypatch.setattr(
+            subprocess,
+            "run",
+            lambda *a, **kw: self._ps_result(header + "1234 D 99.0 claude"),
+        )
+        result = _check_claude_process_state_breakdown()
+        assert result.severity == Severity.WARNING
+        assert "D=1" in result.message
+        assert "99.0" in result.message
+
+    def test_ok_when_no_claude_processes(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        """Empty ps output (no claude rows) → Severity.OK, 'No claude processes running'."""
+        import subprocess
+
+        from autoskillit.cli._doctor import Severity, _check_claude_process_state_breakdown
+
+        header = "PID STAT %CPU COMMAND\n"
+        monkeypatch.setattr(
+            subprocess,
+            "run",
+            lambda *a, **kw: self._ps_result(header + "5678 S 0.1 python"),
+        )
+        result = _check_claude_process_state_breakdown()
+        assert result.severity == Severity.OK
+        assert result.message == "No claude processes running"
+
+    def test_ok_when_ps_missing(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        """FileNotFoundError from ps → Severity.OK explaining ps unavailability."""
+        import subprocess
+
+        from autoskillit.cli._doctor import Severity, _check_claude_process_state_breakdown
+
+        def _raise(*a, **kw):
+            raise FileNotFoundError("ps")
+
+        monkeypatch.setattr(subprocess, "run", _raise)
+        result = _check_claude_process_state_breakdown()
+        assert result.severity == Severity.OK
+        assert "ps unavailable" in result.message
+        assert "FileNotFoundError" in result.message
--- a/tests/docs/test_doc_counts.py
+++ b/tests/docs/test_doc_counts.py
@@ -203,6 +203,14 @@ def test_quota_thresholds_defaults() -> None:
 
 
 def test_doctor_check_count_is_17() -> None:
+    # _count_doctor_checks() counts every "# Check N:" and "# Check Nb:" marker
+    # in run_doctor() — 15 numbered base markers + 2 lettered sub-check markers
+    # (4b, 7b) = 17 total.  test_installation_states_15_doctor_checks checks the
+    # *user-visible* count from docs/installation.md ("13 numbered + 2 lettered
+    # sub-checks 4b and 7b = 15").  The gap of 2 is intentional: Check 4 and
+    # Check 7 each appear as separate implementation markers but the docs
+    # present them as single numbered entries that subsume their b variants.
+    # Update both tests whenever a new doctor check is added.
     assert _count_doctor_checks() == 17, (
         f"Expected 17 doctor checks; found {_count_doctor_checks()}"
     )
--- a/tests/docs/test_doc_counts.py
+++ b/tests/docs/test_doc_counts.py
@@ -202,9 +202,9 @@ def test_quota_thresholds_defaults() -> None:
     assert long_ == pytest.approx(98.0)
 
 
-def test_doctor_check_count_is_16() -> None:
-    assert _count_doctor_checks() == 16, (
-        f"Expected 16 doctor checks; found {_count_doctor_checks()}"
+def test_doctor_check_count_is_17() -> None:
+    assert _count_doctor_checks() == 17, (
+        f"Expected 17 doctor checks; found {_count_doctor_checks()}"
     )
 
 
@@ -277,8 +277,8 @@ def test_configuration_states_quota_thresholds() -> None:
     )
 
 
-def test_installation_states_14_doctor_checks() -> None:
-    _assert_doc_states_number(DOCS_DIR / "installation.md", "doctor checks", 14)
+def test_installation_states_15_doctor_checks() -> None:
+    _assert_doc_states_number(DOCS_DIR / "installation.md", "doctor checks", 15)
 
 
 def test_recipes_overview_states_5_recipes() -> None:
--- a/tests/execution/test_anomaly_detection.py
+++ b/tests/execution/test_anomaly_detection.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from autoskillit.execution.anomaly_detection import (
+    BENIGN_WCHANS,
     AnomalyKind,
     AnomalySeverity,
     detect_anomalies,
@@ -23,6 +24,7 @@ def _snap(
     wchan: str = "",
     ctx_switches_voluntary: int = 500,
     ctx_switches_involuntary: int = 20,
+    cpu_percent: float = 0.0,
 ) -> dict[str, object]:
     return {
         "state": state,
@@ -37,6 +39,7 @@ def _snap(
         "wchan": wchan,
         "ctx_switches_voluntary": ctx_switches_voluntary,
         "ctx_switches_involuntary": ctx_switches_involuntary,
+        "cpu_percent": cpu_percent,
     }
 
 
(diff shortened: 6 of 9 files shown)


src/autoskillit/cli/_doctor.py
```diff
@@ -426,6 +423,67 @@ def _check_quota_cache_schema(cache_path: Path | None = None) -> DoctorResult:
     )
 
 
+def _check_claude_process_state_breakdown() -> DoctorResult:
```
Trecek, line 435: [warning] arch: Process-state inspection logic (ps parsing, D-state detection, CPU breakdown) is implemented directly in L3 CLI instead of delegating to the existing L1 execution abstraction in linux_tracing.py. This duplicates domain knowledge about process states across layers: L1 already has read_proc_snapshot() / psutil-based process inspection. The doctor check should call an L1 helper rather than re-implementing ps-based process parsing in the CLI layer.

src/autoskillit/execution/linux_tracing.py
```diff
@@ -101,20 +103,27 @@ def _parse_proc_status(content: str) -> dict[str, str]:
...
-        p = psutil.Process(pid)
+        p = process if process is not None else psutil.Process(pid)
         with p.oneshot():
             state = p.status()
             mem = p.memory_info()
             num_threads = p.num_threads()
             num_fds = p.num_fds()
             fd_soft_limit = p.rlimit(psutil.RLIMIT_NOFILE)[0]
             ctx = p.num_ctx_switches()
+            cpu_pct = p.cpu_percent(interval=0) if process is not None else 0.0
```
Trecek, line 126: [warning] defense: cpu_percent(interval=0) is called inside a p.oneshot() context manager. psutil's cpu_percent uses its own internal timing state and calling it inside oneshot() may yield unreliable results — oneshot() is documented for caching status/memory/etc, not for cpu_percent which depends on wall-clock delta. The call should be moved outside the oneshot() block to guarantee a correct measurement.

src/autoskillit/execution/anomaly_detection.py
```diff
@@ -129,6 +148,49 @@ def detect_anomalies(
         else:
             consecutive_zombie = 0
 
+        # D-state sustained: process stuck in uninterruptible sleep
+        wchan = snap.get("wchan", "")
+        if state == "disk-sleep" and isinstance(wchan, str) and wchan not in BENIGN_WCHANS:
```
Trecek, line 153: [warning] defense: The D-state branch checks isinstance(wchan, str) but only after reading snap.get('wchan', ''). If wchan is not a str (e.g. None stored in the snapshot dict), the condition short-circuits to consecutive_d_state = 0 silently. A non-string wchan from a malformed snapshot suppresses detection without any diagnostic. The guard should explicitly validate and log unexpected wchan types.

Trecek (commented):

(Note: posting as COMMENT since reviewer is PR author; verdict is changes_requested based on findings below.)

tests/execution/test_linux_tracing.py
```diff
@@ -314,3 +317,48 @@ async def test_proc_monitor_snapshots_have_distinct_timestamps():
...
+
+    This is only possible when a single psutil.Process is reused across iterations
+    and its baseline is primed before the loop. A fresh Process per iteration always
+    returns 0.0 on the first cpu_percent(interval=0) call.
+    """
+    import subprocess
+
+    from autoskillit.execution.linux_tracing import proc_monitor
+
+    proc = subprocess.Popen(
```
Trecek, line 334: [warning] defense: The CPU-bound subprocess test spawns an infinite loop process with no timeout guard on the async-for loop. If proc_monitor never yields 5 snapshots, the test blocks indefinitely. Wrap the async iteration with asyncio.wait_for or an anyio timeout to prevent unbounded hang in CI.

src/autoskillit/cli/_doctor.py
```diff
@@ -426,6 +423,67 @@ def _check_quota_cache_schema(cache_path: Path | None = None) -> DoctorResult:
...
+    if not claude_rows:
+        return DoctorResult(Severity.OK, check_name, "No claude processes running")
+
+    breakdown: dict[str, int] = {}
+    for _, state, _ in claude_rows:
+        breakdown[state] = breakdown.get(state, 0) + 1
+
+    summary = ", ".join(f"{s}={c}" for s, c in sorted(breakdown.items()))
+
+    d_rows = [f"pid={pid} pcpu={pcpu}" for pid, state, pcpu in claude_rows if state == "D"]
```
Trecek, line 481: [warning] bugs: D-state filter uses exact equality state == "D" but ps -axo state can emit multi-character state strings such as "D+" (foreground) or "Ds" (swapped D). These are missed, silently reporting OK when a foreground or swapped claude process is stuck in uninterruptible sleep. Use state.startswith("D") or "D" in state instead.

src/autoskillit/execution/anomaly_detection.py
```diff
@@ -26,6 +28,21 @@ class AnomalySeverity(StrEnum):
     CRITICAL = "critical"
 
 
+# Kernel wait-channel values that are normal for a healthy process in
+# uninterruptible sleep.  D-state snapshots whose wchan matches any of these
+# are NOT counted toward the D_STATE_SUSTAINED threshold.
+# "" and "0" guard against missing-data snapshots (unreadable /proc/wchan).
+BENIGN_WCHANS: frozenset[str] = frozenset(
```
Trecek, line 35: [warning] cohesion: BENIGN_WCHANS constant is defined mid-file between the AnomalySeverity enum and the _anomaly() helper. Module-level configuration constants in this file should be grouped near the top (with imports/enums) rather than interleaved with function definitions.

tests/execution/test_session_log_integration.py
```diff
@@ -70,3 +70,68 @@ async def test_full_tracing_pipeline_writes_distinct_timestamps(tmp_path):
         "duration_seconds in summary.json must reflect monotonic elapsed, "
         "not wall-clock subtraction"
     )
+
+
+# ---------------------------------------------------------------------------
+# REQ-DIAG-001 — new anomaly kinds surface in anomalies.jsonl
+# ---------------------------------------------------------------------------
+
+_BASE_SNAP: dict[str, object] = {
```
Trecek, line 79: [warning] cohesion: _BASE_SNAP dict duplicates the role of the _snap() helper factory already present in tests/execution/test_anomaly_detection.py. Having two partial base dicts for the same snapshot shape creates two sources of truth. The session-log integration tests should reuse _snap() or extract the helper to a shared conftest.

tests/execution/test_anomaly_detection.py
```diff
@@ -137,3 +140,217 @@ def test_anomaly_record_structure():
     assert "detail" in anomaly
     assert "snapshot" in anomaly
     assert anomaly["pid"] == 1234
+
+
```
Trecek, line 146: [warning] slop: Excessive section separator comment blocks: 13 separator blocks are used for 13 individual tests, one-per-test. Separators should group related clusters of tests, not wrap every single test function. This is AI-generated over-annotation.

tests/execution/test_linux_tracing.py
```diff
@@ -314,3 +317,48 @@ async def test_proc_monitor_snapshots_have_distinct_timestamps():
...
     timestamps = [s.captured_at for s in result]
     assert len(set(timestamps)) == len(timestamps), "All captured_at must be unique"
+
+
+# ---------------------------------------------------------------------------
+# design prerequisite — cpu_percent field
+# ---------------------------------------------------------------------------
+
+
+def test_read_proc_snapshot_returns_cpu_percent_field():
```
Trecek: [warning] slop: test_read_proc_snapshot_returns_cpu_percent_field duplicates assertions already present in the pre-existing test_read_proc_snapshot_has_all_fields (lines L75-L77 in the diff hunk). The new test adds no additional coverage and should be removed.

tests/execution/test_linux_tracing.py
```diff
@@ -314,3 +317,48 @@ async def test_proc_monitor_snapshots_have_distinct_timestamps():
...
+        [sys.executable, "-c", "while True: pass"],
+    )
+    try:
+        snaps = []
+        async for snap in proc_monitor(proc.pid, interval=0.1):
+            snaps.append(snap)
+            if len(snaps) >= 5:
+                break
+        assert len(snaps) >= 3, "Need at least 3 snapshots"
+        assert any(s.cpu_percent > 0.0 for s in snaps), (
```
Trecek, line 345: [warning] tests: The assertion any(s.cpu_percent > 0.0 for s in snaps) is environment-dependent: on a loaded CI host or under QEMU/WSL cpu_percent(interval=0) may legitimately return 0.0 for all snapshots even for a CPU-bound process, making this a potentially flaky test. Consider whether this assertion should be guarded with a @pytest.mark.flaky or made conditional on the platform.

tests/docs/test_doc_counts.py
```diff
@@ -202,9 +202,9 @@ def test_quota_thresholds_defaults() -> None:
     assert long_ == pytest.approx(98.0)
 
 
-def test_doctor_check_count_is_16() -> None:
-    assert _count_doctor_checks() == 16, (
-        f"Expected 16 doctor checks; found {_count_doctor_checks()}"
+def test_doctor_check_count_is_17() -> None:
```
Trecek, line 205: [warning] tests: test_doctor_check_count_is_17 asserts 17 implementation check functions while test_installation_states_15_doctor_checks asserts the doc states 15. If these track different things (implementation vs user-visible count), this should be documented with a comment explaining the distinction — the mismatch is non-obvious and risks future maintainers updating one without the other.

src/autoskillit/execution/linux_tracing.py
```diff
@@ -161,8 +171,13 @@ async def proc_monitor(pid: int, interval: float = 5.0) -> AsyncIterator[ProcSna
     maintain the monotonic ordering invariant at the production site.
     """
     _last_captured_at: str = ""
+    try:
+        _proc = psutil.Process(pid)
+        _proc.cpu_percent(interval=0)  # prime the psutil-internal baseline
+    except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
```
Trecek, line 177: [warning] defense: When psutil raises NoSuchProcess/AccessDenied/ZombieProcess during initial _proc construction, the generator silently returns without any logging. A caller cannot distinguish 'process never existed' from 'process died immediately'. Consider emitting a debug-level log before returning so transient failures are visible in diagnostics.

tests/execution/test_anomaly_detection.py
```diff
@@ -137,3 +140,217 @@ def test_anomaly_record_structure():
...
+    d_state = [a for a in anomalies if a["kind"] == AnomalyKind.D_STATE_SUSTAINED]
+    assert len(d_state) == 0
+
+
+# ---------------------------------------------------------------------------
+# REQ-WCHAN-002, REQ-TEST-005 — benign wchans are skipped
+# ---------------------------------------------------------------------------
+
+
+def test_d_state_sustained_benign_wchan_do_nanosleep_skipped():
```
Trecek: [info] tests: test_d_state_sustained_benign_wchan_epoll_skipped and test_d_state_sustained_benign_wchan_hrtimeout_skipped are structurally identical to the do_nanosleep variant. Consider consolidating into a single parametrized test over BENIGN_WCHANS members to reduce duplication.

tests/execution/test_anomaly_detection.py
```diff
@@ -137,3 +140,217 @@ def test_anomaly_record_structure():
...
     assert "snapshot" in anomaly
     assert anomaly["pid"] == 1234
+
+
+# ---------------------------------------------------------------------------
+# REQ-ENUM-001, REQ-ENUM-002
+# ---------------------------------------------------------------------------
+
+
+def test_anomaly_kind_has_d_state_sustained():
```
Trecek, line 152: [info] tests: test_anomaly_kind_has_d_state_sustained and test_anomaly_kind_has_high_cpu_sustained test the string value of StrEnum members. For StrEnum, the string value is auto-derived from the member name, so these assert a language-level tautology rather than application logic.

tests/cli/test_doctor.py
```diff
@@ -1431,3 +1431,83 @@ def test_check_quota_cache_schema_warning_includes_cache_path_and_observed_value
...
+
+# ---------------------------------------------------------------------------
+# REQ-DOCTOR-001 — _check_claude_process_state_breakdown
+# ---------------------------------------------------------------------------
+
+
+class TestCheckClaudeProcessStateBreakdown:
+    """Tests for the claude_process_state doctor check (Check 15)."""
+
+    def _ps_result(self, stdout: str, returncode: int = 0):
```
Trecek, line 1479: [info] tests: The _ps_result helper constructs a dynamic class using type() to simulate subprocess.CompletedProcess. subprocess.CompletedProcess can be instantiated directly, which would be cleaner and less fragile.

src/autoskillit/cli/_doctor.py
```diff
@@ -426,6 +423,67 @@ def _check_quota_cache_schema(cache_path: Path | None = None) -> DoctorResult:
...
+
+    if result.returncode != 0:
+        return DoctorResult(
+            Severity.OK,
+            check_name,
+            f"ps exited {result.returncode}; skipping claude process check",
+        )
+
+    claude_rows: list[tuple[int, str, float]] = []
+    for line in result.stdout.splitlines()[1:]:
```
Trecek, line 460: [info] defense: ps output parser skips result.stdout.splitlines()[1:] assuming the first line is always a header. On some platforms or non-standard ps implementations, the header may be absent or multi-line. A more defensive approach would match lines where parts[0] is a digit.

src/autoskillit/cli/_doctor.py
```diff
@@ -426,6 +423,67 @@ def _check_quota_cache_schema(cache_path: Path | None = None) -> DoctorResult:
...
+        parts = line.split(maxsplit=3)
+        if len(parts) < 4:
+            continue
+        comm = parts[3]
+        if comm != "claude" and not comm.endswith("/claude"):
```
Trecek: [info] defense: The filter comm.endswith('/claude') is unreachable in practice: ps -axo comm returns only the base name (max 15 chars), never a path with slashes. The check can be simplified to comm == 'claude' only.

src/autoskillit/execution/linux_tracing.py
```diff
@@ -78,6 +78,8 @@ class ProcSnapshot:
     sig_cgt: str
     oom_score: int
     wchan: str
+    # CPU utilisation (0.0 when process arg is not supplied to read_proc_snapshot)
+    cpu_percent: float
```
Trecek, line 82: [info] defense: cpu_percent field defaults to 0.0 when process=None, but 0.0 is also a valid measurement. The 'no measurement taken' state is indistinguishable from an idle process. Consider using Optional[float] = None as a sentinel or adding a cpu_percent_available: bool field.

tests/execution/test_linux_tracing.py
```diff
@@ -314,3 +317,48 @@ async def test_proc_monitor_snapshots_have_distinct_timestamps():
     assert len(result) >= 2
     timestamps = [s.captured_at for s in result]
     assert len(set(timestamps)) == len(timestamps), "All captured_at must be unique"
+
+
+# ---------------------------------------------------------------------------
```
Trecek: [info] slop: Section separator comment block added for a single new test. One test does not warrant a full separator block.

tests/execution/test_session_log_integration.py
```diff
@@ -70,3 +70,68 @@ async def test_full_tracing_pipeline_writes_distinct_timestamps(tmp_path):
         "duration_seconds in summary.json must reflect monotonic elapsed, "
         "not wall-clock subtraction"
     )
+
+
+# ---------------------------------------------------------------------------
```
Trecek, line 75: [info] slop: Section separator comment block with REQ-DIAG-001 reference wrapping two related tests. The REQ tag is untracked verbosity; the separator is disproportionate for two tests.

tests/cli/test_doctor.py
```diff
@@ -1431,3 +1431,83 @@ def test_check_quota_cache_schema_warning_includes_cache_path_and_observed_value
         assert result.severity == Severity.WARNING
         assert str(cache) in result.message
         assert "observed=1" in result.message
+
+
+# ---------------------------------------------------------------------------
```
Trecek, line 1471: [info] slop: Section separator comment block with REQ-DOCTOR-001 reference wrapping a single test class. The REQ tag is untracked verbosity.

src/autoskillit/execution/linux_tracing.py
```diff
@@ -78,6 +78,8 @@ class ProcSnapshot:
     sig_cgt: str
     oom_score: int
     wchan: str
+    # CPU utilisation (0.0 when process arg is not supplied to read_proc_snapshot)
```
Trecek, line 81: [info] slop: Inline field comment '# CPU utilisation (0.0 when process arg is not supplied to read_proc_snapshot)' restates implementation detail already covered by the updated docstring on read_proc_snapshot; redundant.

src/autoskillit/cli/_doctor.py
Trecek: Agreed — fixed in 07db28a. `ps -axo comm` returns only the base executable name (max 15 chars, no slashes), making `comm.endswith('/claude')` structurally unreachable. Simplified to `comm == 'claude'` only.

tests/execution/test_linux_tracing.py
Trecek: Agreed — fixed in 79e6b1f. Wrapped the async-for loop with `anyio.fail_after(10)` to bound the test runtime in CI.

tests/execution/test_anomaly_detection.py
Trecek: Agreed — fixed in ae55e9d. Collapsed the 11 one-per-test separator blocks, keeping only the two cluster-level separators (REQ-ENUM and REQ-WCHAN-002). The pre-existing tests have no per-test banners, so the new tests now follow the same style.

tests/execution/test_linux_tracing.py
Trecek: Agreed — fixed in 79e6b1f. `test_read_proc_snapshot_returns_cpu_percent_field` was removed; the `isinstance(snap.cpu_percent, float)` and `snap.cpu_percent >= 0.0` assertions were already present at lines 75-77 in `test_read_proc_snapshot_has_all_fields`.

tests/docs/test_doc_counts.py
Trecek: Agreed — fixed in 91ca0f8. Added a comment above `test_doctor_check_count_is_17` explaining the 17 vs 15 distinction: 17 = 15 numbered `# Check N:` markers + 2 lettered `# Check Nb:` markers; 15 (docs) = 13 user-visible numbered entries + 2 lettered sub-check entries (4b, 7b), since Check 4/4b and Check 7/7b are each presented as a single user-visible check in the docs.

tests/execution/test_anomaly_detection.py
Trecek: Agreed — fixed in ae55e9d. Consolidated the three structurally identical benign-wchan tests into a single `@pytest.mark.parametrize` test over `['do_nanosleep', 'do_epoll_wait', 'schedule_hrtimeout_range']`.

src/autoskillit/cli/_doctor.py
Trecek: Investigated — this is intentional. `read_proc_snapshot(pid: int, ...)` in `linux_tracing.py` is a single-PID deep-snapshot tool for session monitoring; it cannot enumerate all claude processes by name. The doctor check's job is system-wide process-table enumeration, which requires `ps -axo` (cross-platform, no psutil dependency). Additionally, `linux_tracing.py` is gated behind `LINUX_TRACING_AVAILABLE` (`sys.platform == 'linux'`), making it unsuitable for a general doctor check. The two code paths serve categorically different purposes.

src/autoskillit/execution/linux_tracing.py
Trecek: Investigated — this is intentional. psutil's own `oneshot()` docstring explicitly shows `p.cpu_percent()` as an example call inside the context manager, and the non-blocking path (`interval=0`) reads from instance-level fields (`_last_sys_cpu_times`, `_last_proc_cpu_times`) then calls `p.cpu_times()` — which IS cached under `oneshot()`. The baseline is primed at `proc_monitor` line 176 before the loop, so the delta is always correct. This is consistent with psutil's documented and recommended usage pattern.

src/autoskillit/execution/anomaly_detection.py
Trecek: Investigated — this is intentional. `ProcSnapshot.wchan: str` is typed as non-optional at `linux_tracing.py:80`. The field is always assigned a string (stripped `/proc/wchan` content or `''` on `OSError`). `dataclasses.asdict()` preserves the type, so a non-string `wchan` in a snapshot dict is structurally impossible. The `isinstance(wchan, str)` guard is a consistent defensive pattern applied uniformly to all dynamically-typed dict field reads in `detect_anomalies()` (same pattern at lines 96 and 175). Adding a log for a theoretically-impossible case would be dead code.

src/autoskillit/cli/_doctor.py
Trecek: Investigated — this is intentional. The invocation uses `ps -axo pid,state,pcpu,comm`. The `state` format specifier is an alias for `s` (single-character process state per the ps man page). Multi-character suffixes like `D+` or `Ds` only appear with the `stat` specifier. With `state`, the column is always exactly one character, so `state == 'D'` is correct and complete. The reviewer conflated the `state` (single-char) and `stat` (multi-char) ps format specifiers.

tests/execution/test_session_log_integration.py
Trecek: Investigated — this is intentional. `_BASE_SNAP` in `test_session_log_integration.py` includes a `captured_at` field (`'2026-01-01T00:00:00+00:00'`) that `_snap()` in `test_anomaly_detection.py` deliberately omits — `_snap()` produces dicts for `detect_anomalies()` which does not require `captured_at`, while `_BASE_SNAP` feeds `flush_session_log()` which writes `captured_at` as the `ts` field in `proc_trace.jsonl`. The two helpers serve different call-site contracts and are not duplicates of the same shape.

src/autoskillit/execution/linux_tracing.py
Trecek: Investigated — this is intentional. The `cpu_percent=0.0` sentinel is explicitly documented in two places: the field comment at `linux_tracing.py:81` ('`CPU utilisation (0.0 when process arg is not supplied to read_proc_snapshot)`') and the `read_proc_snapshot()` docstring ('When *process* is None, cpu_percent defaults to 0.0'). In practice, `proc_monitor()` always passes `process=_proc`, so real 0.0 values from a genuinely idle process come from the monitor path (not from one-shot calls with `process=None`). The design is intentionally documented.

src/autoskillit/execution/anomaly_detection.py
Trecek: Valid observation — flagged for design decision. Two reasonable conventions conflict: (1) group all module-level constants near the top with imports/enums, (2) place constants immediately above the code that uses them for locality. Both are valid. A human maintainer should decide which convention this module should follow before any change is made.

tests/execution/test_linux_tracing.py
Trecek: Valid observation — flagged for design decision. The assertion `any(s.cpu_percent > 0.0 for s in snaps)` is the functional specification being tested (psutil Process reuse with primed baseline). Removing it would eliminate the only test that verifies the core design. However, the WSL2 environment concern about `cpu_percent(interval=0)` returning 0.0 for a CPU-bound process is real. A human should decide whether to add `@pytest.mark.flaky`, a platform guard, or accept the risk.

src/autoskillit/execution/linux_tracing.py
Trecek: Valid observation — flagged for design decision. The diagnostic value of logging the early-return is real (a caller cannot distinguish 'process never existed' from 'process died immediately'). However, `linux_tracing.py` has no logging import or logger setup — adding one requires a design decision about which framework to use (stdlib `logging` vs `structlog` used elsewhere in the project). A human should decide before any change is made.

PR #808, opened 2026-04-13
Rectify: `channel_won` Branch Unconditionally SIGKILLs Claude CLI

### LINKED ISSUE #804: run_skill: channel_won branch unconditionally SIGKILLs Claude CLI, producing systematic exit_code=-9 + success=true ambiguity
## Observed symptom

- `resolve-review` session for PR #787 returned `exit_code: -9` (SIGKILL) but with a complete, successful result dict.
- Session ID: `46cbdc5b-1631-41db-a356-72c090c6f198`
- `summary.json` recorded simultaneously: `exit_code=-9`, `subtype="success"`, `cli_subtype="success"`, `termination_reason="completed"`, `success=true`, `write_call_count=2`, `write_path_warnings=[]`.
- The result content showed fixes applied (commit `2a6097c` on PR #799), tests passing, threads resolved (2/2), inline replies posted (8/0 failed) — session ran to completion.
- **Orchestrator-facing ambiguity:** pipeline-mode `run_skill` formatter emits `run_skill: OK [success]` immediately followed by `exit_code: -9` with no annotation, creating a cognitive contradiction for the LLM orchestrator. The orchestrator has to parse result content to decide what happened because the exit code says otherwise.

## Scope — systemic, not one-off

5 of the last 500 sessions in `~/.local/share/autoskillit/logs/sessions.jsonl` show the same `(exit_code=-9, success=true)` pattern across five different skills:

| step_name | skill |
|---|---|
| `compose_pr` | `/compose-pr` |
| `fix` | `/resolve-failures` |
| `verify` | `/dry-walkthrough` |
| `implement` | `/implement-worktree-no-merge` |
| `resolve_review` | `/resolve-review` |

All 5 share identical `peak_rss_kb=2176`, `peak_oom_score=666` (which is a **separate** tracer-wrong-PID issue — flagged below for its own issue).

## Root cause

`src/autoskillit/execution/process.py:289-297` — the `channel_won` branch in `run_managed_async`:

```python
else:
    proc_log.debug(
        "kill_decision",
        reason="channel_won",
        channel_a=signals.channel_a_confirmed,
        channel_b=signals.channel_b_status,
    )
    # Channel A or B won; process still alive — kill immediately.
    await async_kill_process_tree(proc.pid)
```

When Channel A (stdout heartbeat, 0.5 s poll for `type=result`) fires before `_watch_process` has observed natural exit (which is ~always, because the Claude CLI takes longer than 500 ms to flush, close fds, and return from `main()`), this branch runs `async_kill_process_tree` immediately. That sends `SIGTERM` → 2 s `psutil.wait_procs` → `SIGKILL`, so `proc.returncode == -9` for every monitored successful session.

**Git blame:** the unconditional kill was inherited from pre-anyio code in commit `f35c87690` (#72, "Anyio concurrency refactor", 2026-03-01). The pre-anyio code had separate `elif channel_a_confirmed:` and `elif channel_b_status == "completion":` branches, both ending in the kill. The refactor collapsed them into a single `else:` preserving semantics. **No commit message explained why immediate kill is required** — it was treated as mechanical preservation, not a deliberate design decision.

## Why the kill exists at all (important constraint)

The Claude CLI does **not** self-exit promptly after writing its final `type=result` output without `CLAUDE_CODE_EXIT_AFTER_STOP_DELAY`. `src/autoskillit/config/defaults.yaml:46` sets `exit_after_stop_delay_ms: 120000` (2 minutes). That 2-minute timer is the CLI's self-exit mechanism. Waiting 120 s per `run_skill` call is prohibitive, so AutoSkillit force-kills the CLI once channels confirm completion. **Any fix must preserve this constraint.**

## Adjudication is already correct — this is NOT a `_compute_success` bug

`src/autoskillit/execution/session.py:579-682` `_compute_success` explicitly handles `COMPLETED + nonzero returncode`. The authoritative comment at `session.py:647-649`:

> "The process was killed by our own `async_kill_process_tree` (signal -15 or -9), so a non-zero returncode is expected and trustworthy when the session envelope says 'success'."

And `session.py:799-801` in `_compute_retry`:

> "Infrastructure killed the process. SIGTERM/SIGKILL produce nonzero returncode by design — do not gate on returncode here."

So `success=True` + `exit_code=-9` is documented, intended behavior. **The ambiguity is NOT in adjudication** — it is in (a) surface presentation in the run_skill formatter and (b) the absence of a natural-exit grace window in the kill path.

## Session 46cbdc5b timeline (UTC)

```
10:18:30.104  start_ts; subprocess spawned
10:18:32.047  queue-operation enqueue (prompt with %%ORDER_UP::8221c391%%)
...           714 s of work; 97 assistant messages; 43 tool-use round-trips
10:29:30.864  Bash tool_use
10:29:33.477  TodoWrite tool_use
10:29:35.508  assistant message stop_reason=end_turn (thinking block only)
10:29:39.333  assistant text message — "resolve-review complete / Status: PASS"
              + %%ORDER_UP::8221c391%% token
10:29:40.009  seq=143 final proc_trace snapshot
  └─ Channel A heartbeat detects %%ORDER_UP:: in stdout within next 0.5 s
  └─ trigger.set(); await trigger.wait() returns
  └─ tg.cancel_scope.cancel() (process.py:249)
  └─ tracing_handle.stop()
  └─ resolve_termination() → COMPLETED + CHANNEL_A
  └─ process.py:289 else-branch → async_kill_process_tree(pid)
  └─ SIGTERM → 2 s wait → SIGKILL → 1 s kernel reap
  └─ stdout_file.close(); stderr_file.close()
  └─ read_temp_output() drains 714 s of PTY output
  └─ SubprocessResult(returncode=-9, termination=COMPLETED,
                      channel_confirmation=CHANNEL_A)
  └─ _build_skill_result → _compute_outcome → _compute_success → True
  └─ SkillResult(success=True, exit_code=-9)
10:30:25.041  end_ts (44.7 s after last snapshot)
              flush_session_log writes summary.json + appends sessions.jsonl
```

## Affected components (file:line)

- `src/autoskillit/execution/process.py:289-297` — unconditional kill; **the primary fix site**
- `src/autoskillit/execution/process.py:269-274` — existing `signals.process_exited` branch; demonstrates correct "no-kill-needed" shape
- `src/autoskillit/execution/_process_kill.py:16-58` — `kill_process_tree` SIGTERM→2s→SIGKILL→1s
- `src/autoskillit/execution/_process_race.py:236-250` — existing `completion_drain_timeout` pattern (precedent for bounded waits)
- `src/autoskillit/execution/_process_monitor.py:25-64` — `_heartbeat` / Channel A 0.5 s stdout poll
- `src/autoskillit/execution/session.py:579-682` — `_compute_success` (**already correct; do NOT change logic**)
- `src/autoskillit/execution/session.py:647-649` — documented `-15/-9` contract
- `src/autoskillit/execution/headless.py:539-876` — `_build_skill_result` / `_compute_outcome`
- `src/autoskillit/execution/headless.py:1062-1063` — `end_ts` computation from `time.monotonic()`
- `src/autoskillit/execution/headless.py:1121, 1136` — `flush_session_log` writes `termination_reason`
- `src/autoskillit/execution/commands.py:231` — injects `CLAUDE_CODE_EXIT_AFTER_STOP_DELAY` from config
- `src/autoskillit/config/defaults.yaml:46` — `exit_after_stop_delay_ms: 120000`
- `src/autoskillit/core/_type_results.py:130-157` — `SkillResult.exit_code` always serialized
- `src/autoskillit/hooks/_fmt_execution.py:30,52,90,104` — formatter unconditionally surfaces `exit_code` to orchestrator LLM
- `tests/execution/test_headless.py:1401,1420,1654` — COMPLETED+nonzero tests exist but only use `-15`, not `-9`
- `tests/execution/test_session_adjudication.py:118-140` — CHANNEL_A/B COMPLETED tests only use `-15`

## Test gaps

1. **No test asserts the `(success=True, exit_code=-9)` combination.** Every COMPLETED+nonzero returncode test uses `-15` (SIGTERM); `-9` is logically equivalent but unguarded against future drift (e.g. a misguided "treat -9 as OOM-killed, retry" refactor).
2. **No test exercises the `channel_won` branch at `process.py:289`.** Existing process tests use synthetic subprocesses that exit naturally; none simulate the race where Channel A wins before `_watch_process` observes exit.
3. **No test asserts formatter behavior for `run_skill` success + `exit_code=-9`** — the ambiguity is effectively "by spec" at the formatter layer.
4. **`CLAUDE_CODE_EXIT_AFTER_STOP_DELAY=120000` interaction with the kill path is untested** — whether the CLI's internal 2-minute timer could deliver a delayed SIGKILL racing with AutoSkillit's monitoring is not verified.

## Related prior commits (for context, NOT recurrence)

- `f35c87690` (#72, 2026-03-01) — Anyio concurrency refactor; introduced current unconditional kill shape
- `d15db94ef` — "fix(process): add symmetric drain window to eliminate asymmetric drain race" (drain-window precedent)
- `fa8d98032` — "fix(session): strengthen Channel A predicate to eliminate drain-race false negative"
- `5e46614c` (#568) — "Session Adjudication False-Positive — Recovery/Channel-B Contract Breach" (most similar rectify; tightened CHANNEL_B bypass invariant)
- `ae7b95a5` (#358) — "eliminate SkillResult dual-source subtype contradiction" (closest precedent for "same thing reported two different ways" bug)
- `225d904d` (#735) — "Stdout Idle Watchdog + Bounded Suppression"
- `a7ad04a4` (#750) — "Channel B Blind Spot" (expanded ChannelBStatus state space)
- `974920ed` (#793) — "Per-Step idle_output_timeout Override"

No prior `/autoskillit:investigate` found this specific root cause. **This is NOT a recurring bug** — standard remediation path is appropriate.

### Layer 1 — Architectural fix (primary; HIGH value)

Insert a bounded natural-exit grace window in `process.py:289-297` before the kill:

```python
else:  # channel_won
    proc_log.debug("kill_decision", reason="channel_won", ...)
    with anyio.move_on_after(natural_exit_grace):  # new config, default 3.0 s
        await process_exited_event.wait()
    if proc.returncode is not None:
        proc_log.debug("natural_exit_after_channel_won", returncode=proc.returncode)
        # fall through; no kill; returncode will be 0 for clean sessions
    else:
        await async_kill_process_tree(proc.pid)
```

Requirements:

- `_watch_process` must expose an `anyio.Event` set when `proc.wait()` returns (it already observes this internally — just needs to wire an event into the accumulator).
- Add `natural_exit_grace_seconds: 3.0` to `defaults.yaml`.
- **Must pair with** reduced `exit_after_stop_delay_ms` (e.g. 2000 ms), since the current 120000 ms default would block the CLI from self-exiting within any reasonable grace window.
- Validate against recording/replay fixtures in `tests/execution/recording/` — the paired change may break existing scenario recordings.

**Benefit:** most sessions record `exit_code=0` (true natural exit), eliminating the ambiguity at its source.
**Cost:** adds up to 3 s to every `run_skill` where Channel A won but the process would have exited naturally anyway.
**Risk:** if the Claude CLI genuinely hangs post-finalization for a path we haven't mapped, this extends session wall-clock by 3 s before the kill. Acceptable.

### Layer 2 — Surface fix (LOW effort, LOW value; optional)

In `src/autoskillit/hooks/_fmt_execution.py:30,52` annotate or suppress `exit_code` when `success=True`:

```
exit_code: -9 (infrastructure kill after completion signal — expected)
```

Or suppress entirely when `success=True`, since the orchestrator routes on `success`, not `exit_code`, per `tools_recipe.py:150-151`. Minimal cost; removes the cognitive contradiction immediately without any execution-layer changes. **Downside:** hides diagnostic info when the combination IS pathological.

### Layer 3 — Test gap (LOW effort, MEDIUM value)

Add three regression tests:

1. `test_compute_success_completed_returncode_minus_9_is_success` in `tests/execution/test_session_adjudication.py` — unit-level guard for the `-9` path.
2. `test_build_skill_result_channel_a_win_preserves_success_with_minus_9` in `tests/execution/test_headless.py` — end-to-end assertion.
3. `test_run_managed_async_channel_won_grace_window_permits_natural_exit` in `tests/execution/test_process_run.py` — if Layer 1 adopted, assert grace-window behavior.

## What NOT to do

- **Do NOT remove the kill entirely.** 120 s `CLAUDE_CODE_EXIT_AFTER_STOP_DELAY` default means sessions would hang post-completion.
- **Do NOT change `_compute_success`** to unconditionally accept COMPLETED+success-subtype regardless of content. Existing content check guards against false positives (`session.py:650-663`); PR #568 history shows the cost of loosening this.
- **Do NOT transform `exit_code` in `SkillResult.to_json()`.** Wide blast radius for a formatter-layer cosmetic issue.
- **Do NOT conflate with the tracer-wrong-PID issue** (separate defect below).

## Separate defect flagged (should be its own issue)

The Linux process monitor tracks the **wrong PID**. Session 46cbdc5b `proc_trace` shows `vm_rss_kb=2176`, `threads=1`, `state=sleeping`, `wchan=do_sys_poll`, `cpu_percent=0.0` across all 144 snapshots — consistent with a Python/shell wrapper waiting on a PTY, NOT the Node.js Claude CLI (which would be hundreds of MB RSS, multi-threaded, active CPU). All 5 sessions with the `(exit_code=-9, success=true)` pattern share identical `peak_rss_kb=2176` and `peak_oom_score=666`, confirming systematic tracer attachment to the wrong PID. Consequences:

- `anomaly_count` is effectively meaningless — the detector has zero visibility into actual workload.
- `peak_oom_score=666` alongside `peak_rss_kb=2176` is physically implausible (OOM killer would not target a 2 MB process).
- Token/resource telemetry for the Claude CLI process is unavailable for diagnostics.

This is orthogonal to the `exit_code=-9` issue and should be filed as a separate issue. Suggested title: *"Linux process monitor tracks launcher wrapper PID instead of Claude CLI child PID"*. (Related but distinct from #771 which covers test-code `/dev/shm` pollution.)

## Recommended minimal action (priority order)

1. **Layer 3 regression tests first** (catches regressions when anyone touches this code): ~30 min
2. **Layer 2 formatter annotation** (immediate orchestrator-side win): ~15 min
3. **Layer 1 architectural fix:** plan via `/autoskillit:make-plan`. Combined change to `defaults.yaml` + `_process_race.py` + `_process_monitor.py` + `process.py` + validation against recording/replay fixtures.

Trecek:
## Summary

The immediate defect is a single branch in `run_managed_async`: when either race channel confirms completion, `process.py:289-297` unconditionally calls `async_kill_process_tree(proc.pid)` with no grace window, producing `exit_code=-9` with `success=true` for every monitored successful session. Rather than a minimal patch, this PR replaces the architectural pattern that made the bug writable in the first place.

The fix introduces a pure `decide_termination_action` decision table, a single `execute_termination_action` executor as the sole authorized kill caller (enforced by AST guard), a `DRAIN_THEN_KILL_IF_ALIVE` drain window for the `channel_won` path, first-class `kill_reason: KillReason` metadata surfaced to `SubprocessResult`/`SkillResult`, and a settings coherence guard ensuring the grace window is never smaller than the CLI self-exit delay. Together these changes make the original bug class — "a termination branch forgets to drain" — syntactically unreachable.

### Concurrency Diagram

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 42, 'rankSpacing': 52, 'curve': 'basis'}}}%%
flowchart TB
    %% CLASS DEFINITIONS %%

    START([START: run_headless_core])
    END([END: SkillResult returned])

    %% ── CALLER ──────────────────────────────────────────────────────── %%
    subgraph Caller ["● CALLER — headless.py / process.py"]
        direction TB
        HC["● run_headless_core<br/>━━━━━━━━━━<br/>Sequential await chain<br/>await runner → blocks until subprocess done"]
        RUNNER["● run_managed_async<br/>━━━━━━━━━━<br/>DefaultSubprocessRunner<br/>Spawns task group + all watch coroutines"]
    end

    %% ── TASK GROUP ───────────────────────────────────────────────────── %%
    subgraph TaskGroup ["● ANYIO TASK GROUP — _process_race.py (5 parallel watchers)"]
        direction LR
        WP["_watch_process<br/>━━━━━━━━━━<br/>await proc.wait()<br/>Writes: process_exited<br/>Sets: process_exited_event THEN trigger"]
        WH["_watch_heartbeat<br/>━━━━━━━━━━<br/>Channel A: poll stdout NDJSON<br/>Writes: channel_a_confirmed<br/>Sets: trigger"]
        WI["_watch_stdout_idle<br/>━━━━━━━━━━<br/>Poll stdout file size<br/>Writes: idle_stall<br/>Sets: trigger"]
        WID["_extract_stdout_session_id<br/>━━━━━━━━━━<br/>Scan stdout JSON records<br/>Writes: stdout_session_id<br/>Sets: stdout_session_id_ready"]
        WS["_watch_session_log<br/>━━━━━━━━━━<br/>Channel B: JSONL monitor<br/>Internal drain: await trigger<br/>inside move_on_after window<br/>Writes: channel_b_status<br/>Sets: channel_b_ready THEN trigger"]
    end

    %% ── SHARED STATE ─────────────────────────────────────────────────── %%
    subgraph SharedState ["● SHARED STATE — RaceAccumulator (cooperative, no locks)"]
        direction LR
        ACC["● RaceAccumulator<br/>━━━━━━━━━━<br/>process_exited / process_returncode<br/>channel_a_confirmed<br/>channel_b_status / channel_b_session_id<br/>stdout_session_id / idle_stall<br/>process_exited_event: anyio.Event"]
        EVENTS["anyio.Event primitives<br/>━━━━━━━━━━<br/>trigger — first-wins race signal<br/>process_exited_event — cross-boundary<br/>stdout_session_id_ready — gates Ch-B Phase 1<br/>channel_b_ready — symmetric drain barrier"]
    end

    %% ── BARRIER + DECISION ───────────────────────────────────────────── %%
    subgraph Join ["● BARRIER + RACE RESOLUTION — process.py"]
        direction TB
        BARRIER["await trigger.wait<br/>━━━━━━━━━━<br/>move_on_after timeout<br/>First-wins barrier — unblocks main coroutine"]
        SYMDRAIN["Symmetric drain window<br/>━━━━━━━━━━<br/>If process_exited but channel_b_status is None:<br/>await channel_b_ready inside move_on_after<br/>Prevents Channel B signal loss"]
        CANCEL["tg.cancel_scope.cancel<br/>━━━━━━━━━━<br/>Structured teardown of all remaining tasks<br/>Single explicit cancellation point"]
        RESOLVE["● resolve_termination<br/>━━━━━━━━━━<br/>Pure fn: RaceSignals snapshot<br/>Priority: process_exit > idle_stall > ch_b_stale > completed<br/>Returns: TerminationReason + ChannelConfirmation"]
    end

    %% ── TERMINATION DECISION ─────────────────────────────────────────── %%
    subgraph TermDec ["● TERMINATION DECISION — process.py (channel_won SIGKILL fix)"]
        direction TB
        DECIDE["● decide_termination_action<br/>━━━━━━━━━━<br/>Pure decision table — testable, no I/O<br/>Priority: timeout_fired > process_exited > reason<br/>Returns: TerminationAction enum"]
        EXECUTE["● execute_termination_action<br/>━━━━━━━━━━<br/>SOLE authorized kill caller<br/>Awaits process_exited_event for drain path<br/>Returns: KillReason"]
    end

    %% ── DRAIN WINDOW ─────────────────────────────────────────────────── %%
    subgraph DrainPaths ["● DRAIN WINDOW — 3 TerminationAction paths"]
        direction LR
        NOKILL["NO_KILL<br/>━━━━━━━━━━<br/>process_exited=True, no timeout<br/>Return NATURAL_EXIT — no kill sent"]
        DRAIN["● DRAIN_THEN_KILL_IF_ALIVE<br/>━━━━━━━━━━<br/>channel_won + process still alive<br/>await process_exited_event<br/>inside move_on_after grace_seconds<br/>Kill only if process survives drain"]
        IMMKILL["IMMEDIATE_KILL<br/>━━━━━━━━━━<br/>timeout / stall / stale<br/>async_kill_process_tree — no drain"]
    end

    %% ── RESULT TYPES ─────────────────────────────────────────────────── %%
    subgraph ResultTypes ["● RESULT TYPES — _type_subprocess.py / _type_results.py"]
        direction LR
        SUBRES["● SubprocessResult<br/>━━━━━━━━━━<br/>termination: TerminationReason<br/>channel_confirmation: ChannelConfirmation<br/>+kill_reason: KillReason (NEW)<br/>exit_code may be -9 on completed+killed"]
        SKILL["● SkillResult<br/>━━━━━━━━━━<br/>+kill_reason: KillReason (NEW)<br/>Resolves success=True + exit_code=-9<br/>contradiction for formatter/CI"]
    end

    %% ── SESSION LOG ──────────────────────────────────────────────────── %%
    subgraph SessionLog ["● SESSION LOG — session_log.py (synchronous, post-execution)"]
        direction LR
        FLUSH["● flush_session_log<br/>━━━━━━━━━━<br/>Sync I/O after runner completes<br/>Writes proc_trace, summary, sessions.jsonl<br/>Accepted race on sessions.jsonl append vs atomic_write"]
        RECOVER["● recover_crashed_sessions<br/>━━━━━━━━━━<br/>Startup-only SIGKILL orphan recovery<br/>Scans /dev/shm for orphaned trace files<br/>PID liveness + boot-ID gates"]
    end

    %% ── NEW TEST COVERAGE ────────────────────────────────────────────── %%
    subgraph Tests ["★ NEW TESTS — tests/execution/"]
        direction LR
        TACT["★ test_termination_action.py<br/>━━━━━━━━━━<br/>11-case parametrized matrix for<br/>decide_termination_action<br/>Invariant guards: timeout always IMMEDIATE"]
        TEXEC["★ test_termination_executor.py<br/>━━━━━━━━━━<br/>Real subprocess integration tests<br/>Drain window: exits inside grace → NO_KILL<br/>Drain window: survives grace → KILL_AFTER_COMPLETION"]
    end

    %% ── FLOWS ────────────────────────────────────────────────────────── %%
    START --> HC
    HC --> RUNNER

    RUNNER -->|"tg.start_soon"| WP
    RUNNER -->|"tg.start_soon"| WH
    RUNNER -->|"tg.start_soon"| WI
    RUNNER -->|"tg.start_soon"| WID
    RUNNER -->|"tg.start_soon"| WS

    WP --> ACC
    WH --> ACC
    WI --> ACC
    WID --> ACC
    WS --> ACC

    WP -->|"set before trigger"| EVENTS
    WH -->|"set trigger"| EVENTS
    WI -->|"set trigger"| EVENTS
    WID -->|"set stdout_session_id_ready"| EVENTS
    WS -->|"set channel_b_ready + trigger"| EVENTS

    EVENTS -.->|"stdout_session_id_ready gates Phase 1"| WS

    ACC -->|"to_race_signals snapshot"| BARRIER
    EVENTS -->|"trigger.wait"| BARRIER

    BARRIER --> SYMDRAIN
    SYMDRAIN -.->|"await channel_b_ready"| EVENTS
    SYMDRAIN --> CANCEL
    CANCEL --> RESOLVE
    ACC -->|"snapshot"| RESOLVE

    RESOLVE -->|"TerminationReason"| DECIDE
    DECIDE -->|"TerminationAction"| EXECUTE

    EXECUTE --> NOKILL
    EXECUTE --> DRAIN
    EXECUTE --> IMMKILL

    DRAIN -.->|"await process_exited_event"| EVENTS

    NOKILL -->|"KillReason.NATURAL_EXIT"| SUBRES
    DRAIN -->|"KillReason.KILL_AFTER_COMPLETION"| SUBRES
    IMMKILL -->|"KillReason.INFRA_KILL"| SUBRES

    RESOLVE -->|"TerminationReason + ChannelConfirmation"| SUBRES
    SUBRES --> SKILL
    SKILL --> FLUSH
    FLUSH --> END

    RECOVER -.->|"startup: re-uses flush_session_log for crashed sessions"| FLUSH

    DECIDE -.->|"tested by"| TACT
    EXECUTE -.->|"tested by"| TEXEC

    %% ── CLASS ASSIGNMENTS ────────────────────────────────────────────── %%
    class START,END terminal;
    class HC,RUNNER phase;
    class WP,WH,WI,WID,WS handler;
    class ACC,EVENTS stateNode;
    class BARRIER,SYMDRAIN,CANCEL,RESOLVE phase;
    class DECIDE,EXECUTE newComponent;
    class NOKILL,IMMKILL output;
    class DRAIN newComponent;
    class SUBRES,SKILL output;
    class FLUSH,RECOVER phase;
    class TACT,TEXEC newComponent;
```

### Error/Resilience Diagram

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 40, 'rankSpacing': 52, 'curve': 'basis'}}}%%
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINALS %%
    T_START([START: run_headless_core])
    T_COMPLETE([SUCCESS])
    T_RETRY([RETRIABLE])
    T_FAILED([FAILED])
    T_CIRCUIT([BUDGET_EXHAUSTED])
    T_GATE_ERR([GATE_ERROR])

    %% ── STARTUP VALIDATION GATES ── %%
    subgraph ConfigGate ["STARTUP VALIDATION GATES"]
        CFG_GUARD["● RunSkillConfig.__post_init__<br/>━━━━━━━━━━<br/>grace_secs * 1000 ≥ exit_delay_ms + 500<br/>ValueError → aborts startup<br/>src: config/settings.py"]
        GATE_CHECK["● DefaultGateState<br/>━━━━━━━━━━<br/>Kitchen open/closed check<br/>gate_error_result or headless_error_result<br/>src: pipeline/gate.py"]
    end

    %% ── RACE MONITORING ── %%
    subgraph RaceMonitor ["RACE MONITORING (Concurrent Watchers)"]
        CH_A["Channel A: Heartbeat<br/>━━━━━━━━━━<br/>NDJSON type=result in stdout<br/>channel_a_confirmed = True<br/>src: execution/_process_race.py"]
        CH_B["● Channel B: JSONL Marker<br/>━━━━━━━━━━<br/>Watches session log dir<br/>completion_drain_timeout: 5.0s window<br/>lets Channel A preempt<br/>src: ● execution/_process_race.py"]
        IDLE_W["● Idle Stall Watcher<br/>━━━━━━━━━━<br/>Polls stdout byte growth<br/>idle_output_timeout: 600s<br/>OSError on stat → silent continue<br/>src: ● execution/_process_race.py"]
        TIMEOUT_W["Outer Timeout Fence<br/>━━━━━━━━━━<br/>anyio.move_on_after<br/>run_skill.timeout: 7200s<br/>overrides all race outcomes<br/>src: execution/process.py"]
    end

    %% ── TERMINATION DECISION + ACTION ── %%
    subgraph TermDecision ["TERMINATION DECISION + ACTION"]
        RESOLVE["● resolve_termination<br/>━━━━━━━━━━<br/>Priority: NATURAL_EXIT → IDLE_STALL<br/>→ STALE → COMPLETED<br/>assert_never enforces exhaustive match<br/>src: ● execution/_process_race.py"]
        DECIDE["★ decide_termination_action<br/>━━━━━━━━━━<br/>timeout_fired overrides everything<br/>proc_exited → NO_KILL<br/>COMPLETED+alive → DRAIN_THEN_KILL<br/>src: ● execution/process.py"]
        NOKILL["NO_KILL<br/>━━━━━━━━━━<br/>Process already exited naturally<br/>KillReason: NATURAL_EXIT"]
        DRAIN_WIN["★ DRAIN_THEN_KILL_IF_ALIVE<br/>━━━━━━━━━━<br/>Await natural_exit_grace_seconds: 3.0s<br/>process_exited_event secondary race<br/>KillReason: KILL_AFTER_COMPLETION if killed<br/>KillReason: NATURAL_EXIT if self-exits<br/>src: ● execution/process.py"]
        IMM_KILL["IMMEDIATE_KILL<br/>━━━━━━━━━━<br/>Trigger: IDLE_STALL / STALE<br/>/ TIMED_OUT / timeout_fired<br/>KillReason: INFRA_KILL<br/>src: execution/process.py"]
        EXEC_TERM["★ execute_termination_action<br/>━━━━━━━━━━<br/>Sole authorized caller of<br/>async_kill_process_tree<br/>Sets KillReason on SubprocessResult<br/>src: ● execution/process.py"]
    end

    %% ── RESULT ROUTING ── %%
    subgraph ResultRoute ["RESULT ROUTING (_build_skill_result)"]
        STALE_REC["● Stale Recovery<br/>━━━━━━━━━━<br/>Parse stdout for complete record<br/>Win → subtype=recovered_from_stale<br/>Lose → needs_retry=True STALE<br/>src: ● execution/headless.py"]
        NORMAL_REC["● Normal Recovery Sequence<br/>━━━━━━━━━━<br/>Ch-B drain-race recovery<br/>→ separate-marker recovery<br/>→ block pattern recovery<br/>→ write artifact synthesis<br/>src: ● execution/headless.py"]
        PATH_GATE["● PATH_CONTAMINATION Gate<br/>━━━━━━━━━━<br/>Output paths escape cwd?<br/>→ retry_reason=PATH_CONTAMINATION<br/>src: ● execution/headless.py"]
        ZERO_GATE["● ZERO_WRITES Gate<br/>━━━━━━━━━━<br/>success=True + write_call_count=0<br/>when writes expected<br/>→ retry_reason=ZERO_WRITES<br/>src: ● execution/headless.py"]
        CONTRACT_GATE["● CONTRACT_RECOVERY Gate<br/>━━━━━━━━━━<br/>adjudicated_failure +<br/>write_call_count ≥ 1<br/>→ retry_reason=CONTRACT_RECOVERY<br/>src: ● execution/headless.py"]
        BUDGET["● _apply_budget_guard<br/>━━━━━━━━━━<br/>consecutive_failures > max_retries<br/>→ needs_retry=False BUDGET_EXHAUSTED<br/>Applied twice on CONTRACT path<br/>src: ● execution/headless.py"]
    end

    %% ── SESSION DIAGNOSTICS ── %%
    subgraph DiagLog ["SESSION DIAGNOSTICS (Write-Only Path)"]
        FLUSH["● flush_session_log<br/>━━━━━━━━━━<br/>Writes: proc_trace.jsonl, summary.json<br/>sessions.jsonl (index)<br/>Runner crash → subtype=crashed exit=-1<br/>All flush failures: silent debug log<br/>src: ● execution/session_log.py"]
        CRASH_REC["● recover_crashed_sessions<br/>━━━━━━━━━━<br/>Gate 1: enrollment sidecar exists<br/>Gate 2: boot_id matches current boot<br/>Gate 3: PID liveness + starttime_ticks<br/>PID recycled → recovery proceeds<br/>src: ● execution/session_log.py"]
    end

    %% ── CONNECTIONS ── %%
    T_START --> CFG_GUARD
    CFG_GUARD -->|"ValueError: aborts startup"| T_FAILED
    CFG_GUARD -->|"valid config"| GATE_CHECK
    GATE_CHECK -->|"gate closed"| T_GATE_ERR
    GATE_CHECK -->|"gate open"| CH_A
    GATE_CHECK -->|"gate open"| CH_B
    GATE_CHECK -->|"gate open"| IDLE_W
    GATE_CHECK -->|"gate open"| TIMEOUT_W

    CH_A --> RESOLVE
    CH_B --> RESOLVE
    IDLE_W --> RESOLVE
    TIMEOUT_W --> RESOLVE

    RESOLVE --> DECIDE
    DECIDE -->|"COMPLETED + alive"| DRAIN_WIN
    DECIDE -->|"NATURAL_EXIT"| NOKILL
    DECIDE -->|"STALE / IDLE_STALL / timeout"| IMM_KILL

    NOKILL --> EXEC_TERM
    DRAIN_WIN --> EXEC_TERM
    IMM_KILL --> EXEC_TERM

    EXEC_TERM -->|"TerminationReason=STALE"| STALE_REC
    EXEC_TERM -->|"TerminationReason=TIMED_OUT"| T_FAILED
    EXEC_TERM -->|"COMPLETED"| NORMAL_REC
    EXEC_TERM -.->|"always"| FLUSH

    STALE_REC -->|"stdout recovered"| T_COMPLETE
    STALE_REC -->|"no recovery"| BUDGET

    NORMAL_REC --> PATH_GATE
    PATH_GATE -->|"contaminated"| BUDGET
    PATH_GATE -->|"clean"| ZERO_GATE
    ZERO_GATE -->|"zero writes"| BUDGET
    ZERO_GATE -->|"ok"| CONTRACT_GATE
    CONTRACT_GATE -->|"adjudicated failure"| BUDGET
    CONTRACT_GATE -->|"ok"| T_COMPLETE

    BUDGET -->|"under limit"| T_RETRY
    BUDGET -->|"limit exceeded"| T_CIRCUIT

    FLUSH -->|"crashed session"| CRASH_REC

    %% ── CLASS ASSIGNMENTS ── %%
    class CFG_GUARD,GATE_CHECK,PATH_GATE,ZERO_GATE,CONTRACT_GATE detector;
    class CH_A,CH_B stateNode;
    class IDLE_W,TIMEOUT_W,NOKILL,IMM_KILL,STALE_REC,NORMAL_REC handler;
    class RESOLVE,DECIDE phase;
    class DRAIN_WIN,EXEC_TERM newComponent;
    class BUDGET gap;
    class FLUSH,CRASH_REC output;
    class T_START,T_COMPLETE,T_RETRY,T_FAILED,T_CIRCUIT,T_GATE_ERR terminal;
```

### State Lifecycle Diagram

```mermaid
flowchart TB

    START([Process Launch])

    subgraph Accum ["● RACE ACCUMULATOR — Write-Once Fields (_process_race.py)"]
        direction LR
        AccProc["process_exited<br/>━━━━━━━━━━<br/>_watch_process<br/>INIT_ONLY"]
        AccChanA["channel_a_confirmed<br/>━━━━━━━━━━<br/>_watch_heartbeat<br/>INIT_ONLY"]
        AccChanB["channel_b_status<br/>━━━━━━━━━━<br/>_watch_session_log<br/>INIT_ONLY"]
        AccIdle["idle_stall<br/>━━━━━━━━━━<br/>_watch_stdout_idle<br/>INIT_ONLY"]
        AccSID["stdout_session_id<br/>━━━━━━━━━━<br/>_extract_stdout_session_id<br/>INIT_ONLY"]
    end

    AccEvent["● process_exited_event<br/>━━━━━━━━━━<br/>anyio.Event; shared ref<br/>Set BEFORE trigger · INIT_PRESERVE"]

    subgraph DrainWindows ["DRAIN WINDOWS — Validation Gates (● _process_race.py, ● settings.py)"]
        direction TB
        DrainB["Channel B Completion Drain<br/>━━━━━━━━━━<br/>After COMPLETION: await trigger<br/>completion_drain_timeout (5s)"]
        DrainSym["Symmetric Drain<br/>━━━━━━━━━━<br/>After process_exited: await chan_b_ready<br/>completion_drain_timeout (5s)"]
        ConfigGate["● RunSkillConfig.__post_init__<br/>━━━━━━━━━━<br/>FAIL-FAST: grace_secs ≥ exit_after_ms/1000 + 0.5<br/>Config validation gate"]
    end

    subgraph Resolve ["SIGNAL RESOLUTION (● _process_race.py, ● process.py)"]
        direction TB
        FrozenSnap["RaceSignals<br/>━━━━━━━━━━<br/>frozen=True snapshot<br/>INIT_ONLY post-construction"]
        ResolveFn["resolve_termination()<br/>━━━━━━━━━━<br/>Pure fn; priority: exit > idle > stale > chan_won<br/>→ TerminationReason + ChannelConfirmation"]
        DecideFn["● decide_termination_action()<br/>━━━━━━━━━━<br/>Priority table; timeout overrides all<br/>COMPLETED → DRAIN_THEN_KILL_IF_ALIVE"]
    end

    subgraph TermActions ["● TERMINATION ACTION ENUM (_type_enums.py)"]
        direction LR
        NoKill["● NO_KILL<br/>━━━━━━━━━━<br/>Process already exited naturally<br/>No syscall needed"]
        DrainKill["● DRAIN_THEN_KILL_IF_ALIVE<br/>━━━━━━━━━━<br/>channel_won path<br/>Grace drain → conditional SIGKILL"]
        ImmKill["● IMMEDIATE_KILL<br/>━━━━━━━━━━<br/>Timeout / stall / stale<br/>Immediate SIGTERM→SIGKILL"]
    end

    subgraph GraceDrain ["GRACE DRAIN (● process.py, ● settings.py)"]
        direction TB
        GraceWindow["execute_termination_action()<br/>━━━━━━━━━━<br/>Await process_exited_event<br/>up to natural_exit_grace_secs (3s)"]
        KillDecision{"proc.returncode<br/>not None?"}
    end

    subgraph KillResult ["● KILL REASON ENUM (_type_enums.py)"]
        direction LR
        KRNat["● NATURAL_EXIT<br/>━━━━━━━━━━<br/>No kill performed<br/>Hard-coded in gate error envelopes"]
        KRComp["● KILL_AFTER_COMPLETION<br/>━━━━━━━━━━<br/>channel_won + grace expired<br/>SIGTERM→SIGKILL performed"]
        KRInfra["● INFRA_KILL<br/>━━━━━━━━━━<br/>Timeout / stall forced kill<br/>SIGTERM→SIGKILL performed"]
    end

    subgraph ResultProp ["● KILL REASON PROPAGATION (_type_subprocess.py → _type_results.py)"]
        direction LR
        SubRes["● SubprocessResult.kill_reason<br/>━━━━━━━━━━<br/>Set by execute_termination_action<br/>Immutable post-construction"]
        SkillRes["● SkillResult.kill_reason<br/>━━━━━━━━━━<br/>Read-through from SubprocessResult<br/>dataclasses.replace in gate chain"]
    end

    subgraph SessionGates ["SESSION OUTCOME GATES (● headless.py, ● gate.py)"]
        direction TB
        KitGate["● DefaultGateState.enabled<br/>━━━━━━━━━━<br/>MUTABLE; kitchen open/close<br/>Error path: kill_reason=natural_exit hardcoded"]
        OutGate["_build_skill_result() gate chain<br/>━━━━━━━━━━<br/>10 sequential gates; success adjudication<br/>CONTRACT_RECOVERY · zero-write · budget guard"]
    end

    subgraph ResumeDetect ["RESUME DETECTION — Session ID Precedence (● headless.py)"]
        direction LR
        Tier1["Tier 1: stdout_session_id<br/>━━━━━━━━━━<br/>type=system record · highest priority<br/>INIT_PRESERVE"]
        Tier2["Tier 2: channel_b_session_id<br/>━━━━━━━━━━<br/>JSONL filename stem<br/>INIT_PRESERVE"]
        Tier3["Tier 3: transport session_id<br/>━━━━━━━━━━<br/>Fallback identity<br/>INIT_PRESERVE"]
    end

    subgraph Persist ["AUDIT LOG — Append + Checkpoint (● session_log.py)"]
        direction LR
        SessLog["sessions.jsonl<br/>━━━━━━━━━━<br/>APPEND_ONLY · never truncated<br/>kill_reason flushed per session"]
        SummJson["● summary.json<br/>━━━━━━━━━━<br/>atomic_write · kill_reason stamped<br/>Diagnostic checkpoint"]
        TokJson["token_usage.json<br/>━━━━━━━━━━<br/>atomic_write · INIT_PRESERVE<br/>Read at server startup"]
    end

    END([SkillResult → caller])

    %% CONNECTIONS %%
    START --> AccProc
    START --> AccChanA
    START --> AccChanB
    START --> AccIdle
    START --> AccSID
    START --> AccEvent

    AccChanB -->|COMPLETION| DrainB
    AccProc -->|process_exited=True| DrainSym
    DrainB --> FrozenSnap
    DrainSym --> FrozenSnap
    AccProc --> FrozenSnap
    AccChanA --> FrozenSnap
    AccIdle --> FrozenSnap
    AccSID --> FrozenSnap
    AccEvent --> FrozenSnap

    ConfigGate -->|init guard| DecideFn
    FrozenSnap --> ResolveFn
    ResolveFn --> DecideFn
    DecideFn --> NoKill
    DecideFn --> DrainKill
    DecideFn --> ImmKill

    NoKill --> KRNat
    DrainKill --> GraceWindow
    AccEvent -.->|shared ref; awaited in drain| GraceWindow
    GraceWindow --> KillDecision
    KillDecision -->|Yes: natural exit| KRNat
    KillDecision -->|No: grace expired| KRComp
    ImmKill --> KRInfra

    KRNat --> SubRes
    KRComp --> SubRes
    KRInfra --> SubRes
    SubRes --> SkillRes

    SkillRes --> KitGate
    KitGate --> OutGate

    AccSID --> Tier1
    AccChanB -->|channel_b_session_id| Tier2
    Tier1 --> SessLog
    Tier2 --> SessLog
    Tier3 --> SessLog

    OutGate --> SummJson
    OutGate --> SessLog
    OutGate --> TokJson
    SummJson --> END
    SessLog --> END
    TokJson --> END

    %% CLASS ASSIGNMENTS %%
    class AccProc,AccChanA,AccChanB,AccIdle,AccSID detector;
    class AccEvent gap;
    class FrozenSnap detector;
    class DrainB,DrainSym,ConfigGate stateNode;
    class ResolveFn,DecideFn phase;
    class NoKill,DrainKill,ImmKill handler;
    class GraceWindow stateNode;
    class KillDecision phase;
    class KRNat,KRComp,KRInfra output;
    class SubRes,SkillRes handler;
    class KitGate,OutGate stateNode;
    class Tier1,Tier2,Tier3 cli;
    class SessLog,SummJson,TokJson output;
    class START,END terminal;
```

Closes #804

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-804-20260413-075452-722374/.autoskillit/temp/rectify/rectify_channel_won_sigkill_2026-04-13_110657.md`

## Token Usage Summary

| investigate | 145 | 16.7k | 632.6k | 89.4k | 1 | 7m 46s |
| rectify | 10.2k | 47.9k | 1.3M | 222.9k | 2 | 19m 47s |
| review | 10.4k | 8.2k | 163.8k | 53.3k | 1 | 5m 31s |
| dry_walkthrough | 5.0k | 22.1k | 664.3k | 65.6k | 1 | 8m 0s |
| implement | 646 | 51.9k | 9.6M | 164.0k | 1 | 16m 55s |
| retry_worktree | 966 | 48.4k | 13.6M | 147.9k | 1 | 24m 11s |
| prepare_pr | 70 | 9.1k | 220.6k | 30.1k | 1 | 2m 32s |
| run_arch_lenses | 5.8k | 42.8k | 582.2k | 132.0k | 3 | 15m 17s |
| compose_pr | 59 | 12.5k | 244.3k | 40.5k | 1 | 2m 55s |
| **Total** | 33.3k | 259.7k | 27.0M | 945.7k | | 1h 42m |


Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (13 actionable warnings found)

src/autoskillit/core/_type_enums.py
```diff
@@ -183,6 +185,36 @@ class TerminationReason(StrEnum):
     TIMED_OUT = "timed_out"
 
 
+class TerminationAction(StrEnum):
+    """What execute_termination_action should do with a subprocess after the race loop.
+
+    Produced by decide_termination_action (pure function) and consumed by
+    execute_termination_action (the single authorized kill caller in process.py).
```
Trecek, line 192: [info] arch: TerminationAction docstring references L1 implementation details (execute_termination_action, decide_termination_action in process.py). L0 types should be layer-agnostic; remove the function-name references and describe the protocol in terms of producer/consumer roles without naming specific L1 functions.

src/autoskillit/pipeline/gate.py
```diff
@@ -56,6 +56,7 @@ def gate_error_result(message: str | None = None) -> str:
             "cli_subtype": "",
             "is_error": True,
             "exit_code": -1,
+            "kill_reason": "natural_exit",
```
Trecek: [warning] arch: gate_error_result hardcodes the string "natural_exit" instead of KillReason.NATURAL_EXIT.value. pipeline/gate.py already imports from autoskillit.core (L0), so KillReason is reachable. The hardcoded string will silently diverge if the enum value ever changes.

src/autoskillit/pipeline/gate.py
```diff
@@ -89,6 +90,7 @@ def headless_error_result(message: str | None = None) -> str:
...
```
Trecek: [warning] arch: headless_error_result hardcodes the string "natural_exit" instead of KillReason.NATURAL_EXIT.value. Same issue as gate_error_result — gate.py already imports from autoskillit.core, so the enum is accessible without any new dependency.

tests/execution/test_termination_executor.py
```diff
@@ -0,0 +1,197 @@
...
+            proc_exited_event.set()
+
+        kill_calls: list[int] = []
+
+        async def _mock_kill(pid: int, timeout: float = 2.0) -> None:
+            kill_calls.append(pid)
+
+        with patch(
+            "autoskillit.execution.process.async_kill_process_tree",
+            new_callable=lambda: lambda: _mock_kill,
```
Trecek: [warning] tests: Inconsistent patch style: uses `new_callable=lambda: lambda: _mock_kill` (double-lambda factory) instead of `side_effect=_mock_kill` as used in all other tests in the same file. This double-lambda form is fragile and non-idiomatic; it passes a factory that returns a factory, which may silently fail to route calls to the mock.

tests/execution/test_termination_executor.py
```diff
@@ -0,0 +1,197 @@
...
+                kill_reason = await execute_termination_action(
+                    TerminationAction.DRAIN_THEN_KILL_IF_ALIVE,
+                    proc=proc,
+                    process_exited_event=proc_exited_event,
+                    grace_seconds=3.0,
+                    proc_log=proc_log,
+                )
+                tg.cancel_scope.cancel()
+
+        assert kill_reason == KillReason.NATURAL_EXIT
```
Trecek, line 83: [warning] tests: `kill_calls == []` assertion is vacuously true because the mock is wired via the broken double-lambda form at line 66. If the patch does not actually intercept `async_kill_process_tree`, a spurious kill could occur and kill_calls would still be empty since the mock was never installed. Use `side_effect=_mock_kill` to make this assertion meaningful.

tests/execution/test_termination_executor.py
```diff
@@ -0,0 +1,197 @@
...
+
+        with patch(
+            "autoskillit.execution.process.async_kill_process_tree",
+            new_callable=lambda: lambda: _mock_kill,
+        ):
+            import structlog
+
+            proc_log = structlog.get_logger().bind(pid=proc.pid)
+
+            async with anyio.create_task_group() as tg:
```
Trecek, line 72: [warning] tests: `execute_termination_action` is run inside `anyio.create_task_group` that is immediately cancelled after it returns. If the task group is cancelled before `_watch` completes, `proc.wait()` is interrupted and the process may be left as a zombie. The test should `await proc.wait()` or `await proc.aclose()` after the task group exits.

tests/execution/test_termination_action.py
```diff
@@ -0,0 +1,91 @@
...
+        # (TIMED_OUT, timeout=True, exited=False) → IMMEDIATE_KILL
+        (TerminationReason.TIMED_OUT, True, False, TerminationAction.IMMEDIATE_KILL),
+        # (IDLE_STALL, timeout=False, exited=False) → IMMEDIATE_KILL
+        (TerminationReason.IDLE_STALL, False, False, TerminationAction.IMMEDIATE_KILL),
+        # (STALE, timeout=False, exited=False) → IMMEDIATE_KILL
+        (TerminationReason.STALE, False, False, TerminationAction.IMMEDIATE_KILL),
+        # timeout_fired beats process_exited
+        (TerminationReason.COMPLETED, True, True, TerminationAction.IMMEDIATE_KILL),
+        # IDLE_STALL with process_exited=False, no timeout → IMMEDIATE_KILL
+        (TerminationReason.IDLE_STALL, False, True, TerminationAction.NO_KILL),
```
Trecek, line 38: [warning] tests: Comment says 'IDLE_STALL with process_exited=False, no timeout → IMMEDIATE_KILL' but the parametrize row is `(TerminationReason.IDLE_STALL, False, True, TerminationAction.NO_KILL)` — i.e., `process_exited=True`, expecting NO_KILL. The comment describes the wrong case and could mask a future copy-paste error.

tests/execution/test_session_adjudication.py
```diff
@@ -1648,3 +1648,81 @@ def test_parse_session_result_non_write_tools_no_file_path(make_ndjson):
...
+        """_compute_retry must not raise for any (TerminationReason, returncode) pair."""
+        session = ClaudeSessionResult(
+            subtype="success",
+            result="done. %%ORDER_UP%%",
+            is_error=False,
+            session_id="rc9",
+            errors=[],
+        )
+        result = _compute_retry(session, returncode=returncode, termination=termination)
+        assert isinstance(result, tuple) and len(result) == 2
```
Trecek, line 1705: [info] tests: `test_compute_retry_handles_minus_9_for_all_termination_reasons` only asserts `isinstance(result, tuple) and len(result) == 2` — structural shape without checking the actual retry decision values. A stronger assertion would verify the specific (needs_retry, RetryReason) values for the COMPLETED case.

tests/execution/test_session_adjudication.py
```diff
@@ -1648,3 +1648,81 @@ def test_parse_session_result_non_write_tools_no_file_path(make_ndjson):
...
+
+    @pytest.mark.parametrize("returncode", [0, -15, -9])
+    @pytest.mark.parametrize("termination", list(TerminationReason))
+    def test_compute_retry_handles_minus_9_for_all_termination_reasons(
+        self,
+        termination: TerminationReason,
+        returncode: int,
+    ) -> None:
+        """_compute_retry must not raise for any (TerminationReason, returncode) pair."""
+        session = ClaudeSessionResult(
```
Trecek, line 1697: [info] tests: `test_compute_success_handles_minus_9_for_all_termination_reasons` passes no `channel_confirmation` argument. Depending on the default, this may silently test a non-representative path. Consider explicitly passing `channel_confirmation=ChannelConfirmation.CHANNEL_A`.

tests/execution/test_process_race.py
```diff
@@ -244,7 +244,74 @@ class TestRaceSignalsFieldCount:
...
+        async with anyio.create_task_group() as tg:
+            tg.start_soon(_watch_process, proc, acc, trigger)
+            await trigger.wait()
+            tg.cancel_scope.cancel()
+
+        assert acc.process_exited is True
+        assert acc.process_exited_event.is_set() is True
+
+    @pytest.mark.anyio
+    async def test_process_exited_event_fires_before_trigger(self, tmp_path) -> None:
```
Trecek, line 282: [info] tests: `test_process_exited_event_fires_before_trigger` doesn't assert `acc.process_exited is True`. Adding this assertion makes the post-condition complete and consistent with `test_watch_process_sets_both_event_and_flag`.

tests/config/test_defaults.py
```diff
@@ -0,0 +1,44 @@
...
+import pytest
+
+from autoskillit.config import AutomationConfig
+from autoskillit.config.settings import RunSkillConfig
+
+
+class TestGraceWindowCoherence:
+    """natural_exit_grace_seconds must cover exit_after_stop_delay_ms + margin."""
+
+    def test_natural_exit_grace_covers_exit_after_stop_delay(self) -> None:
```
Trecek, line 18: [info] tests: `test_natural_exit_grace_covers_exit_after_stop_delay` duplicates the constraint already enforced by `RunSkillConfig.__post_init__`. The test is meaningful as a defaults.yaml regression guard, but its assertion message reformats arithmetic already in `__post_init__`, creating two parallel error strings to maintain.

src/autoskillit/execution/process.py
```diff
@@ -94,6 +104,74 @@ def _resolve_session_id(
...
+            assert_never(unreachable)
+
+
+async def execute_termination_action(
+    action: TerminationAction,
+    *,
+    proc: anyio.abc.Process,
+    process_exited_event: anyio.Event,
+    grace_seconds: float,
+    proc_log: Any,
```
Trecek: [warning] defense: `execute_termination_action` accepts `proc_log: Any` — bypasses static checking at this public boundary. Should be typed as `structlog.BoundLogger` or a Protocol with a `debug` method to prevent silent misuse (e.g., passing None or an incompatible logger that fails at runtime with AttributeError).

src/autoskillit/execution/process.py
```diff
@@ -94,6 +104,74 @@ def _resolve_session_id(
...
+
+    Returns the KillReason that surfaces to SubprocessResult.kill_reason.
+    """
+    match action:
+        case TerminationAction.NO_KILL:
+            return KillReason.NATURAL_EXIT
+        case TerminationAction.DRAIN_THEN_KILL_IF_ALIVE:
+            with anyio.move_on_after(grace_seconds):
+                await process_exited_event.wait()
+            if proc.returncode is not None:
```
Trecek, line 164: [warning] defense: In DRAIN_THEN_KILL_IF_ALIVE, `async_kill_process_tree` at line 166 is called without exception guard. If the process already exited between `decide_termination_action` and the kill call, the helper may raise ProcessLookupError/OSError that propagates uncaught, leaving `kill_reason` unset on `SubprocessResult`.

src/autoskillit/execution/process.py
```diff
@@ -94,6 +104,74 @@ def _resolve_session_id(
...
+                proc_log.debug("natural_exit_after_drain", returncode=proc.returncode)
+                return KillReason.NATURAL_EXIT
+            proc_log.debug("grace_expired_killing", grace_seconds=grace_seconds)
+            await async_kill_process_tree(proc.pid)
+            return KillReason.KILL_AFTER_COMPLETION
+        case TerminationAction.IMMEDIATE_KILL:
+            await async_kill_process_tree(proc.pid)
```
Trecek, line 171: [warning] defense: IMMEDIATE_KILL path calls `async_kill_process_tree` without any exception guard. If the process has already exited between `decide_termination_action` (which saw `process_exited=False`) and this call, an OS-level exception propagates uncaught through `execute_termination_action` and into `run_managed_async`.

src/autoskillit/config/settings.py
```diff
@@ -92,10 +92,22 @@ class RunSkillConfig:
...
     completion_marker: str = "%%ORDER_UP%%"
     completion_drain_timeout: float = 5.0
-    exit_after_stop_delay_ms: int = 120000
+    exit_after_stop_delay_ms: int = 2000
+    natural_exit_grace_seconds: float = 3.0
     idle_output_timeout: int = 600
     max_suppression_seconds: int = 1800
 
+    def __post_init__(self) -> None:
+        required_ms = self.exit_after_stop_delay_ms + 500
```
Trecek, line 101: [info] defense: `__post_init__` validates the grace window coherence but does not guard against non-positive or zero `natural_exit_grace_seconds`. A caller can set `exit_after_stop_delay_ms=0` with `natural_exit_grace_seconds=0.5` (500ms >= 0+500=500), which passes — this edge case (zero exit delay, 500ms grace) is valid but may be worth documenting.

src/autoskillit/hooks/_fmt_execution.py
```diff
@@ -13,6 +13,28 @@
...
+    """
+    exit_code = data.get("exit_code", "")
+    kill_reason = data.get("kill_reason")
+    if kill_reason is None:
+        # Legacy payload — no annotation
+        return f"exit_code: {exit_code}"
+    if kill_reason == "kill_after_completion":
+        return f"exit_code: {exit_code} (infra-terminated after completion — grace exceeded)"
+    if kill_reason == "infra_kill":
+        termination_reason = data.get("termination_reason") or data.get("subtype") or ""
```
Trecek, line 30: [warning] bugs: Key name mismatch: `_format_exit_code_line` looks for `data.get("subtype")` to annotate infra_kill exit codes, but `SkillResult.to_json()` serializes the field as `"cli_subtype"` (not `"subtype"`). In production, `data.get("subtype")` returns None and `data.get("termination_reason")` is also absent, so the annotation always degrades to the generic `(infra-killed)` form. Only the test passes because its fixture dict uses `"subtype"` directly.

src/autoskillit/execution/_process_race.py
```diff
@@ -32,6 +36,7 @@ class RaceSignals:
     channel_b_session_id: str = ""  # Claude Code session ID from JSONL filename stem, or ""
     stdout_session_id: str | None = None  # Session ID extracted from stdout type=system record
     idle_stall: bool = False
+    process_exited_event: anyio.Event = field(default_factory=anyio.Event)
```
Trecek, line 39: [info] defense: `RaceSignals` is frozen, but `process_exited_event: anyio.Event` is a mutable object. Code holding a `RaceSignals` reference can call `process_exited_event.set()`, mutating shared state through what is supposed to be a read-only snapshot. The field should be documented as intentionally shared-by-reference, or the frozen constraint should acknowledge this carve-out.

src/autoskillit/execution/process.py
```diff
@@ -263,38 +342,37 @@ async def run_managed_async(
...
-                    channel_b=signals.channel_b_status,
-                )
-                # Channel A or B won; process still alive — kill immediately.
-                await async_kill_process_tree(proc.pid)
+            action = decide_termination_action(
+                termination,
+                timeout_fired=timeout_scope.cancelled_caught,
+                process_exited=signals.process_exited,
+            )
+            _timeout_fired = timeout_scope.cancelled_caught
```
Trecek: [warning] cohesion: `_kill_reason_hint` is a re-derivation of information already determined by `decide_termination_action`. It re-maps `TerminationAction` values back to string labels that must stay in sync with the enum. The debug log should use `str(action)` directly, or defer until `kill_reason` is available after `execute_termination_action` returns.

src/autoskillit/hooks/_fmt_execution.py
```diff
@@ -13,6 +13,28 @@
...
+
+    Legacy payloads without kill_reason render as bare exit_code values so that
+    JSON-line replay of old session logs still parses correctly.
+    """
+    exit_code = data.get("exit_code", "")
+    kill_reason = data.get("kill_reason")
+    if kill_reason is None:
+        # Legacy payload — no annotation
+        return f"exit_code: {exit_code}"
+    if kill_reason == "kill_after_completion":
```
Trecek, line 27: [warning] cohesion: `_format_exit_code_line` compares `kill_reason` against raw string literals (`'kill_after_completion'`, `'infra_kill'`) instead of `KillReason` enum values. A rename of `KillReason` members would not be caught by mypy. Import and use `KillReason.KILL_AFTER_COMPLETION.value` etc.

src/autoskillit/core/_type_enums.py
```diff
@@ -183,6 +185,36 @@ class TerminationReason(StrEnum):
     TIMED_OUT = "timed_out"
 
 
+class TerminationAction(StrEnum):
```
Trecek, line 188: [info] cohesion: `TerminationAction` docstring references `decide_termination_action` and `execute_termination_action` by function name, tying the L0 enum definition to L1 implementation functions. L0 should be independent of L1 callers; describe the enum's contract without naming specific consumers.

src/autoskillit/core/_type_subprocess.py
```diff
@@ -103,6 +103,12 @@ class SubprocessResult:
     Consumers (session_log, tokens) must use this float directly — never re-derive
     duration from start_ts/end_ts ISO strings.
     """
+    kill_reason: KillReason = KillReason.NATURAL_EXIT
```
Trecek, line 106: [info] cohesion: `SubprocessResult.kill_reason` docstring says 'Set by run_managed_async via execute_termination_action', naming a specific L1 function from within a core L0 type. Embeds L1 implementation details in the L0 type definition.

tests/execution/test_termination_executor.py
```diff
@@ -0,0 +1,197 @@
...
+        proc_exited_event = anyio.Event()
+
+        # Start a task that waits for the process and sets the event
+        async def _watch() -> None:
+            await proc.wait()
+            proc_exited_event.set()
+
+        kill_calls: list[int] = []
+
+        async def _mock_kill(pid: int, timeout: float = 2.0) -> None:
```
Trecek, line 61: [warning] cohesion: Asymmetric patching pattern: `TestDrainWindowPermitsNaturalExit` uses `new_callable=lambda: lambda: _mock_kill` while all other test classes in this file use `side_effect=_mock_kill`. Inconsistent mock style across tests that cover the same function.

src/autoskillit/core/_type_enums.py
```diff
@@ -183,6 +185,36 @@ class TerminationReason(StrEnum):
...
+    """What execute_termination_action should do with a subprocess after the race loop.
```
Trecek, line 189: [info] slop: TerminationAction class-level docstring over-explains the mapping to consuming functions. The docstring body repeats what the enum member names already convey. Class docstring should state the invariant, not narrate the call graph.

src/autoskillit/core/_type_enums.py
```diff
@@ -183,6 +185,36 @@ class TerminationReason(StrEnum):
...
+      alive; wait up to grace_seconds for natural exit, then kill if still running.
+    - IMMEDIATE_KILL: timeout, stall, or stale — kill without draining.
+    """
+
+    NO_KILL = "no_kill"
+    DRAIN_THEN_KILL_IF_ALIVE = "drain_then_kill_if_alive"
+    IMMEDIATE_KILL = "immediate_kill"
+
+
+class KillReason(StrEnum):
```
Trecek, line 205: [info] slop: KillReason docstring contains motivation text ('resolving the cognitive contradiction "success=True + exit_code=-9"') that belongs in a commit message or PR description, not in source-level enum documentation.

src/autoskillit/core/_type_subprocess.py
```diff
@@ -103,6 +103,12 @@ class SubprocessResult:
...
+    """Why the subprocess was (or was not) killed after the race loop.
```
Trecek, line 107: [info] slop: SubprocessResult.kill_reason field docstring narrates the write path ('Set by run_managed_async via execute_termination_action') rather than the field's semantic meaning. Implementation-path narration belongs in function comments, not on the field.

src/autoskillit/core/_type_results.py
```diff
@@ -145,6 +145,12 @@ class SkillResult:
     write_path_warnings: list[str] = field(default_factory=list)
     write_call_count: int = 0
     order_id: str = ""
+    kill_reason: KillReason = KillReason.NATURAL_EXIT
+    """Why the subprocess was (or was not) killed after the race loop.
```
Trecek, line 149: [info] slop: SkillResult.kill_reason field docstring repeats the same 'resolving the success=True + exit_code=-9 contradiction' framing as the KillReason enum docstring. Duplicate motivation prose across two docstrings; one should be removed.

src/autoskillit/execution/process.py
```diff
@@ -94,6 +104,74 @@ def _resolve_session_id(
     return stdout_session_id or channel_b_session_id or ""
 
 
+def decide_termination_action(
+    termination: TerminationReason,
+    *,
+    timeout_fired: bool,
+    process_exited: bool,
+) -> TerminationAction:
+    """Pure decision function: maps race signals to a TerminationAction.
```
Trecek, line 115: [info] slop: `decide_termination_action` docstring restates the priority rules in a 'Priority: 1/2/3' list that duplicates the match statement below it. The ordered list is a parallel narration of the code — the match is self-documenting.

src/autoskillit/execution/process.py
```diff
@@ -94,6 +104,74 @@ def _resolve_session_id(
...
+    action: TerminationAction,
+    *,
+    proc: anyio.abc.Process,
+    process_exited_event: anyio.Event,
+    grace_seconds: float,
+    proc_log: Any,
+) -> KillReason:
+    """Single authorized executor for all kill decisions in run_managed_async.
+
+    This is the ONLY function in process.py permitted to call
```
Trecek, line 153: [info] slop: `execute_termination_action` docstring cross-references a test by name ('enforced by test_no_direct_async_kill_process_tree_outside_executor'). Cross-referencing tests in production docstrings is brittle (test renames break the comment) and conflates enforcement with documentation.

src/autoskillit/execution/_process_race.py
```diff
@@ -23,6 +23,10 @@ class RaceSignals:
     Captures what happened without making any decisions about what it means.
     All fields are independent: multiple can be True simultaneously when tasks
     complete in the same event loop tick.
+
+    ``process_exited_event`` is the same object as on RaceAccumulator — setting
```
Trecek, line 27: [info] slop: Addition to RaceSignals docstring narrates implementation details of the shared-reference invariant and cross-references `execute_termination_action`. This is coupling data-class documentation to implementation consumers — the invariant is already tested.

src/autoskillit/execution/_process_race.py
```diff
@@ -42,6 +47,10 @@ class RaceAccumulator:
     In async cooperative concurrency there are no concurrent writes — only one
     coroutine runs between yield points. to_race_signals() converts to the
     frozen RaceSignals consumed by resolve_termination.
+
+    ``process_exited_event`` is set by _watch_process BEFORE setting trigger,
```
Trecek, line 51: [info] slop: RaceAccumulator docstring addition narrates which downstream function uses `process_exited_event`. Same anti-pattern: coupling data class documentation to implementation consumers.

tests/execution/test_termination_action.py
```diff
@@ -0,0 +1,91 @@
+"""Unit tests for decide_termination_action — pure decision function (1a).
+
+Tests the decision table that replaces the if/elif/else dispatch in process.py.
+All tests here must FAIL before Phase 2 implementation introduces
```
Trecek, line 4: [info] slop: Module docstring contains dead TDD phase-gate annotation: 'All tests here must FAIL before Phase 2 implementation introduces...' Implementation is complete and tests pass — remove this stale annotation.

tests/execution/test_termination_executor.py
```diff
@@ -0,0 +1,197 @@
+"""Integration tests for execute_termination_action (1b).
+
+Uses real subprocesses to verify drain-window behavior. All tests here must
```
Trecek, line 3: [info] slop: Module docstring contains dead TDD phase-gate annotation: 'All tests here must FAIL before Phase 2 implements execute_termination_action...' Implementation is complete — remove this stale annotation.

src/autoskillit/hooks/_fmt_execution.py
```diff
@@ -13,6 +13,28 @@
...
+def _format_exit_code_line(data: dict) -> str:
+    """Return the formatted exit_code line, annotated with kill_reason when present.
+
+    Legacy payloads without kill_reason render as bare exit_code values so that
+    JSON-line replay of old session logs still parses correctly.
+    """
+    exit_code = data.get("exit_code", "")
+    kill_reason = data.get("kill_reason")
+    if kill_reason is None:
+        # Legacy payload — no annotation
```
Trecek, line 25: [info] slop: Inline comment '# Legacy payload — no annotation' is obvious from the surrounding `if kill_reason is None` guard. The comment adds no information beyond what the condition already states.

src/autoskillit/hooks/_fmt_execution.py
```diff
@@ -13,6 +13,28 @@
...
+        return f"exit_code: {exit_code}"
+    if kill_reason == "kill_after_completion":
+        return f"exit_code: {exit_code} (infra-terminated after completion — grace exceeded)"
+    if kill_reason == "infra_kill":
+        termination_reason = data.get("termination_reason") or data.get("subtype") or ""
+        if termination_reason:
+            return f"exit_code: {exit_code} (infra-killed: {termination_reason})"
+        return f"exit_code: {exit_code} (infra-killed)"
+    # natural_exit or any unknown value → bare
```
Trecek, line 34: [info] slop: Comment '# natural_exit or any unknown value → bare' documents a silent fallback for unexpected enum values. A default `assert_never` or raise for unknown values would be more defensive than a silent fallback that could mask future bugs.

tests/config/test_defaults.py
```diff
@@ -0,0 +1,44 @@
...
+                exit_after_stop_delay_ms=2000,
+            )
+
+    def test_settings_validator_accepts_coherent_config(self) -> None:
+        """RunSkillConfig with sufficient grace window constructs without error."""
+        cfg = RunSkillConfig(
+            natural_exit_grace_seconds=3.0,
+            exit_after_stop_delay_ms=2000,
+        )
+        # 3.0 * 1000 = 3000 >= 2000 + 500 = 2500 ✓
```
Trecek, line 43: [info] slop: Inline comment '# 3.0 * 1000 = 3000 >= 2000 + 500 = 2500 ✓' narrates arithmetic any reader can verify in one second. AI-generated arithmetic annotation — remove it.

tests/infra/test_pretty_output.py
```diff
@@ -1779,3 +1779,93 @@ def test_pretty_output_public_surface_unchanged() -> None:
 
     assert callable(p.main)
     assert callable(p._format_response)
+
+
+# PHK-kill-reason: formatter branches on kill_reason (1e)
```
Trecek, line 1784: [info] slop: Section marker comment '# PHK-kill-reason: formatter branches on kill_reason (1e)' is a phase-tracking annotation (PHK prefix, parenthetical phase number). This belongs in commit messages or plan docs, not source files.

Trecek (commented):
AutoSkillit review found 13 blocking issues (warnings). See inline comments for details. Verdict: changes_requested.

src/autoskillit/pipeline/gate.py
Trecek: Agreed — fixed in e16f5e0. gate.py already imports from autoskillit.core (L1 module); KillReason directly reachable. Fixed in e16f5e0.

src/autoskillit/pipeline/gate.py
Trecek: Agreed — fixed in e16f5e0. Same root cause — headless_error_result also hardcoded 'natural_exit'. KillReason reachable from existing core import. Fixed in e16f5e0.

tests/execution/test_termination_executor.py
Trecek: Agreed — fixed in fc517da. Confirmed: 3 other test classes use side_effect=_mock_kill. Fixed to use side_effect=_mock_kill for consistency in fc517da.

tests/execution/test_termination_executor.py
Trecek: Investigated — this is intentional. Investigated — this is intentional. The double-lambda form correctly installs _mock_kill: patch calls outer lambda → inner lambda → _mock_kill itself. The assertion is not vacuous; the process genuinely exits naturally within the grace window.

tests/execution/test_termination_executor.py
Trecek: Investigated — this is intentional. Investigated — this is intentional. execute_termination_action blocks until proc_exited_event is set, which only fires after _watch() completes proc.wait(). Cancel after the function returns cannot interrupt an already-completed wait. No zombie risk in this flow.

tests/execution/test_termination_action.py
Trecek: Agreed — fixed in 926bb78. Comment described completely wrong case. Fixed in 926bb78: 'IDLE_STALL with process_exited=True, no timeout → NO_KILL (process exit wins)'.

src/autoskillit/execution/process.py
Trecek: Agreed — fixed in bb860c9. Changed proc_log: Any to proc_log: structlog.BoundLogger under TYPE_CHECKING guard in bb860c9. Only call site already passes logger.bind(pid=proc.pid).

src/autoskillit/execution/process.py
Trecek: Investigated — this is intentional. Investigated — this is intentional. DRAIN_THEN_KILL_IF_ALIVE has explicit guard at lines 162-164: 'if proc.returncode is not None: return KillReason.NATURAL_EXIT'. Beyond that, async_kill_process_tree catches psutil.NoSuchProcess immediately (_process_kill.py lines 27-29). No uncaught exception escapes.

src/autoskillit/execution/process.py
Trecek: Investigated — this is intentional. Investigated — this is intentional. kill_process_tree begins with 'try: parent = psutil.Process(pid) except psutil.NoSuchProcess: return' (_process_kill.py lines 27-29). If the process already exited, exception is caught and function returns silently. No ProcessLookupError or OSError reaches the caller.

src/autoskillit/hooks/_fmt_execution.py
Trecek: Investigated — this is intentional. Investigated — this is intentional. SkillResult.to_json() serializes BOTH 'subtype' (self.subtype) AND 'cli_subtype' (self.cli_subtype) as separate keys (_type_results.py lines 160-161). data.get('subtype') returns the CLI session subtype string, not None. No key mismatch exists.

src/autoskillit/execution/process.py
Trecek: Agreed — fixed in 0aa08b8. _kill_reason_hint manual string mapping removed in 0aa08b8. reason field now uses str(action) (canonical TerminationAction enum value). Debug logging tests updated to expect new canonical values.

src/autoskillit/hooks/_fmt_execution.py
Trecek: Investigated — this is intentional. Investigated — this is intentional. _fmt_execution.py is explicitly stdlib-only at runtime (enforced by import-linter contracts and arch gate tests). Importing KillReason at runtime violates the architectural isolation contract for hooks. String literals exactly match KillReason enum .value strings serialized by to_json().

tests/execution/test_termination_executor.py
Trecek: Agreed — fixed in fc517da. Same fix as 3075876452 — replaced double-lambda with side_effect=_mock_kill in fc517da for consistency with all other test classes in the file.

Issue #771, opened 2026-04-12
recover_crashed_sessions + test-code /dev/shm pollution produce fake crashed rows in sessions.jsonl


Trecek:
## Summary

Two linked bugs combine to make the AutoSkillit session dashboard unreliable:

1. **Test code writes `autoskillit_trace_<pid>.jsonl` files into the real `/dev/shm`** because several tests construct `LinuxTracingConfig` without overriding `tmpfs_path` (default: `/dev/shm`). The traced PIDs are test-helper processes (`sleep`, the pytest worker itself), not Claude Code subprocesses.

2. **`session_log.py::recover_crashed_sessions` globs `/dev/shm/autoskillit_trace_*.jsonl`** on server startup, and for every file older than 30s, unconditionally writes a `subtype=crashed` row to `sessions.jsonl` with `session_id=crashed_<pid>_<ts>` and empty `cwd` / `skill_command` / `kitchen_id` / `claude_code_log`. No check that the traced PID was ever a Claude Code process.

The failure dashboard is unusable as a health signal because ~92% of "failures" are these fake rows.

Whatever combination of fixes produces a clean, stable, accurate dashboard is in scope — the specifics are for whoever picks this up.

### Numbers

`~/.local/share/autoskillit/logs/sessions.jsonl` over the last 14 days (2026-03-29 → 2026-04-12), verified independently:

- **500 sessions total**
- **248 classified as failures** (`success=false`)
- **229 of those are fake `crashed` rows** with empty `skill_command`, empty `cwd`, empty `kitchen_id`, `session_id` starting with `crashed_`, `exit_code=-1`
- **19 are real failures** — real failure rate is **3.80%**, not the 49.6% the raw count suggests

### Concrete example

`~/.local/share/autoskillit/logs/sessions/crashed_1019554_2026-04-12T16-50-46.443560+00-00/proc_trace.jsonl` — 5 snapshots of PID 1019554:

```json
{"pid": 1019554, "vm_rss_kb": 256,  "fd_count": 3, "state": "running",  "wchan": "0",                  "threads": 1}
{"pid": 1019554, "vm_rss_kb": 1920, "fd_count": 3, "state": "sleeping", "wchan": "hrtimer_nanosleep", "threads": 1}
{"pid": 1019554, "vm_rss_kb": 1920, "fd_count": 3, "state": "sleeping", "wchan": "hrtimer_nanosleep", "threads": 1}
```

[...]

PR #809, opened 2026-04-13
Rectify: PTY Wrapper Tracer PID — Target-Resolution & Self-Identifying Snapshots (Issue #806)

### LINKED ISSUE #806: Linux process tracer monitors script(1) PTY wrapper instead of claude grandchild — all proc_trace snapshots report impossible 2 MB RSS / 1 thread
## Observed symptom — universal across all sessions

Every `proc_trace.jsonl` file under `~/.local/share/autoskillit/logs/sessions/` records the same implausible resource profile regardless of actual workload:

- `vm_rss_kb`: 1792–2176 KB (constant, never grows)
- `threads`: 1 (never more)
- `fd_count`: 4–7 (never more)
- `wchan`: `do_sys_poll` (always sleeping in poll)
- `cpu_percent`: 0.0 (zero CPU across hundreds of snapshots)
- `sig_blk`: `fffffffe7ffbfeff` (nearly all signals blocked)
- `oom_score`: 666 (high, paradoxical for 2 MB RSS)

This is not session-specific. Randomly sampled 5 non-flagged sessions (varying skills, 180 s–637 s durations, all `exit_code=0 success=true`):

| session | skill | duration | snapshots | vm_rss_kb | threads | fd_count |
|---|---|---|---|---|---|---|
| 29c359f3 | /arch-lens-concurrency | 180.9 s | 37 | 1792–2176 | 1 | 6–7 |
| bcc8e8c5 | /resolve-failures | 363.3 s | 74 | 2176 | 1 | 7 |
| 31896797 | /dry-walkthrough | 263.9 s | 54 | 2176 | 1 | 7 |
| 30e3f3f8 | /rectify | 636.6 s | 128 | 2176 | 1 | 7 |
| f13912bb | /dry-walkthrough | 326.8 s | 66 | 1920–2176 | 1 | 7 |

Plus the 5 `exit_code=-9` sessions flagged in the companion issue (#804), all with identical profile.

## What the real claude process looks like

Two currently-running claude processes on the same host:

| PID | cmdline | VmRSS | Threads | FDs | State |
|---|---|---|---|---|---|
| 5446 | `claude --dangerously-skip-permissions --resume 7e268930-...` | **915472 KB** (≈ 895 MB) | **24** | **67** | S (sleeping) |
| 1366786 | `claude --dangerously-skip-permissions ... --plugin-dir .../autoskillit ...` | **494204 KB** (≈ 483 MB) | **20** | **62** | S (sleeping) |

Native binary: `/home/talon/.local/share/claude/versions/2.1.104` — 233 MB ELF 64-bit executable (Node SEA). This is the real resource profile for a Claude CLI session: hundreds of MB RSS, 20+ threads, 60+ FDs.

**The proc_trace values cannot possibly be the claude process.**

## Root cause — PTY wrapping via script(1)

`src/autoskillit/execution/process.py:130-131` unconditionally wraps the command in `script(1)` when `pty_mode=True`:

```python
if pty_mode:
    cmd = pty_wrap_command(cmd)
```

`src/autoskillit/execution/_process_pty.py:10-27`:

```python
def pty_wrap_command(cmd: list[str]) -> list[str]:
    """Wrap a command with ``script`` to provide a PTY."""
    script_path = shutil.which("script")
    if script_path is None:
        return cmd
    if sys.platform == "darwin":
        return [script_path, "-q", "/dev/null"] + cmd
    escaped = " ".join(shlex.quote(c) for c in cmd)
    return [script_path, "-qefc", escaped, "/dev/null"]
```

So when claude is invoked via a Linux headless session, the final command becomes:

```
script -qefc "claude -p ... --dangerously-skip-permissions" /dev/null
```

Then at `process.py:144-152`:

```python
proc = await anyio.open_process(
    cmd,
    stdout=stdout_file,
    stderr=stderr_file,
    stdin=stdin_handle if stdin_handle is not None else subprocess.DEVNULL,
    cwd=cwd,
    env=env,
    start_new_session=True,
)
```

`anyio.open_process` spawns `script`, not `claude`. `proc.pid` is `script`'s PID.

At `process.py:222-226`:

```python
tracing_handle = start_linux_tracing(
    pid=proc.pid,
    config=linux_tracing_config,
    tg=tg,
)
```

`start_linux_tracing` receives `script`'s PID. The tracer loop at `linux_tracing.py:224-251` (`proc_monitor`) holds a single `psutil.Process(pid)` for the entire session lifetime and snapshots **only that PID** — no descendant walk. The tracer never sees the actual claude grandchild.

Process hierarchy at runtime:

```
autoskillit (parent)
  └── script (direct child; PID captured by anyio.open_process; tracked by proc_monitor)
        └── sh -c "claude ..." (fork from script)
              └── claude (real Node.js process — ~500-900 MB RSS, 20+ threads; NEVER tracked)
```

The `script` process sits in `do_sys_poll` forwarding bytes between the PTY master and stdout. It has 1 thread, ~2 MB RSS, 4–7 FDs (stdin, stdout, stderr, pty master, /dev/null, possibly signalfd) — an exact match for what proc_trace records.

## Why this wasn't caught

1. **Tests use synthetic processes that ARE the direct child.** `tests/execution/test_linux_tracing.py` spawns `subprocess.Popen(["sleep", "2"])` or uses `os.getpid()`. In both cases the PID *is* the process being observed — there is no PTY wrapper inserted in tests. The test suite never exercises the production code path that calls `pty_wrap_command`.
2. **No assertion compares tracer identity to expected process.** Tests assert snapshots are non-empty and timestamps are distinct. No test asserts `vm_rss_kb > 50_000` or `threads > 1` for a Claude-like workload, so the 2 MB / 1 thread signature has never tripped an assertion.
3. **`anomaly_count=0` looks healthy in telemetry.** The dashboard uses `sessions.jsonl` as a health signal. With the tracer blind to the real process, there are never any anomalies to report, and the dashboard looks green. The broken subsystem is silently broken.
4. **The `sig_blk=fffffffe7ffbfeff` signature matches script's sigprocmask behavior.** Anyone inspecting individual snapshots could have read this as "Claude masks signals" instead of "this isn't Claude".

## Consequences

- **`anomaly_count` is meaningless across all sessions.** The post-hoc anomaly detection in `src/autoskillit/execution/anomaly_detection.py` scans snapshots for D-state, high-CPU, or memory pressure events — but it only ever sees `script` idling in poll. Entire subsystem is non-functional for its intended purpose.
- **`peak_rss_kb`, `peak_fd_ratio`, `peak_oom_score` in `summary.json` are all wrong.** They describe `script`, not `claude`.
- **Resource telemetry in `sessions.jsonl` is unusable for capacity planning, leak detection, or diagnosing OOM/thrashing events.** Anyone grepping for high-RSS sessions to investigate a memory issue will find nothing.
- **Recent anomaly-related work (issues #631 "false stale kills during long-running Bash", #760 "D-state / high-CPU detection") operates on blind telemetry.** Those systems may be making decisions based on data that never reflects the real Claude process state.
- **The wrong-PID profile is the reason the companion issue #804 (`exit_code=-9 + success=true` ambiguity) shows `peak_oom_score=666` alongside `peak_rss_kb=2176`** — the implausible combination was a tell, but only because an investigation happened to look.

### Option A — Walk to the grandchild after spawn (simplest, least invasive)

After `anyio.open_process` returns, before calling `start_linux_tracing`, walk `script`'s descendants to find the claude PID. A short retry loop accommodates the fork/exec race:

```python
# pseudocode
import psutil, time
def find_claude_descendant(parent_pid: int, timeout: float = 2.0) -> int:
    deadline = time.monotonic() + timeout
    while time.monotonic() < deadline:
        try:
            parent = psutil.Process(parent_pid)
            for desc in parent.children(recursive=True):
                if desc.name() == "claude" or "claude" in " ".join(desc.cmdline()):
                    return desc.pid
        except psutil.NoSuchProcess:
            pass
        time.sleep(0.05)
    return parent_pid  # fall back to wrapper PID
```

Then: `start_linux_tracing(pid=find_claude_descendant(proc.pid), ...)`.

Pros: surgical, no architectural change, preserves PTY wrapping.
Cons: the 2-second retry loop adds startup latency; grandchild PID selection is heuristic.

### Option B — Track both PIDs

Change `proc_monitor` to accept a list of PIDs (or walk descendants) and snapshot all of them, tagging each snapshot with its PID. Anomaly detection and summary computation pick the Claude-like PID (highest RSS, or matched by name). Schema changes to `ProcSnapshot` and `sessions.jsonl`.

Pros: no race to resolve; captures script behavior too (useful for its own diagnostics).
Cons: breaks backward compatibility of `proc_trace.jsonl` schema and any downstream consumers; heavier refactor.

### Option C — Replace `script(1)` with a Python-native PTY fork

Use `pty.fork()` or `ptyprocess.PtyProcess` to allocate a PTY and fork claude directly from the parent, cutting out `script` entirely. The returned PID is the claude process itself.

Pros: architecturally cleanest; no descendant walk; tracer is correct by construction.
Cons: largest change; `pty.fork()` has Windows portability issues (already non-issue since tracing is Linux-only); needs careful handling of stdin/stdout/stderr inheritance that `script` currently handles for us.

### Option D — Add a PID-selection hook to `pty_wrap_command`

Have `pty_wrap_command` return both the wrapped command AND a "pid resolver" callback. The default resolver returns the direct child PID (current behavior). On Linux, the PTY-wrapping path returns a resolver that walks to the first grandchild matching the name in `cmd[0]`. Architectural middle ground between A and C.

## Affected components (file:line)

- `src/autoskillit/execution/process.py:130-131` — unconditional `pty_wrap_command` call
- `src/autoskillit/execution/process.py:144-152` — `anyio.open_process` captures wrapper PID
- `src/autoskillit/execution/process.py:222-226` — `start_linux_tracing(pid=proc.pid, ...)` passes wrapper PID to tracer
- `src/autoskillit/execution/_process_pty.py:10-27` — `pty_wrap_command` wraps in `script -qefc ... /dev/null`
- `src/autoskillit/execution/linux_tracing.py:224-251` — `proc_monitor` holds a single `psutil.Process` for the session lifetime; no descendant walk
- `src/autoskillit/execution/linux_tracing.py:120-141` — `ProcSnapshot` dataclass; no `comm`/`cmdline`/`exe` field, so identity of the tracked PID is never captured (blocks post-hoc validation)
- `src/autoskillit/execution/anomaly_detection.py` — consumer of wrong data
- `src/autoskillit/execution/session_log.py` — `flush_session_log` writes `peak_rss_kb`, `peak_fd_ratio`, `peak_oom_score` from the wrong PID

## Related / NOT-duplicate issues

- **#771 — recover_crashed_sessions + test-code /dev/shm pollution produces fake crashed rows.** That issue covers (a) test fixtures writing trace files to real `/dev/shm` and (b) `recover_crashed_sessions` producing fake `crashed_*` rows for orphaned non-Claude PIDs. It is in the same general area (proc tracer correctness) but addresses a **different leg**: orphan-file gating and test-suite tmpfs hygiene. This issue is about the **production path** where tracing IS happening for a real session but attaches to the wrong PID in the hierarchy.
- **#631 — false stale kills during long-running background Bash tasks.** Consumer of the same broken telemetry.
- **#760 — D-state / high-CPU anomaly detection work.** Operates on the same blind snapshots.
- **#804 — `run_skill: channel_won branch unconditionally SIGKILLs ... exit_code=-9 + success=true ambiguity`.** Companion issue; the proc_trace evidence in #804 is what surfaced this defect. Filing these separately because the root causes are orthogonal: #804 is about the kill path decision logic; this issue is about the tracer's PID selection.

## Test gaps

1. **No test asserts the tracer PID identity matches the intended target process.** `tests/execution/test_linux_tracing.py` uses `sleep` and `os.getpid()` directly without any PTY wrapper; the production interaction between `pty_wrap_command` and `start_linux_tracing` is untested.
2. **No test asserts snapshot sanity bounds.** E.g., `assert peak_rss_kb > some_floor` for a workload that clearly exceeds that floor.
3. **No test captures the identity (`comm`/`cmdline`/`exe`) of the tracked PID in `ProcSnapshot`.** Cannot post-hoc distinguish real claude from `script` wrapper.
4. **No integration test spawns a short-lived claude-equivalent via the PTY path and verifies the tracer sees the grandchild, not the wrapper.** A mock binary that prints "hello" and allocates 50 MB would be enough.

## Acceptance criteria for a fix

- `proc_trace.jsonl` for a real Claude CLI session records `vm_rss_kb > 100_000` (100 MB) and `threads >= 4` at least once during a 60+ second session.
- Unit test: `test_pty_wrapped_command_tracer_targets_grandchild` — spawns a PTY-wrapped helper that forks a memory-allocating child, asserts the tracer records the child's RSS signature, not the wrapper's.
- `ProcSnapshot` includes a `comm` field (or equivalent identity) in every snapshot so future drift is immediately diagnosable from `proc_trace.jsonl` alone.
- No regression in existing `tests/execution/test_linux_tracing.py`.
- `anomaly_count` in `sessions.jsonl` becomes non-zero for at least some sessions (not universally zero) — a coarse liveness check that anomaly detection is actually seeing the workload.
- `peak_oom_score=666` alongside `peak_rss_kb=2176` should no longer appear; the two fields should be internally consistent for the tracked process.

## Recommendation

Option A (grandchild walk) for speed of remediation, paired with adding `comm`/`cmdline` to `ProcSnapshot` so we have self-validating telemetry going forward. Option C (Python-native PTY fork) is the right long-term direction but is a bigger change; it can follow in a separate plan once Option A unblocks the dashboard.

Trecek:
## Summary

`anyio.open_process([script, -qefc, "<claude cmd>", /dev/null])` returns the PID of the `script(1)` PTY wrapper, not the intended claude binary. That wrapper PID was used as the subject of observation across ten call sites in `process.py` and downstream through the entire telemetry pipeline (tracer, anomaly detection, peak aggregation, `sessions.jsonl`, GitHub issue bodies). The broken subsystem has been silently wrong since PTY wrapping was introduced — and two prior investigations (#771/#776) saw the same `~2 MB RSS / 1 thread / do_sys_poll` fingerprint without identifying the PTY wrapper as the root cause.

This rectify produces immunity against the entire bug class — not just the `script(1)` instance — by introducing a **`TraceTarget`** newtype that can only be produced by a `resolve_trace_target()` resolver (which walks descendants and returns the workload process, not the spawn process), adding a **`comm` field to `ProcSnapshot`** so every row in `proc_trace.jsonl` self-identifies the process it describes, changing **`start_linux_tracing(pid: int)` to `start_linux_tracing(target: TraceTarget)`** so the raw-int path becomes unrepresentable, and adding an **AST architectural test** (ARCH-008) that forbids passing a raw `Attribute` PID node as the `target` argument to `start_linux_tracing`.

### Concurrency Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    START([● run_headless_core / DefaultHeadlessExecutor])

    subgraph PreSpawn ["SEQUENTIAL PRE-SPAWN  —  ● headless.py · ● process.py"]
        direction TB
        PTY_WRAP["● pty_wrap_command()<br/>━━━━━━━━━━<br/>Rewrites cmd: script -qefc<br/>Stores _workload_basename BEFORE rewrite"]
        SPAWN["● anyio.open_process()<br/>━━━━━━━━━━<br/>Returns script(1) wrapper PID<br/>(not the workload PID)"]
        PTY_DEC{"PTY<br/>mode?"}
    end

    subgraph PIDRect ["PID RECTIFICATION  —  ● linux_tracing.py  (BLOCKING SYNC before task group)"]
        direction TB
        RESOLVE["● resolve_trace_target()<br/>━━━━━━━━━━<br/>Blocks event loop: time.sleep(0.05) × ≤40<br/>Walks /proc descendants from wrapper PID<br/>Matches expected_basename (workload)"]
        RESOLVE_ERR["● TraceTargetResolutionError<br/>━━━━━━━━━━<br/>Hard fail — never silently falls back<br/>Carries .root_pid + .expected_basename"]
        DIRECT["● trace_target_from_pid()<br/>━━━━━━━━━━<br/>Single /proc read<br/>Non-PTY direct path"]
        TRACE_TARGET["● TraceTarget (frozen dataclass)<br/>━━━━━━━━━━<br/>pid · comm · cmdline<br/>starttime_ticks (PID-recycle guard)<br/>boot_id collision-resistant triple"]
        ENROLL["● _write_enrollment_atomic()<br/>━━━━━━━━━━<br/>tempfile.mkstemp + os.replace<br/>Crash-recovery identity record"]
    end

    FORK["● anyio.create_task_group() — FORK<br/>━━━━━━━━━━<br/>process.py: starts 5–6 concurrent tasks"]

    subgraph TaskGroup ["CONCURRENT ASYNC WATCHERS  —  anyio cooperative multitasking (single thread)"]
        direction LR
        W_PROC["_watch_process<br/>━━━━━━━━━━<br/>Awaits subprocess exit<br/>CHANNEL: natural exit"]
        W_HEART["_watch_heartbeat<br/>━━━━━━━━━━<br/>Polls stdout JSONL<br/>Completion marker records<br/>CHANNEL A"]
        W_SESS["_watch_session_log<br/>━━━━━━━━━━<br/>Polls Claude session JSONL dir<br/>CHANNEL B"]
        W_IDLE["_watch_stdout_idle<br/>━━━━━━━━━━<br/>Fires if stdout silent<br/>idle_output_timeout guard"]
        MONITOR["● _run_monitor / proc_monitor<br/>━━━━━━━━━━<br/>anyio.sleep(proc_interval)<br/>Reads /proc at each tick<br/>Accumulates ProcSnapshot list<br/>Injected into same TaskGroup"]
    end

    subgraph RaceSignals ["RACE SIGNALS  —  ● process.py (shared, anyio-safe)"]
        direction TB
        TRIGGER["● anyio.Event: trigger<br/>━━━━━━━━━━<br/>First watcher to fire wins"]
        CB_READY["anyio.Event: channel_b_ready<br/>━━━━━━━━━━<br/>Symmetric Channel B drain guard"]
        SID_READY["anyio.Event: stdout_session_id_ready<br/>━━━━━━━━━━<br/>Session ID from stdout JSONL"]
    end

    BARRIER["trigger.wait() + move_on_after(timeout)<br/>━━━━━━━━━━<br/>● process.py — awaits first race signal<br/>Drain window: move_on_after(completion_drain_timeout)"]
    CANCEL["tg.cancel_scope.cancel()<br/>━━━━━━━━━━<br/>Terminates all remaining tasks<br/>Including _run_monitor CancelScope"]

    subgraph PostHoc ["POST-HOC ANALYSIS  —  sequential, after task group exits"]
        direction TB
        STOP["● tracing_handle.stop()<br/>━━━━━━━━━━<br/>Cancels LinuxTracingHandle._monitor_cancel_scope<br/>Returns accumulated ProcSnapshot list"]
        RESULT["● SubprocessResult<br/>━━━━━━━━━━<br/>proc_snapshots: list[dict] | None<br/>tracked_comm: str|None  ← Issue #806 fix<br/>session_id · channel_confirmation · pid"]
        FLUSH["● flush_session_log()<br/>━━━━━━━━━━<br/>Writes proc_trace.jsonl to session dir"]
        ANOMALY["● detect_anomalies()<br/>━━━━━━━━━━<br/>Pure sync · post-hoc only<br/>OOM spike/critical · zombie · D-state<br/>high CPU · RSS growth · FDs"]
        DRIFT["● detect_identity_drift()<br/>━━━━━━━━━━<br/>Pure sync · checks each snapshot<br/>comm matches TraceTarget.comm"]
    end

    subgraph NewTests ["★ NEW INTEGRATION TESTS"]
        direction LR
        T_RESOLVER["★ test_trace_target_resolver.py<br/>━━━━━━━━━━<br/>test_resolve_trace_target_walks_from_wrapper_to_workload<br/>Asserts target.pid ≠ wrapper_pid<br/>test_resolve_trace_target_raises_on_miss<br/>Asserts TraceTargetResolutionError raised"]
        T_PTY["★ test_linux_tracing_pty_integration.py<br/>━━━━━━━━━━<br/>test_pty_wrapped_command_is_traced_as_grandchild<br/>Asserts peak RSS > 30,000 KB (not script wrapper 2 MB)<br/>test_pty_wrapped_tracing_produces_no_script_snapshots<br/>Asserts no comm='script' in proc_trace.jsonl"]
    end

    %% MAIN SEQUENTIAL FLOW %%
    START --> PTY_WRAP
    PTY_WRAP --> SPAWN
    SPAWN --> PTY_DEC
    PTY_DEC -->|"pty_mode=True"| RESOLVE
    PTY_DEC -->|"pty_mode=False"| DIRECT
    RESOLVE -->|"found within timeout"| TRACE_TARGET
    RESOLVE -->|"timeout exceeded"| RESOLVE_ERR
    DIRECT --> TRACE_TARGET
    TRACE_TARGET --> ENROLL
    ENROLL --> FORK

    %% FORK TO PARALLEL WATCHERS %%
    FORK --> W_PROC
    FORK --> W_HEART
    FORK --> W_SESS
    FORK --> W_IDLE
    FORK --> MONITOR

    %% WATCHERS FIRE TRIGGER %%
    W_PROC --> TRIGGER
    W_HEART --> TRIGGER
    W_SESS --> TRIGGER
    W_IDLE --> TRIGGER

    %% MONITOR ACCUMULATES (does not fire trigger) %%
    MONITOR -.->|"ProcSnapshot accumulation"| STOP

    %% RACE BARRIER %%
    TRIGGER --> BARRIER
    CB_READY -.->|"drain guard"| BARRIER
    SID_READY -.->|"drain guard"| BARRIER

    BARRIER --> CANCEL
    CANCEL --> STOP

    %% POST-HOC CHAIN %%
    STOP --> RESULT
    RESULT --> FLUSH
    FLUSH --> ANOMALY
    FLUSH --> DRIFT

    %% TEST COVERAGE (dashed) %%
    T_RESOLVER -.->|"exercises"| RESOLVE
    T_PTY -.->|"exercises"| MONITOR
    T_PTY -.->|"exercises"| FLUSH

    %% CLASS ASSIGNMENTS %%
    class START terminal;
    class PTY_WRAP,SPAWN handler;
    class PTY_DEC stateNode;
    class RESOLVE phase;
    class RESOLVE_ERR detector;
    class DIRECT phase;
    class TRACE_TARGET stateNode;
    class ENROLL output;
    class FORK detector;
    class W_PROC,W_HEART,W_SESS,W_IDLE handler;
    class MONITOR phase;
    class TRIGGER,CB_READY,SID_READY stateNode;
    class BARRIER,CANCEL detector;
    class STOP,RESULT output;
    class FLUSH,ANOMALY,DRIFT phase;
    class T_RESOLVER,T_PTY newComponent;
```

### Process Flow Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINALS %%
    START([START])
    COMPLETE([COMPLETE])
    ERROR([ERROR — no fallback])

    %% ── PHASE 1: SESSION ORCHESTRATION ── %%
    subgraph Orchestration ["Phase 1 — Session Orchestration (● headless.py)"]
        direction TB
        RHC["● run_headless_core<br/>━━━━━━━━━━<br/>Entry: skill + LinuxTracingConfig"]
        PTYMode["pty_mode = True<br/>━━━━━━━━━━<br/>Hardcoded for all headless sessions"]
    end

    %% ── PHASE 2: PTY PROCESS SPAWN ── %%
    subgraph Spawn ["Phase 2 — PTY Process Spawn (● process.py)"]
        direction TB
        CaptureBasename["● Capture _workload_basename<br/>━━━━━━━━━━<br/>Path(cmd[0]).name BEFORE PTY wrap<br/>(#806 guard — preserves real target name)"]
        WrapCmd["pty_wrap_command(cmd)<br/>━━━━━━━━━━<br/>Rewrites cmd → script(1) wrapper<br/>cmd[0] is now the PTY shim"]
        OpenProcess["anyio.open_process(cmd)<br/>━━━━━━━━━━<br/>Spawns; proc.pid = script(1) PID<br/>(NOT the claude workload)"]
    end

    %% ── PHASE 3: PID RECTIFICATION ── %%
    subgraph Rectification ["Phase 3 — PTY PID Rectification (● process.py + ● linux_tracing.py)"]
        direction TB
        TracingEnabled{"linux_tracing_config<br/>not None?"}
        PtyLinux{"pty_mode AND<br/>LINUX_TRACING_AVAILABLE?"}
        ResolveTarget["● resolve_trace_target<br/>━━━━━━━━━━<br/>root_pid=proc.pid, basename=workload<br/>Poll children() every 50 ms, timeout=2s<br/>Match name or cmdline[0] basename"]
        FoundTarget{"workload child<br/>found in time?"}
        TraceTargetDirect["● trace_target_from_pid<br/>━━━━━━━━━━<br/>Non-PTY: direct /proc read<br/>Builds TraceTarget; never raises"]
        RawPID["_observed_pid = proc.pid<br/>━━━━━━━━━━<br/>Tracing disabled path<br/>Raw wrapper PID used"]
        RectifiedPID["● _observed_pid = target.pid<br/>━━━━━━━━━━<br/>Rectified workload PID<br/>_tracked_comm = target.comm"]
        RTError["● TraceTargetResolutionError<br/>━━━━━━━━━━<br/>No silent fallback to wrapper PID<br/>(fallback would recreate #806)"]
    end

    %% ── PHASE 4: EXECUTION RACE ── %%
    subgraph ExecRace ["Phase 4 — Execution Race (● process.py)"]
        direction TB
        StartTracing["start_linux_tracing(target)<br/>━━━━━━━━━━<br/>target must be TraceTarget<br/>TypeError guard blocks raw int PIDs"]
        Race["Race Watchers<br/>━━━━━━━━━━<br/>_watch_process | _watch_heartbeat<br/>_watch_session_log | _watch_stdout_idle<br/>First to signal trigger wins"]
        ResolveTerm{"resolve_termination<br/>NATURAL_EXIT?"}
        KillTree["async_kill_process_tree<br/>━━━━━━━━━━<br/>Reaps PTY wrapper + workload<br/>for TIMED_OUT / STALL / STALE"]
        SubprocResult["● SubprocessResult<br/>━━━━━━━━━━<br/>pid=_observed_pid (rectified)<br/>tracked_comm=_tracked_comm"]
    end

    %% ── PHASE 5: SESSION LOGGING ── %%
    subgraph SessionLog ["Phase 5 — Session Logging (● session_log.py + ● anomaly_detection.py)"]
        direction TB
        FlushLog["● flush_session_log<br/>━━━━━━━━━━<br/>tracked_comm=result.tracked_comm<br/>Propagated from SubprocessResult"]
        ModalComm["● Modal comm resolution<br/>━━━━━━━━━━<br/>Compute from snapshots if absent<br/>Sets tracked_comm_drift flag"]
        DetectAnomalies["● detect_anomalies<br/>━━━━━━━━━━<br/>OOM / zombie / D-state<br/>CPU / FD / RSS thresholds"]
        IdentityDrift["● detect_identity_drift<br/>━━━━━━━━━━<br/>Emit IDENTITY_DRIFT CRITICAL<br/>if any snap.comm != expected_comm<br/>(immunity check for #806 regression)"]
        WriteArtifacts["● Write session artifacts<br/>━━━━━━━━━━<br/>summary.json (tracer_version=2)<br/>sessions.jsonl (tracked_comm)<br/>anomalies.jsonl (if non-empty)"]
    end

    %% ── PHASE 6: CRASH RECOVERY ── %%
    subgraph CrashRecovery ["Phase 6 — Crash Recovery (● session_log.py)"]
        direction TB
        RecoverScan["recover_crashed_sessions<br/>━━━━━━━━━━<br/>Scan tmpfs sidecars (age > 30s)"]
        Gate1{"Gate 1: enrollment<br/>sidecar exists?"}
        Gate2{"Gate 2: boot_id<br/>matches current boot?"}
        Gate3{"Gate 3: PID dead or<br/>starttime_ticks mismatch?"}
        Gate4{"● Gate 4: snap[0].comm<br/>non-empty AND != 'claude'?"}
        AlienReject["● Alien file rejection<br/>━━━━━━━━━━<br/>PTY wrapper trace detected<br/>Delete sidecar files, skip recovery<br/>(comm = script/pty shim name)"]
        RecoverFlush["flush_session_log<br/>━━━━━━━━━━<br/>subtype='crashed', exit_code=-1<br/>termination='CRASHED'"]
    end

    %% ── MAIN FLOW ── %%
    START --> RHC
    RHC --> PTYMode
    PTYMode --> CaptureBasename
    CaptureBasename --> WrapCmd
    WrapCmd --> OpenProcess
    OpenProcess --> TracingEnabled
    TracingEnabled -->|"YES"| PtyLinux
    TracingEnabled -->|"NO"| RawPID
    PtyLinux -->|"YES: PTY + Linux"| ResolveTarget
    PtyLinux -->|"NO: non-PTY or unavailable"| TraceTargetDirect
    ResolveTarget --> FoundTarget
    FoundTarget -->|"YES: TraceTarget built"| RectifiedPID
    FoundTarget -->|"NO: deadline exceeded"| RTError
    TraceTargetDirect --> RectifiedPID
    RawPID --> StartTracing
    RTError --> ERROR
    RectifiedPID --> StartTracing
    StartTracing --> Race
    Race --> ResolveTerm
    ResolveTerm -->|"YES: process exited cleanly"| SubprocResult
    ResolveTerm -->|"NO: TIMED_OUT / STALL / STALE"| KillTree
    KillTree --> SubprocResult
    SubprocResult --> FlushLog
    FlushLog --> ModalComm
    ModalComm --> DetectAnomalies
    DetectAnomalies --> IdentityDrift
    IdentityDrift --> WriteArtifacts
    WriteArtifacts --> COMPLETE

    %% ── CRASH RECOVERY FLOW (parallel operational path) ── %%
    RecoverScan --> Gate1
    Gate1 -->|"NO"| SKIP_G1["skip"]
    Gate1 -->|"YES"| Gate2
    Gate2 -->|"NO: stale boot"| SKIP_G2["delete + skip"]
    Gate2 -->|"YES"| Gate3
    Gate3 -->|"NO: still alive"| SKIP_G3["still running, skip"]
    Gate3 -->|"YES: process gone"| Gate4
    Gate4 -->|"YES: alien comm"| AlienReject
    Gate4 -->|"NO: claude or empty comm"| RecoverFlush
    AlienReject --> SKIP_G4["deleted, skipped"]
    RecoverFlush --> COMPLETE

    %% ── CLASS ASSIGNMENTS ── %%
    class START,COMPLETE,ERROR terminal;
    class SKIP_G1,SKIP_G2,SKIP_G3,SKIP_G4 terminal;
    class RHC,CaptureBasename,WrapCmd,OpenProcess phase;
    class PTYMode stateNode;
    class TracingEnabled,PtyLinux,FoundTarget,ResolveTerm stateNode;
    class ResolveTarget,TraceTargetDirect handler;
    class RectifiedPID,RawPID stateNode;
    class RTError detector;
    class StartTracing,Race,KillTree handler;
    class SubprocResult stateNode;
    class FlushLog,ModalComm,DetectAnomalies handler;
    class IdentityDrift detector;
    class WriteArtifacts output;
    class RecoverScan phase;
    class Gate1,Gate2,Gate3,Gate4 stateNode;
    class AlienReject detector;
    class RecoverFlush handler;
```

### State Lifecycle Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINALS %%
    START([PROCESS SPAWN])
    CRASH_END([CRASH RECOVERY COMPLETE])
    STOP_END([SESSION STOP])

    %% ──────────────────────────────────────── %%
    subgraph PIDResolution ["● PTY WRAPPER PID RESOLUTION  ·  linux_tracing.py"]
        PTYGate{"PTY mode?"}
        ResolvePTY["● resolve_trace_target<br/>━━━━━━━━━━<br/>Walk descendants at 50ms polls<br/>Match expected_basename<br/>Capture comm + cmdline + starttime_ticks<br/>HARD FAIL on timeout — NO wrapper fallback"]
        ResolveDirect["trace_target_from_pid<br/>━━━━━━━━━━<br/>Direct /proc read<br/>Returns empty strings on OSError<br/>Never raises"]
        ResolutionError["TraceTargetResolutionError<br/>━━━━━━━━━━<br/>root_pid + expected_basename carried<br/>Cannot trace: wrapper PID forbidden<br/>Abort — recreates #806 otherwise"]
    end

    %% ──────────────────────────────────────── %%
    subgraph InitOnlyFields ["INIT_ONLY FIELDS  ·  frozen=True  ·  linux_tracing.py ● / _type_subprocess.py ●"]
        TraceTarget["● TraceTarget  INIT_ONLY<br/>━━━━━━━━━━<br/>pid: workload PID only (not script wrapper)<br/>comm: /proc/pid/comm (max 15 chars)<br/>cmdline: tuple[str,...] immutable<br/>starttime_ticks: PID-reuse guard<br/>resolved_at: UTC datetime"]
        EnrollRecord["● TraceEnrollmentRecord  INIT_ONLY<br/>━━━━━━━━━━<br/>schema_version: 2 (post-#806)<br/>pid, boot_id, starttime_ticks<br/>session_id, enrolled_at, kitchen_id<br/>comm: new field — alien file rejection"]
        ProcSnap["● ProcSnapshot  INIT_ONLY<br/>━━━━━━━━━━<br/>comm: /proc/pid/comm  NEW field (#806)<br/>captured_at: unique UTC str (monotonic)<br/>vm_rss_kb, threads, fd_count, fd_soft_limit<br/>sig_pnd/blk/cgt, oom_score, wchan<br/>ctx_switches_voluntary/involuntary, cpu_percent"]
    end

    %% ──────────────────────────────────────── %%
    subgraph ContractGates ["CONTRACT ENFORCEMENT GATES  ·  linux_tracing.py ●"]
        TypeCheck{"isinstance(target,<br/>TraceTarget)?"}
        TypeFail["TypeError  HARD FAIL<br/>━━━━━━━━━━<br/>ARCH-008 / issue #806<br/>Raw int PID rejected explicitly<br/>No recovery path"]
        PlatformGate{"Linux + enabled<br/>+ tg not None?"}
        NoopReturn["Returns None<br/>━━━━━━━━━━<br/>Platform/config gate<br/>Tracing silently disabled"]
        MonoGate["Monotonic Timestamp Gate<br/>━━━━━━━━━━<br/>proc_monitor ensures captured_at<br/>strictly increases per snapshot<br/>+1μs applied on NTP/WSL2 regression"]
    end

    %% ──────────────────────────────────────── %%
    subgraph MutableHandle ["MUTABLE STATE  ·  LinuxTracingHandle  ·  linux_tracing.py ●"]
        HandleInit["Handle constructed<br/>━━━━━━━━━━<br/>_snapshots: []<br/>_trace_path: None<br/>_trace_file: None<br/>_enrollment_path: None<br/>_monitor_cancel_scope: None"]
        PostInit["Post-init mutation<br/>━━━━━━━━━━<br/>_monitor_cancel_scope set by start()<br/>_trace_path, _trace_file opened in tmpfs<br/>_enrollment_path written atomically<br/>All set inside start_linux_tracing()"]
        TraceDegraded["Degraded mode<br/>━━━━━━━━━━<br/>tmpfs unavailable → _trace_path=None<br/>write error → _trace_file closed/None<br/>Snapshots accumulate in-memory only"]
        HandleStop["stop() teardown<br/>━━━━━━━━━━<br/>Cancel scope cancelled<br/>_trace_file flushed + closed → None<br/>trace JSONL deleted from tmpfs<br/>enrollment sidecar deleted from tmpfs"]
    end

    %% ──────────────────────────────────────── %%
    subgraph RaceTransition ["MUTABLE → FROZEN TRANSITION  ·  process.py ●"]
        RaceAcc["● RaceAccumulator  MUTABLE<br/>━━━━━━━━━━<br/>process_exited, process_returncode<br/>channel_a_confirmed, channel_b_status<br/>channel_b_session_id, stdout_session_id, idle_stall<br/>Each field written exactly once by one coroutine"]
        RaceSignals["RaceSignals  INIT_ONLY after freeze<br/>━━━━━━━━━━<br/>Produced by to_race_signals()<br/>Immutable result consumed by resolve_termination()<br/>basis for SubprocessResult construction"]
    end

    %% ──────────────────────────────────────── %%
    subgraph ResultFields ["● SubprocessResult  MIXED-LIFECYCLE  ·  _type_subprocess.py ●"]
        ResultInit["Required at construction<br/>━━━━━━━━━━<br/>returncode, stdout, stderr<br/>termination, pid (workload PID post-#806)"]
        ResultPostSet["● Set post-construction by headless.py<br/>━━━━━━━━━━<br/>start_ts, end_ts: ISO timestamps<br/>elapsed_seconds: monotonic pre-computed<br/>proc_snapshots: list[dict] or None<br/>tracked_comm: TraceTarget.comm propagated<br/>channel_confirmation, session_id"]
    end

    %% ──────────────────────────────────────── %%
    subgraph CrashRecovery ["RESUME DETECTION  ·  recover_crashed_sessions()  ·  session_log.py ●"]
        AgeGate["Gate 1: Age Guard<br/>━━━━━━━━━━<br/>mtime within 30s → SKIP<br/>Active session protection"]
        EnrollGate["Gate 2: Enrollment Sidecar<br/>━━━━━━━━━━<br/>read_enrollment() = None → SKIP<br/>Alien/test file rejection"]
        BootGate["Gate 3: Boot ID Match<br/>━━━━━━━━━━<br/>boot_id mismatch → DELETE BOTH<br/>Pre-reboot stale rejection"]
        PIDGate["Gate 4: PID Liveness + starttime_ticks<br/>━━━━━━━━━━<br/>alive + ticks match → SKIP (still running)<br/>alive + ticks differ → CRASH (PID recycled)<br/>not alive → CRASH"]
        CommGate["● Gate 5: comm Alien Rejection  NEW<br/>━━━━━━━━━━<br/>snapshots[0].comm non-empty AND != claude<br/>→ DELETE trace + enrollment<br/>Pre-fix (no comm field) → RECOVER as-is"]
        CrashFlush["flush_session_log  CRASH path<br/>━━━━━━━━━━<br/>subtype=crashed, exit_code=-1<br/>detect_identity_drift() appended<br/>Deletes trace + enrollment after write"]
    end

    %% ──────────────────────────────────────── %%
    subgraph WriteOnlyArtifacts ["WRITE-ONLY SESSION ARTIFACTS  ·  session_log.py ●"]
        SummaryJson["summary.json<br/>━━━━━━━━━━<br/>Atomic write (tempfile+replace)<br/>tracer_target_resolution_version=2<br/>tracked_comm_drift flag<br/>peak_rss_kb, peak_oom_score, peak_fd_ratio"]
        ProcTrace["proc_trace.jsonl<br/>━━━━━━━━━━<br/>Every ProcSnapshot line-buffered<br/>Survives crash (stop() not called)<br/>Deleted after recovery or stop()"]
        SessionsJsonl["sessions.jsonl<br/>━━━━━━━━━━<br/>Append-only index (known race)<br/>Retention sweeps at 500 sessions<br/>WRITE-ONLY for logic decisions"]
        AnomaliesJsonl["● anomalies.jsonl<br/>━━━━━━━━━━<br/>detect_anomalies() output<br/>detect_identity_drift() — IDENTITY_DRIFT kind<br/>WRITE-ONLY — never read for decisions"]
    end

    %% ──────────────────────────────────────── %%
    %% NEW TEST NODES %%
    PTYIntegTest["★ test_linux_tracing_pty_integration.py<br/>━━━━━━━━━━<br/>No snapshot has comm=script<br/>max(vm_rss_kb) > 30,000 KB<br/>proc_trace.jsonl: comm on every row"]
    ResolverTest["★ test_trace_target_resolver.py<br/>━━━━━━━━━━<br/>resolved pid != wrapper pid<br/>TraceTargetResolutionError on miss<br/>error carries root_pid + expected_basename"]

    %% ──────────────────────────────────────── %%
    %% FLOW CONNECTIONS %%

    START --> PTYGate
    PTYGate -- "PTY=True" --> ResolvePTY
    PTYGate -- "PTY=False" --> ResolveDirect
    ResolvePTY -- "timeout" --> ResolutionError
    ResolvePTY -- "match found" --> TraceTarget
    ResolveDirect --> TraceTarget

    TraceTarget --> EnrollRecord
    TraceTarget --> TypeCheck

    TypeCheck -- "raw int PID" --> TypeFail
    TypeCheck -- "TraceTarget OK" --> PlatformGate
    PlatformGate -- "disabled/non-Linux" --> NoopReturn
    PlatformGate -- "enabled" --> HandleInit
    HandleInit --> PostInit
    PostInit -- "tmpfs ok" --> MonoGate
    PostInit -- "tmpfs missing" --> TraceDegraded
    MonoGate --> ProcSnap
    TraceDegraded --> ProcSnap

    ProcSnap --> RaceAcc
    ProcSnap --> ProcTrace
    RaceAcc --> RaceSignals
    RaceSignals --> ResultInit
    ResultInit --> ResultPostSet

    PostInit -- "enrolled_at" --> EnrollRecord
    ResultPostSet --> HandleStop
    HandleStop --> STOP_END

    %% Crash recovery path %%
    ProcTrace -- "process crashes (stop not called)" --> AgeGate
    AgeGate -- "age > 30s" --> EnrollGate
    EnrollGate -- "sidecar exists" --> BootGate
    BootGate -- "boot_id match" --> PIDGate
    PIDGate -- "crash detected" --> CommGate
    CommGate -- "comm=claude or legacy" --> CrashFlush
    CrashFlush --> SummaryJson
    CrashFlush --> SessionsJsonl
    CrashFlush --> AnomaliesJsonl
    CrashFlush --> CRASH_END

    %% Normal session log flush %%
    HandleStop --> SummaryJson
    HandleStop --> SessionsJsonl

    %% Test coverage arrows %%
    ResolvePTY -. "tested by" .-> PTYIntegTest
    ProcSnap -. "comm invariant" .-> PTYIntegTest
    TraceTarget -. "tested by" .-> ResolverTest
    ResolutionError -. "hard-fail contract" .-> ResolverTest

    %% CLASS ASSIGNMENTS %%
    class START,CRASH_END,STOP_END terminal;
    class PTYGate,TypeCheck,PlatformGate phase;
    class ResolvePTY,ResolveDirect handler;
    class ResolutionError,TypeFail detector;
    class TraceTarget,EnrollRecord,ProcSnap stateNode;
    class HandleInit,PostInit,TraceDegraded,HandleStop,MonoGate handler;
    class RaceAcc handler;
    class RaceSignals stateNode;
    class ResultInit,ResultPostSet stateNode;
    class AgeGate,EnrollGate,BootGate,PIDGate,CommGate detector;
    class CrashFlush handler;
    class SummaryJson,ProcTrace,SessionsJsonl,AnomaliesJsonl output;
    class PTYIntegTest,ResolverTest newComponent;
    class NoopReturn,TraceDegraded gap;
```

Closes #806

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-806-20260413-075454-098046/.autoskillit/temp/rectify/rectify_pty_wrapper_tracer_pid_806_2026-04-13_110443.md`

## Token Usage Summary

| investigate | 2.3k | 19.9k | 1.3M | 113.8k | 1 | 31m 1s |
| rectify | 3.4k | 24.7k | 540.5k | 76.0k | 1 | 9m 1s |
| review | 78 | 10.2k | 289.0k | 48.7k | 1 | 17m 29s |
| dry_walkthrough | 168 | 16.0k | 1.0M | 56.1k | 1 | 6m 33s |
| implement | 454 | 60.8k | 6.4M | 166.7k | 1 | 18m 30s |
| retry_worktree | 624 | 34.8k | 6.3M | 103.9k | 1 | 19m 7s |
| prepare_pr | 100 | 8.1k | 350.0k | 27.8k | 1 | 2m 29s |
| run_arch_lenses | 3.1k | 35.2k | 598.4k | 115.0k | 3 | 13m 45s |
| compose_pr | 67 | 12.8k | 251.8k | 40.6k | 1 | 3m 14s |
| **Total** | 10.3k | 222.6k | 17.1M | 748.6k | | 2h 1m |

### DIFF
--- a/src/autoskillit/execution/anomaly_detection.py
+++ b/src/autoskillit/execution/anomaly_detection.py
@@ -46,7 +46,7 @@ BENIGN_WCHANS: frozenset[str] = frozenset(
 
 def _safe_int(value: object, *, default: int) -> int:
     try:
-        return int(value)  # type: ignore[arg-type]
+        return int(value)  # type: ignore[call-overload]
     except (ValueError, TypeError):
         return default
 
--- a/src/autoskillit/execution/anomaly_detection.py
+++ b/src/autoskillit/execution/anomaly_detection.py
@@ -20,6 +20,7 @@ class AnomalyKind(StrEnum):
     FD_HIGH = "fd_high"
     D_STATE_SUSTAINED = "d_state_sustained"
     HIGH_CPU_SUSTAINED = "high_cpu_sustained"
+    IDENTITY_DRIFT = "identity_drift"
 
 
 class AnomalySeverity(StrEnum):
@@ -250,5 +251,50 @@ def detect_anomalies(
                     pid,
                 )
             )
+
+    return anomalies
+
+
+def detect_identity_drift(
+    snapshots: list[dict[str, object]],
+    expected_comm: str,
+) -> list[dict[str, object]]:
+    """Detect process identity drift: snapshots whose comm != expected_comm.
+
+    This is the architectural immunity check introduced in #806. If PTY wrapping
+    somehow bypasses the TraceTarget resolver, every snapshot will carry the wrong
+    comm (e.g., 'script' instead of 'claude'). This detector surfaces it immediately
+    rather than letting wrong telemetry accumulate silently for months.
+
+    Args:
+        snapshots: List of snapshot dicts (each may have a 'comm' field).
+        expected_comm: The process name we expect every snapshot to describe.
+
+    Returns:
+        List of IDENTITY_DRIFT anomaly records (one per first-seen mismatch).
+    """
+    anomalies: list[dict[str, object]] = []
+    if not snapshots or not expected_comm:
+        return anomalies
+
+    for seq, snap in enumerate(snapshots):
+        actual_comm = snap.get("comm", "")
+        if actual_comm and actual_comm != expected_comm:
+            anomalies.append(
+                _anomaly(
+                    AnomalyKind.IDENTITY_DRIFT,
+                    AnomalySeverity.CRITICAL,
+                    {
+                        "expected_comm": expected_comm,
+                        "actual_comm": actual_comm,
+                        "seq": seq,
+                    },
+                    snap,
+                    seq,
+                    int(snap.get("pid") or 0),  # type: ignore[call-overload]
+                )
+            )
+            # Report on first occurrence only — one anomaly is enough to diagnose the drift
+            break
 
     return anomalies
--- a/src/autoskillit/execution/anomaly_detection.py
+++ b/src/autoskillit/execution/anomaly_detection.py
@@ -44,6 +44,13 @@ BENIGN_WCHANS: frozenset[str] = frozenset(
 )
 
 
+def _safe_int(value: object, *, default: int) -> int:
+    try:
+        return int(value)  # type: ignore[arg-type]
+    except (ValueError, TypeError):
+        return default
+
+
 def _anomaly(
     kind: AnomalyKind,
     severity: AnomalySeverity,
@@ -291,7 +298,7 @@ def detect_identity_drift(
                     },
                     snap,
                     seq,
-                    int(snap.get("pid") or 0),  # type: ignore[call-overload]
+                    _safe_int(snap.get("pid"), default=0),
                 )
             )
             # Report on first occurrence only — one anomaly is enough to diagnose the drift
--- a/src/autoskillit/execution/linux_tracing.py
+++ b/src/autoskillit/execution/linux_tracing.py
@@ -21,6 +21,7 @@ import json
 import os
 import sys
 import tempfile
+import time
 from collections.abc import AsyncIterator
 from dataclasses import dataclass, field
 from datetime import UTC, datetime, timedelta
@@ -50,18 +51,166 @@ def read_boot_id() -> str | None:
 
 
 def read_starttime_ticks(pid: int) -> int | None:
-    """Read process starttime ticks from /proc/pid/stat."""
+    """Read process starttime ticks from /proc/pid/stat.
+
+    Uses rfind(")") to correctly locate the field boundary even when the
+    process comm contains a ")" character. Matches psutil's own _parse_stat_file()
+    which uses rfind(b")") for the same reason.
+    """
     try:
         stat = Path(f"/proc/{pid}/stat").read_text()
-        # comm may contain spaces; find the closing paren to parse fields after it
-        after_paren = stat.split(")", 1)
-        if len(after_paren) >= 2:
-            fields = after_paren[1].strip().split()
-            # starttime is field 22 in /proc/stat (0-indexed position 19 after state)
-            return int(fields[19])
+        # comm may contain ")" — use rfind to find the *last* ")" as the boundary
+        rpar = stat.rfind(")")
+        if rpar == -1:
+            return None
+        fields = stat[rpar + 2 :].split()
+        # starttime is field 22 (1-indexed per man page), offset 19 from the field after ")"
+        return int(fields[19])
     except (OSError, ValueError, IndexError):
         pass
     return None
+
+
+# ---------------------------------------------------------------------------
+# TraceTarget: workload process identity with provenance
+# ---------------------------------------------------------------------------
+
+
+class TraceTargetResolutionError(RuntimeError):
+    """Raised when resolve_trace_target cannot find the expected workload within timeout.
+
+    Attributes:
+        root_pid: The spawn PID (e.g., script(1)) that was walked.
+        expected_basename: The basename we were looking for (e.g., 'claude').
+    """
+
+    def __init__(self, root_pid: int, expected_basename: str) -> None:
+        self.root_pid = root_pid
+        self.expected_basename = expected_basename
+        super().__init__(
+            f"resolve_trace_target: timeout waiting for '{expected_basename}' "
+            f"to appear as a descendant of PID {root_pid}. "
+            f"The workload process did not start within the resolution window. "
+            f"Cannot trace: falling back to wrapper PID would recreate issue #806."
+        )
+
+
+@dataclass(frozen=True)
+class TraceTarget:
+    """Workload process identity with provenance — the correct target for the tracer.
+
+    Can only be produced by resolve_trace_target() (PTY mode, walks descendants)
+    or trace_target_from_pid() (non-PTY mode, direct PID). Never from a raw int.
+
+    Fields:
+        pid: PID of the workload process (not the spawn wrapper).
+        comm: /proc/{pid}/comm value — process name, max 15 chars.
+        cmdline: Full command line as a tuple of strings.
+        starttime_ticks: /proc/{pid}/stat field 22 — collision-resistant identity.
+        resolved_at: UTC datetime when this target was resolved.
+    """
+
+    pid: int
+    comm: str
+    cmdline: tuple[str, ...]
+    starttime_ticks: int
+    resolved_at: datetime
+
+
+def resolve_trace_target(
+    root_pid: int,
+    expected_basename: str,
+    timeout: float = 2.0,
+) -> TraceTarget:
+    """Walk descendants of root_pid to find the workload process by basename.
+
+    Used after anyio.open_process() in PTY mode: root_pid is the script(1) wrapper;
+    we need to find the actual workload (e.g., 'claude') in its subtree.
+
+    Polls at 50 ms intervals up to timeout. Raises TraceTargetResolutionError on miss
+    — never falls back to root_pid to prevent silent re-introduction of issue #806.
+
+    Args:
+        root_pid: Spawn PID (e.g., script(1) when PTY mode is active).
+        expected_basename: Basename to match (e.g., 'claude', 'python3').
+        timeout: Maximum seconds to wait for the workload to appear.
+
+    Returns:
+        TraceTarget with the workload's pid, comm, cmdline, starttime_ticks.
+
+    Raises:
+        TraceTargetResolutionError: When workload not found within timeout.
+    """
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        try:
+            root_proc = psutil.Process(root_pid)
+            children = root_proc.children(recursive=True)
+        except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
+            break
+
+        for child in children:
+            try:
+                name = child.name()
+                cmdline = child.cmdline()
+                basename_matches = name == expected_basename or (
+                    cmdline and Path(cmdline[0]).name == expected_basename
+                )
+                if not basename_matches:
+                    continue
+                # Found the workload — read identity fields from /proc
+                try:
+                    comm = Path(f"/proc/{child.pid}/comm").read_text().strip()
+                except OSError:
+                    comm = name
+                starttime = read_starttime_ticks(child.pid) or 0
+                return TraceTarget(
+                    pid=child.pid,
+                    comm=comm,
+                    cmdline=tuple(cmdline),
+                    starttime_ticks=starttime,
+                    resolved_at=datetime.now(UTC),
+                )
+            except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess, OSError):
+                continue
+
+        time.sleep(0.05)
+
+    raise TraceTargetResolutionError(root_pid=root_pid, expected_basename=expected_basename)
+
+
+def trace_target_from_pid(pid: int) -> TraceTarget:
+    """Build a TraceTarget directly from a PID without child walking.
+
+    Used for pty_mode=False (direct child, no wrapper) and in tests where the
+    spawn PID is already the workload.
+
+    Does a single /proc read to populate comm, cmdline, and starttime_ticks.
+    Never raises — returns empty strings/tuples on /proc read errors (process
+    may have already exited when this is called).
+    """
+    try:
+        comm = Path(f"/proc/{pid}/comm").read_text().strip()
+    except OSError:
+        comm = ""
+    try:
+        proc = psutil.Process(pid)
+        cmdline: tuple[str, ...] = tuple(proc.cmdline())
+    except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess, OSError):
+        cmdline = ()
+    starttime = read_starttime_ticks(pid) or 0
+    return TraceTarget(
+        pid=pid,
+        comm=comm,
+        cmdline=cmdline,
+        starttime_ticks=starttime,
+        resolved_at=datetime.now(UTC),
+    )
+
+
+# ---------------------------------------------------------------------------
+# TraceEnrollmentRecord: schema version 2 (adds comm field)
+# ---------------------------------------------------------------------------
 
 
 @dataclass(frozen=True)
@@ -71,9 +220,12 @@ class TraceEnrollmentRecord:
     (boot_id, pid, starttime_ticks) together form a collision-resistant identity:
     - boot_id rejects pre-reboot stale files
     - starttime_ticks detects PID recycling
+
+    Schema version 2 adds 'comm' to the enrollment record so that crash recovery
+    can verify the process identity (not just its PID) and reject alien trace files.
     """
 
-    schema_version: int  # always 1
+    schema_version: int  # 2 for post-#806 records; 1 for pre-fix records
     pid: int
     boot_id: str | None  # read_boot_id(); None if /proc unavailable
     starttime_ticks: int | None  # read_starttime_ticks(pid); None if unavailable
@@ -81,6 +233,7 @@ class TraceEnrollmentRecord:
     enrolled_at: str  # ISO 8601 UTC
     kitchen_id: str  # ""
     order_id: str  # ""
+    comm: str = ""  # /proc/{pid}/comm value; "" for pre-fix (schema_version=1) records
 
 
 def _write_enrollment_atomic(path: Path, record: TraceEnrollmentRecord) -> None:
@@ -112,17 +265,30 @@ def read_enrollment(path: Path) -> TraceEnrollmentRecord | None:
             enrolled_at=data.get("enrolled_at", ""),
             kitchen_id=data.get("kitchen_id", ""),
             order_id=data.get("order_id", ""),
+            comm=data.get("comm", ""),  # "" for pre-fix schema_version=1 records
         )
     except (OSError, KeyError, TypeError, ValueError, json.JSONDecodeError):
         return None
 
 
+# ---------------------------------------------------------------------------
+# ProcSnapshot: point-in-time snapshot with self-identifying comm field
+# ---------------------------------------------------------------------------
+
+
 @dataclass(frozen=True)
 class ProcSnapshot:
-    """Point-in-time snapshot of process state."""
+    """Point-in-time snapshot of process state.
+
+    The 'comm' field (added in #806 fix) self-identifies the process each snapshot
+    describes. Post-hoc drift detection becomes trivial: any row in proc_trace.jsonl
+    where comm != expected_comm is a drift indicator.
+    """
 
     # Temporal anchor — set at capture time, never reassigned
     captured_at: str
+    # Process identity — populated from /proc/{pid}/comm (max 15 chars kernel truncation)
+    comm: str
     # psutil-sourced fields
     state: str
     vm_rss_kb: int
@@ -186,6 +352,12 @@ def read_proc_snapshot(pid: int, *, process: psutil.Process | None = None) -> Pr
     except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
         return None
 
+    # Read process name from /proc/{pid}/comm (max 15 chars, kernel-truncated)
+    try:
+        comm = Path(f"/proc/{pid}/comm").read_text().strip()
+    except (FileNotFoundError, PermissionError):
+        comm = ""
+
     # Hand-rolled /proc reads for fields psutil doesn't expose
     try:
         status_content = Path(f"/proc/{pid}/status").read_text()
@@ -205,6 +377,7 @@ def read_proc_snapshot(pid: int, *, process: psutil.Process | None = None) -> Pr
 
     return ProcSnapshot(
         captured_at=captured_at,
+        comm=comm,
         state=state,
         vm_rss_kb=mem.rss // 1024,
         threads=num_threads,
@@ -284,7 +457,7 @@ class LinuxTracingHandle:
 
 
 def start_linux_tracing(
-    pid: int,
+    target: TraceTarget,
     config: LinuxTracingConfig,
     tg: anyio.abc.TaskGroup | None,
     *,
@@ -292,12 +465,26 @@ def start_linux_tracing(
     kitchen_id: str = "",
     order_id: str = "",
 ) -> LinuxTracingHandle | None:
-    """Start Linux tracing if all gates pass. Returns handle or None."""
+    """Start Linux tracing if all gates pass. Returns handle or None.
+
+    Args:
+        target: TraceTarget produced by resolve_trace_target() (PTY mode) or
+                trace_target_from_pid() (non-PTY mode). Never pass a raw int PID —
+                use the appropriate resolver to get a TraceTarget first (ARCH-008).
+    """
+    if not isinstance(target, TraceTarget):
+        raise TypeError(
+            f"start_linux_tracing: 'target' must be a TraceTarget, "
+            f"got {type(target).__name__!r}. "
+            "Use resolve_trace_target() (PTY mode) or trace_target_from_pid() (non-PTY mode) "
+            "instead of passing a raw int pid. (ARCH-008 / issue #806)"
+        )
     if not LINUX_TRACING_AVAILABLE or not config.enabled:
         return None
     if tg is None:
         return None
 
+    pid = target.pid
     handle = LinuxTracingHandle()
     scope = anyio.CancelScope()
 
@@ -313,17 +500,19 @@ def start_linux_tracing(
             handle._trace_file = None
 
         # Write enrollment sidecar atomically for crash-recovery identity contract
+        # Schema version 2: includes comm field for alien-file rejection in recovery
         enrollment_path = tmpfs / f"autoskillit_enrollment_{pid}.json"
         try:
             record = TraceEnrollmentRecord(
-                schema_version=1,
+                schema_version=2,
                 pid=pid,
                 boot_id=read_boot_id(),
-                starttime_ticks=read_starttime_ticks(pid),
+                starttime_ticks=target.starttime_ticks,
                 session_id=session_id,
                 enrolled_at=datetime.now(UTC).isoformat(),
                 kitchen_id=kitchen_id,
                 order_id=order_id,
+                comm=target.comm,
             )
             _write_enrollment_atomic(enrollment_path, record)
             handle._enrollment_path = enrollment_path
--- a/src/autoskillit/execution/process.py
+++ b/src/autoskillit/execution/process.py
@@ -61,6 +61,7 @@ if TYPE_CHECKING:
     import structlog
 
     from autoskillit.config import LinuxTracingConfig
+    from autoskillit.execution.linux_tracing import TraceTarget
 
 logger = get_logger(__name__)
 
@@ -240,7 +241,7 @@ async def run_managed_async(
             # script(1) wrapper, not claude. resolve_trace_target walks descendants
             # to find the actual workload by basename. Raising here (on miss) is
             # intentional: a silent fallback to proc.pid recreates issue #806.
-            _target: object = None
+            _target: TraceTarget | None = None
             _observed_pid: int = proc.pid
             _tracked_comm: str | None = None
             if linux_tracing_config is not None:
@@ -261,8 +262,9 @@ async def run_managed_async(
                 else:
                     # Non-PTY mode: proc.pid IS the workload (direct child)
                     _target = trace_target_from_pid(proc.pid)
-                _observed_pid = _target.pid  # type: ignore[union-attr]
-                _tracked_comm = _target.comm  # type: ignore[union-attr]
+                assert _target is not None
+                _observed_pid = _target.pid
+                _tracked_comm = _target.comm
 
             termination = TerminationReason.NATURAL_EXIT
             _channel_confirmation = ChannelConfirmation.UNMONITORED
@@ -333,7 +335,7 @@ async def run_managed_async(
                     from autoskillit.execution.linux_tracing import start_linux_tracing
 
                     tracing_handle = start_linux_tracing(
-                        target=_target,  # type: ignore[arg-type]
+                        target=_target,
                         config=linux_tracing_config,
                         tg=tg,
                     )
--- a/src/autoskillit/execution/process.py
+++ b/src/autoskillit/execution/process.py
@@ -247,7 +247,6 @@ async def run_managed_async(
             if linux_tracing_config is not None:
                 from autoskillit.execution.linux_tracing import (
                     LINUX_TRACING_AVAILABLE,
-                    TraceTargetResolutionError,  # noqa: F401 — used by raise below
                     resolve_trace_target,
                     trace_target_from_pid,
                 )
--- a/src/autoskillit/execution/process.py
+++ b/src/autoskillit/execution/process.py
@@ -208,6 +208,9 @@ async def run_managed_async(
     6. read_temp_output for results
     7. cleanup temp files via context manager
     """
+    # Capture workload basename before PTY wrapping rewrites cmd (#806)
+    _workload_basename = Path(cmd[0]).name if cmd else ""
+
     if pty_mode:
         cmd = pty_wrap_command(cmd)
 
@@ -232,10 +235,39 @@ async def run_managed_async(
                 start_new_session=True,
             )
 
+            # Resolve the workload TraceTarget — the PID that should be observed.
+            # anyio.open_process returns the spawn PID, which in PTY mode is the
+            # script(1) wrapper, not claude. resolve_trace_target walks descendants
+            # to find the actual workload by basename. Raising here (on miss) is
+            # intentional: a silent fallback to proc.pid recreates issue #806.
+            _target: object = None
+            _observed_pid: int = proc.pid
+            _tracked_comm: str | None = None
+            if linux_tracing_config is not None:
+                from autoskillit.execution.linux_tracing import (
+                    LINUX_TRACING_AVAILABLE,
+                    TraceTargetResolutionError,  # noqa: F401 — used by raise below
+                    resolve_trace_target,
+                    trace_target_from_pid,
+                )
+
+                if pty_mode and LINUX_TRACING_AVAILABLE:
+                    # PTY mode: proc.pid is the script(1) wrapper — resolve to workload
+                    _target = resolve_trace_target(
+                        root_pid=proc.pid,
+                        expected_basename=_workload_basename,
+                        timeout=2.0,
+                    )
+                else:
+                    # Non-PTY mode: proc.pid IS the workload (direct child)
+                    _target = trace_target_from_pid(proc.pid)
+                _observed_pid = _target.pid  # type: ignore[union-attr]
+                _tracked_comm = _target.comm  # type: ignore[union-attr]
+
             termination = TerminationReason.NATURAL_EXIT
             _channel_confirmation = ChannelConfirmation.UNMONITORED
 
-            proc_log = logger.bind(pid=proc.pid)
+            proc_log = logger.bind(pid=_observed_pid, comm=_tracked_comm)
             proc_log.debug(
                 "run_managed_async_entry",
                 cmd_summary=cmd[0] if cmd else "<empty>",
@@ -277,7 +309,7 @@ async def run_managed_async(
                         stale_threshold,
                         time.time(),
                         session_record_types,
-                        proc.pid,
+                        _observed_pid,
                         completion_drain_timeout,
                         acc,
                         trigger,
@@ -297,11 +329,11 @@ async def run_managed_async(
                         trigger,
                     )
                 tracing_handle = None
-                if linux_tracing_config is not None:
+                if linux_tracing_config is not None and _target is not None:
                     from autoskillit.execution.linux_tracing import start_linux_tracing
 
                     tracing_handle = start_linux_tracing(
-                        pid=proc.pid,
+                        target=_target,  # type: ignore[arg-type]
                         config=linux_tracing_config,
                         tg=tg,
                     )
@@ -337,7 +369,7 @@ async def run_managed_async(
                 from autoskillit.execution.linux_tracing import read_proc_snapshot
 
                 accumulated = tracing_handle.stop()
-                final_snap = read_proc_snapshot(proc.pid)
+                final_snap = read_proc_snapshot(_observed_pid)
                 if final_snap:
                     accumulated.append(final_snap)
                 snapshots_data = [s.__dict__ for s in accumulated]
@@ -377,7 +409,7 @@ async def run_managed_async(
                 stdout=stdout,
                 stderr=stderr,
                 termination=termination,
-                pid=proc.pid,
+                pid=_observed_pid,
                 channel_confirmation=_channel_confirmation,
                 proc_snapshots=snapshots_data,
                 channel_b_session_id=signals.channel_b_session_id,
@@ -385,6 +417,7 @@ async def run_managed_async(
                     signals.stdout_session_id, signals.channel_b_session_id
                 ),
                 kill_reason=kill_reason,
+                tracked_comm=_tracked_comm,
             )
             proc_log.debug(
                 "run_managed_async_result",
--- a/src/autoskillit/execution/session_log.py
+++ b/src/autoskillit/execution/session_log.py
@@ -404,18 +404,19 @@ def recover_crashed_sessions(tmpfs_path: str = "/dev/shm", log_dir: str = "") ->
             continue
 
         # Gate 4: comm-based alien file rejection (issue #806 immunity)
-        # If the first snapshot has a non-empty comm that is not "claude", this trace
-        # file describes a non-autoskillit process and should be excluded from recovery.
-        # Pre-fix snapshots (comm absent or empty) are recovered as-is to avoid
-        # silently dropping legitimate crash data.
+        # Use enrollment.comm as the expected comm (schema_version=2 records carry the
+        # enrolled binary name). Pre-fix schema_version=1 records have comm="" — skip
+        # the check for those to preserve recovery of legitimate crash data.
         _is_alien = False
-        if snapshots:
+        expected_comm = enrollment.comm
+        if snapshots and expected_comm:
             first_comm = snapshots[0].get("comm", "")
-            if first_comm and isinstance(first_comm, str) and first_comm != "claude":
+            if first_comm and isinstance(first_comm, str) and first_comm != expected_comm:
                 logger.debug(
-                    "Skipping %s: alien comm '%s' (expected 'claude')",
+                    "Skipping %s: alien comm '%s' (expected '%s')",
                     trace_file.name,
                     first_comm,
+                    expected_comm,
                 )
                 _is_alien = True
         if _is_alien:
--- a/src/autoskillit/execution/session_log.py
+++ b/src/autoskillit/execution/session_log.py
@@ -20,7 +20,7 @@ from typing import Any
 import psutil
 
 from autoskillit.core import atomic_write, claude_code_log_path, get_logger
-from autoskillit.execution.anomaly_detection import detect_anomalies
+from autoskillit.execution.anomaly_detection import detect_anomalies, detect_identity_drift
 from autoskillit.execution.linux_tracing import (
     read_boot_id,
     read_enrollment,
@@ -97,6 +97,7 @@ def flush_session_log(
     write_path_warnings: list[str] | None = None,
     write_call_count: int = 0,
     clone_contamination_reverted: bool = False,
+    tracked_comm: str | None = None,
 ) -> None:
     """Flush session diagnostics to disk.
 
@@ -128,6 +129,8 @@ def flush_session_log(
     peak_oom_score = 0
     peak_fd_ratio = 0.0
     anomalies: list[dict[str, object]] = []
+    _effective_tracked_comm: str | None = tracked_comm
+    _tracked_comm_drift: bool = False
 
     # Write proc_trace.jsonl
     if proc_snapshots:
@@ -158,8 +161,29 @@ def flush_session_log(
                     if ratio > peak_fd_ratio:
                         peak_fd_ratio = ratio
 
-        # Anomaly detection
+        # Compute effective tracked_comm from snapshots if not provided by caller
+        if _effective_tracked_comm is None:
+            # Use modal comm value across all snapshots
+            comm_counts: dict[str, int] = {}
+            for snap in proc_snapshots:
+                c = snap.get("comm", "")
+                if c and isinstance(c, str):
+                    comm_counts[c] = comm_counts.get(c, 0) + 1
+            if comm_counts:
+                _effective_tracked_comm = max(comm_counts, key=lambda k: comm_counts[k])
+
+        # Detect identity drift: if snapshots have mixed comm values, flag it
+        if _effective_tracked_comm:
+            comms_seen = {snap.get("comm", "") for snap in proc_snapshots if snap.get("comm", "")}
+            if len(comms_seen) > 1:
+                _tracked_comm_drift = True
+
+        # Anomaly detection (standard)
         anomalies = detect_anomalies(proc_snapshots, pid)
+        # Identity drift anomaly (post-fix immunity check)
+        if _effective_tracked_comm:
+            drift_anomalies = detect_identity_drift(proc_snapshots, _effective_tracked_comm)
+            anomalies.extend(drift_anomalies)
 
     # Write anomalies.jsonl (only if anomalies exist)
     if anomalies:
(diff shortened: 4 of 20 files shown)


tests/execution/test_linux_tracing_pty_integration.py
```diff
@@ -0,0 +1,142 @@
...
+
+import pytest
+
+pytestmark = pytest.mark.skipif(
+    sys.platform != "linux",
+    reason="Linux only — tests PTY wrapping + /proc tracing",
+)
+
+# Skip the entire module when script(1) is absent; no stub needed.
+pytestmark_script = pytest.mark.skipif(
```
Trecek, line 28: [critical] tests: Double `pytestmark` assignment — the module-level `pytestmark` on line 21 (Linux platform guard) is silently overwritten by the second `pytestmark = pytest.mark.skipif(shutil.which('script') is None, ...)` assignment here. Only the `script(1)` availability check survives; the Linux-only guard is lost. Tests in this module will execute on non-Linux platforms and fail. Fix: use a list: `pytestmark = [pytest.mark.skipif(sys.platform != 'linux', reason='Linux only — tests PTY wrapping + /proc tracing'), pytest.mark.skipif(shutil.which('script') is None, reason='script(1) not available on this system')]`

src/autoskillit/execution/session_log.py
```diff
@@ -371,6 +401,27 @@ def recover_crashed_sessions(tmpfs_path: str = "/dev/shm", log_dir: str = "") ->
...
 
+        # Gate 4: comm-based alien file rejection (issue #806 immunity)
+        # If the first snapshot has a non-empty comm that is not "claude", this trace
+        # file describes a non-autoskillit process and should be excluded from recovery.
+        # Pre-fix snapshots (comm absent or empty) are recovered as-is to avoid
+        # silently dropping legitimate crash data.
+        _is_alien = False
+        if snapshots:
+            first_comm = snapshots[0].get("comm", "")
+            if first_comm and isinstance(first_comm, str) and first_comm != "claude":
```
Trecek: [critical] bugs: `recover_crashed_sessions` hardcodes `'claude'` as the only valid comm for alien-file rejection. Any trace whose first snapshot has `comm != 'claude'` is deleted — including legitimate traces from python3, pytest, or any future binary rename. This is a data-loss bug. The expected comm should be read from the enrollment sidecar's own `comm` field (`TraceEnrollmentRecord.comm`, populated in schema_version=2 records) rather than compared against a hardcoded literal. The enrollment object is already in scope from earlier gates.

Trecek (commented):

src/autoskillit/execution/session_log.py
```diff
@@ -371,6 +401,27 @@ def recover_crashed_sessions(tmpfs_path: str = "/dev/shm", log_dir: str = "") ->
...
+                logger.debug(
+                    "Skipping %s: alien comm '%s' (expected 'claude')",
+                    trace_file.name,
+                    first_comm,
+                )
+                _is_alien = True
+        if _is_alien:
```
Trecek, line 422: [critical] bugs: `trace_file.unlink()` and `enrollment_path.unlink()` delete files based solely on `first_comm` from the snapshot — without using the enrollment record's `comm` field. The enrollment object (loaded in Gate 2/3) already carries the expected comm in schema_version=2 records; using it for alien rejection would be both more correct and self-consistent. Also: deleting without re-verifying against the enrollment record creates a TOCTOU window. Use `enrollment.comm` (from the already-loaded enrollment object) as the comparator instead.

tests/execution/test_session_log.py
```diff
@@ -592,6 +592,121 @@ def test_flush_writes_step_timing_json(tmp_path):
...
+    for s in recovered:
+        summary_file = s / "summary.json"
+        if summary_file.exists():
+            session_summaries.append(json.loads(summary_file.read_text()))
+
+    claude_sessions = [s for s in session_summaries if "20001" in s.get("session_id", "")]
+
+    # The claude trace must be recovered (not excluded)
+    assert claude_sessions or count >= 1, (
+        "The claude trace file must be recovered by recover_crashed_sessions"
```
Trecek: [warning] tests: Weak compound assertion `assert claude_sessions or count >= 1` — if the alien session is accidentally recovered and the claude session is not, `count >= 1` satisfies the `or` branch and the test passes without validating the intended invariant. Change to `assert claude_sessions` to make the assertion precise.

tests/execution/test_session_log_integration.py
```diff
@@ -135,3 +135,69 @@ def test_flush_session_log_surfaces_high_cpu_sustained(tmp_path):
...
+_ALLOCATE_60MB_SCRIPT = """\
+import time
+data = bytearray(60 * 1024 * 1024)  # 60 MB resident allocation
+time.sleep(3)
+"""
+
+
+@pytest.mark.anyio
+@pytest.mark.skipif(
+    __import__("shutil").which("script") is None,
```
Trecek, line 149: [warning] tests: `test_peak_rss_kb_above_sanity_floor` skips when `script(1)` is absent but has no Linux-platform guard. On macOS/Windows, `LINUX_TRACING_AVAILABLE` is False so `proc_snapshots` will be `None`, causing the `assert result.proc_snapshots is not None` to fire with a confusing message rather than a clean skip. Add `@pytest.mark.skipif(sys.platform != 'linux', reason='Linux only')` alongside the existing `script` check.

src/autoskillit/execution/process.py
```diff
@@ -151,10 +154,39 @@ async def run_managed_async(
                 start_new_session=True,
             )
 
+            # Resolve the workload TraceTarget — the PID that should be observed.
+            # anyio.open_process returns the spawn PID, which in PTY mode is the
+            # script(1) wrapper, not claude. resolve_trace_target walks descendants
+            # to find the actual workload by basename. Raising here (on miss) is
+            # intentional: a silent fallback to proc.pid recreates issue #806.
+            _target: object = None
```
Trecek: [warning] defense: `_target` is typed as `object = None` rather than `TraceTarget | None = None`, requiring two `# type: ignore[union-attr]` suppressions at lines 183-184 and 255. This erases the typed contract: a future maintainer adding a branch that doesn't assign `_target` gets a silent runtime `AttributeError` instead of a type error. Declare `_target: TraceTarget | None = None` (import is already present inside the block).

src/autoskillit/execution/anomaly_detection.py
```diff
@@ -252,3 +253,48 @@ def detect_anomalies(
...
+                    AnomalyKind.IDENTITY_DRIFT,
+                    AnomalySeverity.CRITICAL,
+                    {
+                        "expected_comm": expected_comm,
+                        "actual_comm": actual_comm,
+                        "seq": seq,
+                    },
+                    snap,
+                    seq,
+                    int(snap.get("pid") or 0),  # type: ignore[call-overload]
```
Trecek: [warning] defense: `int(snap.get('pid') or 0)` suppressed with `# type: ignore[call-overload]` does not guard against a non-numeric `pid` field (e.g. a string from a corrupt snapshot). `int('abc')` raises `ValueError` uncaught at the call site. Wrap in `try/except (ValueError, TypeError)` and fall back to `0`, or validate the field type before casting.

src/autoskillit/execution/linux_tracing.py
```diff
@@ -50,37 +51,189 @@ def read_boot_id() -> str | None:
...
+    Raises:
+        TraceTargetResolutionError: When workload not found within timeout.
+    """
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        try:
+            root_proc = psutil.Process(root_pid)
+            children = root_proc.children(recursive=True)
+        except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
+            break
```
Trecek, line 150: [warning] defense: In `resolve_trace_target`, when `psutil` raises `NoSuchProcess`/`AccessDenied`/`ZombieProcess` on `root_proc.children()`, the loop `break`s immediately and raises `TraceTargetResolutionError`. A single transient OS error causes permanent resolution failure with no retry. Replace `break` with `time.sleep(0.05); continue` so the polling loop can recover within the deadline.

tests/execution/test_session_log_integration.py
```diff
@@ -135,3 +135,69 @@ def test_flush_session_log_surfaces_high_cpu_sustained(tmp_path):
     assert anomalies_path.exists(), "anomalies.jsonl must be created when anomalies are detected"
     kinds = [json.loads(line)["kind"] for line in anomalies_path.read_text().splitlines()]
     assert "high_cpu_sustained" in kinds
+
+
+# ---------------------------------------------------------------------------
+# Test 1.6 — Sanity lower bound on peak RSS for a real workload
+# ---------------------------------------------------------------------------
+
+_ALLOCATE_60MB_SCRIPT = """\
```
Trecek: [warning] cohesion: `_ALLOCATE_60MB_SCRIPT` is duplicated verbatim from `tests/execution/test_linux_tracing_pty_integration.py` (same content, same variable name). Extract to a shared `conftest.py` fixture or `tests/execution/_helpers.py` to avoid silent divergence when the allocation script changes.

src/autoskillit/execution/session_log.py
```diff
@@ -157,8 +160,29 @@ def flush_session_log(
                     if ratio > peak_fd_ratio:
                         peak_fd_ratio = ratio
 
-        # Anomaly detection
+        # Compute effective tracked_comm from snapshots if not provided by caller
+        if _effective_tracked_comm is None:
```
Trecek, line 165: [warning] cohesion: Drift detection uses two parallel mechanisms in the same flush path: (1) inline modal-comm computation + set-cardinality check produces `_tracked_comm_drift` boolean (lines 164-178), and (2) `detect_identity_drift()` produces `IDENTITY_DRIFT` anomaly records. Both detect mismatched comm values but through different algorithms with different outputs. Consider collapsing the boolean drift flag into the anomaly detection path.

src/autoskillit/execution/process.py
```diff
@@ -151,10 +154,39 @@ async def run_managed_async(
...
+            _observed_pid: int = proc.pid
+            _tracked_comm: str | None = None
+            if linux_tracing_config is not None:
+                from autoskillit.execution.linux_tracing import (
+                    LINUX_TRACING_AVAILABLE,
+                    TraceTargetResolutionError,  # noqa: F401 — used by raise below
```
Trecek: [warning] slop: `TraceTargetResolutionError` is imported with `# noqa: F401 — used by raise below` but no `raise TraceTargetResolutionError(...)` appears in this scope — the exception propagates from `resolve_trace_target()` internally. The noqa comment is misleading. If the import is intentional (re-export), make it explicit; if unused, remove it.

Trecek (commented):
AutoSkillit review found 11 blocking issues (3 critical, 8 warning). See inline comments.

src/autoskillit/execution/session_log.py
Trecek: Agreed — fixed in 990d3a8. Use enrollment.comm (schema_version=2 field, set to the enrolled binary's comm) instead of hardcoded 'claude'. Gate 4 now uses enrollment.comm as expected_comm; pre-fix schema_version=1 records with comm='' skip the check to preserve backward compatibility.

src/autoskillit/execution/session_log.py
Trecek: Agreed — fixed in 990d3a8. Gate 4 alien rejection now uses enrollment.comm (loaded from enrollment record at line 373) as expected_comm, consistent with the enrollment data already available. Deletion decision follows the same comm comparison.

tests/execution/test_session_log.py
Trecek: Agreed — fixed in fa261ea. Changed to assert claude_sessions to pin the assertion directly to the intended invariant. Also updated the test helper to write schema_version=2 enrollment records (comm='claude') so Gate 4 correctly rejects the alien trace under the new enrollment.comm-based check.

src/autoskillit/execution/process.py
Trecek: Agreed — fixed in e977936. Added TraceTarget to the TYPE_CHECKING import, changed _target annotation to TraceTarget | None = None, added assert _target is not None guard before .pid/.comm access, and removed all three # type: ignore suppressions.

src/autoskillit/execution/anomaly_detection.py
Trecek: Agreed — fixed in 3e8248e. Added _safe_int(value, *, default: int) helper that wraps int() in try/except (ValueError, TypeError) and falls back to the default. Replaced the bare int(snap.get('pid') or 0) call with _safe_int(snap.get('pid'), default=0).

tests/execution/test_session_log_integration.py
Trecek: Agreed — fixed in 953d84d. Extracted _ALLOCATE_60MB_SCRIPT to tests/execution/conftest.py as a shared module-level constant. Removed the duplicate definitions from test_session_log_integration.py and test_linux_tracing_pty_integration.py, and imported from conftest in both.

src/autoskillit/execution/process.py
Trecek: Agreed — fixed in b129340. No raise TraceTargetResolutionError appears in process.py scope — the exception propagates from resolve_trace_target() internally. Removed the unused import entirely.

tests/execution/test_linux_tracing_pty_integration.py
Trecek: Investigated — this is intentional. Line 27 assigns to pytestmark_script (a distinct variable name), NOT pytestmark. The module-level pytestmark on line 21 (Linux platform guard) is NOT overwritten. pytestmark_script is a separate per-test decorator applied via @pytestmark_script at lines 40 and 94. Both guards coexist: the Linux guard applies module-wide, the script(1) guard applies per-test via decorator.

tests/execution/test_session_log_integration.py
Trecek: Investigated — this is intentional. The file has pytestmark = pytest.mark.skipif(sys.platform != 'linux', reason='Linux only') at line 14 (module-level), which applies to ALL tests in the module including test_peak_rss_kb_above_sanity_floor. On macOS/Windows, the entire module is skipped before any test body executes, so the assert result.proc_snapshots is not None line is never reached.

src/autoskillit/execution/linux_tracing.py
Trecek: Investigated — this is intentional. The except (NoSuchProcess, AccessDenied, ZombieProcess): break fires when root_proc.children() fails because the root process (root_pid, i.e. the script(1) wrapper) has disappeared. If the root is gone, all its descendants are gone too — continue would immediately re-fail on the next psutil.Process(root_pid) call, consuming the polling window uselessly. break + raise TraceTargetResolutionError is correct when the root disappears. Individual child errors are already handled by the inner except at line 174 with continue.

src/autoskillit/execution/session_log.py
Trecek: Valid observation — flagged for design decision. The inline modal-comm computation (lines 164-178) and detect_identity_drift() serve different consumers: the boolean _tracked_comm_drift feeds into the session summary JSON, while IDENTITY_DRIFT anomaly records go to anomalies.jsonl. Consolidating them requires deciding whether the boolean flag is redundant given anomaly records, which is a design trade-off for a future pass.

PR #818, opened 2026-04-13
Rectify: Pre-Queue Routing Block — Architectural Immunity

### LINKED ISSUE #805: Wasteful pre-queue routing: consolidate 3 merge-state checks and skip ci_watch on merge-group-only repos
## Summary

Two closely-related inefficiencies sit back-to-back on the critical path of every queue-capable recipe run. Both were flagged during a live pipeline run and both live in the pre-queue routing block of `implementation.yaml`, `remediation.yaml`, and `implementation-groups.yaml`. I'm filing them together because they touch the same steps, the same three recipe files, and the fix for one naturally composes with the fix for the other.

1. **Three sequential repo-state checks that should be one round-trip.** `check_merge_queue`, `check_merge_group_trigger`, and `check_auto_merge` run as three separate `run_cmd` / `gh` invocations in order, each hitting the same repo. Two are single GraphQL queries on `repository(owner, name)`. The third is an N+1 REST loop that lists every file in `.github/workflows/` and fetches each file's base64-encoded content until one contains the string `merge_group`. All three can collapse into a single GraphQL round-trip.
2. **`ci_watch` with `event: push` is a no-op on merge-group-only repos.** `wait_for_ci` returned `conclusion: no_runs` eight times across one parallel pipeline run because the target repo only triggers CI on `merge_group`. The push-event REST filter returns zero runs, the watcher burns its entire poll budget, and eventually times out. Worse: `check_merge_group_trigger` — which detects this exact condition — runs **after** `ci_watch`, so by the time the answer is in `context.merge_group_trigger`, the damage is already done.

## Relation to #802

This issue and #802 touch the **same three recipe files** (`implementation.yaml`, `remediation.yaml`, `implementation-groups.yaml`) and the **same module** (`src/autoskillit/execution/merge_queue.py`), but they sit in different subgraphs of the queue-capable recipes:

- **#802** is about the **post-enqueue classifier** (`wait_for_merge_queue` / `DefaultMergeQueueWatcher.wait`). It extends `_QUERY` at `merge_queue.py:36-61` to select `mergeable` and adds a new `pr_state: dropped_healthy` terminal.
- **This issue** is about the **pre-enqueue routing block** (`check_merge_queue` / `check_merge_group_trigger` / `check_auto_merge` / `ci_watch`). Option B below also touches `merge_queue.py` but adds a *new* query constant and a *new* helper function — it does not modify `_QUERY`.

The two issues do not conflict structurally. However, they both edit the same YAML regions in the three recipes, so implementing them in the same branch (or in strict serial order) will avoid merge-conflict churn. Either sequence works; this issue's changes are additive with respect to #802's and vice versa.

### Current state (verified line numbers)

All three recipe files contain an identical block:

- `src/autoskillit/recipes/implementation.yaml` lines 612-695
- `src/autoskillit/recipes/remediation.yaml` lines 587-656
- `src/autoskillit/recipes/implementation-groups.yaml` lines 662-731

Step graph inside that block:

```
check_merge_queue → check_merge_group_trigger → check_auto_merge → route_queue_mode
```

All three capture into `context` (`queue_available`, `merge_group_trigger`, `auto_merge_available`) and the results are only consumed together by `route_queue_mode` immediately after.

### What each step actually does

**`check_merge_queue`** — one `gh api graphql` call. Queries `repository(owner, name).mergeQueue(branch).id`. Single GraphQL round-trip.

**`check_merge_group_trigger`** — this is the hidden cost. Runs a shell script that:
1. `gh api /repos/{owner}/{repo}/contents/.github/workflows` to list workflow files (REST).
2. For each workflow file, `gh api /repos/{owner}/{repo}/contents/.github/workflows/{name}` to fetch base64-encoded content (REST).
3. Decodes base64 and greps for `merge_group`.
4. Stops at the first match.

This is an **N+1 REST loop**, not the \"GraphQL or REST call\" the original complaint described. On a repo with 15 workflow files that don't contain `merge_group` until the last one, that's 16 REST calls to resolve one boolean.

**`check_auto_merge`** — one `gh api graphql` call. Queries `repository(owner, name).autoMergeAllowed`. Single GraphQL round-trip.

All three resolve `REPO` from `git remote get-url origin` in `${{ context.work_dir }}`, so yes, all three hit the same repository.

### Why the obvious merge is only partial

`check_merge_queue` and `check_auto_merge` both select fields on the same `repository` node and are trivially mergeable into one GraphQL query. `check_merge_group_trigger` is a different beast — but GitHub's GraphQL API does support fetching workflow tree contents in a single query via `repository.object(expression: ...)`, which means all three **can** collapse to one round-trip if the fetch moves to GraphQL.

### Solution options

**Option A — Recipe-only merge of the two GraphQL steps.** Replace `check_merge_queue` + `check_auto_merge` with a single step that runs:

```graphql
query(\$owner: String!, \$repo: String!, \$branch: String!) {
  repository(owner: \$owner, name: \$repo) {
    mergeQueue(branch: \$branch) { id }
    autoMergeAllowed
  }
}
```

Both fields live on `repository` with no conflict. `jq` extracts two captures from one JSON blob. Zero Python. Saves one round-trip. Leaves the N+1 REST loop intact. Six YAML hunks total (lines 612-637 and 669-695 in each of the three recipe files).

Caveat: `check_merge_queue`'s current `on_failure` routes to `release_issue_success` (queue availability is a prerequisite); `check_auto_merge` falls through. The merged step must adopt the stricter `check_merge_queue` failure behavior.

**Option B — New `check_repo_merge_state` MCP tool. (Recommended.)** Add a new `@mcp.tool` to `src/autoskillit/server/tools_ci.py` (after the existing handlers ending around line 444, following the exact pattern of `toggle_auto_merge`/`wait_for_merge_queue`). Implementation lives in `src/autoskillit/execution/merge_queue.py` as a new `_REPO_STATE_QUERY` constant and a new `check_repo_merge_state(owner, repo, branch, token)` async function — mirroring the precedent at `merge_queue.py:36-61` where `_QUERY` already consolidates multi-field GraphQL for the post-enqueue watcher. (Do **not** modify the existing `_QUERY` — #802 edits that one.)

The combined GraphQL query:

```graphql
query(\$owner: String!, \$repo: String!, \$branch: String!) {
  repository(owner: \$owner, name: \$repo) {
    mergeQueue(branch: \$branch) { id }
    autoMergeAllowed
    object(expression: \"HEAD:.github/workflows\") {
      ... on Tree {
        entries {
          name
          object { ... on Blob { text } }
        }
      }
    }
  }
}
```

`Blob.text` returns raw decoded file content (not base64). Server-side the tool scans `entries[*].object.text` for `merge_group`. Returns:

```json
{\"queue_available\": bool, \"merge_group_trigger\": bool, \"auto_merge_available\": bool}
```

**Savings: 3+ REST calls → 1 GraphQL call.** Every recipe invocation saves N+1 round-trips where N is the workflow file count before the `merge_group` hit.

Recipe change: replace the three-step block in each of the three recipe files with a single `tool: check_repo_merge_state` step that captures all three fields from structured JSON. The `check_merge_queue` \"fail fast\" semantics encode as a structured error from the new tool; the recipe's `on_failure` handles it.

**Option C — New MCP tool for just the GraphQL fields, leave workflow scan as separate step.** Worst of both worlds: introduces Python machinery but leaves the N+1 REST problem. Not recommended.

### Current state (verified line numbers)

- `src/autoskillit/execution/ci.py` lines 95-107 (Phase 1 look-back) and 126-145 (active-runs fetch) both gate the event filter as `if scope.event: params[\"event\"] = scope.event`. If `scope.event` is unset, the REST query matches runs from any trigger.
- `src/autoskillit/execution/ci.py` lines 327-329: after the poll loop exhausts the timeout, returns unconditionally: `return {\"run_id\": None, \"conclusion\": \"no_runs\", \"failed_jobs\": []}`.
- `src/autoskillit/server/tools_ci.py` line 36 exposes `event: str | None = None`; line 103 builds `CIRunScope(event=event or tool_ctx.default_ci_scope.event)`.

### Affected call sites (all five)

| File | Step name | Line | Context |
|---|---|---|---|
| `src/autoskillit/recipes/implementation.yaml` | `ci_watch` | 592 | Pre-queue, after push, before `check_merge_queue` block |
| `src/autoskillit/recipes/remediation.yaml` | `ci_watch` | 568 | Same |
| `src/autoskillit/recipes/implementation-groups.yaml` | `ci_watch` | 591 | Same |
| `src/autoskillit/recipes/merge-prs.yaml` | `wait_for_conflict_ci` | 517 | Conflict-resolution worktree branch after push |
| `src/autoskillit/recipes/merge-prs.yaml` | `ci_watch_pr` | 878 | Integration branch after a review-fix push |

All five use `event: \"push\"`.

### Why it's a no-op on merge-group-only repos

When the target repo's workflows only trigger on `merge_group`, the REST query `GET /repos/.../actions/runs?event=push&branch=...&status=completed` returns zero rows. The watcher polls until the deadline and times out into `no_runs`. Observed eight times in a single parallel pipeline run — sitting on the critical path of every affected pipeline.

### Why the obvious skip doesn't work

`check_merge_group_trigger` **already detects** the answer and stores it in `context.merge_group_trigger`, but in the three non-`merge-prs` recipes it runs at line 639 — **three steps after `ci_watch` at line 588/568/591**. By the time the answer exists, the damage is done.

The naive fix (\"add `skip_when_false: context.merge_group_trigger` to `ci_watch`\") is blocked by a validator: `src/autoskillit/recipe/rules_bypass.py` lines 54-65 require `skip_when_false` to start with `inputs.` and reference a declared ingredient. `context.*` references are rejected at validation time.

In `merge-prs.yaml`, `wait_for_conflict_ci` and `ci_watch_pr` have **no surrounding merge-state detection at all** — there is no `check_merge_group_trigger` anywhere in their flow, so step reordering isn't sufficient to cover them.

### Solution options

**Option A — Recipe-level `on_result` bypass.** In `implementation.yaml`, `remediation.yaml`, and `implementation-groups.yaml`, reorder so `check_merge_group_trigger` (or, if Part 1 Option B ships, `check_repo_merge_state`) runs **before** `ci_watch`. Use an `on_result` routing clause on the detection step: when `merge_group_trigger == \"true\"`, route directly past `ci_watch` to the next step; otherwise fall through to `ci_watch` as today. Zero Python changes. Zero schema changes. Fully auditable in the recipe graph.

Only downside: moves the N+1 REST cost (or the single GraphQL cost under Option B) to the happy path for every recipe run. With Option B from Part 1, this is essentially free — one GraphQL call instead of three is a net win even on repos that do have push CI.

**Option B — Preflight in the CI watcher.** Teach `DefaultCIWatcher.wait` in `src/autoskillit/execution/ci.py` to check `.github/workflows/` for a push trigger before entering Phase 1. Self-healing across all five call sites with zero recipe changes. Cost: 1-2 extra API calls on every single `wait_for_ci` invocation, even for repos that do have push CI. Also duplicates logic that `check_merge_group_trigger` already owns.

**Option C — `event=None` + client-side filter, drop `event: push` from merge-prs.yaml.** Since `ci.py:106` and `:144` already gate the event filter (`if scope.event:`), omitting the `event` parameter causes the REST query to match runs from any trigger. Simply removing `event: \"push\"` from `wait_for_conflict_ci` (line 517) and `ci_watch_pr` (line 878) in `merge-prs.yaml` lets the watcher find whatever's running. No Python changes. This handles the two `merge-prs.yaml` call sites that Option A cannot reach.

**Recommended combined fix:**
- **Option A** for the three pre-queue `ci_watch` call sites in `implementation.yaml` / `remediation.yaml` / `implementation-groups.yaml` — with Part 1 Option B, `check_repo_merge_state` becomes the single early step that produces `merge_group_trigger`, and the `on_result` bypass on that step routes around `ci_watch` when `merge_group_trigger == \"true\"`.
- **Option C** for the two `merge-prs.yaml` call sites — simply drop `event: \"push\"`.

Five recipe locations, zero Python changes, all five call sites covered.

## Why this is one issue, not two

Both concerns sit in the same ~80-line block of three recipe files. Both are critical-path inefficiencies in the pre-queue routing. The recommended fix for Part 2 (reorder `check_merge_group_trigger` before `ci_watch`) composes naturally with Part 1 Option B (which replaces `check_merge_group_trigger` with `check_repo_merge_state`) — the reordered step becomes the new consolidated step. Splitting them would mean editing the same YAML regions twice in quick succession.

## Test plan

- [ ] `tests/server/test_tools_ci.py` — new test for `check_repo_merge_state` tool: GraphQL response with all three fields mocked, asserts returned dict shape.
- [ ] `tests/execution/test_merge_queue.py` — test new `_REPO_STATE_QUERY` constant exists and that `check_repo_merge_state` helper correctly parses the GraphQL tree response (including the case where `.github/workflows` is empty or missing — `object` returns `null`).
- [ ] `tests/execution/test_merge_queue.py` — test that workflow files containing `merge_group` cause `merge_group_trigger: true` and files without it cause `merge_group_trigger: false`.
- [ ] `tests/execution/test_merge_queue.py` — structural anti-recurrence assertion: `\"mergeQueue\"`, `\"autoMergeAllowed\"`, and `\"object(expression:\"` all appear in the new `_REPO_STATE_QUERY` literal, so a future refactor can't silently drop one of the three checks.
- [ ] `tests/recipe/test_*.py` — parametrised test over the three queue-capable recipes asserting the pre-queue block now contains exactly one step (`check_repo_merge_state`) instead of three, and that `route_queue_mode` consumes all three context captures from that one step.
- [ ] `tests/recipe/test_*.py` — test that `ci_watch` does not appear in the step chain when the recipe's precomputed `merge_group_trigger` is true (use a fixture that mocks the detection step's output and traces the routing).
- [ ] `tests/recipe/test_*.py` — test that `merge-prs.yaml`'s `wait_for_conflict_ci` and `ci_watch_pr` steps no longer set `event: \"push\"`.
- [ ] `tests/execution/test_ci.py` — sanity test that `CIRunScope(event=None)` omits the `event` REST parameter (already the behavior at `ci.py:106,144`, but lock it in).
- [ ] **Regression guard for #802 compatibility:** verify `_QUERY` in `merge_queue.py` is unchanged by this ticket's diff. The new `_REPO_STATE_QUERY` is a separate constant. #802 may ship before or after this ticket with no conflict in `merge_queue.py`.

## Files to change

- `src/autoskillit/execution/merge_queue.py` — add `_REPO_STATE_QUERY` constant, add `check_repo_merge_state(owner, repo, branch, token)` async function. Do **not** touch the existing `_QUERY` (that's #802's territory).
- `src/autoskillit/server/tools_ci.py` — register new `check_repo_merge_state` `@mcp.tool` after existing handlers (around line 444), follow the `toggle_auto_merge` pattern.
- `src/autoskillit/recipes/implementation.yaml` — replace three-step block (lines 612-695) with one `check_repo_merge_state` step; move it before `ci_watch` (currently line 588); add `on_result` routing on the new step that bypasses `ci_watch` when `merge_group_trigger == \"true\"`.
- `src/autoskillit/recipes/remediation.yaml` — same changes (three-step block at lines 587-656, `ci_watch` at line 568).
- `src/autoskillit/recipes/implementation-groups.yaml` — same changes (three-step block at lines 662-731, `ci_watch` at line 591).
- `src/autoskillit/recipes/contracts/implementation.yaml` — contract update: new step `check_repo_merge_state` produces `queue_available` + `merge_group_trigger` + `auto_merge_available`; old three steps removed.
- `src/autoskillit/recipes/contracts/remediation.yaml` — same contract update.
- `src/autoskillit/recipes/contracts/implementation-groups.yaml` — same contract update.
- `src/autoskillit/recipes/merge-prs.yaml` — remove `event: \"push\"` from `wait_for_conflict_ci` (line 517) and `ci_watch_pr` (line 878). No other changes.
- `tests/execution/test_merge_queue.py` — new tests for `_REPO_STATE_QUERY` parsing and helper behavior; anti-recurrence structural assertion.
- `tests/server/test_tools_ci.py` — MCP round-trip test for new tool.
- `tests/recipe/test_*.py` — step-graph tests per above.
- `tests/execution/test_ci.py` — `event=None` REST parameter sanity test.

**Do not touch:** `src/autoskillit/execution/merge_queue.py` `_QUERY`, `PRFetchState`, `_fetch_pr_and_queue_state`, or `DefaultMergeQueueWatcher.wait` — those are the territory of #802. This ticket only *adds* new symbols to `merge_queue.py`.

## Non-goals

- Webhook-driven repo-state change notifications. Out of scope.
- Refactoring `DefaultCIWatcher.wait` into a dispatched classifier. Out of scope.
- Any change to non-queue-capable recipe paths. The `check_repo_merge_state` consolidation only replaces steps that already exist in the queue-capable recipes.
- Any change to #802's post-enqueue misclassification fix. The two tickets are independent and can ship in either order.

## Note to the implementing agent

You will not have access to the local investigation that produced this ticket — everything needed is inline above: file paths, line numbers, verified GraphQL fragments, the exact five recipe call sites, the validator constraint that blocks the naive `skip_when_false` fix, and the compatibility boundary with #802. Treat this as the single source of truth.

The critical constraint to respect: **this ticket's changes to `src/autoskillit/execution/merge_queue.py` must be strictly additive.** Do not modify `_QUERY`, `PRFetchState`, or `_fetch_pr_and_queue_state`. Add a new query constant and a new helper function alongside them. This lets #802 and this ticket land independently in either order.

For Part 2 Option A's `on_result` bypass in the three pre-queue recipes, the step to consult is the *new* `check_repo_merge_state` step that replaces `check_merge_group_trigger`. The bypass routes around `ci_watch` directly to whatever step currently follows `ci_watch` in each recipe (verify per-recipe; the graphs are identical but the next-step name may differ). For Part 2 Option C, removing `event: \"push\"` from the two `merge-prs.yaml` steps is a two-line diff — the CI watcher already handles the unset case correctly at `ci.py:106,144`, no code changes needed there.

Run the full test suite before merging and confirm that **no tests outside the files listed above change behavior**, and specifically that `tests/execution/test_merge_queue.py` tests targeting `DefaultMergeQueueWatcher.wait` / `_QUERY` / `PRFetchState` remain green and unchanged — those are #802's territory.

Trecek:
## Summary

Issue #805 is not two bugs, it is two instances of **one** architectural class: *uninspected block-level properties of the recipe routing graph*. Part A of this rectify delivers the architectural primitive that makes the *budget* instance of the class impossible: a first-class `RecipeBlock` schema primitive, a new `rules_blocks.py` semantic-rule family that enforces per-block budgets on `run_cmd`, MCP tool counts, and forbidden shell patterns, and the consolidated `check_repo_merge_state` MCP tool that replaces the three sequential `run_cmd` steps in the pre-queue routing block.

After Part A lands, any future ticket that extends the `pre_queue_gate` block (or any other named block) with an additional `run_cmd`, additional `gh api` call, or additional MCP tool beyond its declared budget will fail `task test-all` at review time with a message naming the block, the budget, the actual count, and the remediation hint. No human reviewer vigilance is required.

**Part A scope:** block primitive, block-level budget rules, pre-queue block consolidation (the direct fix), and the failing tests that lock in the budgets. Part A is self-contained and can ship alone; every one of its tests is green at end of Part A with no dependency on later parts.

Part B (separate task) covers symbolic condition-aware reachability and capture-inversion detection, which addresses the *ordering* instance of the same class (the `ci_watch` event-hardcode inversion generalised across all four recipe files and the `diagnose-ci` hardcoded-push argument). Part B extends the primitives introduced in Part A but does not modify them — Part A's work is stable and complete on its own.

### Process Flow Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINALS %%
    START_KITCHEN([Kitchen Open])
    START_RECIPE([Recipe Validate])
    START_MERGE([Merge State Check])
    COMPLETE([COMPLETE])
    ERROR([ERROR])

    %% ── FLOW A: KITCHEN GATE ── %%
    subgraph KitchenGate ["● Kitchen Gate / Tool Visibility"]
        direction TB
        ServerStart["● server/__init__.py<br/>━━━━━━━━━━<br/>mcp.disable(tags=kitchen)<br/>41 tools hidden at startup"]
        OpenKitchen["● open_kitchen<br/>━━━━━━━━━━<br/>gate.enable()<br/>writes hook config + primes quota"]
        TriageGate["● _apply_triage_gate<br/>━━━━━━━━━━<br/>check staleness cache<br/>compute BlockFingerprints"]
        BlockFP["● BlockFingerprint<br/>━━━━━━━━━━<br/>member_count · tool_counts<br/>capture_names_hash · entry/exit"]
        FPDrift{"FP drift<br/>detected?"}
        LLMTriage["LLM triage<br/>━━━━━━━━━━<br/>Haiku subprocess ~30s<br/>result cached to disk"]
        RevealTools["● ctx.enable_components<br/>━━━━━━━━━━<br/>reveal 41 kitchen tools<br/>re-hide disabled subsets"]
    end

    %% ── FLOW B: RECIPE LOAD & BLOCK EXTRACTION ── %%
    subgraph RecipeLoad ["● Recipe Load & Block Extraction"]
        direction TB
        LoadRecipe["● load_recipe · io.py<br/>━━━━━━━━━━<br/>YAML → Recipe<br/>_parse_step: block field parsed"]
        BuildGraph["_build_step_graph<br/>━━━━━━━━━━<br/>adjacency dict<br/>_analysis.py"]
        ExtractBlocks["● extract_blocks<br/>━━━━━━━━━━<br/>_analysis.py<br/>RecipeBlock tuple → recipe.blocks<br/>entry/exit/tool_counts/gh_api"]
    end

    %% ── FLOW C: SEMANTIC VALIDATION ── %%
    subgraph SemanticVal ["● Semantic Validation Pipeline"]
        direction TB
        ValidateStructural["validate_recipe<br/>━━━━━━━━━━<br/>structural checks<br/>routing targets / capture refs"]
        RunSemanticRules["● run_semantic_rules<br/>━━━━━━━━━━<br/>registry.py<br/>make_validation_context"]
        BFSAnalysis["● _bfs_with_facts<br/>━━━━━━━━━━<br/>fact-propagating BFS<br/>intersects fact sets at joins"]
        StdRules["std _RULE_REGISTRY loop<br/>━━━━━━━━━━<br/>all @semantic_rule checks<br/>dataflow · merge · CI · etc."]
        HasBlocks{"ctx.blocks<br/>non-empty?"}
    end

    %% ── FLOW D: BLOCK-LEVEL RULES ★ ── %%
    subgraph BlockRules ["★ Block-Level Budget Rules (NEW)"]
        direction TB
        BlockCtxLoop["● BlockContext loop<br/>━━━━━━━━━━<br/>per RecipeBlock<br/>_BLOCK_RULE_REGISTRY dispatch"]
        RulesBlocks["★ rules_blocks.py<br/>━━━━━━━━━━<br/>5 rules: run_cmd/mcp_tools budget<br/>gh_api_forbidden · single_producer<br/>entry_exit_reachable"]
        LoadBudgets["★ block_budgets.yaml<br/>━━━━━━━━━━<br/>pre_queue_gate: run_cmd=0<br/>mcp_tools=1 · forbid_gh_api=true<br/>DEFAULT fallback (lru_cache)"]
        BudgetViolation{"budget<br/>violated?"}
        BudgetError["ERROR finding<br/>━━━━━━━━━━<br/>run_cmd / mcp_tools exceeded<br/>or gh_api_forbidden breach"]
    end

    %% ── FLOW E: SYMBOLIC REACHABILITY RULES ★ ── %%
    subgraph ReachabilityRules ["★ Symbolic Reachability Rules (NEW)"]
        direction TB
        RulesReachability["★ rules_reachability.py<br/>━━━━━━━━━━<br/>_check_capture_inversion<br/>_check_event_scope"]
        CaptureInvCheck{"context ref X<br/>in intersected<br/>fact set?"}
        EventScopeCheck{"merge_group_trigger<br/>producer is<br/>ancestor?"}
        ReachabilityError["ERROR finding<br/>━━━━━━━━━━<br/>capture inversion detected<br/>or missing upstream trigger"]
    end

    %% ── FLOW F: MERGE STATE MCP TOOL ★ ── %%
    subgraph MergeStateTool ["★ check_repo_merge_state — New MCP Tool"]
        direction TB
        CheckMergeState["★ check_repo_merge_state<br/>━━━━━━━━━━<br/>tools_ci.py<br/>consolidates pre_queue_gate block logic"]
        InferRepo["infer_repo_from_remote<br/>━━━━━━━━━━<br/>remote_resolver.py<br/>upstream › origin precedence"]
        RepoResolved{"owner/repo<br/>resolved?"}
        FetchMergeState["● fetch_repo_merge_state<br/>━━━━━━━━━━<br/>merge_queue.py<br/>single GraphQL round-trip"]
        ScanWorkflows["scan workflow YAMLs<br/>━━━━━━━━━━<br/>detect merge_group_trigger<br/>ci_event: push › merge_group › None"]
        MergeStateResult["★ state JSON result<br/>━━━━━━━━━━<br/>queue_available<br/>merge_group_trigger<br/>auto_merge_available · ci_event"]
    end

    %% ══════════════════════════════════════════
    %%  CONNECTIONS — FLOW A: KITCHEN GATE
    %% ══════════════════════════════════════════
    START_KITCHEN --> ServerStart
    ServerStart --> OpenKitchen
    OpenKitchen --> TriageGate
    TriageGate --> BlockFP
    BlockFP --> FPDrift
    FPDrift -->|"drift detected"| LLMTriage
    FPDrift -->|"no drift / cached"| RevealTools
    LLMTriage --> RevealTools
    RevealTools --> COMPLETE

    %% ══════════════════════════════════════════
    %%  CONNECTIONS — FLOW B+C+D+E: RECIPE VALIDATION
    %% ══════════════════════════════════════════
    START_RECIPE --> LoadRecipe
    LoadRecipe --> BuildGraph
    BuildGraph --> ExtractBlocks
    ExtractBlocks --> ValidateStructural
    ValidateStructural --> RunSemanticRules
    RunSemanticRules --> BFSAnalysis
    BFSAnalysis --> StdRules
    StdRules --> HasBlocks
    HasBlocks -->|"no blocks"| COMPLETE
    HasBlocks -->|"blocks present"| BlockCtxLoop
    BlockCtxLoop --> RulesBlocks
    RulesBlocks --> LoadBudgets
    LoadBudgets --> BudgetViolation
    BudgetViolation -->|"violation"| BudgetError
    BudgetViolation -->|"within budget"| RulesReachability
    BudgetError --> RulesReachability
    RulesReachability --> CaptureInvCheck
    CaptureInvCheck -->|"ref absent from fact set"| ReachabilityError
    CaptureInvCheck -->|"ok"| EventScopeCheck
    EventScopeCheck -->|"no merge_group ancestor"| ReachabilityError
    EventScopeCheck -->|"ok"| COMPLETE
    ReachabilityError --> COMPLETE

    %% ══════════════════════════════════════════
    %%  CONNECTIONS — FLOW F: MERGE STATE
    %% ══════════════════════════════════════════
    START_MERGE --> CheckMergeState
    CheckMergeState --> InferRepo
    InferRepo --> RepoResolved
    RepoResolved -->|"resolution failed"| ERROR
    RepoResolved -->|"owner/repo ok"| FetchMergeState
    FetchMergeState --> ScanWorkflows
    ScanWorkflows --> MergeStateResult
    MergeStateResult --> COMPLETE

    %% ══════════════════════════════════════════
    %%  CLASS ASSIGNMENTS
    %% ══════════════════════════════════════════
    class START_KITCHEN,START_RECIPE,START_MERGE,COMPLETE,ERROR terminal;
    class ServerStart,OpenKitchen,RevealTools phase;
    class TriageGate,BlockFP handler;
    class FPDrift,HasBlocks,BudgetViolation,CaptureInvCheck,EventScopeCheck,RepoResolved stateNode;
    class LLMTriage handler;
    class LoadRecipe,BuildGraph,ExtractBlocks,ValidateStructural,RunSemanticRules,BFSAnalysis,StdRules handler;
    class BlockCtxLoop phase;
    class RulesBlocks,LoadBudgets,RulesReachability,CheckMergeState,MergeStateResult newComponent;
    class BudgetError,ReachabilityError detector;
    class InferRepo,FetchMergeState,ScanWorkflows handler;
```

### State Lifecycle Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% ENTRY %%
    YAML["Recipe YAML<br/>━━━━━━━━━━<br/>name, steps, blocks<br/>ingredients"]

    %% REGISTRY (APPEND_ONLY at import time) %%
    subgraph Registries ["APPEND_ONLY REGISTRIES (import-time)"]
        direction LR
        RULE_REG["_RULE_REGISTRY<br/>━━━━━━━━━━<br/>Recipe-level RuleSpec list<br/>@semantic_rule decorator"]
        BLOCK_REG["★ _BLOCK_RULE_REGISTRY<br/>━━━━━━━━━━<br/>Block-level BlockRuleSpec list<br/>@block_rule decorator"]
    end

    %% GATE 1: STRUCTURAL %%
    subgraph Gate1 ["GATE 1 — STRUCTURAL VALIDATION (hard fail)"]
        STRUCT["validate_recipe()<br/>━━━━━━━━━━<br/>Discriminator exclusivity<br/>Routing target existence<br/>Capture expr format<br/>context-ref ordering<br/>→ list[str] errors"]
    end

    %% GATE 2: BLOCK EXTRACTION %%
    subgraph Gate2 ["GATE 2 — BLOCK EXTRACTION (INIT_ONLY production)"]
        direction LR
        EXTRACT["extract_blocks()<br/>━━━━━━━━━━<br/>Scans step.block annotations<br/>Populates Recipe.blocks<br/>RecipeBlock is frozen=True"]
        BUDGET_YAML["★ block_budgets.yaml<br/>━━━━━━━━━━<br/>lru_cache process-lifetime const<br/>run_cmd / mcp_tools / forbid_gh_api<br/>pre_queue_gate: rc=0, mcp=1, fga=true"]
    end

    %% GATE 3: CONTEXT %%
    subgraph Gate3 ["GATE 3 — VALIDATION CONTEXT (derived state)"]
        CTX["make_validation_context()<br/>━━━━━━━━━━<br/>step_graph built once<br/>DataFlowReport (non-blocking warns)<br/>★ BlockContext frozen per block"]
    end

    %% GATE 4: RECIPE-LEVEL SEMANTIC RULES %%
    subgraph Gate4 ["GATE 4 — RECIPE-LEVEL SEMANTIC RULES"]
        direction LR
        REACH["★ capture-inversion-detection<br/>━━━━━━━━━━<br/>BFS-with-facts per variable<br/>ERROR if context.X read<br/>but not guaranteed on all paths"]
        EVSCOPE["★ event-scope-requires-upstream-capture<br/>━━━━━━━━━━<br/>wait_for_ci literal event check<br/>ERROR if no upstream<br/>merge_group_trigger capture"]
    end

    %% GATE 5: BLOCK-LEVEL RULES %%
    subgraph Gate5 ["★ GATE 5 — BLOCK-LEVEL BUDGET RULES (per RecipeBlock)"]
        direction TB
        BUDGET_ERR["★ block-run-cmd-budget / block-mcp-tool-budget<br/>━━━━━━━━━━<br/>Reads RecipeBlock.tool_counts<br/>Compare to budget YAML entry<br/>ERROR if over budget"]
        GH_BAN["★ block-gh-api-forbidden<br/>━━━━━━━━━━<br/>Reads RecipeBlock.gh_api_occurrences<br/>ERROR if forbid_gh_api=true and > 0<br/>Enforces shell-free blocks"]
        WARN_RULES["★ block-single-producer / entry-exit-reachable<br/>━━━━━━━━━━<br/>WARNING — non-blocking<br/>Reads bctx.parent.step_graph<br/>Block connectivity integrity"]
    end

    %% GATE 6: CONTRACTS %%
    subgraph Gate6 ["GATE 6 — CONTRACT VALIDATION"]
        CONTRACTS["● validate_recipe_cards()<br/>━━━━━━━━━━<br/>contract-unreferenced-required<br/>contract-unsatisfied-input<br/>ERROR severity — blocks validity"]
    end

    %% GATE 7: STALENESS + FINGERPRINT DRIFT %%
    subgraph Gate7 ["★ GATE 7 — STALENESS / BlockFingerprint DRIFT"]
        CACHE_HIT["staleness_cache<br/>━━━━━━━━━━<br/>Key: recipe_hash + manifest_version<br/>Tier 1: HIT clean → return []<br/>Tier 2: stale → recompute<br/>Tier 3: miss → full check"]
        FP_DRIFT["★ block_composition_drift<br/>━━━━━━━━━━<br/>BlockFingerprint frozen equality<br/>7-field comparison per block<br/>member_count, tool_counts_sorted<br/>gh_api_occurrences, capture_names_hash<br/>entry_step, exit_step"]
        HASH_CHK["● skill hash / manifest version<br/>━━━━━━━━━━<br/>SHA256(SKILL.md) per skill<br/>bundled_manifest_version check<br/>WARNING — non-blocking"]
    end

    %% AGGREGATION %%
    VALID["● compute_recipe_validity()<br/>━━━━━━━━━━<br/>Aggregates: schema errors<br/>+ Severity.ERROR findings<br/>+ contract 'error' severity<br/>→ bool"]

    %% PRE-QUEUE GATE SUBSYSTEM %%
    subgraph PreQueue ["● PRE-QUEUE GATE: check_repo_merge_state consolidation"]
        direction TB
        REMOVED["(removed) 3× run_cmd steps<br/>━━━━━━━━━━<br/>check_merge_queue<br/>check_merge_group_trigger<br/>check_auto_merge<br/>Violated block budget rules"]
        NEW_TOOL["● check_repo_merge_state MCP tool<br/>━━━━━━━━━━<br/>Single GraphQL round-trip<br/>run_cmd=0 enforced by budget rule<br/>forbid_gh_api enforced by gh-ban rule"]
        PRFETCH["● PRFetchState (INIT_ONLY snapshot)<br/>━━━━━━━━━━<br/>merged · state · merge_state_status<br/>auto_merge_enabled_at · pr_node_id<br/>in_queue · queue_state · checks_state"]
        CI_EVT["● ci_event (DERIVED)<br/>━━━━━━━━━━<br/>push > merge_group > null<br/>workflow scan: on:push / merge_group<br/>→ queue_available, merge_group_trigger<br/>auto_merge_available"]
        POLL["● Poll Gate Sequence D1–D7<br/>━━━━━━━━━━<br/>D1 merged → terminal success<br/>D2 CLOSED → ejected/ci_failure<br/>D3 in_queue reset counter<br/>D5 stall: auto_merge toggle + retry<br/>D6 PENDING guard<br/>D7 final ejection classification"]
    end

    %% CONNECTIONS — main pipeline %%
    YAML --> STRUCT
    YAML -.->|"Recipe.steps parsed"| RULE_REG
    YAML -.->|"RecipeBlock registered"| BLOCK_REG
    STRUCT -->|"errors == []"| EXTRACT
    EXTRACT --> CTX
    BUDGET_YAML -.->|"process-const budget"| BUDGET_ERR
    CTX -->|"step_graph + DataFlowReport"| REACH
    CTX -->|"step_graph + DataFlowReport"| EVSCOPE
    REACH --> BUDGET_ERR
    EVSCOPE --> BUDGET_ERR
    BUDGET_ERR --> GH_BAN
    GH_BAN --> WARN_RULES
    WARN_RULES --> CONTRACTS
    CONTRACTS --> CACHE_HIT
    CACHE_HIT -->|"miss or stale"| FP_DRIFT
    CACHE_HIT -->|"clean hit"| VALID
    FP_DRIFT --> HASH_CHK
    HASH_CHK --> VALID

    %% PRE-QUEUE subsystem %%
    VALID --> REMOVED
    REMOVED -. "consolidated into" .-> NEW_TOOL
    NEW_TOOL --> PRFETCH
    PRFETCH --> CI_EVT
    PRFETCH --> POLL

    %% CLASS ASSIGNMENTS %%
    class YAML stateNode;
    class RULE_REG handler;
    class BLOCK_REG newComponent;
    class STRUCT detector;
    class EXTRACT phase;
    class BUDGET_YAML newComponent;
    class CTX phase;
    class REACH newComponent;
    class EVSCOPE newComponent;
    class BUDGET_ERR newComponent;
    class GH_BAN newComponent;
    class WARN_RULES newComponent;
    class CONTRACTS stateNode;
    class CACHE_HIT output;
    class FP_DRIFT newComponent;
    class HASH_CHK stateNode;
    class VALID detector;
    class REMOVED gap;
    class NEW_TOOL handler;
    class PRFETCH stateNode;
    class CI_EVT phase;
    class POLL handler;
```

### Module Dependency Diagram

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 50, 'rankSpacing': 80, 'curve': 'basis'}}}%%
graph TB
    %% CLASS DEFINITIONS %%

    %% ══════════════════════════════════════════════════
    %% LAYER 3 — APPLICATION
    %% ══════════════════════════════════════════════════
    subgraph L3 ["LAYER 3 — APPLICATION  (server/ · cli/)"]
        direction LR

        SRV_INIT["● server/__init__<br/>━━━━━━━━━━<br/>FastMCP app · tool registration<br/>Side-effect: imports all tools_*"]

        subgraph ServerTools ["Server Tool Handlers"]
            TOOLS_CI["● server/tools_ci<br/>━━━━━━━━━━<br/>check_repo_merge_state MCP tool<br/>pre_queue_gate consolidation"]
            TOOLS_KITCHEN["● server/tools_kitchen<br/>━━━━━━━━━━<br/>open_kitchen · kitchen gating<br/>Triggers recipe init"]
        end

        SRV_HELPERS["● server/helpers<br/>━━━━━━━━━━<br/>fetch_repo_merge_state re-export<br/>Wide fan-out: exec+recipe+pipeline+workspace"]

        CLI_COOK["● cli/_cook<br/>━━━━━━━━━━<br/>Ephemeral skill session launcher<br/>Imports: cli._terminal only"]
    end

    %% ══════════════════════════════════════════════════
    %% LAYER 2 — BUSINESS LOGIC
    %% ══════════════════════════════════════════════════
    subgraph L2 ["LAYER 2 — BUSINESS LOGIC  (recipe/)"]
        direction TB

        subgraph RulePlugins ["★ New Rule Modules (PR-added)"]
            direction LR
            RULES_BLOCKS["★ recipe/rules_blocks<br/>━━━━━━━━━━<br/>5 block-level rules<br/>block-run-cmd-budget · block-mcp-tool-budget<br/>block-gh-api-forbidden · block-single-producer<br/>block-entry-exit-reachable"]
            RULES_REACH["★ recipe/rules_reachability<br/>━━━━━━━━━━<br/>2 symbolic reachability rules<br/>capture-inversion-detection<br/>event-scope-requires-upstream-capture"]
            BLOCK_BUDGETS["★ recipe/block_budgets.yaml<br/>━━━━━━━━━━<br/>Budget config (pre_queue_gate · DEFAULT)<br/>Loaded via lru_cache"]
        end

        subgraph RecipeCore ["● Recipe Core (modified)"]
            direction TB
            RECIPE_INIT["● recipe/__init__<br/>━━━━━━━━━━<br/>Public surface re-exporter<br/>Side-effect: imports all rules_* plugins<br/>Fan-in: 3+ (server, tests, validator)"]
            REGISTRY["● recipe/registry<br/>━━━━━━━━━━<br/>RuleFinding · @semantic_rule · @block_rule<br/>Fan-in: all rule modules"]
            ANALYSIS["● recipe/_analysis<br/>━━━━━━━━━━<br/>ValidationContext · _bfs_reachable<br/>_build_step_graph · extract_blocks<br/>Fan-in: 4 (registry · rules_reach · contracts · io)"]
            SCHEMA["● recipe/schema<br/>━━━━━━━━━━<br/>Recipe · RecipeBlock · RecipeStep<br/>Leaf — imports core only"]
            CONTRACTS["● recipe/contracts<br/>━━━━━━━━━━<br/>Contract card generation + staleness<br/>Deferred import of _analysis (cycle guard)"]
            IO["● recipe/io<br/>━━━━━━━━━━<br/>load_recipe · iter_steps_with_context<br/>Deferred import of _analysis (cycle guard)"]
        end
    end

    %% ══════════════════════════════════════════════════
    %% LAYER 1 — INFRASTRUCTURE
    %% ══════════════════════════════════════════════════
    subgraph L1 ["LAYER 1 — INFRASTRUCTURE  (execution/ · pipeline/)"]
        direction LR
        EXEC_INIT["● execution/__init__<br/>━━━━━━━━━━<br/>Re-export aggregator<br/>Re-exports fetch_repo_merge_state<br/>Fan-in: server/helpers, recipe/contracts"]
        MERGE_QUEUE["● execution/merge_queue<br/>━━━━━━━━━━<br/>fetch_repo_merge_state()<br/>DefaultMergeQueueWatcher<br/>Imports: core · execution.github"]
        EXEC_GITHUB["execution/github<br/>━━━━━━━━━━<br/>github_headers<br/>parse_merge_queue_response"]
        PIPELINE["pipeline/<br/>━━━━━━━━━━<br/>ToolContext · DefaultGateState<br/>gate_error_result"]
    end

    %% ══════════════════════════════════════════════════
    %% LAYER 0 — FOUNDATION
    %% ══════════════════════════════════════════════════
    subgraph L0 ["LAYER 0 — FOUNDATION  (core/)"]
        direction LR
        CORE["core/<br/>━━━━━━━━━━<br/>Severity · load_yaml · pkg_root<br/>atomic_write · get_logger · types"]
        CORE_CONSTS["● core/_type_constants<br/>━━━━━━━━━━<br/>GATED_TOOLS += check_repo_merge_state<br/>Zero autoskillit imports — true L0 leaf"]
    end

    %% STANDALONE (no layer — stdlib-only)
    HOOKS["● hooks/pretty_output_hook<br/>━━━━━━━━━━<br/>Stdlib-only MCP output formatter<br/>No autoskillit package imports"]

    %% ══════════════════════════════════════════════════
    %% CONNECTIONS — L3 internal
    %% ══════════════════════════════════════════════════
    SRV_INIT -->|"side-effect import"| TOOLS_CI
    SRV_INIT -->|"side-effect import"| TOOLS_KITCHEN
    TOOLS_CI -->|"imports"| SRV_HELPERS
    TOOLS_KITCHEN -->|"imports"| SRV_HELPERS

    %% ══════════════════════════════════════════════════
    %% CONNECTIONS — L3 → L2
    %% ══════════════════════════════════════════════════
    SRV_HELPERS -->|"imports recipe.*"| RECIPE_INIT

    %% ══════════════════════════════════════════════════
    %% CONNECTIONS — L3 → L1
    %% ══════════════════════════════════════════════════
    SRV_HELPERS -->|"imports fetch_repo_merge_state"| EXEC_INIT
    SRV_HELPERS -->|"imports gate_error_result"| PIPELINE

    %% ══════════════════════════════════════════════════
    %% CONNECTIONS — L2 rule plugin registration
    %% ══════════════════════════════════════════════════
    RECIPE_INIT -->|"side-effect import ★"| RULES_BLOCKS
    RECIPE_INIT -->|"side-effect import ★"| RULES_REACH
    RULES_BLOCKS -->|"@block_rule decorator"| REGISTRY
    RULES_REACH -->|"@semantic_rule decorator"| REGISTRY
    RULES_BLOCKS -->|"lru_cache load"| BLOCK_BUDGETS
    RULES_REACH -->|"BFS helpers"| ANALYSIS

    %% ══════════════════════════════════════════════════
    %% CONNECTIONS — L2 recipe core internals
    %% ══════════════════════════════════════════════════
    REGISTRY -->|"imports ValidationContext"| ANALYSIS
    REGISTRY -->|"imports RecipeBlock"| SCHEMA
    ANALYSIS -->|"imports Recipe/RecipeStep"| SCHEMA
    ANALYSIS -->|"imports _CONTEXT_REF_RE etc."| CONTRACTS
    CONTRACTS -->|"imports _parse_recipe"| IO
    CONTRACTS -.->|"deferred import<br/>(cycle avoidance)"| ANALYSIS
    IO -.->|"deferred import<br/>(cycle avoidance)"| ANALYSIS

    %% ══════════════════════════════════════════════════
    %% CONNECTIONS — L2 → L1
    %% ══════════════════════════════════════════════════
    CONTRACTS -->|"deferred: DefaultSkillResolver"| EXEC_INIT

    %% ══════════════════════════════════════════════════
    %% CONNECTIONS — L1 internal
    %% ══════════════════════════════════════════════════
    EXEC_INIT -->|"re-exports"| MERGE_QUEUE
    MERGE_QUEUE -->|"imports github_headers"| EXEC_GITHUB

    %% ══════════════════════════════════════════════════
    %% CONNECTIONS — L2/L1 → L0
    %% ══════════════════════════════════════════════════
    RULES_BLOCKS -->|"Severity · load_yaml · pkg_root"| CORE
    RULES_REACH -->|"Severity"| CORE
    SCHEMA -->|"RecipeSource"| CORE
    ANALYSIS -->|"SKILL_TOOLS · get_logger"| CORE
    CONTRACTS -->|"atomic_write · get_logger"| CORE
    IO -->|"load_yaml · pkg_root"| CORE
    MERGE_QUEUE -->|"get_logger"| CORE
    PIPELINE -->|"types · logging"| CORE
    CORE_CONSTS -.->|"part of"| CORE

    %% CLASS ASSIGNMENTS %%
    class SRV_INIT,TOOLS_CI,TOOLS_KITCHEN,CLI_COOK cli;
    class SRV_HELPERS handler;
    class RECIPE_INIT,REGISTRY,ANALYSIS phase;
    class SCHEMA,IO,CORE,CORE_CONSTS stateNode;
    class CONTRACTS handler;
    class RULES_BLOCKS,RULES_REACH,BLOCK_BUDGETS newComponent;
    class EXEC_INIT,MERGE_QUEUE,EXEC_GITHUB,PIPELINE handler;
    class HOOKS integration;
```

Closes #805

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-805-20260413-075453-538827/.autoskillit/temp/rectify/rectify_pre_queue_block_immunity_2026-04-13_082500_part_a.md`

## Token Usage Summary

| investigate | 165 | 25.6k | 931.8k | 108.1k | 1 | 12m 5s |
| rectify | 13.2k | 86.9k | 3.2M | 303.3k | 2 | 31m 27s |
| review | 52 | 6.7k | 182.3k | 40.6k | 1 | 7m 45s |
| dry_walkthrough | 1.6k | 65.4k | 4.9M | 145.9k | 2 | 26m 52s |
| implement | 6.8k | 156.9k | 24.4M | 334.2k | 2 | 43m 31s |
| retry_worktree | 2.7k | 168.0k | 34.1M | 412.2k | 3 | 1h 5m |
| prepare_pr | 70 | 9.9k | 246.4k | 35.9k | 1 | 2m 36s |
| run_arch_lenses | 8.0k | 35.2k | 540.3k | 107.0k | 3 | 12m 6s |
| compose_pr | 51 | 12.3k | 190.5k | 39.6k | 1 | 2m 55s |
| **Total** | 32.7k | 566.9k | 68.6M | 1.5M | | 3h 24m |

### DIFF
--- a/src/autoskillit/execution/merge_queue.py
+++ b/src/autoskillit/execution/merge_queue.py
@@ -405,28 +405,51 @@ class DefaultMergeQueueWatcher:
 # ---------------------------------------------------------------------------
 
 
+def _text_has_push_trigger(text: str) -> bool:
+    """Return True if a workflow file text declares a push trigger.
+
+    Checks for the three YAML forms GitHub supports for a push trigger:
+    - ``on: [push, ...]``  (list form)
+    - ``on: push``         (scalar form)
+    - ``push:``            (mapping form under ``on:``)
+
+    Note: does not trigger on ``git push`` in comments or step bodies
+    because those forms do not start with ``on:`` or appear as bare YAML keys.
+    """
+    return any(pat in text for pat in ("on: [push", "on: push", "push:"))
+
+
 async def fetch_repo_merge_state(
     owner: str,
     repo: str,
     branch: str,
     token: str | None,
-) -> dict[str, bool]:
+) -> dict[str, bool | str | None]:
     """Fetch repository merge-state in a single GraphQL round-trip.
 
-    Returns a dict with three boolean keys:
-    - ``queue_available``: the branch has an active GitHub merge queue.
+    Returns a dict with four keys:
+    - ``queue_available``: the branch has an active GitHub merge queue (bool).
     - ``merge_group_trigger``: at least one CI workflow declares the
-      ``merge_group`` event trigger.
-    - ``auto_merge_available``: the repository has auto-merge enabled.
+      ``merge_group`` event trigger (bool).
+    - ``auto_merge_available``: the repository has auto-merge enabled (bool).
+    - ``ci_event``: the recommended CI event to poll — ``"push"`` when any
+      workflow declares a push trigger, ``"merge_group"`` when only merge_group
+      triggers are found, or ``None`` when no push/merge_group triggers exist
+      (ci.py scope.event=None matches any trigger).
 
     Null-handling:
     - ``mergeQueue is null`` → ``queue_available: False``  (no queue)
-    - ``object is null`` → ``merge_group_trigger: False``  (no workflows dir)
+    - ``object is null`` → ``merge_group_trigger: False``, ``ci_event: None``  (no workflows dir)
     - ``entry.object.text is null`` → skip entry (binary/large file)
     - GraphQL ``autoMergeAllowed`` field error (GHES 3.0.x) → ``auto_merge_available: False``
 
     Only transport-level failures (network timeout, non-200 HTTP status) are
     allowed to propagate; callers are expected to handle them.
+
+    Historical note: Issue #498 ("Merge queue detection should validate workflow has
+    merge_group trigger") established the merge_group_trigger field. The ci_event
+    field is a closely related extension — verify that the push-trigger scan does
+    not regress the merge_group-only detection that #498 established.
     """
     async with httpx.AsyncClient(
         headers=github_headers(token),
@@ -453,19 +476,33 @@ async def fetch_repo_merge_state(
         False if auto_merge_field_missing else bool(repo_data.get("autoMergeAllowed", False))
     )
 
-    # Detect merge_group trigger: check each workflow file's text content.
+    # Scan workflow files for push and merge_group trigger declarations.
+    # Both flags are derived from the same Blob.text scan — no extra round-trips.
     merge_group_trigger = False
+    has_push_trigger = False
     workflows_tree = repo_data.get("object")
     if workflows_tree is not None:
         for entry in workflows_tree.get("entries", []):
             blob = entry.get("object") or {}
             text = blob.get("text")
-            if text is not None and "merge_group" in text:
+            if text is None:
+                continue  # binary or oversized blob — skip
+            if "merge_group" in text:
                 merge_group_trigger = True
-                break
+            if _text_has_push_trigger(text):
+                has_push_trigger = True
+
+    # Derive ci_event: prefer push for historical compatibility when both are present.
+    if has_push_trigger:
+        ci_event: str | None = "push"
+    elif merge_group_trigger:
+        ci_event = "merge_group"
+    else:
+        ci_event = None  # schedule-only or workflow_dispatch-only: match any
 
     return {
         "queue_available": queue_available,
         "merge_group_trigger": merge_group_trigger,
         "auto_merge_available": auto_merge_available,
+        "ci_event": ci_event,
     }
--- a/src/autoskillit/execution/merge_queue.py
+++ b/src/autoskillit/execution/merge_queue.py
@@ -76,6 +76,33 @@ mutation EnableAutoMerge($prId: ID!, $mergeMethod: PullRequestMergeMethod!) {
 }
 """
 
+# Repo-level state query: consolidates three former run_cmd steps into one HTTP round-trip.
+# Distinct from _QUERY (PR-level + queue-entries); do not merge these constants.
+# Returns: mergeQueue presence, autoMergeAllowed flag, and workflow file texts for
+# merge_group trigger detection — all in a single GraphQL call.
+_REPO_STATE_QUERY = """
+query GetRepoMergeState($owner: String!, $repo: String!, $branch: String!) {
+  repository(owner: $owner, name: $repo) {
+    mergeQueue(branch: $branch) {
+      id
+    }
+    autoMergeAllowed
+    object(expression: "HEAD:.github/workflows") {
+      ... on Tree {
+        entries {
+          name
+          object {
+            ... on Blob {
+              text
+            }
+          }
+        }
+      }
+    }
+  }
+}
+"""
+
 
 class DefaultMergeQueueWatcher:
     """Polls GitHub merge queue state until merged, ejected, stalled, or timed out.
@@ -371,3 +398,74 @@ class DefaultMergeQueueWatcher:
                 raise RuntimeError(f"GraphQL mutation error: {body['errors']}")
             if i < len(mutations) - 1:
                 await asyncio.sleep(2)
+
+
+# ---------------------------------------------------------------------------
+# Repo-level state helper (used by check_repo_merge_state MCP tool)
+# ---------------------------------------------------------------------------
+
+
+async def fetch_repo_merge_state(
+    owner: str,
+    repo: str,
+    branch: str,
+    token: str | None,
+) -> dict[str, bool]:
+    """Fetch repository merge-state in a single GraphQL round-trip.
+
+    Returns a dict with three boolean keys:
+    - ``queue_available``: the branch has an active GitHub merge queue.
+    - ``merge_group_trigger``: at least one CI workflow declares the
+      ``merge_group`` event trigger.
+    - ``auto_merge_available``: the repository has auto-merge enabled.
+
+    Null-handling:
+    - ``mergeQueue is null`` → ``queue_available: False``  (no queue)
+    - ``object is null`` → ``merge_group_trigger: False``  (no workflows dir)
+    - ``entry.object.text is null`` → skip entry (binary/large file)
+    - GraphQL ``autoMergeAllowed`` field error (GHES 3.0.x) → ``auto_merge_available: False``
+
+    Only transport-level failures (network timeout, non-200 HTTP status) are
+    allowed to propagate; callers are expected to handle them.
+    """
+    async with httpx.AsyncClient(
+        headers=github_headers(token),
+        timeout=30.0,
+    ) as client:
+        resp = await client.post(
+            _GRAPHQL_ENDPOINT,
+            json={
+                "query": _REPO_STATE_QUERY,
+                "variables": {"owner": owner, "repo": repo, "branch": branch},
+            },
+        )
+        resp.raise_for_status()
+        body = resp.json()
+
+    # Gracefully handle GHES 3.0.x where autoMergeAllowed doesn't exist.
+    auto_merge_field_missing = any(
+        "autoMergeAllowed" in str(e.get("message", "")) for e in body.get("errors", [])
+    )
+
+    repo_data: dict[str, Any] = (body.get("data") or {}).get("repository") or {}
+    queue_available = repo_data.get("mergeQueue") is not None
+    auto_merge_available = (
+        False if auto_merge_field_missing else bool(repo_data.get("autoMergeAllowed", False))
+    )
+
+    # Detect merge_group trigger: check each workflow file's text content.
+    merge_group_trigger = False
+    workflows_tree = repo_data.get("object")
+    if workflows_tree is not None:
+        for entry in workflows_tree.get("entries", []):
+            blob = entry.get("object") or {}
+            text = blob.get("text")
+            if text is not None and "merge_group" in text:
+                merge_group_trigger = True
+                break
+
+    return {
+        "queue_available": queue_available,
+        "merge_group_trigger": merge_group_trigger,
+        "auto_merge_available": auto_merge_available,
+    }
--- a/src/autoskillit/execution/merge_queue.py
+++ b/src/autoskillit/execution/merge_queue.py
@@ -465,6 +465,10 @@ async def fetch_repo_merge_state(
         resp.raise_for_status()
         body = resp.json()
 
+    # GitHub GraphQL always returns a JSON object; guard against unexpected shapes.
+    if not isinstance(body, dict):
+        body = {}
+
     # Gracefully handle GHES 3.0.x where autoMergeAllowed doesn't exist.
     auto_merge_field_missing = any(
         "autoMergeAllowed" in str(e.get("message", "")) for e in body.get("errors", [])
--- a/src/autoskillit/recipe/contracts.py
+++ b/src/autoskillit/recipe/contracts.py
@@ -330,7 +330,8 @@ def generate_recipe_card(
     if isinstance(pipeline_path, Recipe):
         return _generate_recipe_card_for_recipe(pipeline_path)
 
-    assert recipes_dir is not None, "recipes_dir required when pipeline_path is a file path"
+    if recipes_dir is None:
+        raise ValueError("recipes_dir required when pipeline_path is a file path")
     pipeline_path = Path(pipeline_path)
     recipes_dir = Path(recipes_dir)
 
@@ -499,6 +500,10 @@ def check_contract_staleness(
 
     When ``skills_dir`` is None, the bundled skills directory is used for hash
     comparison.
+
+    When ``contract`` is a ``Recipe`` but ``stored_card`` is ``None``, no
+    comparison baseline is available and [] is returned immediately.  This is
+    expected during initial card generation before a stored card exists.
 
     Returns a list of StaleItem entries indicating what changed.
     """
--- a/src/autoskillit/recipe/contracts.py
+++ b/src/autoskillit/recipe/contracts.py
@@ -23,6 +23,7 @@ from autoskillit.core import (
     pkg_root,
 )
 from autoskillit.recipe.io import _parse_recipe
+from autoskillit.recipe.schema import Recipe, RecipeBlock
 from autoskillit.recipe.staleness_cache import (
     StalenessEntry,
     compute_recipe_hash,
@@ -78,6 +79,24 @@ class DataFlowEntry:
     produced: list[str]
 
 
+@dataclasses.dataclass(frozen=True)
+class BlockFingerprint:
+    """Structural fingerprint for a named recipe block.
+
+    Used by ``check_contract_staleness`` to detect silent composition drift:
+    any change to a block's member count, tool usage, gh api call count, or
+    capture names produces a fingerprint mismatch (reason='block_composition_drift').
+    """
+
+    name: str
+    member_count: int
+    tool_counts_sorted: tuple[tuple[str, int], ...]  # sorted by tool name for stable comparison
+    gh_api_occurrences: int
+    capture_names_hash: str  # sha256hex of sorted capture key names across all members
+    entry_step: str
+    exit_step: str
+
+
 @dataclasses.dataclass
 class RecipeCard:
     generated_at: str
@@ -85,6 +104,9 @@ class RecipeCard:
     skill_hashes: dict[str, str]
     skills: dict[str, SkillContract]
     dataflow: list[DataFlowEntry]
+    block_fingerprints: tuple[BlockFingerprint, ...] = dataclasses.field(
+        default_factory=tuple  # type: ignore[arg-type]
+    )
 
 
 # ---------------------------------------------------------------------------
@@ -222,16 +244,76 @@ def count_positional_args(skill_command: str) -> int:
 
 
 # ---------------------------------------------------------------------------
+# Block fingerprint helpers
+# ---------------------------------------------------------------------------
+
+
+def _compute_block_fingerprint(block: RecipeBlock) -> BlockFingerprint:
+    """Compute a structural fingerprint for a RecipeBlock.
+
+    The fingerprint captures:
+    - ``member_count``: total number of member steps
+    - ``tool_counts_sorted``: per-tool counts sorted by tool name (stable comparison)
+    - ``gh_api_occurrences``: total 'gh api' shell substring occurrences
+    - ``capture_names_hash``: sha256 of the sorted set of capture key names across members
+    - ``entry_step`` / ``exit_step``: the block's entry and exit step names
+    """
+    all_capture_names: list[str] = []
+    for step in block.members:
+        all_capture_names.extend(sorted((step.capture or {}).keys()))
+    sorted_capture_names = sorted(all_capture_names)
+    capture_names_hash = (
+        f"sha256:{hashlib.sha256(' '.join(sorted_capture_names).encode()).hexdigest()}"
+    )
+    tool_counts_sorted = tuple(sorted(block.tool_counts.items()))
+    return BlockFingerprint(
+        name=block.name,
+        member_count=len(block.members),
+        tool_counts_sorted=tool_counts_sorted,
+        gh_api_occurrences=block.gh_api_occurrences,
+        capture_names_hash=capture_names_hash,
+        entry_step=block.entry,
+        exit_step=block.exit,
+    )
+
+
+def _generate_recipe_card_for_recipe(recipe: Recipe) -> RecipeCard:
+    """Generate a RecipeCard from a Recipe object (no disk write).
+
+    Used by the block fingerprint drift detection path in ``check_contract_staleness``
+    and by tests that want a ``RecipeCard`` with populated ``block_fingerprints``.
+    Uses a deferred import of ``_build_step_graph`` and ``extract_blocks`` to avoid
+    a circular import (``_analysis.py`` imports from ``contracts.py``).
+    """
+    # Deferred import to avoid circular dependency:
+    # contracts.py → _analysis.py (already safe via _analysis.py → contracts.py)
+    from autoskillit.recipe._analysis import _build_step_graph, extract_blocks  # noqa: PLC0415
+
+    step_graph = _build_step_graph(recipe)
+    blocks = extract_blocks(recipe, step_graph)
+    fingerprints = tuple(_compute_block_fingerprint(b) for b in blocks)
+    manifest = load_bundled_manifest()
+    return RecipeCard(
+        generated_at=datetime.now(UTC).isoformat(),
+        bundled_manifest_version=manifest.get("version", ""),
+        skill_hashes={},
+        skills={},
+        dataflow=[],
+        block_fingerprints=fingerprints,
+    )
+
+
+# ---------------------------------------------------------------------------
 # Pipeline contract generation, loading, and validation
 # ---------------------------------------------------------------------------
 
 
 def generate_recipe_card(
-    pipeline_path: Path | str,
-    recipes_dir: Path | str,
+    pipeline_path: Path | str | Recipe,
+    recipes_dir: Path | str | None = None,
     *,
     skills_dir: Path | None = None,
-) -> dict:
+) -> dict | RecipeCard:
     """Generate a recipe card file for a recipe.
 
     Walks each step, resolves skill names, looks up contracts in the manifest,
@@ -241,8 +323,14 @@ def generate_recipe_card(
     When ``skills_dir`` is None, skill hashes are not computed and ``skill_hashes``
     in the generated card will be empty.
 
-    Returns the contract data dict directly (no disk re-read required by callers).
+    When ``pipeline_path`` is a ``Recipe`` object, returns a ``RecipeCard`` with
+    populated ``block_fingerprints`` (no disk write).  The path-based form returns
+    the contract data dict directly (no disk re-read required by callers).
     """
+    if isinstance(pipeline_path, Recipe):
+        return _generate_recipe_card_for_recipe(pipeline_path)
+
+    assert recipes_dir is not None, "recipes_dir required when pipeline_path is a file path"
     pipeline_path = Path(pipeline_path)
     recipes_dir = Path(recipes_dir)
 
@@ -387,14 +475,21 @@ def validate_recipe_cards(recipe: Any, contract: dict[str, Any]) -> list[dict[st
 
 
 def check_contract_staleness(
-    contract: dict[str, Any],
+    contract: dict[str, Any] | Recipe,
     *,
     recipe_path: Path | None = None,
     cache_path: Path | None = None,
     skills_dir: Path | None = None,
     resolver: SkillResolver | None = None,
+    stored_card: RecipeCard | None = None,
 ) -> list[StaleItem]:
     """Check a pipeline contract for staleness against the current manifest.
+
+    When ``stored_card`` is provided and ``contract`` is a ``Recipe``, compares
+    block fingerprints from the stored card against the current recipe's blocks.
+    Returns ``StaleItem`` entries with ``reason='block_composition_drift'`` for
+    any block whose fingerprint has changed.  This path does not perform manifest
+    or skill-hash checks — it is a pure structural comparison.
 
     When ``recipe_path`` and ``cache_path`` are both provided, a disk-backed
     cache keyed by recipe content hash + manifest version is consulted first.
@@ -407,6 +502,37 @@ def check_contract_staleness(
 
     Returns a list of StaleItem entries indicating what changed.
     """
+    if stored_card is not None:
+        recipe_obj = contract if isinstance(contract, Recipe) else None
+        if recipe_obj is not None:
+            current_card = _generate_recipe_card_for_recipe(recipe_obj)
+            current_fps = {fp.name: fp for fp in current_card.block_fingerprints}
+            stale_items: list[StaleItem] = []
+            for stored_fp in stored_card.block_fingerprints:
+                current_fp = current_fps.get(stored_fp.name)
+                if current_fp is None:
+                    stale_items.append(
+                        StaleItem(
+                            skill=stored_fp.name,
+                            reason="block_composition_drift",
+                            stored_value=repr(stored_fp),
+                            current_value="(block removed)",
+                        )
+                    )
+                elif current_fp != stored_fp:
+                    stale_items.append(
+                        StaleItem(
+                            skill=stored_fp.name,
+                            reason="block_composition_drift",
+                            stored_value=repr(stored_fp),
+                            current_value=repr(current_fp),
+                        )
+                    )
+            return stale_items
+
+    if isinstance(contract, Recipe):
+        return []
+
     manifest = load_bundled_manifest()
     current_version = manifest["version"]
     cached: StalenessEntry | None = None
--- a/src/autoskillit/recipe/rules_blocks.py
+++ b/src/autoskillit/recipe/rules_blocks.py
@@ -22,7 +22,10 @@ from autoskillit.recipe.registry import BlockContext, RuleFinding, block_rule
 def _block_budgets() -> Mapping[str, Mapping[str, Any]]:
     """Load block_budgets.yaml, cached for the lifetime of the process."""
     path = pkg_root() / "recipe" / "block_budgets.yaml"
-    data = load_yaml(path)
+    try:
+        data = load_yaml(path)
+    except FileNotFoundError:
+        return {}
     if not isinstance(data, dict):
         return {}
     return data  # type: ignore[return-value]
@@ -40,7 +43,10 @@ def _budget_for(block_name: str) -> dict[str, Any]:
     severity=Severity.ERROR,
 )
 def _check_block_run_cmd_budget(bctx: BlockContext) -> list[RuleFinding]:
-    budget = _budget_for(bctx.block.name).get("run_cmd", 1)
+    budget_entry = _budget_for(bctx.block.name)
+    if "run_cmd" not in budget_entry:
+        return []  # No run_cmd budget declared for this block — skip check
+    budget = int(budget_entry["run_cmd"])
     actual = bctx.block.tool_counts.get("run_cmd", 0)
     if actual <= budget:
         return []
@@ -154,11 +160,8 @@ def _check_block_single_producer(bctx: BlockContext) -> list[RuleFinding]:
 )
 def _check_block_entry_exit_reachable(bctx: BlockContext) -> list[RuleFinding]:
     step_graph = bctx.parent.step_graph
-    # Build predecessor map over the block's step_graph slice
-    predecessors: dict[str, set[str]] = {}
-    for src, successors in step_graph.items():
-        for dst in successors:
-            predecessors.setdefault(dst, set()).add(src)
+    # Use the pre-computed predecessor map from ValidationContext (built once per pass).
+    predecessors = bctx.parent.predecessors
 
     member_names = {s.name for s in bctx.block.members}
     entry_candidates = [
--- a/src/autoskillit/recipe/rules_blocks.py
+++ b/src/autoskillit/recipe/rules_blocks.py
@@ -14,9 +14,7 @@ from collections.abc import Mapping
 from functools import lru_cache
 from typing import Any
 
-from autoskillit.core import Severity
-from autoskillit.core.io import load_yaml
-from autoskillit.core.paths import pkg_root
+from autoskillit.core import Severity, load_yaml, pkg_root
 from autoskillit.recipe.registry import BlockContext, RuleFinding, block_rule
 
 
--- a/src/autoskillit/recipe/rules_blocks.py
+++ b/src/autoskillit/recipe/rules_blocks.py
@@ -10,8 +10,9 @@ Budget values are loaded from block_budgets.yaml at import time (lru_cache).
 
 from __future__ import annotations
 
+from collections.abc import Mapping
 from functools import lru_cache
-from typing import Any, Mapping
+from typing import Any
 
 from autoskillit.core import Severity
 from autoskillit.core.io import load_yaml
@@ -70,9 +71,7 @@ def _check_block_mcp_tool_budget(bctx: BlockContext) -> list[RuleFinding]:
         return []  # No mcp_tools budget declared for this block — skip check
     budget = int(budget_entry["mcp_tools"])
     # MCP tool calls: any tool step that is not run_cmd
-    actual = sum(
-        count for tool, count in bctx.block.tool_counts.items() if tool != "run_cmd"
-    )
+    actual = sum(count for tool, count in bctx.block.tool_counts.items() if tool != "run_cmd")
     if actual <= budget:
         return []
     return [
@@ -127,7 +126,7 @@ def _check_block_single_producer(bctx: BlockContext) -> list[RuleFinding]:
     # Build a map: capture_key → list of step names that produce it
     producers: dict[str, list[str]] = {}
     for step in bctx.block.members:
-        for cap_key in (step.capture or {}):
+        for cap_key in step.capture or {}:
             producers.setdefault(cap_key, []).append(step.name)
     findings = []
     for cap_key, producer_names in producers.items():
@@ -165,12 +164,10 @@ def _check_block_entry_exit_reachable(bctx: BlockContext) -> list[RuleFinding]:
 
     member_names = {s.name for s in bctx.block.members}
     entry_candidates = [
-        s for s in bctx.block.members
-        if not (predecessors.get(s.name, set()) & member_names)
+        s for s in bctx.block.members if not (predecessors.get(s.name, set()) & member_names)
     ]
     exit_candidates = [
-        s for s in bctx.block.members
-        if not (step_graph.get(s.name, set()) & member_names)
+        s for s in bctx.block.members if not (step_graph.get(s.name, set()) & member_names)
     ]
     findings = []
     if len(entry_candidates) != 1:
--- a/src/autoskillit/recipe/rules_blocks.py
+++ b/src/autoskillit/recipe/rules_blocks.py
@@ -1,0 +1,204 @@
+"""Block-level semantic validation rules for recipe pipelines.
+
+Each rule is registered via @block_rule and dispatched by run_semantic_rules
+for every RecipeBlock in a recipe.  Block rules operate on one block at a time
+via BlockContext — they must not access ctx.recipe.steps directly to express
+cross-block constraints; use bctx.parent for read-only access when needed.
+
+Budget values are loaded from block_budgets.yaml at import time (lru_cache).
+"""
+
+from __future__ import annotations
+
+from functools import lru_cache
+from typing import Any, Mapping
+
+from autoskillit.core import Severity
+from autoskillit.core.io import load_yaml
+from autoskillit.core.paths import pkg_root
+from autoskillit.recipe.registry import BlockContext, RuleFinding, block_rule
+
+
+@lru_cache(maxsize=1)
+def _block_budgets() -> Mapping[str, Mapping[str, Any]]:
+    """Load block_budgets.yaml, cached for the lifetime of the process."""
+    path = pkg_root() / "recipe" / "block_budgets.yaml"
+    data = load_yaml(path)
+    if not isinstance(data, dict):
+        return {}
+    return data  # type: ignore[return-value]
+
+
+def _budget_for(block_name: str) -> dict[str, Any]:
+    """Return the budget dict for a named block, falling back to DEFAULT."""
+    budgets = _block_budgets()
+    return dict(budgets.get(block_name, budgets.get("DEFAULT", {})))
+
+
+@block_rule(
+    name="block-run-cmd-budget",
+    description="Block members contain at most the declared number of run_cmd steps.",
+    severity=Severity.ERROR,
+)
+def _check_block_run_cmd_budget(bctx: BlockContext) -> list[RuleFinding]:
+    budget = _budget_for(bctx.block.name).get("run_cmd", 1)
+    actual = bctx.block.tool_counts.get("run_cmd", 0)
+    if actual <= budget:
+        return []
+    return [
+        RuleFinding(
+            rule="block-run-cmd-budget",
+            severity=Severity.ERROR,
+            step_name=bctx.block.entry,
+            message=(
+                f"Block {bctx.block.name!r} contains {actual} run_cmd step"
+                f"{'s' if actual != 1 else ''} (budget: {budget}). "
+                f"Consolidate into an MCP tool."
+            ),
+        )
+    ]
+
+
+@block_rule(
+    name="block-mcp-tool-budget",
+    description="Block members contain at most the declared number of non-run_cmd MCP tool calls.",
+    severity=Severity.ERROR,
+)
+def _check_block_mcp_tool_budget(bctx: BlockContext) -> list[RuleFinding]:
+    budget_entry = _budget_for(bctx.block.name)
+    if "mcp_tools" not in budget_entry:
+        return []  # No mcp_tools budget declared for this block — skip check
+    budget = int(budget_entry["mcp_tools"])
+    # MCP tool calls: any tool step that is not run_cmd
+    actual = sum(
+        count for tool, count in bctx.block.tool_counts.items() if tool != "run_cmd"
+    )
+    if actual <= budget:
+        return []
+    return [
+        RuleFinding(
+            rule="block-mcp-tool-budget",
+            severity=Severity.ERROR,
+            step_name=bctx.block.entry,
+            message=(
+                f"Block {bctx.block.name!r} contains {actual} MCP tool call"
+                f"{'s' if actual != 1 else ''} (budget: {budget}). "
+                f"Reduce to a single consolidated tool call."
+            ),
+        )
+    ]
+
+
+@block_rule(
+    name="block-gh-api-forbidden",
+    description=(
+        "Blocks with forbid_gh_api: true must not contain 'gh api' shell patterns "
+        "in any run_cmd step."
+    ),
+    severity=Severity.ERROR,
+)
+def _check_block_gh_api_forbidden(bctx: BlockContext) -> list[RuleFinding]:
+    if not _budget_for(bctx.block.name).get("forbid_gh_api", False):
+        return []  # This block does not prohibit gh api shell patterns
+    if bctx.block.gh_api_occurrences == 0:
+        return []
+    return [
+        RuleFinding(
+            rule="block-gh-api-forbidden",
+            severity=Severity.ERROR,
+            step_name=bctx.block.entry,
+            message=(
+                f"Block {bctx.block.name!r} has {bctx.block.gh_api_occurrences} "
+                f"'gh api' shell invocation"
+                f"{'s' if bctx.block.gh_api_occurrences != 1 else ''}. "
+                f"MCP tools must own all GitHub API calls in declared blocks; "
+                f"consolidate into a single MCP tool call."
+            ),
+        )
+    ]
+
+
+@block_rule(
+    name="block-single-producer",
+    description="Each capture key produced within a block must originate from exactly one step.",
+    severity=Severity.WARNING,
+)
+def _check_block_single_producer(bctx: BlockContext) -> list[RuleFinding]:
+    # Build a map: capture_key → list of step names that produce it
+    producers: dict[str, list[str]] = {}
+    for step in bctx.block.members:
+        for cap_key in (step.capture or {}):
+            producers.setdefault(cap_key, []).append(step.name)
+    findings = []
+    for cap_key, producer_names in producers.items():
+        if len(producer_names) > 1:
+            findings.append(
+                RuleFinding(
+                    rule="block-single-producer",
+                    severity=Severity.WARNING,
+                    step_name=bctx.block.entry,
+                    message=(
+                        f"Block {bctx.block.name!r}: capture key {cap_key!r} is produced by "
+                        f"{len(producer_names)} steps ({', '.join(producer_names)}); "
+                        f"expected exactly one producer."
+                    ),
+                )
+            )
+    return findings
+
+
+@block_rule(
+    name="block-entry-exit-reachable",
+    description=(
+        "Block members must form a single connected region with exactly one "
+        "entry step and one exit step."
+    ),
+    severity=Severity.WARNING,
+)
+def _check_block_entry_exit_reachable(bctx: BlockContext) -> list[RuleFinding]:
+    step_graph = bctx.parent.step_graph
+    # Build predecessor map over the block's step_graph slice
+    predecessors: dict[str, set[str]] = {}
+    for src, successors in step_graph.items():
+        for dst in successors:
+            predecessors.setdefault(dst, set()).add(src)
+
+    member_names = {s.name for s in bctx.block.members}
+    entry_candidates = [
+        s for s in bctx.block.members
+        if not (predecessors.get(s.name, set()) & member_names)
+    ]
+    exit_candidates = [
+        s for s in bctx.block.members
+        if not (step_graph.get(s.name, set()) & member_names)
+    ]
+    findings = []
+    if len(entry_candidates) != 1:
+        findings.append(
+            RuleFinding(
+                rule="block-entry-exit-reachable",
+                severity=Severity.WARNING,
+                step_name=bctx.block.entry,
+                message=(
+                    f"Block {bctx.block.name!r} has {len(entry_candidates)} entry candidates "
+                    f"(expected 1): "
+                    f"{[s.name for s in entry_candidates]}. "
+                    f"Block members must form a single connected region."
+                ),
+            )
+        )
+    if len(exit_candidates) != 1:
+        findings.append(
+            RuleFinding(
+                rule="block-entry-exit-reachable",
+                severity=Severity.WARNING,
+                step_name=bctx.block.exit,
+                message=(
+                    f"Block {bctx.block.name!r} has {len(exit_candidates)} exit candidates "
+                    f"(expected 1): "
+                    f"{[s.name for s in exit_candidates]}. "
+                    f"Block members must form a single connected region."
+                ),
+            )
+        )
+    return findings
--- a/src/autoskillit/recipe/rules_reachability.py
+++ b/src/autoskillit/recipe/rules_reachability.py
@@ -26,7 +26,6 @@ from autoskillit.recipe._analysis import (
     ValidationContext,
     _bfs_reachable,
     _bfs_with_facts,
-    _build_step_graph,
 )
 from autoskillit.recipe.registry import RuleFinding, semantic_rule
 
@@ -44,7 +43,9 @@ def _find_context_refs_in_step(step_name: str, ctx: ValidationContext) -> list[s
     if step is None:
         return []
     refs: list[str] = []
-    for val in (step.with_args or {}).values():
+    for key, val in (step.with_args or {}).items():
+        if key == "skill_command":
+            continue  # handled below to avoid double-scan via values()
         if isinstance(val, str):
             refs.extend(_CTX_REF_RE.findall(val))
     skill_cmd = (step.with_args or {}).get("skill_command", "")
@@ -103,6 +104,13 @@ def _check_capture_inversion(ctx: ValidationContext) -> list[RuleFinding]:
 
     facts = _bfs_with_facts(ctx.step_graph, ctx.recipe, start=entry)
 
+    # Pre-compute the full conditional fact domain once — not inside the inner loop.
+    all_facts_in_recipe: set[tuple[str, str]] = set()
+    for node_facts in facts.values():
+        for fs in node_facts:
+            all_facts_in_recipe.update(fs)
+    recipe_fact_vars = {v for v, _ in all_facts_in_recipe}
+
     findings: list[RuleFinding] = []
     for step_name, step in ctx.recipe.steps.items():
         context_refs = _find_context_refs_in_step(step_name, ctx)
@@ -117,11 +125,6 @@ def _check_capture_inversion(ctx: ValidationContext) -> list[RuleFinding]:
             # Only flag vars that appear in the conditional fact domain of the recipe.
             # If no step establishes (var, value) via a conditional edge, the var is
             # captured unconditionally — not an inversion.
-            all_facts_in_recipe: set[tuple[str, str]] = set()
-            for node_facts in facts.values():
-                for fs in node_facts:
-                    all_facts_in_recipe.update(fs)
-            recipe_fact_vars = {v for v, _ in all_facts_in_recipe}
 
             if var not in recipe_fact_vars:
                 continue  # var is not conditionally established anywhere — skip
@@ -189,7 +192,9 @@ def _check_event_scope_requires_upstream_capture(ctx: ValidationContext) -> list
             continue  # producer is upstream — context is known
 
         producer_desc = (
-            f"producer step {mg_producer!r}" if mg_producer else "no producer for merge_group_trigger"
+            f"producer step {mg_producer!r}"
+            if mg_producer
+            else "no producer for merge_group_trigger"
         )
         findings.append(
             RuleFinding(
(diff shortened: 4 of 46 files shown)


src/autoskillit/recipe/rules_reachability.py
```diff
@@ -0,0 +1,209 @@
...
+
+Both rules fire at severity ERROR and are registered via ``@semantic_rule``.
+"""
+
+from __future__ import annotations
+
+import re
+
+from autoskillit.core import Severity
+from autoskillit.recipe._analysis import (
```
Trecek, line 25: [warning] arch: rules_reachability.py imports three underscore-prefixed (private) symbols from _analysis.py: _bfs_reachable, _bfs_with_facts, _build_step_graph. Rule modules importing private internals of a sibling module creates intra-L2 coupling to implementation details. These helpers should either be promoted to public API (no underscore prefix) in _analysis.py or re-exported from recipe/__init__.py, so the coupling is explicit and the symbols are part of the module's contract.

tests/recipe/test_rules_reachability.py
```diff
@@ -0,0 +1,185 @@
...
+_QUEUE_CAPABLE = ("implementation.yaml", "remediation.yaml", "implementation-groups.yaml")
+
+
+# ---------------------------------------------------------------------------
+# Bundled recipe guards (post-Part-B: must be clean)
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("recipe_name", _QUEUE_CAPABLE)
+def test_wait_for_ci_event_push_without_upstream_capture_is_flagged(recipe_name):
```
Trecek: [warning] tests: Test name 'test_wait_for_ci_event_push_without_upstream_capture_is_flagged' contradicts its behavior: the test asserts zero findings (post-remediation clean state), not that a finding is produced. A passing test with a misleading 'is_flagged' name will confuse future readers into thinking the rule fires when it should not.

tests/execution/test_check_repo_merge_state.py
```diff
@@ -0,0 +1,159 @@
...
+    assert result["queue_available"] is False
+    assert result["merge_group_trigger"] is True
+    assert result["auto_merge_available"] is True
+    # ci_event: both push and merge_group present — prefer push for historical compat
+    assert result["ci_event"] == "push"
+    assert len(httpx_mock.get_requests()) == 1  # NOT 3, NOT N+2
+
+
+@pytest.mark.anyio
+async def test_check_repo_merge_state_handles_null_blob_text(httpx_mock):
```
Trecek, line 53: [warning] tests: test_check_repo_merge_state_handles_null_blob_text does not assert ci_event. The function returns four fields; only two are checked (merge_group_trigger, queue_available). ci_event should be None (no parseable trigger) in this case — the missing assertion leaves a regression gap for the new ci_event field.

tests/execution/test_check_repo_merge_state.py
```diff
@@ -0,0 +1,159 @@
...
+            }
+        },
+    )
+    result = await fetch_repo_merge_state(owner="o", repo="r", branch="main", token=None)
+    assert result["merge_group_trigger"] is False
+    assert result["queue_available"] is True
+
+
+@pytest.mark.anyio
+async def test_check_repo_merge_state_handles_missing_workflows_dir(httpx_mock):
```
Trecek, line 78: [warning] tests: test_check_repo_merge_state_handles_missing_workflows_dir does not assert ci_event. When object is null, ci_event should be None per the docstring contract. Missing assertion leaves the ci_event=None path unverified for the null-workflows-dir case.

tests/recipe/test_rules_blocks.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def _make_synthetic_recipe_with_two_gh_api_run_cmds_in_block(block_name: str):
+    """Build a minimal Recipe with two run_cmd gh api steps in a declared block.
+
+    Used by the synthetic block budget test below.
+    """
+    from autoskillit.recipe.schema import Recipe, RecipeStep
+
+    step_a = RecipeStep(
```
Trecek, line 64: [critical] tests: RecipeStep instances in _make_synthetic_recipe_with_two_gh_api_run_cmds_in_block are created without setting the 'name' field, so step.name defaults to ''. Steps are added to the recipe dict under keys 'step_a'/'step_b', but extract_blocks iterates recipe.steps.values() and uses s.name for block membership checks. With name='' for both steps, block extraction finds two steps with the same empty name, breaking entry/exit detection and making test_synthetic_block_with_two_run_cmds_emits_block_run_cmd_budget_finding unreliable.

Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (24 blocking findings)

tests/recipe/test_bundled_recipes_no_inversions.py
```diff
@@ -0,0 +1,49 @@
...
+
+@pytest.mark.parametrize("recipe_path", _BUNDLED_RECIPE_PATHS, ids=lambda p: p.stem)
+def test_diagnose_ci_skill_command_uses_captured_ci_event(recipe_path):
+    """No skill_command step may pass a hardcoded trigger event as a positional.
+    All /autoskillit:diagnose-ci invocations must thread ${{ context.ci_event }}."""
+    recipe = yaml.safe_load(recipe_path.read_text())
+    for step_name, step in recipe.get("steps", {}).items():
+        cmd = step.get("skill_command") or step.get("with", {}).get("skill_command", "")
+        if "diagnose-ci" in cmd:
+            assert "${{ context.ci_event }}" in cmd or "push" not in cmd.split(), (
```
Trecek: [warning] tests: The assertion '${{ context.ci_event }}' in cmd or 'push' not in cmd.split() is logically weak: a diagnose-ci invocation that hardcodes 'merge_group' (not 'push') would pass the assertion despite being a literal event. The check should also exclude any hardcoded event literal, not just 'push'.

tests/recipe/test_contracts_block_fingerprint.py
```diff
@@ -0,0 +1,119 @@
...
+    new_step_name = f"_synthetic_run_cmd_{block_name}"
+    new_step = RecipeStep(
+        name=new_step_name,
+        tool="run_cmd",
+        with_args={"cmd": "echo synthetic", "step_name": new_step_name},
+        block=block_name,
+    )
+    new_steps = dict(recipe.steps)
+    new_steps[new_step_name] = new_step
+    return Recipe(
```
Trecek, line 66: [warning] tests: _add_synthetic_run_cmd_to_block constructs a new Recipe without passing the blocks field, so recipe.blocks is empty tuple. check_contract_staleness then calls _generate_recipe_card_for_recipe which re-runs extract_blocks on the mutated steps dict. The helper does not verify that extract_blocks correctly re-derives blocks from the mutated steps, which could mask the test intent.

src/autoskillit/execution/merge_queue.py
```diff
@@ -371,3 +398,111 @@ async def _toggle_auto_merge(self, pr_node_id: str) -> None:
...
+    ) as client:
+        resp = await client.post(
+            _GRAPHQL_ENDPOINT,
+            json={
+                "query": _REPO_STATE_QUERY,
+                "variables": {"owner": owner, "repo": repo, "branch": branch},
+            },
+        )
+        resp.raise_for_status()
+        body = resp.json()
```
Trecek, line 466: [warning] defense: fetch_repo_merge_state calls resp.raise_for_status() and resp.json() after the 'async with' block has exited. While the response body is typically buffered, body.get('data') is accessed without checking whether body is a dict — if the GitHub API returns a non-JSON error body that passes raise_for_status (e.g. 200 with HTML), resp.json() will raise JSONDecodeError which is not caught here, leaking the exception contrary to the docstring's claim.

src/autoskillit/execution/merge_queue.py
```diff
@@ -371,3 +398,111 @@ async def _toggle_auto_merge(self, pr_node_id: str) -> None:
...
+
+    # Gracefully handle GHES 3.0.x where autoMergeAllowed doesn't exist.
+    auto_merge_field_missing = any(
+        "autoMergeAllowed" in str(e.get("message", "")) for e in body.get("errors", [])
+    )
+
+    repo_data: dict[str, Any] = (body.get("data") or {}).get("repository") or {}
```
Trecek, line 477: [warning] defense: repo_data is assigned the result of chained .get() calls with an or {} fallback, but if body itself is not a dict (e.g. json() returns a list), body.get('data') will raise AttributeError. There is no isinstance check on the body before accessing it as a dict.

src/autoskillit/recipe/contracts.py
```diff
@@ -221,17 +243,77 @@ def count_positional_args(skill_command: str) -> int:
...
 
 def generate_recipe_card(
-    pipeline_path: Path | str,
-    recipes_dir: Path | str,
+    pipeline_path: Path | str | Recipe,
+    recipes_dir: Path | str | None = None,
     *,
     skills_dir: Path | None = None,
-) -> dict:
+) -> dict | RecipeCard:
```
Trecek, line 316: [warning] defense: generate_recipe_card uses an assert statement to validate a required parameter (assert recipes_dir is not None). assert is disabled under Python's -O optimized mode. This should be a proper ValueError or TypeError raise to guarantee validation at construction regardless of optimization flags.

src/autoskillit/recipe/rules_blocks.py
```diff
@@ -0,0 +1,199 @@
...
+
+from autoskillit.core import Severity, load_yaml, pkg_root
+from autoskillit.recipe.registry import BlockContext, RuleFinding, block_rule
+
+
+@lru_cache(maxsize=1)
+def _block_budgets() -> Mapping[str, Mapping[str, Any]]:
+    """Load block_budgets.yaml, cached for the lifetime of the process."""
+    path = pkg_root() / "recipe" / "block_budgets.yaml"
+    data = load_yaml(path)
```
Trecek: [warning] defense: _block_budgets() calls load_yaml(path) without catching FileNotFoundError. If block_budgets.yaml is missing (e.g. after a partial install), load_yaml will raise and the lru_cache will cache the exception state, causing all subsequent calls to also raise. A missing-file fallback (return {}) would be more defensive.

src/autoskillit/recipe/rules_reachability.py
```diff
@@ -0,0 +1,209 @@
...
+    ``${{ context.X }}`` patterns.
+    """
+    step = ctx.recipe.steps.get(step_name)
+    if step is None:
+        return []
+    refs: list[str] = []
+    for val in (step.with_args or {}).values():
+        if isinstance(val, str):
+            refs.extend(_CTX_REF_RE.findall(val))
+    skill_cmd = (step.with_args or {}).get("skill_command", "")
```
Trecek, line 51: [warning] bugs: Double-scan of skill_command: the for-loop over with_args.values() (lines L47-L49) already scans the 'skill_command' value, then the explicit skill_cmd block (lines L50-L52) scans it again. This produces duplicate variable entries in refs, which causes _check_capture_inversion to emit duplicate RuleFinding objects for the same step and variable.

src/autoskillit/recipe/rules_blocks.py
```diff
@@ -0,0 +1,199 @@
...
+    return dict(budgets.get(block_name, budgets.get("DEFAULT", {})))
+
+
+@block_rule(
+    name="block-run-cmd-budget",
+    description="Block members contain at most the declared number of run_cmd steps.",
+    severity=Severity.ERROR,
+)
+def _check_block_run_cmd_budget(bctx: BlockContext) -> list[RuleFinding]:
+    budget = _budget_for(bctx.block.name).get("run_cmd", 1)
```
Trecek: [warning] cohesion: _check_block_run_cmd_budget defaults the run_cmd budget to 1 when the block is not in block_budgets.yaml, but _check_block_mcp_tool_budget returns early with no check when 'mcp_tools' is absent. These two budget rules are asymmetric: one has a hardcoded fallback default while the other silently skips.

src/autoskillit/recipe/rules_reachability.py
```diff
@@ -0,0 +1,209 @@
...
+        step_facts_set = facts.get(step_name, {frozenset()})
+        # Intersected fact set at this step (single element after _bfs_with_facts return)
+        intersected = next(iter(step_facts_set), frozenset())
+        known_vars = {var for var, _ in intersected}
+
+        for var in context_refs:
+            # Only flag vars that appear in the conditional fact domain of the recipe.
+            # If no step establishes (var, value) via a conditional edge, the var is
+            # captured unconditionally — not an inversion.
+            all_facts_in_recipe: set[tuple[str, str]] = set()
```
Trecek: [warning] cohesion: The inner loop over context_refs recomputes all_facts_in_recipe and recipe_fact_vars for every variable in every step (O(steps × vars × facts)). This duplicated scan inside the hot loop belongs in a pre-computed set built once before the outer loop.

src/autoskillit/recipe/contracts.py
```diff
@@ -221,17 +243,77 @@ def count_positional_args(skill_command: str) -> int:
...
+        member_count=len(block.members),
+        tool_counts_sorted=tool_counts_sorted,
+        gh_api_occurrences=block.gh_api_occurrences,
+        capture_names_hash=capture_names_hash,
+        entry_step=block.entry,
+        exit_step=block.exit,
+    )
+
+
+def _generate_recipe_card_for_recipe(recipe: Recipe) -> RecipeCard:
```
Trecek, line 280: [warning] cohesion: _generate_recipe_card_for_recipe is a private function, but the public generate_recipe_card can return either a dict or a RecipeCard depending on whether the first argument is a Recipe or a Path. This mixed return type breaks abstraction symmetry: callers on the path-based flow get a raw dict while callers on the Recipe-based flow get a typed object.

src/autoskillit/recipe/rules_blocks.py
```diff
@@ -0,0 +1,199 @@
...
+
+@block_rule(
+    name="block-entry-exit-reachable",
+    description=(
+        "Block members must form a single connected region with exactly one "
+        "entry step and one exit step."
+    ),
+    severity=Severity.WARNING,
+)
+def _check_block_entry_exit_reachable(bctx: BlockContext) -> list[RuleFinding]:
```
Trecek, line 161: [warning] cohesion: _check_block_entry_exit_reachable rebuilds the predecessor map by iterating over bctx.parent.step_graph from scratch for every block (O(blocks × edges)). The same predecessor map inversion was already computed inside extract_blocks. This is duplicate logic that should be shared — either as a field on RecipeBlock or as a helper accessible to both.

src/autoskillit/recipe/contracts.py
```diff
@@ -407,6 +502,37 @@ def check_contract_staleness(
...
+                        StaleItem(
+                            skill=stored_fp.name,
+                            reason="block_composition_drift",
+                            stored_value=repr(stored_fp),
+                            current_value=repr(current_fp),
+                        )
+                    )
+            return stale_items
+
+    if isinstance(contract, Recipe):
```
Trecek, line 538: [warning] cohesion: check_contract_staleness: when stored_card is None and contract is a Recipe, the function returns [] at line L533 without performing any checks. This silent early return is undocumented: the docstring describes the stored_card path but does not explain the Recipe-with-no-stored-card case, making the function's contract difficult to reason about.

src/autoskillit/execution/merge_queue.py
```diff
@@ -76,6 +76,33 @@ class PRFetchState(TypedDict):
 }
 """
 
+# Repo-level state query: consolidates three former run_cmd steps into one HTTP round-trip.
```
Trecek, line 79: [warning] slop: Block comment above _REPO_STATE_QUERY reads 'Distinct from _QUERY (PR-level + queue-entries); do not merge these constants.' — a meta-instruction that restates what the structural test already enforces. The comment explains 'do not merge' rather than 'what the constant does', making it an AI-generated caution note rather than useful documentation.

src/autoskillit/recipe/contracts.py
```diff
@@ -221,17 +243,77 @@ def count_positional_args(skill_command: str) -> int:
...
+
+def _generate_recipe_card_for_recipe(recipe: Recipe) -> RecipeCard:
+    """Generate a RecipeCard from a Recipe object (no disk write).
+
+    Used by the block fingerprint drift detection path in ``check_contract_staleness``
+    and by tests that want a ``RecipeCard`` with populated ``block_fingerprints``.
+    Uses a deferred import of ``_build_step_graph`` and ``extract_blocks`` to avoid
+    a circular import (``_analysis.py`` imports from ``contracts.py``).
+    """
+    # Deferred import to avoid circular dependency:
```
Trecek, line 288: [warning] slop: The inline comment block at lines L288-L290 duplicates the existing docstring. The docstring at L281-L286 already explains the deferred-import motivation and circular dependency. The inline comment is a redundant restatement directly above the import it describes.

tests/execution/test_check_repo_merge_state.py
```diff
@@ -0,0 +1,159 @@
+"""Round-trip budget tests for fetch_repo_merge_state.
+
+Mirrors test_single_graphql_call_per_poll_cycle at the consolidated-tool layer.
+All three booleans (queue_available, merge_group_trigger, auto_merge_available)
+must come from a single HTTP round-trip to the GitHub GraphQL endpoint.
+
+Tests will FAIL until:
```
Trecek: [warning] slop: Lines L7-L10 contain a 'Tests will FAIL until:' block comment in the module docstring. This is stale scaffold commentary: the implementation is complete (the functions it references exist in this very PR). Temporal 'will FAIL until' notes are AI implementation scaffolding that should be removed once the described code lands.

tests/recipe/test_bundled_recipes_no_inversions.py
```diff
@@ -0,0 +1,49 @@
+"""Bundled-recipe regression guard: no capture inversions or hardcoded event literals.
+
+Tests in this file will FAIL against the pre-Part-B state where:
```
Trecek: [warning] slop: Lines L3-L11 contain a 'Tests in this file will FAIL against the pre-Part-B state' block. This is temporal scaffold commentary describing a now-resolved pre-implementation state. The implementation exists in this PR; keeping references to a past broken state as permanent module-level documentation is AI slop.

tests/recipe/test_contracts_block_fingerprint.py
```diff
@@ -0,0 +1,119 @@
+"""Block fingerprint drift detection tests.
+
+Tests in this file will FAIL against the pre-Part-B state where:
```
Trecek: [warning] slop: Lines L3-L11 contain 'Tests in this file will FAIL against the pre-Part-B state' temporal scaffolding. The pre-conditions described are resolved by this PR and should not be immortalized as module documentation.

tests/recipe/test_rules_blocks.py
```diff
@@ -0,0 +1,103 @@
+"""Block-budget regression guards for the pre_queue_gate block.
+
+Tests in this file will FAIL against the current codebase:
```
Trecek: [warning] slop: Lines L3-L8 contain 'Tests in this file will FAIL against the current codebase' with a steps list. This is implementation-phase scaffolding left in the committed test file. The steps (2.1-2.5) have been completed; the comment describes a past state that no longer exists.

tests/recipe/test_rules_reachability.py
```diff
@@ -0,0 +1,185 @@
+"""Capture-inversion and event-scope reachability guards.
+
+Tests verify that:
+1. Bundled recipes produce zero inversion findings after Part B remediation.
+2. Synthetic recipes with inversions are correctly flagged.
+3. _bfs_with_facts correctly intersects facts at join points.
+
+Tests in this file will FAIL against the pre-Part-B state:
```
Trecek: [warning] slop: Lines L8-L12 contain 'Tests in this file will FAIL against the pre-Part-B state'. Same temporal scaffolding pattern present in multiple test files — these should be removed once implementation lands (which it has in this PR).

Trecek (commented):
AutoSkillit review found 24 blocking issues (changes_requested verdict). See inline comments for details. Note: cannot post as REQUEST_CHANGES on own PR — verdict is changes_requested.

src/autoskillit/recipe/rules_reachability.py
Trecek: Investigated — this is intentional. validator.py __all__ already re-exports _build_step_graph; contracts.py and io.py both import _build_step_graph directly from _analysis.py using the same pattern. All rule modules share _analysis.py internals by design.

src/autoskillit/recipe/rules_reachability.py
Trecek: Agreed — fixed in a07320e0. Lines 47-49 iterate with_args.values() including skill_command, then lines 50-52 scan skill_command again — duplicate refs causing duplicate RuleFinding entries.

src/autoskillit/recipe/rules_reachability.py
Trecek: Agreed — fixed in a07320e0. all_facts_in_recipe and recipe_fact_vars now pre-computed once before outer loop, eliminating O(steps×vars×facts) redundant recomputation.

src/autoskillit/recipe/rules_blocks.py
Trecek: Agreed — fixed in 01b63532. _block_budgets() now wraps load_yaml in try/except FileNotFoundError with {} fallback.

src/autoskillit/recipe/rules_blocks.py
Trecek: Agreed — fixed in 01b63532. Added DEFAULT entry to block_budgets.yaml and changed _check_block_run_cmd_budget to skip when run_cmd key absent — both rules now YAML-driven and symmetric.

src/autoskillit/recipe/rules_blocks.py
Trecek: Agreed — fixed in 01b63532. ValidationContext now carries a predecessors field (built once in build_validation_context); _check_block_entry_exit_reachable uses bctx.parent.predecessors directly.

src/autoskillit/recipe/contracts.py
Trecek: Agreed — fixed in 5b93aac1. assert replaced with if recipes_dir is None: raise ValueError(...) — enforced under python -O.

src/autoskillit/recipe/contracts.py
Trecek: Valid observation — flagged for design decision. The mixed dict|RecipeCard return type is an intentional backward-compat design explicitly documented in the signature and docstring. Flagged for design review.

src/autoskillit/recipe/contracts.py
Trecek: Agreed — fixed in 5b93aac1. Docstring now documents the Recipe-with-no-stored-card path: 'no comparison baseline available → [] returned immediately'.

src/autoskillit/recipe/contracts.py
Trecek: Investigated — this is intentional. The inline comment adds directional import chain detail ('contracts.py -> _analysis.py') not present in the docstring — it is not purely redundant.

src/autoskillit/execution/merge_queue.py
Trecek: Investigated — this is intentional. resp.raise_for_status() and resp.json() are inside the async with block (lines 465-466); reviewer misread indentation. JSONDecodeError propagation is an intentional design contract (only transport-level failures propagate).

src/autoskillit/execution/merge_queue.py
Trecek: Agreed — fixed in c9229b97. Added isinstance(body, dict) guard — non-dict response body now falls back to {} instead of raising AttributeError.

src/autoskillit/execution/merge_queue.py
Trecek: Investigated — this is intentional. Comment distinguishes _REPO_STATE_QUERY from _QUERY by scope and prevents confusion between two similar GraphQL constants — structural documentation, not AI caution note.

tests/recipe/test_rules_reachability.py
Trecek: Agreed — fixed in 7f9bf05e. Renamed to test_bundled_recipe_wait_for_ci_has_no_inversion_findings — accurately reflects the test asserting zero findings.

tests/recipe/test_rules_blocks.py
Trecek: Valid observation — flagged for design decision. Step name='' does not break the specific assertions (finding embeds bctx.block.name, not step.name) but the structural precision concern is valid. Flagged for design decision.

tests/recipe/test_bundled_recipes_no_inversions.py
Trecek: Agreed — fixed in 7f9bf05e. Assertion now guards against all hardcoded event literals (push, merge_group) via a set membership check.

tests/recipe/test_contracts_block_fingerprint.py
Trecek: Investigated — this is intentional. _generate_recipe_card_for_recipe calls extract_blocks freshly from recipe.steps — recipe.blocks field is never consulted. Test works correctly; reviewer misread the data flow.

tests/recipe/test_bundled_recipes_no_inversions.py
Trecek: Agreed — fixed in 7f9bf05e. Removed stale 'Tests in this file will FAIL against the pre-Part-B state' scaffold from module docstring.

tests/recipe/test_contracts_block_fingerprint.py
Trecek: Agreed — fixed in 7f9bf05e. Removed stale temporal scaffold from module docstring.

tests/recipe/test_rules_blocks.py
Trecek: Agreed — fixed in 7f9bf05e. Removed stale scaffold listing Steps 2.1-2.5 from module docstring.

tests/recipe/test_rules_reachability.py
Trecek: Agreed — fixed in 7f9bf05e. Removed stale 'will FAIL against the pre-Part-B state' scaffold from module docstring.

tests/execution/test_check_repo_merge_state.py
Trecek: Agreed — fixed in 7f9bf05e. Added assert result['ci_event'] is None — null blob text path (no parseable trigger) now verified.

tests/execution/test_check_repo_merge_state.py
Trecek: Agreed — fixed in 7f9bf05e. Added assert result['ci_event'] is None — null-workflows-dir path (ci_event=None contract) now verified.

tests/execution/test_check_repo_merge_state.py
Trecek: Agreed — fixed in 7f9bf05e. Removed stale 'Tests will FAIL until:' scaffold from module docstring.

PR #896, opened 2026-04-14
Rectify: Stale MCP Direct Entry — Symmetric Registration Lifecycle

### LINKED ISSUE #859: Stale direct MCP entry in ~/.claude.json causes dual-server spawning after marketplace install
## Problem

When a user runs `autoskillit init` (which writes a direct `mcpServers.autoskillit` entry to `~/.claude.json`) and later runs `autoskillit install` (which registers the marketplace plugin), the direct entry is **never cleaned up**. This causes **two separate autoskillit MCP server processes** to spawn in every Claude Code session — one from the direct entry (prefix `mcp__autoskillit__`) and one from the marketplace plugin (prefix `mcp__plugin_autoskillit_autoskillit__`).

### Symptoms

- Claude's first tool call attempt may target the wrong prefix, producing: *"The plugin tool isn't available under that name"*
- The model self-corrects on retry, but this wastes a turn and confuses users
- Two independent MCP server processes run with independent gate state and kitchen context
- `autoskillit doctor` reports OK because it short-circuits on the first registration it finds

### Reproduction

1. `autoskillit init` — writes `mcpServers.autoskillit` to `~/.claude.json`
2. `autoskillit install` — registers marketplace plugin but leaves the direct entry
3. Start a Claude Code session — two autoskillit servers spawn
4. Try to call `open_kitchen` — first attempt may fail with wrong prefix

### No eviction code exists anywhere

There is **no code path in the entire codebase** that removes a stale `mcpServers.autoskillit` entry from `~/.claude.json`. The only code that mutates `mcpServers` is `_register_mcp_server()` in `_init_helpers.py:336`, which only **adds** entries.

### Affected code paths

| Entry Point | File:Line | Current Behavior | Gap |
|---|---|---|---|
| `install()` | `_marketplace.py:99-175` | Registers marketplace, sweeps hooks only | Never touches `~/.claude.json` at all |
| `_register_all()` | `_init_helpers.py:433-435` | Skips write when plugin installed | Does not remove pre-existing direct entry |
| `cook()` | `_cook.py:91-173` | No MCP interaction | No check for dual registration |
| `order()` | `app.py:546-735` | No MCP interaction | No check for dual registration |
| `main()` | `app.py:738-745` | Calls `run_update_checks()` only | No dual-MCP signal in update checks |
| `run_update_checks()` | `_update_checks.py:606-648` | 3 signals: binary, hooks, source_drift | No signal for dual MCP registration |

### Doctor check gap

`_check_mcp_server_registered()` in `_doctor.py:34-89` uses a short-circuit OR pattern:
- **Branch 1** (line 44): If `mcpServers.autoskillit` exists in `~/.claude.json` → return OK immediately
- **Branch 2** (line 72): If `claude plugin list` shows autoskillit → return OK immediately
- **Fallthrough** (line 82): WARNING only if neither exists

It never checks for **both** existing simultaneously. The function returns OK on the first match and stops.

### Additional dual-server vector: `--plugin-dir` in cook/order

Both `cook()` and `order()` unconditionally pass `--plugin-dir <pkg_root()>` to the `claude` subprocess:
- `_cook.py:162`: `build_interactive_cmd(plugin_dir=pkg_root(), ...)`
- `app.py:462-463`: `cmd += [ClaudeFlags.PLUGIN_DIR, str(pkg_root())]`

`pkg_root()` contains `.mcp.json` declaring `mcpServers.autoskillit`. If the marketplace plugin is also installed (pointing to the same `pkg_root()` via symlink), Claude Code may spawn a second MCP server depending on its internal deduplication behavior. There is no guard in autoskillit that suppresses `--plugin-dir` when the marketplace plugin already covers it.

### 1. Eviction utility

Add a function (e.g., `evict_direct_mcp_entry()`) to `_init_helpers.py` that:
- Reads `~/.claude.json`
- Checks if `mcpServers.autoskillit` exists AND the marketplace plugin is installed (`_is_plugin_installed()`)
- If both exist, removes the `mcpServers.autoskillit` entry and writes back atomically
- Returns a bool indicating whether eviction occurred (for logging/messaging)

### 2. Call from every entry point

| Call Site | When to Call | User Feedback |
|---|---|---|
| `_marketplace.py:install()` | After successful `claude plugin install` | Print "Removed stale direct MCP entry" |
| `_init_helpers.py:_register_all()` | When `_is_plugin_installed()` is True | Print in summary block |
| `app.py:main()` | For all non-serve commands (alongside `run_update_checks`) | Silent — just fix it |

### 3. Doctor diagnostic

Add `_check_dual_mcp_registration()` to `_doctor.py` that:
- Checks for both `mcpServers.autoskillit` in `~/.claude.json` AND marketplace plugin
- Reports `Severity.WARNING` with actionable message when both exist
- The existing `_check_mcp_server_registered()` should be updated to not return OK on the first match when dual registration exists

### 4. Update check signal (optional)

Add a fourth signal kind `"dual_mcp"` to `run_update_checks()` that detects dual registration and auto-fixes silently, similar to how hook drift is auto-healed.

## Key File References

- `src/autoskillit/cli/_init_helpers.py` — `_register_mcp_server()` (L336), `_register_all()` (L377), `_is_plugin_installed()` (L279), `_user_claude_json_path()` (L331)
- `src/autoskillit/cli/_marketplace.py` — `install()` (L99)
- `src/autoskillit/cli/_doctor.py` — `_check_mcp_server_registered()` (L34)
- `src/autoskillit/cli/_update_checks.py` — `run_update_checks()` (L606)
- `src/autoskillit/cli/_mcp_names.py` — `detect_autoskillit_mcp_prefix()` (prefix detection, read-only)
- `src/autoskillit/cli/app.py` — `main()` (L738), `_launch_cook_session()` (L445), `order()` (L546)
- `src/autoskillit/cli/_cook.py` — `cook()` (L91)
- `src/autoskillit/.mcp.json` — declares `mcpServers.autoskillit`
- `src/autoskillit/execution/commands.py` — `build_interactive_cmd()` (L78, `--plugin-dir` flag)

## Prior Art

- PR #542 ("Rectify: MCP Tool Name Prefix Non-Determinism in Orchestrator Prompts") addressed prefix confusion in headless orchestrator prompts by adding `detect_autoskillit_mcp_prefix()`. That fix handles prompt generation but does NOT prevent dual-server spawning — it is a read-only detection mechanism.
- The hooks system already has analogous eviction patterns: `_evict_stale_autoskillit_hooks()` and `sweep_all_scopes_for_orphans()` clean up stale hook registrations across scopes. The same pattern should be applied to MCP server registrations.

Trecek:
## Summary

The MCP server registration in `~/.claude.json` is a write-only operation with no complementary eviction path. When a user runs `autoskillit init` (which writes `mcpServers.autoskillit`) and later runs `autoskillit install` (which installs the marketplace plugin), the direct entry is never removed — causing two independent autoskillit server processes to spawn per Claude Code session with isolated in-process gate state and shared disk-based kitchen-state, producing split-brain behavior.

The hooks system has already solved the analogous problem: `_evict_stale_autoskillit_hooks()` is always called before `sync_hooks_to_settings()`, in both `install()` and `_register_all()`. MCP registration has no such symmetric lifecycle. The architectural solution is to mirror the hooks eviction pattern exactly: introduce `evict_direct_mcp_entry()` alongside `_register_mcp_server()`, and call it from both `install()` and `_register_all()` — making the registration lifecycle bidirectional by design. A new independent doctor check exposes the dual-registration condition when it slips through.

Part B (separate task) covers the `_dual_mcp_signal` in `_update_checks.py` and conditional `--plugin-dir` suppression in cook/order/headless sessions.

### State Lifecycle Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINAL NODES %%
    START([startup / init / install])
    END([session launched])

    %% ── MCP CONFIG STATE ── %%
    subgraph MCPConfig ["● MCP REGISTRATION (External Config State)"]
        direction LR
        DirectMCP["DirectMCP<br/>━━━━━━━━━━<br/>~/.claude.json<br/>mcpServers[autoskillit]<br/>MUTABLE: write OR evict"]
        PluginMCP["PluginMCP<br/>━━━━━━━━━━<br/>installed_plugins.json<br/>autoskillit@local<br/>MUTABLE: install / uninstall"]
        DualBug["DUAL STATE<br/>━━━━━━━━━━<br/>DirectMCP AND PluginMCP<br/>both present simultaneously<br/>→ 2 server procs + split gate state"]
    end

    %% ── DETECTION GATES ── %%
    subgraph DetectionGates ["● DETECTION GATES (read-only, DERIVED)"]
        direction LR
        DualCheck["● _check_dual_mcp_registration<br/>━━━━━━━━━━<br/>reads both config files<br/>has_direct AND has_marketplace<br/>→ DoctorResult WARNING"]
        DualSignal["● _dual_mcp_signal<br/>━━━━━━━━━━<br/>wraps _is_dual_mcp_registered<br/>→ Signal fed to update checks<br/>alongside binary/hooks/drift"]
    end

    %% ── EVICTION GATES ── %%
    subgraph EvictionGates ["● EVICTION GATES (DirectMCP mutation control)"]
        direction TB
        RegisterEvict["● _register_all<br/>━━━━━━━━━━<br/>GATE: plugin_ok = True?<br/>→ evict_direct_mcp_entry<br/>FAIL-OPEN on IO/JSON error"]
        InstallEvict["● install<br/>━━━━━━━━━━<br/>POST-INSTALL GATE<br/>unconditional evict after<br/>claude plugin install succeeds"]
        StartupEvict["● main startup<br/>━━━━━━━━━━<br/>HEAL GATE<br/>evict_direct_mcp_entry<br/>every non-serve invocation"]
    end

    %% ── PLUGIN-DIR RESOLUTION ── %%
    subgraph PluginDirRes ["● PLUGIN-DIR RESOLUTION (DERIVED → INIT_ONLY)"]
        direction TB
        IsInstalled{"● _is_plugin_installed<br/>━━━━━━━━━━<br/>DERIVED gate<br/>reads installed_plugins.json<br/>at resolution time"}
        PDResolved["● resolved_dir<br/>━━━━━━━━━━<br/>INIT_ONLY after resolution<br/>None → omit --plugin-dir flag<br/>pkg_root() → inject --plugin-dir"]
    end

    %% ── TOOL CONTEXT PHASE 1 ── %%
    subgraph CtxP1 ["● TOOL CONTEXT Phase 1 — Construction"]
        direction LR
        PluginDirField["● ctx.plugin_dir<br/>━━━━━━━━━━<br/>MUTABLE<br/>str or None<br/>resolved in make_context"]
        KitchenId["ctx.kitchen_id<br/>━━━━━━━━━━<br/>MUTABLE<br/>empty string = closed<br/>UUID = open"]
        ActivePacks["ctx.active_recipe_packs<br/>━━━━━━━━━━<br/>MUTABLE<br/>None = kitchen closed<br/>frozenset = open"]
    end

    %% ── TOOL CONTEXT PHASE 2 ── %%
    subgraph CtxP2 ["● TOOL CONTEXT Phase 2 — Post-Construction Wiring"]
        direction LR
        ExecutorWire["ctx.executor<br/>━━━━━━━━━━<br/>MUTABLE two-phase<br/>None → DefaultHeadlessExecutor<br/>wired once in make_context"]
        MigrationsWire["ctx.migrations<br/>━━━━━━━━━━<br/>MUTABLE two-phase<br/>None → DefaultMigrationService<br/>wired once in make_context"]
        ResolverWire["● ctx output/write resolvers<br/>━━━━━━━━━━<br/>MUTABLE two-phase<br/>None → closures<br/>wired once in make_context"]
    end

    %% ── APPEND-ONLY STATE ── %%
    subgraph AppendOnly ["APPEND-ONLY STATE"]
        direction LR
        AuditLog["ctx.audit / token_log / timing_log<br/>━━━━━━━━━━<br/>APPEND_ONLY<br/>never cleared during session"]
        DoctorAccum["list DoctorResult<br/>━━━━━━━━━━<br/>APPEND_ONLY<br/>17 checks always accumulated"]
    end

    %% ── FROZEN COMMANDS ── %%
    subgraph FrozenCmds ["FROZEN COMMAND STATE (INIT_ONLY)"]
        direction LR
        ICmd["● ClaudeInteractiveCmd<br/>━━━━━━━━━━<br/>frozen dataclass<br/>cmd: list, env: Mapping<br/>immutable after build"]
        HCmd["ClaudeHeadlessCmd<br/>━━━━━━━━━━<br/>frozen dataclass<br/>cmd: list, env: Mapping<br/>immutable after build"]
    end

    %% ── DISMISS STATE ── %%
    subgraph DismissState ["● DISMISS STATE (~/.autoskillit/update_check.json)"]
        direction TB
        DismissFields["● update_prompt entry<br/>━━━━━━━━━━<br/>MUTABLE persisted JSON<br/>dismissed_at, dismissed_version<br/>conditions: list of str"]
        DismissGate["● _is_dismissed<br/>━━━━━━━━━━<br/>3-AXIS GATE:<br/>time window AND version AND<br/>condition membership — ALL must hold"]
    end

    %% ── CONNECTIONS ── %%

    %% startup flows into detection and resolution
    START --> DirectMCP
    START --> PluginMCP
    START --> IsInstalled
    START --> StartupEvict

    %% detection reads config
    DirectMCP --> DualCheck
    PluginMCP --> DualCheck
    DirectMCP --> DualSignal
    PluginMCP --> DualSignal
    DualCheck --> DualBug
    DualSignal --> DismissGate
    DismissFields --> DismissGate

    %% eviction gates mutate DirectMCP
    RegisterEvict -->|plugin_ok=True: evict| DirectMCP
    RegisterEvict -->|plugin_ok=False: write| DirectMCP
    InstallEvict -->|post-install: evict| DirectMCP
    StartupEvict -->|heal: evict| DirectMCP
    PluginMCP --> RegisterEvict
    PluginMCP --> InstallEvict

    %% plugin-dir resolution flows
    IsInstalled -->|not installed: pkg_root| PDResolved
    IsInstalled -->|installed: None| PDResolved
    PluginMCP --> IsInstalled
    PDResolved --> PluginDirField
    PDResolved --> ICmd
    PDResolved --> HCmd

    %% context construction flows
    PluginDirField --> ExecutorWire
    PluginDirField --> MigrationsWire
    PluginDirField --> ResolverWire
    KitchenId --> ExecutorWire
    ActivePacks --> ExecutorWire

    %% context → append-only → end
    ExecutorWire --> AuditLog
    MigrationsWire --> AuditLog
    ResolverWire --> AuditLog
    DoctorAccum --> AuditLog
    AuditLog --> END
    ICmd --> END
    HCmd --> END

    %% CLASS ASSIGNMENTS %%
    class DirectMCP,PluginMCP stateNode;
    class DualBug gap;
    class DualCheck,DualSignal,IsInstalled detector;
    class RegisterEvict,InstallEvict,StartupEvict handler;
    class PDResolved phase;
    class PluginDirField,KitchenId,ActivePacks stateNode;
    class ExecutorWire,MigrationsWire,ResolverWire stateNode;
    class AuditLog,DoctorAccum handler;
    class ICmd,HCmd output;
    class DismissFields stateNode;
    class DismissGate detector;
    class START,END terminal;
```

### Deployment Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    subgraph InitPhase ["INIT PHASE — autoskillit init (_init_helpers.py)"]
        direction TB
        REGISTER_ALL["● _register_all()<br/>━━━━━━━━━━<br/>Registration orchestrator<br/>_init_helpers.py"]
        IS_PLUGIN_INIT["● _is_plugin_installed()<br/>━━━━━━━━━━<br/>claude plugin list subprocess<br/>_init_helpers.py:281"]
        EVICT["● evict_direct_mcp_entry()<br/>━━━━━━━━━━<br/>Removes stale ~/.claude.json entry<br/>_init_helpers.py:359"]
        REGISTER_DIRECT["_register_mcp_server()<br/>━━━━━━━━━━<br/>Writes mcpServers.autoskillit<br/>_init_helpers.py:337"]
    end

    subgraph RegistrationStorage ["MCP REGISTRATION — LOCAL FILESYSTEM"]
        direction LR
        CLAUDE_JSON[("● ~/.claude.json<br/>━━━━━━━━━━<br/>Direct MCP entry<br/>mcpServers.autoskillit")]
        PLUGIN_REG[("Claude Plugin Registry<br/>━━━━━━━━━━<br/>~/.claude/plugins/<br/>Marketplace install")]
    end

    subgraph ClaudeCodeHost ["CLAUDE CODE HOST PROCESS"]
        direction TB
        CLAUDE_CODE["Claude Code<br/>━━━━━━━━━━<br/>Parent host process<br/>Reads both registrations"]
        DUAL_RISK["⚠ Dual-Server Risk (pre-fix)<br/>━━━━━━━━━━<br/>Both entries present →<br/>two MCP server instances spawned"]
    end

    subgraph MCPServer ["MCP SERVER — autoskillit serve (stdio)"]
        direction TB
        FACTORY["● make_context()<br/>━━━━━━━━━━<br/>Composition root<br/>server/_factory.py:124"]
        IS_PLUGIN_FACTORY["● _is_plugin_installed()<br/>━━━━━━━━━━<br/>Detects install mode<br/>server/_factory.py:207"]
        PLUGIN_DIR_RESOLVE["● plugin_dir resolution<br/>━━━━━━━━━━<br/>None if plugin mode<br/>pkg_root() if direct mode"]
        GATE["DefaultGateState<br/>━━━━━━━━━━<br/>Kitchen closed at startup<br/>Tag-based tool visibility"]
        BG_SUPER["DefaultBackgroundSupervisor<br/>━━━━━━━━━━<br/>Async task queue<br/>pipeline/background.py"]
    end

    subgraph ExecutionPhase ["SKILL EXECUTION — run_skill (MCP tool)"]
        direction TB
        BUILD_CMD["● build_full_headless_cmd()<br/>━━━━━━━━━━<br/>Builds claude --print argv<br/>execution/commands.py:180"]
        PLUGIN_DIR_FLAG["● Conditional --plugin-dir<br/>━━━━━━━━━━<br/>Only injected if plugin_dir≠None<br/>commands.py:238"]
        SESSION_SKILLS["DefaultSessionSkillManager<br/>━━━━━━━━━━<br/>Stages skill dirs<br/>--add-dir injection"]
    end

    subgraph HeadlessProcess ["HEADLESS SUBPROCESS — claude --print"]
        direction TB
        PTY_WRAP["script(1) PTY wrapper<br/>━━━━━━━━━━<br/>Linux: script -qefc<br/>New process group"]
        HEADLESS_CLAUDE["claude --print (headless)<br/>━━━━━━━━━━<br/>Skill execution session<br/>--output-format stream-json"]
    end

    subgraph LocalStorage ["LOCAL STORAGE"]
        direction TB
        TEMP_DIR[(".autoskillit/temp/<br/>━━━━━━━━━━<br/>Kitchen state, sentinel<br/>JSON / filesystem")]
        SESSION_LOGS[("~/.local/share/autoskillit/logs/<br/>━━━━━━━━━━<br/>Session diagnostics<br/>JSONL append")]
        CLAUDE_LOGS[("~/.claude/projects/.../SESSION.jsonl<br/>━━━━━━━━━━<br/>Conversation logs<br/>Polled for completion")]
        EPHEMERAL_SKILLS[("/dev/shm/autoskillit-sessions/<br/>━━━━━━━━━━<br/>Ephemeral skill dirs<br/>tmpfs / add-dir targets")]
        CREDENTIALS[("~/.claude/.credentials.json<br/>━━━━━━━━━━<br/>OAuth token<br/>Quota API auth")]
    end

    subgraph ExternalServices ["EXTERNAL SERVICES (HTTPS:443)"]
        direction TB
        GITHUB["GitHub REST + GraphQL API<br/>━━━━━━━━━━<br/>api.github.com<br/>Issues · PRs · CI · Merge queue"]
        ANTHROPIC["Anthropic Quota API<br/>━━━━━━━━━━<br/>api.anthropic.com/api/oauth/usage<br/>Rate-limit guard"]
    end

    %% INIT PHASE FLOW %%
    REGISTER_ALL --> IS_PLUGIN_INIT
    IS_PLUGIN_INIT -->|"plugin installed → fix"| EVICT
    IS_PLUGIN_INIT -->|"not installed"| REGISTER_DIRECT
    EVICT -->|"atomic delete"| CLAUDE_JSON
    REGISTER_DIRECT -->|"atomic write"| CLAUDE_JSON

    %% REGISTRATION → HOST %%
    CLAUDE_JSON -.->|"direct entry (stale if plugin)"| DUAL_RISK
    PLUGIN_REG -->|"plugin entry"| DUAL_RISK
    DUAL_RISK -->|"eviction fixes: one path only"| CLAUDE_CODE

    %% HOST → MCP SERVER %%
    CLAUDE_CODE -->|"spawns via stdio"| FACTORY

    %% MCP SERVER INTERNALS %%
    FACTORY --> IS_PLUGIN_FACTORY
    IS_PLUGIN_FACTORY -->|"plugin mode: None"| PLUGIN_DIR_RESOLVE
    IS_PLUGIN_FACTORY -->|"direct mode: pkg_root()"| PLUGIN_DIR_RESOLVE
    FACTORY --> GATE
    FACTORY --> BG_SUPER

    %% FACTORY → STORAGE %%
    FACTORY -->|"reads/writes"| TEMP_DIR
    FACTORY -->|"writes session data"| SESSION_LOGS
    FACTORY -->|"reads OAuth token"| CREDENTIALS
    BG_SUPER -->|"quota check"| ANTHROPIC
    FACTORY -->|"GitHub tools"| GITHUB

    %% EXECUTION PHASE %%
    PLUGIN_DIR_RESOLVE -->|"plugin_dir value"| BUILD_CMD
    SESSION_SKILLS -->|"stages skill dirs"| EPHEMERAL_SKILLS
    BUILD_CMD --> PLUGIN_DIR_FLAG
    PLUGIN_DIR_FLAG -->|"--plugin-dir only if direct mode"| PTY_WRAP
    SESSION_SKILLS -->|"--add-dir paths"| PTY_WRAP

    %% HEADLESS %%
    PTY_WRAP -->|"wraps"| HEADLESS_CLAUDE
    HEADLESS_CLAUDE -->|"writes"| CLAUDE_LOGS
    FACTORY -->|"polls Channel B"| CLAUDE_LOGS
    HEADLESS_CLAUDE -->|"reads skill dirs"| EPHEMERAL_SKILLS

    %% CLASS ASSIGNMENTS %%
    class REGISTER_ALL,IS_PLUGIN_INIT,EVICT,REGISTER_DIRECT handler;
    class CLAUDE_JSON,PLUGIN_REG stateNode;
    class CLAUDE_CODE,DUAL_RISK cli;
    class FACTORY,IS_PLUGIN_FACTORY,PLUGIN_DIR_RESOLVE,GATE,BG_SUPER phase;
    class BUILD_CMD,PLUGIN_DIR_FLAG,SESSION_SKILLS handler;
    class PTY_WRAP,HEADLESS_CLAUDE cli;
    class TEMP_DIR,SESSION_LOGS,CLAUDE_LOGS,EPHEMERAL_SKILLS,CREDENTIALS stateNode;
    class GITHUB,ANTHROPIC integration;
```

### Error/Resilience Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% ── ENTRY ── %%
    ENTRY(["● app.main()"])

    subgraph DISPATCH ["● CLI ENTRY DISPATCH (app.py)"]
        SERVE_CHK{"● command<br/>== 'serve'?"}
        PRE_EVICT["● evict_direct_mcp_entry()<br/>━━━━━━━━━━<br/>Proactive self-heal on<br/>every non-serve invocation<br/>(runs BEFORE app())"]
    end

    subgraph DUAL_MCP ["● DUAL-MCP CIRCUIT BREAKERS (_update_checks.py / _doctor.py)"]
        direction TB
        DETECT{"● _is_dual_mcp_registered()<br/>━━━━━━━━━━<br/>~/.claude.json mcpServers AND<br/>plugins/installed_plugins.json<br/>both present?"}
        DET_FAIL["OSError / JSONDecodeError<br/>━━━━━━━━━━<br/>→ return False<br/>(fail-open: cannot confirm)"]
        DUAL_SIG["● _dual_mcp_signal()<br/>━━━━━━━━━━<br/>Signal(dual_mcp, severity=WARNING)<br/>→ prompt: autoskillit install"]
        SIG_FAIL["except Exception<br/>━━━━━━━━━━<br/>→ return None<br/>(silently skip signal)"]
        DOCTOR["● _check_dual_mcp_registration()<br/>━━━━━━━━━━<br/>DoctorResult(WARNING)<br/>Advisory only — never aborts<br/>(OSError/JSONDecodeError → OK)"]
    end

    subgraph EVICTION ["● EVICTION CHAIN (_init_helpers.py / _marketplace.py)"]
        direction TB
        REG_ROUTE{"● _register_all()<br/>_is_plugin_installed()?<br/>━━━━━━━━━━<br/>Plugin already manages MCP?"}
        EVICT_REG["● evict_direct_mcp_entry()<br/>━━━━━━━━━━<br/>Remove stale mcpServers entry<br/>at init time if plugin present<br/>(OSError/JSONDecodeError → False)"]
        WRITE_MCP["_register_mcp_server()<br/>━━━━━━━━━━<br/>Write direct entry to<br/>~/.claude.json<br/>(no plugin path)"]
        MCP_FAIL["ValueError / OSError<br/>━━━━━━━━━━<br/>→ re-raised as CliError<br/>→ caught in init()<br/>→ SystemExit(1)"]
        POST_EVICT["● _marketplace.install()<br/>post-install eviction<br/>━━━━━━━━━━<br/>evict_direct_mcp_entry()<br/>after plugin successfully installed"]
    end

    subgraph INIT_GATES ["● INIT / INSTALL VALIDATION GATES (Fail-Fast)"]
        direction LR
        STDIN_G{"_require_interactive_stdin()<br/>━━━━━━━━━━<br/>stdin.isatty()?"}
        SECRET_G{"● _check_secret_scanning()<br/>━━━━━━━━━━<br/>Scanner present OR<br/>exact bypass phrase accepted?"}
        SRCTREE_G{"source tree guard<br/>━━━━━━━━━━<br/>project_dir inside<br/>autoskillit source?"}
        WORKTREE_G{"● _ensure_marketplace()<br/>━━━━━━━━━━<br/>is_git_worktree(pkg_dir)?"}
        BINARY_G{"claude binary check<br/>━━━━━━━━━━<br/>shutil.which('claude')?"}
        SCOPE_G{"scope validation<br/>━━━━━━━━━━<br/>scope in<br/>('user','project')?"}
    end

    subgraph CMD_GATE ["● COMMAND-BUILDER GATE (_cook.py / app.py / commands.py)"]
        direction LR
        PLUGIN_CHK{"● _is_plugin_installed()?<br/>━━━━━━━━━━<br/>plugin_dir = None if True"}
        WITH_DIR["--plugin-dir injected<br/>━━━━━━━━━━<br/>No plugin: server launched<br/>via --plugin-dir flag"]
        NO_DIR["● --plugin-dir OMITTED<br/>━━━━━━━━━━<br/>Plugin present:<br/>prevents 2nd server spawn<br/>(core dual-MCP fix)"]
    end

    subgraph FACTORY ["● FACTORY / CONTEXT GUARDS (_factory.py / context.py)"]
        direction TB
        TOKEN_G["● _gh_cli_token()<br/>━━━━━━━━━━<br/>except Exception → None<br/>Server starts without token"]
        REC_CHK{"RECORD_SCENARIO set?<br/>━━━━━━━━━━<br/>ImportError on api_simulator?"}
        REC_FAIL["ImportError<br/>━━━━━━━━━━<br/>→ re-raised as RuntimeError<br/>(explicit message, hard abort)"]
        REPLAY_G{"REPLAY_SCENARIO_DIR<br/>━━━━━━━━━━<br/>path exists?"}
        BG_G["● context.__post_init__<br/>━━━━━━━━━━<br/>background=None →<br/>auto-construct DefaultBackgroundSupervisor"]
    end

    subgraph SIGNALS ["● UPDATE-CHECK SIGNAL GATHERING (All Fail-Open)"]
        direction LR
        NET_S["_fetch_with_cache()<br/>━━━━━━━━━━<br/>network + JSON parse<br/>except Exception → None"]
        DRIFT_S["_source_drift_signal()<br/>━━━━━━━━━━<br/>except Exception → None"]
        HOOK_S["_hooks_signal()<br/>━━━━━━━━━━<br/>except Exception → None"]
        BIN_S["_binary_signal()<br/>━━━━━━━━━━<br/>except Exception → None"]
    end

    %% ── TERMINALS ── %%
    T_SERVE(["● Server starts<br/>(single-process, MCP safe)"])
    T_CMD(["● Claude cmd built<br/>(dual-MCP safe)"])
    T_WARN(["● Advisory WARNING<br/>surfaced to user"])
    T_SKIP(["Signal/check = None<br/>(silently skipped)"])
    T_ABORT(["SystemExit(1)<br/>ABORT"])

    %% ── CONNECTIONS ── %%

    %% Entry dispatch
    ENTRY --> SERVE_CHK
    SERVE_CHK -->|"serve"| TOKEN_G
    SERVE_CHK -->|"non-serve"| PRE_EVICT
    PRE_EVICT -->|"OSError/JSONDecodeError<br/>→ False"| T_SKIP
    PRE_EVICT -->|"entry removed"| T_WARN
    PRE_EVICT --> DETECT
    PRE_EVICT --> NET_S

    %% Dual-MCP detection
    DETECT -->|"both present"| DUAL_SIG
    DETECT -->|"file error"| DET_FAIL
    DET_FAIL --> T_SKIP
    DUAL_SIG -->|"signal emitted"| T_WARN
    DUAL_SIG -->|"except Exception"| SIG_FAIL
    SIG_FAIL --> T_SKIP
    DOCTOR -->|"both found"| T_WARN
    DOCTOR -->|"file error"| T_SKIP

    %% Eviction chain
    REG_ROUTE -->|"no plugin"| WRITE_MCP
    REG_ROUTE -->|"plugin installed"| EVICT_REG
    WRITE_MCP -->|"ValueError/OSError"| MCP_FAIL
    MCP_FAIL --> T_ABORT
    EVICT_REG -->|"fail-open result"| T_WARN
    POST_EVICT --> T_WARN

    %% Init/install fail-fast gates
    STDIN_G -->|"not tty"| T_ABORT
    SECRET_G -->|"bypass failed"| T_ABORT
    SRCTREE_G -->|"inside source"| T_ABORT
    WORKTREE_G -->|"is worktree"| T_ABORT
    BINARY_G -->|"missing"| T_ABORT
    SCOPE_G -->|"invalid scope"| T_ABORT
    STDIN_G -->|"ok"| REG_ROUTE
    SECRET_G -->|"ok"| SRCTREE_G
    WORKTREE_G -->|"ok"| WRITE_MCP
    BINARY_G -->|"found"| PLUGIN_CHK

    %% Command-builder gate
    PLUGIN_CHK -->|"no plugin"| WITH_DIR
    PLUGIN_CHK -->|"plugin installed"| NO_DIR
    WITH_DIR --> T_CMD
    NO_DIR --> T_CMD

    %% Factory guards
    TOKEN_G --> T_SERVE
    REC_CHK -->|"set + pkg missing"| REC_FAIL
    REC_FAIL --> T_ABORT
    REC_CHK -->|"not set / pkg ok"| T_SERVE
    REPLAY_G -->|"path missing"| T_SKIP
    REPLAY_G -->|"valid"| T_SERVE
    BG_G --> T_SERVE

    %% Fail-open signals
    NET_S --> T_SKIP
    DRIFT_S --> T_SKIP
    HOOK_S --> T_SKIP
    BIN_S --> T_SKIP

    %% ── CLASS ASSIGNMENTS ── %%
    class ENTRY cli;
    class SERVE_CHK,DETECT,REG_ROUTE,STDIN_G,SECRET_G,SRCTREE_G,WORKTREE_G,BINARY_G,SCOPE_G,PLUGIN_CHK,REC_CHK,REPLAY_G detector;
    class PRE_EVICT,EVICT_REG,POST_EVICT,DUAL_SIG handler;
    class WRITE_MCP,TOKEN_G,BG_G phase;
    class DET_FAIL,SIG_FAIL,MCP_FAIL,REC_FAIL gap;
    class NET_S,DRIFT_S,HOOK_S,BIN_S gap;
    class DOCTOR stateNode;
    class WITH_DIR,NO_DIR output;
    class T_SERVE,T_CMD,T_WARN,T_SKIP,T_ABORT terminal;
```

Closes #859

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260413-190109-264954/.autoskillit/temp/rectify/rectify_stale_mcp_dual_server_spawning_2026-04-13_191500_part_a.md`

### DIFF
--- a/src/autoskillit/cli/__init__.py
+++ b/src/autoskillit/cli/__init__.py
@@ -16,7 +16,7 @@ from autoskillit.cli._doctor import (
     _count_hook_registry_drift,
 )
 from autoskillit.cli._hooks import _claude_settings_path
-from autoskillit.cli._init_helpers import _is_plugin_installed, _prompt_recipe_choice
+from autoskillit.cli._init_helpers import _prompt_recipe_choice
 from autoskillit.cli._prompts import (
     _OPEN_KITCHEN_CHOICE,
     _build_open_kitchen_prompt,
@@ -50,6 +50,18 @@ from autoskillit.cli.app import (
     workspace_init,
 )
 from autoskillit.hook_registry import HookDriftResult
+
+
+def _is_plugin_installed() -> bool:
+    """Return True if autoskillit is installed as a Claude marketplace plugin.
+
+    Package-level wrapper over autoskillit.cli._init_helpers._is_plugin_installed,
+    exposed for consumers outside cli/ (e.g. server._factory).
+    """
+    from autoskillit.cli._init_helpers import _is_plugin_installed as _check
+
+    return _check()
+
 
 __all__ = [
     "_OPEN_KITCHEN_CHOICE",
--- a/src/autoskillit/cli/__init__.py
+++ b/src/autoskillit/cli/__init__.py
@@ -16,14 +16,13 @@ from autoskillit.cli._doctor import (
     _count_hook_registry_drift,
 )
 from autoskillit.cli._hooks import _claude_settings_path
-from autoskillit.cli._init_helpers import _prompt_recipe_choice
+from autoskillit.cli._init_helpers import _is_plugin_installed, _prompt_recipe_choice
 from autoskillit.cli._prompts import (
     _OPEN_KITCHEN_CHOICE,
     _build_open_kitchen_prompt,
     _build_orchestrator_prompt,
     _resolve_recipe_input,
 )
-from autoskillit.cli._update_checks import _is_plugin_installed
 from autoskillit.cli.app import (
     _generate_config_yaml,
     _prompt_test_command,
--- a/src/autoskillit/cli/__init__.py
+++ b/src/autoskillit/cli/__init__.py
@@ -23,6 +23,7 @@ from autoskillit.cli._prompts import (
     _build_orchestrator_prompt,
     _resolve_recipe_input,
 )
+from autoskillit.cli._update_checks import _is_plugin_installed
 from autoskillit.cli.app import (
     _generate_config_yaml,
     _prompt_test_command,
@@ -50,18 +51,6 @@ from autoskillit.cli.app import (
     workspace_init,
 )
 from autoskillit.hook_registry import HookDriftResult
-
-
-def _is_plugin_installed() -> bool:
-    """Return True if autoskillit is installed as a Claude marketplace plugin.
-
-    Package-level wrapper over autoskillit.cli._init_helpers._is_plugin_installed,
-    exposed for consumers outside cli/ (e.g. server._factory).
-    """
-    from autoskillit.cli._init_helpers import _is_plugin_installed as _check
-
-    return _check()
-
 
 __all__ = [
     "_OPEN_KITCHEN_CHOICE",
--- a/src/autoskillit/cli/__init__.py
+++ b/src/autoskillit/cli/__init__.py
@@ -16,7 +16,7 @@ from autoskillit.cli._doctor import (
     _count_hook_registry_drift,
 )
 from autoskillit.cli._hooks import _claude_settings_path
-from autoskillit.cli._init_helpers import _prompt_recipe_choice
+from autoskillit.cli._init_helpers import _is_plugin_installed, _prompt_recipe_choice
 from autoskillit.cli._prompts import (
     _OPEN_KITCHEN_CHOICE,
     _build_open_kitchen_prompt,
@@ -53,6 +53,7 @@ from autoskillit.hook_registry import HookDriftResult
 
 __all__ = [
     "_OPEN_KITCHEN_CHOICE",
+    "_is_plugin_installed",
     "_build_open_kitchen_prompt",
     "_build_orchestrator_prompt",
     "DoctorResult",
--- a/src/autoskillit/cli/__init__.py
+++ b/src/autoskillit/cli/__init__.py
@@ -16,7 +16,7 @@ from autoskillit.cli._doctor import (
     _count_hook_registry_drift,
 )
 from autoskillit.cli._hooks import _claude_settings_path
-from autoskillit.cli._init_helpers import _is_plugin_installed, _prompt_recipe_choice
+from autoskillit.cli._init_helpers import _prompt_recipe_choice
 from autoskillit.cli._prompts import (
     _OPEN_KITCHEN_CHOICE,
     _build_open_kitchen_prompt,
@@ -53,7 +53,6 @@ from autoskillit.hook_registry import HookDriftResult
 
 __all__ = [
     "_OPEN_KITCHEN_CHOICE",
-    "_is_plugin_installed",
     "_build_open_kitchen_prompt",
     "_build_orchestrator_prompt",
     "DoctorResult",
--- a/src/autoskillit/cli/__init__.py
+++ b/src/autoskillit/cli/__init__.py
@@ -25,6 +25,7 @@ from autoskillit.cli._prompts import (
 )
 from autoskillit.cli.app import (
     _generate_config_yaml,
+    _is_plugin_installed,
     _prompt_test_command,
     app,
     config_app,
@@ -53,6 +54,7 @@ from autoskillit.hook_registry import HookDriftResult
 
 __all__ = [
     "_OPEN_KITCHEN_CHOICE",
+    "_is_plugin_installed",
     "_build_open_kitchen_prompt",
     "_build_orchestrator_prompt",
     "DoctorResult",
--- a/src/autoskillit/cli/_doctor.py
+++ b/src/autoskillit/cli/_doctor.py
@@ -10,7 +10,11 @@ from dataclasses import dataclass
 from pathlib import Path
 
 from autoskillit.cli._hooks import _claude_settings_path, _load_settings_data
-from autoskillit.cli._init_helpers import _KNOWN_SCANNERS, _detect_secret_scanner
+from autoskillit.cli._init_helpers import (
+    _KNOWN_SCANNERS,
+    _check_dual_mcp_files,
+    _detect_secret_scanner,
+)
 from autoskillit.core import Severity, get_logger
 from autoskillit.execution import QUOTA_CACHE_SCHEMA_VERSION
 from autoskillit.hook_registry import (
@@ -102,26 +106,17 @@ def _check_dual_mcp_registration(
     _plugins_json = plugins_json_path or (
         Path.home() / ".claude" / "plugins" / "installed_plugins.json"
     )
-    try:
-        has_direct = _claude_json.exists() and "autoskillit" in json.loads(
-            _claude_json.read_text()
-        ).get("mcpServers", {})
-        has_marketplace = _plugins_json.exists() and "autoskillit@autoskillit-local" in json.loads(
-            _plugins_json.read_text()
-        ).get("plugins", {})
-        if has_direct and has_marketplace:
-            return DoctorResult(
-                severity=Severity.WARNING,
-                check="dual_mcp_registration",
-                message=(
-                    "autoskillit is registered both as a direct mcpServers entry "
-                    "(~/.claude.json) and as a marketplace plugin. This spawns two "
-                    "independent server processes per session with split gate state. "
-                    "Run `autoskillit install` to remove the stale direct entry."
-                ),
-            )
-    except (OSError, json.JSONDecodeError):
-        pass  # fail-open: unreadable files → cannot confirm dual, treat as OK
+    if _check_dual_mcp_files(_claude_json, _plugins_json):
+        return DoctorResult(
+            severity=Severity.WARNING,
+            check="dual_mcp_registration",
+            message=(
+                "autoskillit is registered both as a direct mcpServers entry "
+                "(~/.claude.json) and as a marketplace plugin. This spawns two "
+                "independent server processes per session with split gate state. "
+                "Run `autoskillit install` to remove the stale direct entry."
+            ),
+        )
     return DoctorResult(
         severity=Severity.OK,
         check="dual_mcp_registration",
--- a/src/autoskillit/cli/_doctor.py
+++ b/src/autoskillit/cli/_doctor.py
@@ -89,6 +89,46 @@ def _check_mcp_server_registered(claude_json_path: Path | None = None) -> Doctor
     )
 
 
+def _check_dual_mcp_registration(
+    claude_json_path: Path | None = None,
+    plugins_json_path: Path | None = None,
+) -> DoctorResult:
+    """Check that autoskillit is not registered both as a direct entry and as a marketplace plugin.
+
+    Returns WARNING if both registrations are simultaneously present (split-brain condition).
+    Fail-open: unreadable files → cannot confirm dual registration, return OK.
+    """
+    _claude_json = claude_json_path or (Path.home() / ".claude.json")
+    _plugins_json = plugins_json_path or (
+        Path.home() / ".claude" / "plugins" / "installed_plugins.json"
+    )
+    try:
+        has_direct = _claude_json.exists() and "autoskillit" in json.loads(
+            _claude_json.read_text()
+        ).get("mcpServers", {})
+        has_marketplace = _plugins_json.exists() and "autoskillit@autoskillit-local" in json.loads(
+            _plugins_json.read_text()
+        ).get("plugins", {})
+        if has_direct and has_marketplace:
+            return DoctorResult(
+                severity=Severity.WARNING,
+                check="dual_mcp_registration",
+                message=(
+                    "autoskillit is registered both as a direct mcpServers entry "
+                    "(~/.claude.json) and as a marketplace plugin. This spawns two "
+                    "independent server processes per session with split gate state. "
+                    "Run `autoskillit install` to remove the stale direct entry."
+                ),
+            )
+    except (OSError, json.JSONDecodeError):
+        pass  # fail-open: unreadable files → cannot confirm dual, treat as OK
+    return DoctorResult(
+        severity=Severity.OK,
+        check="dual_mcp_registration",
+        message="",
+    )
+
+
 def _check_hook_registration(settings_path: Path) -> DoctorResult:
     data = _load_settings_data(settings_path)
     registered = " ".join(
@@ -602,6 +642,9 @@ def run_doctor(*, output_json: bool = False) -> None:
 
     # Check 2: MCP server registered in ~/.claude.json or via plugin
     results.append(_check_mcp_server_registered(claude_json_path=Path.home() / ".claude.json"))
+
+    # Check 2b: Dual MCP registration — direct entry and marketplace plugin both present
+    results.append(_check_dual_mcp_registration())
 
     # Check 3: autoskillit command on PATH
     if shutil.which("autoskillit") is None:
--- a/src/autoskillit/cli/_init_helpers.py
+++ b/src/autoskillit/cli/_init_helpers.py
@@ -329,6 +329,26 @@ safety:
 """
 
 
+def _check_dual_mcp_files(claude_json: Path, plugins_json: Path) -> bool:
+    """Return True if both direct mcpServers entry and marketplace plugin are active.
+
+    Checks ~/.claude.json for an 'autoskillit' mcpServers key and
+    installed_plugins.json for an 'autoskillit@autoskillit-local' plugin entry.
+
+    Fail-open: any I/O or parse error returns False without raising.
+    """
+    try:
+        has_direct = claude_json.exists() and "autoskillit" in json.loads(
+            claude_json.read_text()
+        ).get("mcpServers", {})
+        has_marketplace = plugins_json.exists() and "autoskillit@autoskillit-local" in json.loads(
+            plugins_json.read_text()
+        ).get("plugins", {})
+        return has_direct and has_marketplace
+    except (OSError, json.JSONDecodeError):
+        return False
+
+
 def _user_claude_json_path() -> Path:
     """Return path to ~/.claude.json (user-scoped MCP server config)."""
     return Path.home() / ".claude.json"
--- a/src/autoskillit/cli/_init_helpers.py
+++ b/src/autoskillit/cli/_init_helpers.py
@@ -289,10 +289,8 @@ def _is_plugin_installed() -> bool:
             timeout=10,
         )
         return result.returncode == 0 and "autoskillit" in result.stdout
-    except FileNotFoundError:
-        return False  # claude CLI not on PATH
-    except (subprocess.TimeoutExpired, OSError):
-        return False  # CLI unavailable or timed out
+    except Exception:
+        return False  # CLI unavailable, timed out, or returned unexpected mock type
 
 
 def _generate_config_yaml(test_command: list[str]) -> str:
--- a/src/autoskillit/cli/_init_helpers.py
+++ b/src/autoskillit/cli/_init_helpers.py
@@ -10,8 +10,10 @@ from pathlib import Path
 from typing import NamedTuple
 
 from autoskillit.config import write_config_layer
-from autoskillit.core import YAMLError, atomic_write, dump_yaml_str, load_yaml
+from autoskillit.core import YAMLError, atomic_write, dump_yaml_str, get_logger, load_yaml
 from autoskillit.recipe import list_recipes
+
+logger = get_logger(__name__)
 
 
 class _ScanResult(NamedTuple):
@@ -290,7 +292,8 @@ def _is_plugin_installed() -> bool:
         )
         return result.returncode == 0 and "autoskillit" in result.stdout
     except Exception:
-        return False  # CLI unavailable, timed out, or returned unexpected mock type
+        logger.debug("plugin install check failed", exc_info=True)
+        return False
 
 
 def _generate_config_yaml(test_command: list[str]) -> str:
--- a/src/autoskillit/cli/_init_helpers.py
+++ b/src/autoskillit/cli/_init_helpers.py
@@ -355,6 +355,28 @@ def _register_mcp_server(claude_json_path: Path) -> None:
     atomic_write(claude_json_path, json.dumps(data, indent=2))
 
 
+def evict_direct_mcp_entry(claude_json_path: Path) -> bool:
+    """Remove mcpServers['autoskillit'] from ~/.claude.json if present.
+
+    Returns True if the entry was removed, False if nothing changed.
+    Fail-open: any I/O or parse error returns False without raising.
+    """
+    if not claude_json_path.exists():
+        return False
+    try:
+        data: dict = json.loads(claude_json_path.read_text())
+    except (OSError, json.JSONDecodeError):
+        return False
+    if "autoskillit" not in data.get("mcpServers", {}):
+        return False
+    del data["mcpServers"]["autoskillit"]
+    try:
+        atomic_write(claude_json_path, json.dumps(data, indent=2))
+    except OSError:
+        return False
+    return True
+
+
 def _print_next_steps(*, context: str = "install") -> None:
     _B, _C, _D, _G, _Y, _R = _colors()
     if context == "install":
@@ -433,6 +455,8 @@ def _register_all(scope: str, project_dir: Path) -> None:
     plugin_ok = _is_plugin_installed()
     if not plugin_ok:
         _register_mcp_server(_user_claude_json_path())
+    else:
+        evict_direct_mcp_entry(_user_claude_json_path())  # remove stale direct entry
 
     # --- Summary block ---
     print()
--- a/src/autoskillit/cli/_update_checks.py
+++ b/src/autoskillit/cli/_update_checks.py
@@ -72,7 +72,7 @@ _GITHUB_INTEGRATION_PYPROJECT_URL = (
 class Signal:
     """A single firing update condition."""
 
-    kind: Literal["binary", "hooks", "source_drift"]
+    kind: Literal["binary", "hooks", "source_drift", "dual_mcp"]
     message: str
 
 
@@ -521,6 +521,37 @@ def _source_drift_signal(info: InstallInfo, home: Path) -> Signal | None:
     return None
 
 
+def _is_dual_mcp_registered(home: Path) -> bool:
+    """Return True if both direct mcpServers entry and marketplace plugin are active."""
+    claude_json = home / ".claude.json"
+    plugins_json = home / ".claude" / "plugins" / "installed_plugins.json"
+    try:
+        has_direct = claude_json.exists() and "autoskillit" in json.loads(
+            claude_json.read_text()
+        ).get("mcpServers", {})
+        has_marketplace = plugins_json.exists() and "autoskillit@autoskillit-local" in json.loads(
+            plugins_json.read_text()
+        ).get("plugins", {})
+        return has_direct and has_marketplace
+    except (OSError, json.JSONDecodeError):
+        return False
+
+
+def _dual_mcp_signal(home: Path | None = None) -> Signal | None:
+    """Return a Signal if both direct mcpServers entry and marketplace plugin are registered."""
+    try:
+        if _is_dual_mcp_registered(home or Path.home()):
+            return Signal(
+                "dual_mcp",
+                "autoskillit is registered as both a direct MCP server (~/.claude.json) "
+                "and a marketplace plugin — two server processes will spawn per session. "
+                "Run `autoskillit install` to remove the stale direct entry.",
+            )
+    except Exception:
+        logger.debug("dual_mcp signal check failed", exc_info=True)
+    return None
+
+
 # ---------------------------------------------------------------------------
 # Unified dismissal
 # ---------------------------------------------------------------------------
@@ -658,6 +689,7 @@ def run_update_checks(home: Path | None = None) -> None:
         _binary_signal(info, _home, current),
         _hooks_signal(_claude_settings_path("user")),
         _source_drift_signal(info, _home),
+        _dual_mcp_signal(_home),
     ]
 
     # Filter by dismissal
--- a/src/autoskillit/cli/_update_checks.py
+++ b/src/autoskillit/cli/_update_checks.py
@@ -28,9 +28,6 @@ from typing import Any, Literal
 import httpx
 
 from autoskillit.cli._hooks import _claude_settings_path
-from autoskillit.cli._init_helpers import (
-    _is_plugin_installed,  # noqa: F401 — relay for autoskillit.cli re-export
-)
 from autoskillit.cli._install_info import (
     InstallInfo,
     InstallType,
--- a/src/autoskillit/cli/_update_checks.py
+++ b/src/autoskillit/cli/_update_checks.py
@@ -532,17 +532,18 @@ def _is_dual_mcp_registered(home: Path) -> bool:
 
 
 def _dual_mcp_signal(home: Path | None = None) -> Signal | None:
-    """Return a Signal if both direct mcpServers entry and marketplace plugin are registered."""
-    try:
-        if _is_dual_mcp_registered(home or Path.home()):
-            return Signal(
-                "dual_mcp",
-                "autoskillit is registered as both a direct MCP server (~/.claude.json) "
-                "and a marketplace plugin — two server processes will spawn per session. "
-                "Run `autoskillit install` to remove the stale direct entry.",
-            )
-    except Exception:
-        logger.debug("dual_mcp signal check failed", exc_info=True)
+    """Return a Signal if both direct mcpServers entry and marketplace plugin are registered.
+
+    _is_dual_mcp_registered() delegates to _check_dual_mcp_files() which is
+    fail-open (catches OSError and json.JSONDecodeError internally, never raises).
+    """
+    if _is_dual_mcp_registered(home or Path.home()):
+        return Signal(
+            "dual_mcp",
+            "autoskillit is registered as both a direct MCP server (~/.claude.json) "
+            "and a marketplace plugin — two server processes will spawn per session. "
+            "Run `autoskillit install` to remove the stale direct entry.",
+        )
     return None
 
 
--- a/src/autoskillit/cli/_update_checks.py
+++ b/src/autoskillit/cli/_update_checks.py
@@ -526,18 +526,12 @@ def _source_drift_signal(info: InstallInfo, home: Path) -> Signal | None:
 
 def _is_dual_mcp_registered(home: Path) -> bool:
     """Return True if both direct mcpServers entry and marketplace plugin are active."""
-    claude_json = home / ".claude.json"
-    plugins_json = home / ".claude" / "plugins" / "installed_plugins.json"
-    try:
-        has_direct = claude_json.exists() and "autoskillit" in json.loads(
-            claude_json.read_text()
-        ).get("mcpServers", {})
-        has_marketplace = plugins_json.exists() and "autoskillit@autoskillit-local" in json.loads(
-            plugins_json.read_text()
-        ).get("plugins", {})
-        return has_direct and has_marketplace
-    except (OSError, json.JSONDecodeError):
-        return False
+    from autoskillit.cli._init_helpers import _check_dual_mcp_files
+
+    return _check_dual_mcp_files(
+        home / ".claude.json",
+        home / ".claude" / "plugins" / "installed_plugins.json",
+    )
 
 
 def _dual_mcp_signal(home: Path | None = None) -> Signal | None:
--- a/src/autoskillit/cli/_update_checks.py
+++ b/src/autoskillit/cli/_update_checks.py
@@ -28,6 +28,9 @@ from typing import Any, Literal
 import httpx
 
 from autoskillit.cli._hooks import _claude_settings_path
+from autoskillit.cli._init_helpers import (
+    _is_plugin_installed,  # noqa: F401 — relay for autoskillit.cli re-export
+)
 from autoskillit.cli._install_info import (
     InstallInfo,
     InstallType,
--- a/src/autoskillit/cli/app.py
+++ b/src/autoskillit/cli/app.py
@@ -27,6 +27,7 @@ from autoskillit.cli._init_helpers import (
     _MARKER_CONTENT,
     _check_secret_scanning,
     _generate_config_yaml,
+    _is_plugin_installed,
     _log_secret_scan_bypass,
     _prompt_test_command,
     _register_all,
@@ -118,7 +119,7 @@ def serve(*, verbose: Annotated[bool, Parameter(name=["--verbose", "-v"])] = Fal
             ",".join(cfg.safety.protected_branches),
         )
 
-    plugin_dir = str(pkg_root())
+    plugin_dir: str | None = None if _is_plugin_installed() else str(pkg_root())
     ctx = make_context(cfg, plugin_dir=plugin_dir)
     _initialize(ctx)
 
@@ -457,14 +458,19 @@ def _launch_cook_session(
         resume_session_id=resume_session_id,
         env_extras=extra_env,
     )
-    cmd = spec.cmd + [
-        ClaudeFlags.PLUGIN_DIR,
-        str(pkg_root()),
-        ClaudeFlags.TOOLS,
-        "AskUserQuestion",
-        ClaudeFlags.APPEND_SYSTEM_PROMPT,
-        system_prompt,
-    ]
+    plugin_dir_flags: list[str] = (
+        [] if _is_plugin_installed() else [ClaudeFlags.PLUGIN_DIR, str(pkg_root())]
+    )
+    cmd = (
+        spec.cmd
+        + plugin_dir_flags
+        + [
+            ClaudeFlags.TOOLS,
+            "AskUserQuestion",
+            ClaudeFlags.APPEND_SYSTEM_PROMPT,
+            system_prompt,
+        ]
+    )
     with terminal_guard():
         result = subprocess.run(cmd, env=spec.env)
     if result.returncode != 0:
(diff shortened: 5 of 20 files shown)


src/autoskillit/server/_factory.py
```diff
@@ -15,6 +15,7 @@
 from pathlib import Path
 from typing import Any
 
+from autoskillit.cli import _is_plugin_installed
```
Trecek, line 18: [critical] arch: `server/_factory.py` (L3 server layer) imports `_is_plugin_installed` from `autoskillit.cli` (L3 cli layer). Server→CLI is an inverted same-layer dependency: server is the composition root for MCP tooling and should not depend on CLI internals. `_is_plugin_installed` lives in `cli/_init_helpers.py` and is then relayed through `cli/_update_checks.py` → `cli/__init__.py` purely to reach here. Relocate the function to a shared lower layer (e.g., `core/` or `execution/`) so both CLI and server can consume it without the server depending on the CLI.

src/autoskillit/cli/_doctor.py
```diff
@@ -89,6 +89,46 @@ def _check_mcp_server_registered(claude_json_path: Path | None = None) -> Doctor
     )
 
 
+def _check_dual_mcp_registration(
```
Trecek, line 96: [critical] cohesion: `_check_dual_mcp_registration()` independently re-implements the same `~/.claude.json` + `installed_plugins.json` parsing logic that `_is_dual_mcp_registered()` in `_update_checks.py` (line ~527) already provides. Both functions open the same two files, check `'autoskillit'` in `mcpServers` and `'autoskillit@autoskillit-local'` in `plugins`, and apply identical error handling. If the marketplace plugin key ever changes, only one site may be updated, causing doctor and update-checks to silently disagree. `_check_dual_mcp_registration()` should delegate to `_is_dual_mcp_registered()` rather than duplicating the detection logic.

src/autoskillit/cli/app.py
```diff
@@ -738,6 +740,9 @@ def main() -> None:
     """Entry point for autoskillit."""
     _first_arg = sys.argv[1] if len(sys.argv) > 1 else "serve"
     if _first_arg != "serve":
```
Trecek, line 745: [critical] bugs: `evict_direct_mcp_entry` is only called when `_first_arg != "serve"`, so running `autoskillit serve` **never** evicts the stale `~/.claude.json` direct entry. Claude Code reads `~/.claude.json` before the MCP server launches, so the stale entry causes two server processes to spawn regardless of the `plugin_dir=None` guard applied inside `serve()`. This is the core defect the PR set out to fix — the eviction should also run on the serve path (move it before the `if _first_arg != "serve":` gate, or call it unconditionally inside `serve()`).

src/autoskillit/cli/_update_checks.py
```diff
@@ -28,6 +28,9 @@
 import httpx
 
 from autoskillit.cli._hooks import _claude_settings_path
+from autoskillit.cli._init_helpers import (
```
Trecek: [warning] arch: `_update_checks.py` imports `_is_plugin_installed` from `_init_helpers.py` with a `# noqa: F401 — relay for autoskillit.cli re-export` comment. This creates a 3-hop relay chain (`_init_helpers` → `_update_checks` → `cli/__init__` → `server/_factory`). Using a module as a passthrough inflates the import graph and forces `_update_checks` to be loaded whenever `_is_plugin_installed` is needed. Remove the relay; `cli/__init__` should import directly from `_init_helpers`.

src/autoskillit/cli/_update_checks.py
```diff
@@ -28,6 +28,9 @@
...
+    _is_plugin_installed,  # noqa: F401 — relay for autoskillit.cli re-export
```
Trecek: [warning] slop: The `# noqa: F401 — relay for autoskillit.cli re-export` inline suppression is a code smell: a suppressed unused-import warning is a signal that the import does not belong here. Any future refactor of `_is_plugin_installed` must touch this unrelated file. Remove the relay by importing the function directly where it is needed.

src/autoskillit/cli/__init__.py
```diff
@@ -23,6 +23,7 @@
     _build_orchestrator_prompt,
     _resolve_recipe_input,
 )
+from autoskillit.cli._update_checks import _is_plugin_installed
```
Trecek: [warning] arch: `_is_plugin_installed` (a private, underscore-prefixed install helper from `cli/_init_helpers.py`) is added to `cli/__init__.__all__`, elevating a CLI internal to the package's public API surface. This makes `server/_factory.py` treat it as a stable public contract. A private function should not be promoted through `__all__`; it should be relocated to a shared lower-level module so callers import it directly from its canonical home.

src/autoskillit/cli/_update_checks.py
```diff
@@ -521,6 +524,37 @@ def _source_drift_signal(info: InstallInfo, home: Path) -> Signal | None:
     return None
 
 
+def _is_dual_mcp_registered(home: Path) -> bool:
```
Trecek, line 524: [warning] bugs: `_is_dual_mcp_registered()` duplicates the file-parsing logic that `_check_dual_mcp_registration()` in `_doctor.py` also implements independently. Both check the same files and the same dict keys. A single shared extraction (e.g., in `_init_helpers.py`) consumed by both callers eliminates the key-divergence risk — if the marketplace plugin key (currently `'autoskillit@autoskillit-local'`) ever changes, both detection sites must be updated in sync.

src/autoskillit/cli/_update_checks.py
```diff
@@ -521,6 +524,37 @@ def _source_drift_signal(info: InstallInfo, home: Path) -> Signal | None:
...
+    except (OSError, json.JSONDecodeError):
+        return False
+
+
+def _dual_mcp_signal(home: Path | None = None) -> Signal | None:
+    """Return a Signal if both direct mcpServers entry and marketplace plugin are registered."""
+    try:
+        if _is_dual_mcp_registered(home or Path.home()):
+            return Signal(
+                "dual_mcp",
```
Trecek: [warning] defense: The outer `except Exception` in `_dual_mcp_signal` (lines ~548-554) is a no-op guard: `_is_dual_mcp_registered()` already catches `(OSError, json.JSONDecodeError)` internally and never raises to callers. The double-guarding hides future regressions — if `_is_dual_mcp_registered` ever grows a new failure mode, `_dual_mcp_signal` will silently swallow it. Either remove the outer guard or narrow it to the specific exception from `Path.home()`.

src/autoskillit/cli/_doctor.py
```diff
@@ -89,6 +89,46 @@ def _check_mcp_server_registered(claude_json_path: Path | None = None) -> Doctor
...
+        has_direct = _claude_json.exists() and "autoskillit" in json.loads(
+            _claude_json.read_text()
+        ).get("mcpServers", {})
+        has_marketplace = _plugins_json.exists() and "autoskillit@autoskillit-local" in json.loads(
+            _plugins_json.read_text()
+        ).get("plugins", {})
+        if has_direct and has_marketplace:
+            return DoctorResult(
+                severity=Severity.WARNING,
+                check="dual_mcp_registration",
```
Trecek: [warning] cohesion: The check name `'dual_mcp_registration'` is consistent with the doctor naming convention, but the parallel detection function in `_update_checks.py` is named `_is_dual_mcp_registered` — an asymmetric verb prefix (`_check_` doctor side vs `_is_` update-checks side) for the same concept. If doctor delegated to `_is_dual_mcp_registered`, the naming asymmetry would dissolve and the duplication would be resolved simultaneously.

src/autoskillit/cli/_init_helpers.py
```diff
@@ -433,6 +456,8 @@ def _register_all(scope: str, project_dir: Path) -> None:
     plugin_ok = _is_plugin_installed()
     if not plugin_ok:
         _register_mcp_server(_user_claude_json_path())
+    else:
+        evict_direct_mcp_entry(_user_claude_json_path())  # remove stale direct entry
```
Trecek, line 480: [warning] defense: `evict_direct_mcp_entry(...)` return value is discarded silently in `_register_all()`. Unlike the `_marketplace.install()` path (which prints `'Removed stale direct MCP entry from ~/.claude.json'`), this eviction during `autoskillit init` gives the user no feedback that their config was mutated. Add a log or print message consistent with the install path.

src/autoskillit/server/_factory.py
```diff
@@ -200,7 +201,11 @@ def make_context(
         lambda: config.github.token or os.environ.get("GITHUB_TOKEN") or _gh_cli_token()
     )
 
-    resolved_dir = plugin_dir if plugin_dir is not None else _default_plugin_dir()
+    resolved_dir = (
```
Trecek, line 206: [warning] bugs: When `plugin_dir=None` is passed by a caller (as `serve()` does when the plugin is installed), `make_context()` discards that intent and re-invokes `_is_plugin_installed()` itself. An explicit `None` (meaning 'plugin is installed, no dir needed') is indistinguishable from 'not provided.' A TOCTOU race — where `_is_plugin_installed()` returns a different result between the caller's check and this re-check — causes `serve()` and `make_context()` to disagree on `resolved_dir`. Use a sentinel (`_UNSET = object()`) to distinguish 'not provided' from `None`, or trust the caller-supplied value and remove the re-check.

src/autoskillit/cli/app.py
```diff
@@ -738,6 +740,9 @@ def main() -> None:
...
+        from autoskillit.cli._init_helpers import _user_claude_json_path, evict_direct_mcp_entry
+
```
Trecek: [warning] cohesion: `evict_direct_mcp_entry` is now called from three separate sites with different conditions and feedback styles: (1) `_marketplace.install()` — prints confirmation; (2) `_register_all()` in `_init_helpers.py` — silent; (3) `main()` here — silent, on every non-serve CLI command. This scattered eviction pattern is hard to reason about. Consider a single authoritative eviction entry point with consistent feedback, invoked from a well-defined lifecycle hook rather than scattered across three callers.

src/autoskillit/cli/app.py
```diff
@@ -738,6 +740,9 @@ def main() -> None:
...
+        evict_direct_mcp_entry(_user_claude_json_path())  # silent heal for stale installs
```
Trecek: [warning] defense: `evict_direct_mcp_entry(...)` silently mutates `~/.claude.json` on every `autoskillit doctor`, `autoskillit migrate`, etc. invocation. The return value is discarded with no user-visible output. A user running `autoskillit doctor` to diagnose problems may have their config silently changed without knowing. The `_marketplace.install()` path prints `'Removed stale direct MCP entry'`; this path should do the same, or the eviction should be conditional (only when the marketplace plugin is detected).

tests/cli/test_cook.py
```diff
@@ -1010,6 +1085,17 @@ def _stub_preview(self, monkeypatch: pytest.MonkeyPatch) -> None:
             lambda *a, **kw: None,
         )
 
+    @pytest.fixture(autouse=True)
```
Trecek, line 1088: [warning] tests: The `_stub_is_plugin_installed` autouse fixture is copy-pasted verbatim across three test classes in this file (lines ~105, ~1088, ~1277). The fixture body is identical in all three. Extract it to a module-level or `conftest.py` autouse fixture to eliminate the duplication and prevent drift if the patching logic changes.

tests/cli/test_cook.py
```diff
@@ -1188,6 +1274,17 @@ def _stub_preview(self, monkeypatch: pytest.MonkeyPatch) -> None:
     def _interactive_stdin(self, monkeypatch: pytest.MonkeyPatch) -> None:
         monkeypatch.setattr("sys.stdin.isatty", lambda: True)
 
+    @pytest.fixture(autouse=True)
```
Trecek, line 1277: [warning] tests: Third identical copy of the `_stub_is_plugin_installed` autouse fixture (same as lines ~105 and ~1088). See finding at L1088 for the fix.

tests/cli/test_terminal.py
```diff
@@ -438,6 +438,7 @@ def test_cook_restores_terminal_on_keyboard_interrupt(self, monkeypatch, tmp_pat
             lambda *a, **kw: (_ for _ in ()).throw(KeyboardInterrupt()),
         )
         monkeypatch.setattr("shutil.which", lambda cmd: "/usr/bin/claude")
+        monkeypatch.setattr("autoskillit.cli._init_helpers._is_plugin_installed", lambda: False)
```
Trecek, line 441: [warning] tests: Inconsistent patch targets for the same guard across two test methods in this class. This method patches `autoskillit.cli._init_helpers._is_plugin_installed` (the source module), while `test_launch_cook_session_restores_terminal_on_keyboard_interrupt` at L487 patches `app_mod._is_plugin_installed` (the consumer module). Both happen to work because `_cook.py` does a late import (making the source-module patch correct for `cook()`) while `app.py` imports at module level (making the consumer-module patch correct for `_launch_cook_session`). Document the rationale for each patch target to prevent future breakage from an 'obvious' cleanup.

Trecek (commented):

Trecek (commented):
AutoSkillit review found 3 critical + 11 warning blocking issues. See inline comments.

**Critical issues:**
- `server/_factory.py:18` — Server layer imports from CLI layer (inverted dependency)
- `cli/_doctor.py:92` — Duplicated dual-MCP detection logic (divergence risk)
- `cli/app.py:742` — **Core bug:** `evict_direct_mcp_entry` not called in the `serve` path; stale `~/.claude.json` entry persists when `autoskillit serve` runs, two servers still spawn

2 additional findings require human decision (app.py L744-L745: scattered eviction sites with inconsistent feedback).

3 unpostable findings (outside diff hunks): test_cook.py over-mocking at lines 1471/1503; test_install.py misleading variable at 1181.

src/autoskillit/cli/_doctor.py
Trecek: Agreed — fixed in 53d0205c. Extracted shared detection logic into `_check_dual_mcp_files()` in `_init_helpers.py`. `_check_dual_mcp_registration()` in `_doctor.py` now delegates to it, eliminating the duplicate file-parsing logic.

src/autoskillit/cli/app.py
Trecek: Agreed — fixed in 65677712. Moved `evict_direct_mcp_entry` outside the `_first_arg != 'serve'` guard so it runs unconditionally at startup, including during `autoskillit serve`.

src/autoskillit/cli/_update_checks.py
Trecek: Agreed — fixed in 9b59ea7e. Removed the relay import from `_update_checks.py`. `_is_plugin_installed` is now re-exported directly via `cli/__init__.py` without the 3-hop chain.

src/autoskillit/cli/_update_checks.py
Trecek: Agreed — fixed in 9b59ea7e. The `# noqa: F401` relay import was removed from `_update_checks.py` entirely.

src/autoskillit/cli/_update_checks.py
Trecek: Agreed — fixed in 53d0205c. `_is_dual_mcp_registered()` now delegates to the shared `_check_dual_mcp_files()` helper in `_init_helpers.py`, eliminating the duplicated file-parsing logic.

src/autoskillit/cli/_update_checks.py
Trecek: Agreed — fixed in fdf8448f. Removed the outer `except Exception` from `_dual_mcp_signal`. `_check_dual_mcp_files` (the inner implementation) already catches `(OSError, json.JSONDecodeError)` and returns `False` — the outer guard was unreachable and hid regressions. Updated the test to verify fail-open behaviour via corrupted input.

src/autoskillit/server/_factory.py
Trecek: Agreed — fixed in 004382a1. Applied the `_UNSET` sentinel pattern (already established for `runner` at L68) to `plugin_dir`. The caller's explicit `None` (plugin installed) is now honoured without triggering a redundant `_is_plugin_installed()` subprocess re-invocation.

src/autoskillit/cli/app.py
Trecek: Agreed — fixed in 65677712. The call site in `main()` is now consistent with the other silent call sites (`_init_helpers.py` L460). The verbose call in `_marketplace.py` is intentionally different (installation flow with user feedback), which is appropriate.

src/autoskillit/server/_factory.py
Trecek: Valid observation — flagged for design decision. The server→cli layer inversion is real: `_is_plugin_installed` (a subprocess probe for `claude plugin list`) has no inherent CLI-layer dependency and could live in `core/` or `config/`. We removed the relay anti-pattern (`_update_checks` relay, commit 9b59ea7e) and now re-export via `cli.app` to satisfy the package gateway and monkeypatch arch tests. Moving to a shared layer requires a broader architectural decision.

src/autoskillit/cli/__init__.py
Trecek: Valid observation — flagged for design decision. `_is_plugin_installed` in `cli.__all__` elevates a private implementation detail to the public API surface. This was required because `server/_factory.py` needs access to it. The correct resolution is to move the function to `core/` or `config/` (eliminating the server→cli dependency), which is a broader architectural change requiring separate design agreement.

src/autoskillit/cli/app.py
Trecek: Valid observation — flagged for design decision. The 'silent heal' comment at this call site is intentional: `main()` runs on every CLI invocation and user-visible output on every startup would be noisy. The `_marketplace.py` site is deliberately verbose (installation flow). We kept this site silent but moved the call outside the serve guard (commit 65677712) so it runs on all startup paths. Whether to surface eviction during `autoskillit doctor` specifically is a UX decision that warrants product discussion.

PR #897, opened 2026-04-14
Rectify: Headless Core Crash Path Raises Instead of Returning Structured Error

### LINKED ISSUE #811: run_headless_core crash path raises instead of returning structured error
## Problem

When `run_headless_core` encounters a pre-launch failure (subprocess never spawns), it writes a crash entry to `sessions.jsonl` correctly via the `finally` block, then raises `RuntimeError("runner() did not return a result — cannot build SkillResult")` at line 1061 of `headless.py`.

This exception propagates uncaught through `tools_execution.py:run_skill` (line 292) to the MCP framework. When the parent session is degraded (e.g., context at 100% capacity), the error response cannot be delivered, and the tool call appears stuck as "Running..." indefinitely.

## Evidence

Two sessions (#801 at 20:52:20, #805 at 20:52:27 on 2026-04-13) showed this exact failure:
- `pid=0`, `session_id=""`, `exit_code=-1`, `0 tokens`, `termination_reason=CRASHED`
- The crash entries were written to `sessions.jsonl` (crash handler worked)
- But the `run_skill` tool call never returned — user saw "Running..." and had to interrupt manually
- The same pattern occurred at 17:36 UTC with 5 sessions crashing simultaneously

The pre-launch crash itself was caused by the MCP startup race (fixed in #803). But the **crash response path** is a separate issue: even when the crash is correctly detected, the tool call doesn't return a parseable result.

### 1. Unhandled RuntimeError on crash path

`headless.py` line 1060-1061:
```python
if _result is None:
    raise RuntimeError("runner() did not return a result — cannot build SkillResult")
```

This should return a structured `SkillResult` with `success=False, subtype="crashed"` instead of raising. The `run_skill` tool handler at `tools_execution.py:292` has no `except` clause for this — the exception escapes to the MCP framework.

### 2. Original exception not logged

`headless.py` lines 1036-1059: the `finally` block catches `_result is None` and writes the crash entry, but the exception that caused `runner()` to raise is not captured or logged. Only `flush_session_log` failures are logged (line 1058-1059). During this investigation, we could not determine the exact exception from any persistent log — the `summary.json` only shows the sentinel values (`pid=0`, `exit_code=-1`).

## Expected Behavior

1. When `runner()` raises, `run_headless_core` should catch the exception, log it with `logger.error(...)`, and return a `SkillResult(success=False, subtype="crashed", ...)` instead of re-raising
2. The original exception type, message, and traceback should be logged to the session diagnostics directory alongside `summary.json`
3. `run_skill` should always return valid JSON to the MCP client, even on pre-launch failures

## Affected Files

- `src/autoskillit/execution/headless.py:1020-1061` — crash path in `run_headless_core`
- `src/autoskillit/server/tools_execution.py:291-307` — `run_skill` try block with no crash handler

## Investigation Report

Full analysis: `.autoskillit/temp/investigate/investigation_stalled_running_headless_sessions_2026-04-13_211500.md`

Trecek:
## Summary

`run_headless_core` raises `RuntimeError` on runner crash instead of returning a structured `SkillResult`. The exception propagates through `DefaultHeadlessExecutor.run()` and `run_skill` (neither has `except`), reaching `@track_response_size` which converts it to a generic 4-field `tool_exception` envelope — not the 13+ field `SkillResult` envelope that pipeline orchestrators parse for `needs_retry`, `session_id`, `subtype`, etc. Additionally, the original exception traceback is never persisted to session diagnostics, making post-mortem analysis impossible.

The existing crash test (`test_headless.py:2151`) explicitly asserts `pytest.raises(RuntimeError)`, validating the broken behavior as correct.

This PR converts the crash path to return a structured `SkillResult` via a new `SkillResult.crashed()` factory, adds exception traceback persistence to `flush_session_log`, and adds defense-in-depth at the `run_skill` boundary. All 14 `@mcp.tool()` handlers gain mandatory `try/except Exception` boundaries enforced by a new AST test in `tests/arch/test_never_raises_contracts.py`.

### Error/Resilience Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    MCP_IN([MCP Tool Request])

    %% ── L4: BACKSTOP ─────────────────────────────────────────────── %%
    BACKSTOP["L4: @track_response_size backstop<br/>━━━━━━━━━━<br/>unreachable in normal operation<br/>→ subtype: tool_exception"]

    %% ── HOOK-LEVEL GUARDS (PreToolUse, outside Python server) ─────── %%
    subgraph HookGates ["HOOK-LEVEL GATES (PreToolUse — Claude Code layer)"]
        direction LR
        HQ["quota_guard<br/>━━━━━━━━━━<br/>should_block?"]
        HH["headless_orchestration_guard<br/>━━━━━━━━━━<br/>run_skill in headless?"]
        HB["branch_protection_guard<br/>━━━━━━━━━━<br/>protected branch?"]
        HSK["skill_command_guard<br/>━━━━━━━━━━<br/>slash prefix check"]
    end

    %% ── L3: @mcp.tool() OUTER EXCEPTION BOUNDARY ─────────────────── %%
    subgraph ToolBoundary ["● L3: @mcp.tool() OUTER EXCEPTION BOUNDARY (tools_*.py × 14)"]
        direction TB
        OUTER_CATCH["outer except Exception<br/>━━━━━━━━━━<br/>pre-executor setup errors<br/>→ SkillResult.crashed().to_json()"]
    end

    %% ── IN-HANDLER VALIDATION ─────────────────────────────────────── %%
    subgraph InHandlerGates ["IN-HANDLER VALIDATION (fail-fast, tools_*.py)"]
        direction LR
        IH1["_require_not_headless()<br/>━━━━━━━━━━<br/>AUTOSKILLIT_HEADLESS == 1?"]
        IH2["_require_enabled()<br/>━━━━━━━━━━<br/>kitchen open? (gate state)"]
        IH3["run_skill preconditions<br/>━━━━━━━━━━<br/>absolute cwd · executor set<br/>dry-walkthrough marker check"]
    end

    %% ── L2: EXECUTOR EXCEPTION BOUNDARY ──────────────────────────── %%
    subgraph ExecBoundary ["● L2: EXECUTOR EXCEPTION BOUNDARY (tools_execution.py)"]
        direction TB
        INNER_CATCH["inner except Exception<br/>━━━━━━━━━━<br/>executor.run() errors<br/>→ SkillResult.crashed().to_json()"]
    end

    %% ── L1: run_headless_core() EXCEPTION BOUNDARY ────────────────── %%
    subgraph HeadlessBoundary ["● L1: run_headless_core() EXCEPTION BOUNDARY (headless.py)"]
        direction TB
        SUBPROCESS["subprocess runner<br/>━━━━━━━━━━<br/>claude headless session<br/>JSONL streaming"]
        CRASH_CATCH["except Exception<br/>━━━━━━━━━━<br/>traceback.format_exc()<br/>logger.error()"]
        ATOMIC_WRITE["● atomic_write()<br/>━━━━━━━━━━<br/>session_dir/crash_exception.txt<br/>(write-only post-mortem artifact)"]
        CRASHED_SR["● SkillResult.crashed()<br/>━━━━━━━━━━<br/>subtype=crashed · needs_retry=False<br/>exit_code=-1 · success=False"]
    end

    %% ── SESSION ERROR DETECTION ───────────────────────────────────── %%
    subgraph SessionDetect ["SESSION ERROR DETECTION (headless.py + session.py)"]
        direction TB
        TERM_REASON["termination reason classifier<br/>━━━━━━━━━━<br/>STALE · IDLE_STALL · TIMED_OUT<br/>NATURAL_EXIT"]
        PARSE_RESULT["parse_session_result()<br/>━━━━━━━━━━<br/>empty output · unparseable NDJSON<br/>context exhaustion"]
        CONTENT_EVAL["_evaluate_content_state()<br/>━━━━━━━━━━<br/>COMPLETE · ABSENT · CONTRACT_VIOLATION"]
        COMPUTE_OUT["_compute_outcome()<br/>━━━━━━━━━━<br/>contradiction guard (success+retry → failure)<br/>dead-end guard (ABSENT → DRAIN_RACE retry)"]
    end

    %% ── RESULT OVERRIDE GATES ─────────────────────────────────────── %%
    subgraph ResultGates ["● RESULT OVERRIDE GATES (_type_results.py + headless.py)"]
        direction LR
        PATH_G["path_contamination gate<br/>━━━━━━━━━━<br/>write path outside cwd?<br/>→ RETRIABLE(PATH_CONTAMINATION)"]
        ZERO_G["zero_writes gate<br/>━━━━━━━━━━<br/>write-expected skill, 0 Edit/Write calls?<br/>→ RETRIABLE(ZERO_WRITES)"]
        CONTRACT_G["contract_recovery gate<br/>━━━━━━━━━━<br/>adjudicated_failure + write evidence?<br/>→ promotes to RETRIABLE"]
        BUDGET_G["_apply_budget_guard() — circuit breaker<br/>━━━━━━━━━━<br/>consecutive_failures > max_consecutive_retries<br/>→ BUDGET_EXHAUSTED (terminal)"]
    end

    %% ── RECOVERY MECHANISMS ───────────────────────────────────────── %%
    subgraph Recovery ["RECOVERY MECHANISMS"]
        direction LR
        STALE_REC["stale stdout recovery<br/>━━━━━━━━━━<br/>parse SUCCESS from stdout<br/>→ synthesize recovered_from_stale"]
        CLONE_REV["clone contamination revert<br/>━━━━━━━━━━<br/>git reset --hard + git clean<br/>→ RETRIABLE(CLONE_CONTAMINATION)"]
        QUOTA_SLEEP["quota sleep routing<br/>━━━━━━━━━━<br/>hook: run_cmd sleep N<br/>→ LLM retries run_skill"]
    end

    %% ── AST ENFORCEMENT ───────────────────────────────────────────── %%
    AST_ENFORCE["● tests/arch/test_never_raises_contracts.py<br/>━━━━━━━━━━<br/>AST: all @mcp.tool() handlers must have<br/>except Exception + 'Never raises.' docstring"]

    %% ── TERMINAL STATES ───────────────────────────────────────────── %%
    T_GATE_DENY([GATE_DENIED])
    T_SUCCEED([SUCCEEDED])
    T_RETRY([RETRIABLE])
    T_FAIL([FAILED])
    T_BUDGET([BUDGET_EXHAUSTED])

    %% ═══════════════ CONNECTIONS ═══════════════ %%

    %% Entry → hooks %%
    MCP_IN --> HookGates
    HQ -->|"quota exceeded"| QUOTA_SLEEP
    QUOTA_SLEEP -->|"after sleep"| MCP_IN
    HH -->|"deny"| T_GATE_DENY
    HB -->|"deny"| T_GATE_DENY
    HSK -->|"deny"| T_GATE_DENY

    %% Hooks → L4 backstop → L3 boundary %%
    HookGates -->|"permit"| BACKSTOP
    BACKSTOP --> ToolBoundary
    BACKSTOP -->|"escapes all layers"| T_FAIL

    %% L3 → validation → L2 %%
    ToolBoundary --> InHandlerGates
    ToolBoundary -->|"exception in outer body"| OUTER_CATCH
    OUTER_CATCH --> T_FAIL
    IH1 -->|"headless"| T_GATE_DENY
    IH2 -->|"kitchen closed"| T_GATE_DENY
    IH3 -->|"precondition fail"| T_FAIL
    InHandlerGates --> ExecBoundary

    %% L2 → L1 %%
    ExecBoundary --> HeadlessBoundary
    ExecBoundary -->|"exception in executor.run()"| INNER_CATCH
    INNER_CATCH --> T_FAIL

    %% L1 crash path %%
    SUBPROCESS -->|"exception during run"| CRASH_CATCH
    CRASH_CATCH --> ATOMIC_WRITE
    CRASH_CATCH --> CRASHED_SR
    CRASHED_SR --> T_FAIL

    %% L1 normal exit → session detection %%
    SUBPROCESS -->|"subprocess exits"| SessionDetect
    TERM_REASON -->|"STALE"| STALE_REC
    STALE_REC -->|"recoverable stdout"| T_SUCCEED
    STALE_REC -->|"not recoverable"| ResultGates
    TERM_REASON -->|"IDLE_STALL / TIMED_OUT"| ResultGates
    TERM_REASON -->|"NATURAL_EXIT"| PARSE_RESULT
    PARSE_RESULT --> CONTENT_EVAL
    CONTENT_EVAL --> COMPUTE_OUT
    COMPUTE_OUT -->|"SUCCEEDED"| T_SUCCEED
    COMPUTE_OUT -->|"RETRIABLE / FAILED"| ResultGates

    %% Override gates %%
    PATH_G -->|"contaminated"| CLONE_REV
    CLONE_REV --> T_RETRY
    PATH_G -->|"clean"| ZERO_G
    ZERO_G -->|"zero writes"| T_RETRY
    ZERO_G -->|"ok"| CONTRACT_G
    CONTRACT_G -->|"promotes"| T_RETRY
    CONTRACT_G -->|"no promotion"| BUDGET_G
    BUDGET_G -->|"budget exhausted"| T_BUDGET
    BUDGET_G -->|"budget ok"| T_RETRY

    %% Enforcement link %%
    ToolBoundary -.->|"structurally enforced by"| AST_ENFORCE

    %% CLASS ASSIGNMENTS %%
    class MCP_IN terminal;
    class HQ,HH,HB,HSK detector;
    class BACKSTOP gap;
    class OUTER_CATCH,INNER_CATCH handler;
    class IH1,IH2,IH3 detector;
    class SUBPROCESS handler;
    class CRASH_CATCH,ATOMIC_WRITE,CRASHED_SR phase;
    class TERM_REASON,PARSE_RESULT,CONTENT_EVAL,COMPUTE_OUT handler;
    class PATH_G,ZERO_G,CONTRACT_G,BUDGET_G detector;
    class STALE_REC,CLONE_REV,QUOTA_SLEEP output;
    class AST_ENFORCE newComponent;
    class T_GATE_DENY,T_SUCCEED,T_RETRY,T_FAIL,T_BUDGET terminal;
```

### State Lifecycle Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINAL NODES %%
    ENTRY([MCP Tool Called])
    RESULT_OUT([SkillResult / JSON → MCP Caller])

    subgraph Lifecycles ["FIELD LIFECYCLE CATEGORIES — ● SkillResult (_type_results.py)"]
        direction LR
        IO["INIT_ONLY — crash factory<br/>━━━━━━━━━━<br/>success=False · is_error=True<br/>exit_code=-1 · subtype='crashed'<br/>needs_retry=False · session_id=''<br/>retry_reason=NONE · stderr=''<br/>result='{ExcType}: {message}'"]
        MU["MUTABLE — reassignable ref<br/>━━━━━━━━━━<br/>skill_result (reference)<br/>may be reassigned by:<br/>• contract nudge (CONTRACT_RECOVERY)<br/>• clone contamination revert<br/>— token_usage · worktree_path"]
        AO["APPEND_ONLY<br/>━━━━━━━━━━<br/>write_path_warnings<br/>grows during run,<br/>never cleared or replaced"]
        DE["DERIVED — not stored<br/>━━━━━━━━━━<br/>outcome property<br/>SUCCEEDED / RETRIABLE / FAILED<br/>excluded from to_json()"]
    end

    subgraph ExcGates ["EXCEPTION BOUNDARY GATES — ● tools_*.py (all server tools)"]
        direction TB
        G1["● Pre-execution Gate<br/>━━━━━━━━━━<br/>Gate policy · kitchen open · quota<br/>Returns error json before any mutation<br/>(FAIL-FAST)"]
        G2["● try: handler body<br/>━━━━━━━━━━<br/>ALL @mcp.tool() handlers<br/>Mandatory top-level try/except Exception<br/>Contract: 'Never raises' in docstring"]
        G3{"exception<br/>raised?"}
        G4["● on exception → error json<br/>━━━━━━━━━━<br/>most tools: json.dumps({'success':False,...})<br/>run_skill: SkillResult.crashed().to_json()<br/>NEVER propagates raw exception"]
        G5["on success → structured json<br/>━━━━━━━━━━<br/>domain-specific response dict<br/>OR SkillResult.to_json()"]
        G6["@track_response_size<br/>━━━━━━━━━━<br/>Tertiary safety net (helpers.py)<br/>subtype='tool_exception'<br/>last-resort catch for any escape"]
    end

    subgraph CrashState ["● CRASH STATE CONSTRUCTION — run_headless_core (headless.py)"]
        direction TB
        C1["● except Exception as exc<br/>━━━━━━━━━━<br/>Runner failure intercepted<br/>NEVER re-raises to caller"]
        C2["● traceback.format_exc()<br/>━━━━━━━━━━<br/>Full traceback captured<br/>into _exc_text string"]
        C3["● flush_session_log(<br/>    exception_text=_exc_text,<br/>    termination_reason='CRASHED',<br/>    exit_code=-1, pid=0<br/>)<br/>━━━━━━━━━━<br/>Best-effort; swallows own failures"]
        C4["● SkillResult.crashed(exc)<br/>━━━━━━━━━━<br/>Factory locks all INIT_ONLY fields<br/>order_id passed through<br/>Returns structured SkillResult"]
    end

    subgraph Artifacts ["● SESSION ARTIFACTS (session_log.py)"]
        direction LR
        AF1["● crash_exception.txt<br/>━━━━━━━━━━<br/>atomic_write()<br/>Full traceback text<br/>WRITE-ONLY audit artifact<br/>(never read back by system)"]
        AF2["SkillResult JSON envelope<br/>━━━━━━━━━━<br/>worktree_path omitted if None<br/>order_id omitted if empty<br/>outcome always excluded"]
    end

    subgraph AST ["● BUILD-TIME ENFORCEMENT — test_never_raises_contracts.py"]
        direction TB
        A1["● test_all_mcp_tool_handlers_have_except_exception<br/>━━━━━━━━━━<br/>AST walk: every @mcp.tool() in server/<br/>must have top-level try/except Exception"]
        A2["● test_all_mcp_tool_handlers_claim_never_raises<br/>━━━━━━━━━━<br/>Docstring must contain 'Never raises'<br/>(case-insensitive)"]
        A3["● test_never_raises_contracts_are_structurally_enforced<br/>━━━━━━━━━━<br/>'Never raises' docstring anywhere in server/<br/>→ must have try/except Exception or BaseException"]
    end

    %% PRIMARY FLOW %%
    ENTRY --> G1
    G1 -->|gate fail: return error| RESULT_OUT
    G1 -->|gate pass| G2
    G2 --> G3
    G3 -->|No| G5
    G3 -->|Yes| G4
    G5 --> RESULT_OUT
    G4 --> RESULT_OUT
    G2 -.->|escape path| G6
    G6 --> RESULT_OUT

    %% CRASH PATH (run_skill delegates into headless) %%
    G2 -.->|"run_skill delegates to<br/>run_headless_core"| C1
    C1 --> C2
    C2 --> C3
    C3 --> C4
    C3 --> AF1
    C4 --> IO
    C4 --> AF2
    AF2 --> G4

    %% FIELD RELATIONSHIPS %%
    IO -.->|"sealed by factory"| AF2
    MU -.->|"may be replaced post-runner"| AF2
    DE -.->|"computed; excluded"| AF2

    %% BUILD-TIME ENFORCEMENT %%
    A1 & A2 & A3 -.->|"enforce at test-time"| G2

    %% CLASS ASSIGNMENTS %%
    class ENTRY,RESULT_OUT terminal;
    class IO detector;
    class MU stateNode;
    class AO gap;
    class DE output;
    class G1 detector;
    class G2,G3 phase;
    class G4 handler;
    class G5 output;
    class G6 gap;
    class C1,C2,C3,C4 handler;
    class A1,A2,A3 cli;
    class AF1,AF2 output;
```

### Operational Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    MCP_CLIENT(["MCP Client / Pipeline Orchestrator"])

    subgraph TOOL_LAYER ["● MCP TOOL HANDLER LAYER  (server/tools_*.py — 12 files)"]
        direction TB
        RUN_SKILL["● run_skill<br/>━━━━━━━━━━<br/>skill_command, cwd, model<br/>→ SkillResult JSON<br/><i>Never raises</i>"]
        OTHER_TOOLS["● 11 other @mcp.tool() handlers<br/>━━━━━━━━━━<br/>run_cmd · run_python · open_kitchen<br/>close_kitchen · kitchen_status<br/>merge_worktree · clone_repo · …<br/><i>Never raises</i>"]
        OUTER_EXCEPT["● Top-level try/except Exception<br/>━━━━━━━━━━<br/>Safety net: catches anything<br/>not caught by inner handlers<br/>Returns domain-specific JSON envelope"]
    end

    subgraph EXEC_LAYER ["● HEADLESS EXECUTION  (execution/headless.py)"]
        direction TB
        RUN_CORE["● run_headless_core()<br/>━━━━━━━━━━<br/>Full subprocess lifecycle<br/>cmd prep → launch → scan → result"]
        SUBPROCESS["claude CLI subprocess<br/>━━━━━━━━━━<br/>JSONL stdout / Channel B JSONL"]
        BUILD_RESULT["_build_skill_result()<br/>━━━━━━━━━━<br/>Normal path: parse session result<br/>subtype · needs_retry · retry_reason"]
        CRASH_EXCEPT["● except Exception (crash path)<br/>━━━━━━━━━━<br/>Pre-launch or mid-flight exception<br/>flush_session_log → SkillResult.crashed()"]
    end

    subgraph RESULT_TYPES ["● RESULT CONTRACTS  (core/_type_results.py)"]
        direction LR
        SKILL_RESULT_OK["SkillResult<br/>━━━━━━━━━━<br/>success · subtype · exit_code<br/>needs_retry · retry_reason<br/>token_usage · kill_reason"]
        CRASHED_FACTORY["● SkillResult.crashed()<br/>━━━━━━━━━━<br/>subtype='crashed'<br/>exit_code=-1 · is_error=True<br/>needs_retry=False<br/>result = ExcType: message"]
    end

    subgraph SESSION_OBS ["● SESSION OBSERVABILITY  (execution/session_log.py)"]
        direction TB
        FLUSH_LOG["● flush_session_log() on crash<br/>━━━━━━━━━━<br/>Writes crash entry to session dir<br/>termination_reason='CRASHED'<br/>exception_text stored"]
        LOG_DIR["Session Logs (write-only)<br/>━━━━━━━━━━<br/>~/.local/share/autoskillit/logs/<br/>{session-id}/  ·  sessions.jsonl<br/>Query: jq 'select(.success==false)'"]
    end

    subgraph ENFORCEMENT ["● STRUCTURAL GUARDS  (tests/arch/test_never_raises_contracts.py)"]
        direction TB
        AST_TOOL_TEST["● test_all_mcp_tool_handlers_have_except_exception<br/>━━━━━━━━━━<br/>AST walk: every @mcp.tool() fn<br/>must have top-level try/except Exception<br/>(stronger than Never raises docstring check)"]
        AST_HELPERS["● _helpers.py · _has_toplevel_except_exception()<br/>━━━━━━━━━━<br/>Shared AST visitor: verifies<br/>first stmt is Try with ExceptException handler"]
    end

    %% MAIN FLOW %%
    MCP_CLIENT --> RUN_SKILL
    MCP_CLIENT --> OTHER_TOOLS

    RUN_SKILL --> OUTER_EXCEPT
    OTHER_TOOLS --> OUTER_EXCEPT

    RUN_SKILL --> RUN_CORE
    RUN_CORE --> SUBPROCESS
    SUBPROCESS --> BUILD_RESULT
    SUBPROCESS --> CRASH_EXCEPT

    BUILD_RESULT --> SKILL_RESULT_OK
    CRASH_EXCEPT --> FLUSH_LOG
    CRASH_EXCEPT --> CRASHED_FACTORY
    OUTER_EXCEPT --> CRASHED_FACTORY

    FLUSH_LOG --> LOG_DIR

    SKILL_RESULT_OK --> MCP_CLIENT
    CRASHED_FACTORY --> MCP_CLIENT

    AST_HELPERS --> AST_TOOL_TEST

    %% CLASS ASSIGNMENTS %%
    class MCP_CLIENT terminal;
    class RUN_SKILL,OTHER_TOOLS handler;
    class OUTER_EXCEPT detector;
    class RUN_CORE,SUBPROCESS phase;
    class BUILD_RESULT phase;
    class CRASH_EXCEPT detector;
    class SKILL_RESULT_OK stateNode;
    class CRASHED_FACTORY newComponent;
    class FLUSH_LOG handler;
    class LOG_DIR output;
    class AST_TOOL_TEST,AST_HELPERS detector;
```

Closes #811

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260413-193250-102132/.autoskillit/temp/rectify/rectify_headless_core_crash_path_2026-04-13_194500_part_a.md`

## Token Usage Summary

| plan | 215 | 19.5k | 1.3M | 77.0k | 1 | 5m 11s |
| verify | 108 | 14.4k | 637.5k | 59.2k | 1 | 4m 3s |
| **Total** | 323 | 33.8k | 1.9M | 136.2k | | 9m 15s |

### DIFF
--- a/src/autoskillit/core/_type_results.py
+++ b/src/autoskillit/core/_type_results.py
@@ -182,7 +182,7 @@ class SkillResult:
     @classmethod
     def crashed(
         cls,
-        exception: BaseException,
+        exception: Exception,
         skill_command: str = "",
         session_id: str = "",
         order_id: str = "",
--- a/src/autoskillit/core/_type_results.py
+++ b/src/autoskillit/core/_type_results.py
@@ -179,6 +179,33 @@ class SkillResult:
             data["order_id"] = self.order_id
         return json.dumps(data, default=lambda o: o.value if isinstance(o, Enum) else str(o))
 
+    @classmethod
+    def crashed(
+        cls,
+        exception: BaseException,
+        skill_command: str = "",
+        session_id: str = "",
+        order_id: str = "",
+    ) -> SkillResult:
+        """Construct a SkillResult for a runner crash (pre-launch or mid-flight exception).
+
+        Produces the same 13+ field envelope as _build_skill_result, ensuring
+        pipeline orchestrators can route crash responses without schema inspection.
+        """
+        return cls(
+            success=False,
+            result=f"{type(exception).__name__}: {exception}",
+            session_id=session_id,
+            subtype="crashed",
+            is_error=True,
+            exit_code=-1,
+            needs_retry=False,
+            retry_reason=RetryReason.NONE,
+            stderr="",
+            kill_reason=KillReason.NATURAL_EXIT,
+            order_id=order_id,
+        )
+
     @property
     def outcome(self) -> SessionOutcome:
         """Classify this result as SUCCEEDED, RETRIABLE, or FAILED.
--- a/src/autoskillit/core/_type_results.py
+++ b/src/autoskillit/core/_type_results.py
@@ -192,9 +192,12 @@ class SkillResult:
         Produces the same 13+ field envelope as _build_skill_result, ensuring
         pipeline orchestrators can route crash responses without schema inspection.
         """
+        _result = f"{type(exception).__name__}: {exception}"
+        if skill_command:
+            _result += f" | skill_command={skill_command!r}"
         return cls(
             success=False,
-            result=f"{type(exception).__name__}: {exception}",
+            result=_result,
             session_id=session_id,
             subtype="crashed",
             is_error=True,
@@ -202,7 +205,7 @@ class SkillResult:
             needs_retry=False,
             retry_reason=RetryReason.NONE,
             stderr="",
-            kill_reason=KillReason.NATURAL_EXIT,
+            kill_reason=KillReason.EXCEPTION,
             order_id=order_id,
         )
 
--- a/src/autoskillit/execution/headless.py
+++ b/src/autoskillit/execution/headless.py
@@ -1040,6 +1040,8 @@ async def run_headless_core(
             _exc_text = traceback.format_exc()
             _log_dir = ctx.config.linux_tracing.log_dir
             try:
+                # Deferred: autoskillit.execution.__init__ imports headless.py (L39-42);
+                # a top-level import of autoskillit.execution would be circular.
                 from autoskillit.execution import flush_session_log
 
                 flush_session_log(
@@ -1063,6 +1065,11 @@ async def run_headless_core(
             return SkillResult.crashed(
                 exception=exc,
                 skill_command=original_skill_command,
+                order_id=order_id,
+            )
+        if _result is None:
+            return SkillResult.crashed(
+                exception=RuntimeError("runner() did not return a result"),
                 order_id=order_id,
             )
         _elapsed = time.monotonic() - _start_mono
--- a/src/autoskillit/execution/headless.py
+++ b/src/autoskillit/execution/headless.py
@@ -14,6 +14,7 @@ import dataclasses
 import os
 import re
 import time
+import traceback
 from collections.abc import Sequence
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
@@ -1034,32 +1035,36 @@ async def run_headless_core(
                 idle_output_timeout=effective_idle,
                 max_suppression_seconds=cfg.max_suppression_seconds,
             )
-        finally:
-            if _result is None:
-                # Runner raised — write a crash entry so sessions.jsonl stays consistent
-                _log_dir = ctx.config.linux_tracing.log_dir
-                try:
-                    from autoskillit.execution import flush_session_log
+        except Exception as exc:
+            logger.error("run_headless_core runner crashed", exc_info=True)
+            _exc_text = traceback.format_exc()
+            _log_dir = ctx.config.linux_tracing.log_dir
+            try:
+                from autoskillit.execution import flush_session_log
 
-                    flush_session_log(
-                        log_dir=_log_dir,
-                        cwd=str(cwd),
-                        kitchen_id=kitchen_id,
-                        order_id=order_id,
-                        session_id="",
-                        pid=0,
-                        skill_command=skill_command,
-                        success=False,
-                        subtype="crashed",
-                        exit_code=-1,
-                        start_ts=_start_ts,
-                        proc_snapshots=None,
-                        termination_reason="CRASHED",
-                    )
-                except Exception:
-                    logger.debug("flush_session_log during crash failed", exc_info=True)
-        if _result is None:
-            raise RuntimeError("runner() did not return a result — cannot build SkillResult")
+                flush_session_log(
+                    log_dir=_log_dir,
+                    cwd=str(cwd),
+                    kitchen_id=kitchen_id,
+                    order_id=order_id,
+                    session_id="",
+                    pid=0,
+                    skill_command=original_skill_command,
+                    success=False,
+                    subtype="crashed",
+                    exit_code=-1,
+                    start_ts=_start_ts,
+                    proc_snapshots=None,
+                    termination_reason="CRASHED",
+                    exception_text=_exc_text,
+                )
+            except Exception:
+                logger.debug("flush_session_log during crash failed", exc_info=True)
+            return SkillResult.crashed(
+                exception=exc,
+                skill_command=original_skill_command,
+                order_id=order_id,
+            )
         _elapsed = time.monotonic() - _start_mono
         _end_ts = (datetime.fromisoformat(_start_ts) + timedelta(seconds=_elapsed)).isoformat()
         result = dataclasses.replace(  # type: ignore[arg-type]
--- a/src/autoskillit/server/tools_ci.py
+++ b/src/autoskillit/server/tools_ci.py
@@ -67,89 +67,95 @@ async def wait_for_ci(
         "action_required", "timed_out", "no_runs", "error", "unknown"),
         and failed_jobs list. Billing limit errors surface as
         conclusion="action_required" with failed_jobs=[].
+
+    Never raises.
     """
     if (gate := _require_enabled()) is not None:
         return gate
-    structlog.contextvars.clear_contextvars()
-    structlog.contextvars.bind_contextvars(tool="wait_for_ci")
-    logger.info("wait_for_ci", branch=branch, repo=repo or "(infer)")
-
-    from autoskillit.server import _get_ctx
-
-    tool_ctx = _get_ctx()
-    if tool_ctx.ci_watcher is None:
-        return json.dumps(
-            {
-                "run_id": None,
-                "conclusion": "error",
-                "failed_jobs": [],
-                "error": "CI watcher not configured",
-            }
-        )
-
-    # Infer head_sha from cwd if not provided
-    if head_sha is None and cwd:
-        try:
-            rc, stdout, _ = await _run_subprocess(
-                ["git", "rev-parse", "HEAD"], cwd=cwd, timeout=5.0
-            )
-            if rc == 0:
-                head_sha = stdout.strip()
-        except Exception:
-            logger.warning("git rev-parse HEAD failed", exc_info=True)
-
-    scope = CIRunScope(
-        workflow=workflow or tool_ctx.default_ci_scope.workflow,
-        head_sha=head_sha,
-        event=event or tool_ctx.default_ci_scope.event,
-    )
-
-    resolved_repo = await infer_repo_from_remote(cwd, hint=remote_url or repo or None)
-
-    await _notify(
-        ctx,
-        "info",
-        f"Watching CI for branch {branch}",
-        "autoskillit.wait_for_ci",
-        extra={
-            "repo": resolved_repo or "(infer)",
-            "head_sha": scope.head_sha or "(any)",
-            "workflow": scope.workflow or "(any)",
-        },
-    )
-
-    _start = time.monotonic()
     try:
-        result = await tool_ctx.ci_watcher.wait(
-            branch,
-            repo=resolved_repo or None,
-            scope=scope,
-            timeout_seconds=timeout_seconds,
-            cwd=cwd,
+        structlog.contextvars.clear_contextvars()
+        structlog.contextvars.bind_contextvars(tool="wait_for_ci")
+        logger.info("wait_for_ci", branch=branch, repo=repo or "(infer)")
+
+        from autoskillit.server import _get_ctx
+
+        tool_ctx = _get_ctx()
+        if tool_ctx.ci_watcher is None:
+            return json.dumps(
+                {
+                    "run_id": None,
+                    "conclusion": "error",
+                    "failed_jobs": [],
+                    "error": "CI watcher not configured",
+                }
+            )
+
+        # Infer head_sha from cwd if not provided
+        if head_sha is None and cwd:
+            try:
+                rc, stdout, _ = await _run_subprocess(
+                    ["git", "rev-parse", "HEAD"], cwd=cwd, timeout=5.0
+                )
+                if rc == 0:
+                    head_sha = stdout.strip()
+            except Exception:
+                logger.warning("git rev-parse HEAD failed", exc_info=True)
+
+        scope = CIRunScope(
+            workflow=workflow or tool_ctx.default_ci_scope.workflow,
+            head_sha=head_sha,
+            event=event or tool_ctx.default_ci_scope.event,
         )
 
-        # Include head_sha used for this CI check so orchestrators can verify
-        # CI results correspond to the current HEAD after a force-push.
-        if scope.head_sha:
-            result = {**result, "head_sha": scope.head_sha}
+        resolved_repo = await infer_repo_from_remote(cwd, hint=remote_url or repo or None)
 
-        conclusion = result.get("conclusion", "unknown")
-        level = "info" if conclusion == "success" else "error"
         await _notify(
             ctx,
-            level,
-            f"CI result: {conclusion}",
+            "info",
+            f"Watching CI for branch {branch}",
             "autoskillit.wait_for_ci",
-            extra={"run_id": result.get("run_id")},
+            extra={
+                "repo": resolved_repo or "(infer)",
+                "head_sha": scope.head_sha or "(any)",
+                "workflow": scope.workflow or "(any)",
+            },
         )
 
-        return json.dumps(result)
+        _start = time.monotonic()
+        try:
+            result = await tool_ctx.ci_watcher.wait(
+                branch,
+                repo=resolved_repo or None,
+                scope=scope,
+                timeout_seconds=timeout_seconds,
+                cwd=cwd,
+            )
+
+            # Include head_sha used for this CI check so orchestrators can verify
+            # CI results correspond to the current HEAD after a force-push.
+            if scope.head_sha:
+                result = {**result, "head_sha": scope.head_sha}
+
+            conclusion = result.get("conclusion", "unknown")
+            level = "info" if conclusion == "success" else "error"
+            await _notify(
+                ctx,
+                level,
+                f"CI result: {conclusion}",
+                "autoskillit.wait_for_ci",
+                extra={"run_id": result.get("run_id")},
+            )
+
+            return json.dumps(result)
+        except Exception as exc:
+            logger.error("autoskillit.wait_for_ci failed", exc_info=exc)
+            return json.dumps({"success": False, "error": f"{type(exc).__name__}: {exc}"})
+        finally:
+            if step_name:
+                tool_ctx.timing_log.record(step_name, time.monotonic() - _start)
     except Exception as exc:
-        logger.error("autoskillit.wait_for_ci failed", exc_info=exc)
+        logger.error("wait_for_ci unhandled exception", exc_info=True)
         return json.dumps({"success": False, "error": f"{type(exc).__name__}: {exc}"})
-    finally:
-        if step_name:
-            tool_ctx.timing_log.record(step_name, time.monotonic() - _start)
 
 
 @mcp.tool(tags={"autoskillit", "kitchen", "github"}, annotations={"readOnlyHint": True})
@@ -177,6 +183,8 @@ async def set_commit_status(
         target_url: Optional URL linking to review details.
         repo: owner/repo format. Inferred from `cwd` git remote if absent.
         cwd: Working directory for repo inference. Defaults to plugin_dir.
+
+    Never raises.
     """
     if (gate := _require_enabled()) is not None:
         return gate
@@ -193,41 +201,45 @@ async def set_commit_status(
             }
         )
 
-    from autoskillit.server import _get_ctx
+    try:
+        from autoskillit.server import _get_ctx
 
-    tool_ctx = _get_ctx()
-    effective_cwd = cwd or tool_ctx.plugin_dir or "."
+        tool_ctx = _get_ctx()
+        effective_cwd = cwd or tool_ctx.plugin_dir or "."
 
-    # Resolve owner/repo if not provided
-    owner_repo = repo
-    if not owner_repo:
-        owner_repo = await infer_repo_from_remote(effective_cwd)
+        # Resolve owner/repo if not provided
+        owner_repo = repo
         if not owner_repo:
-            return json.dumps(
-                {"success": False, "error": "Could not infer owner/repo from git remote"}
-            )
+            owner_repo = await infer_repo_from_remote(effective_cwd)
+            if not owner_repo:
+                return json.dumps(
+                    {"success": False, "error": "Could not infer owner/repo from git remote"}
+                )
 
-    cmd = [
-        "gh",
-        "api",
-        "--method",
-        "POST",
-        f"/repos/{owner_repo}/statuses/{sha}",
-        "-f",
-        f"state={state}",
-        "-f",
-        f"context={context}",
-        "-f",
-        f"description={description}",
-    ]
-    if target_url:
-        cmd += ["-f", f"target_url={target_url}"]
+        cmd = [
+            "gh",
+            "api",
+            "--method",
+            "POST",
+            f"/repos/{owner_repo}/statuses/{sha}",
+            "-f",
+            f"state={state}",
+            "-f",
+            f"context={context}",
+            "-f",
+            f"description={description}",
+        ]
+        if target_url:
+            cmd += ["-f", f"target_url={target_url}"]
 
-    rc, _stdout, stderr = await _run_subprocess(cmd, cwd=effective_cwd, timeout=30.0)
-    if rc != 0:
-        return json.dumps({"success": False, "error": stderr})
+        rc, _stdout, stderr = await _run_subprocess(cmd, cwd=effective_cwd, timeout=30.0)
+        if rc != 0:
+            return json.dumps({"success": False, "error": stderr})
 
-    return json.dumps({"success": True, "sha": sha, "state": state, "context": context})
+        return json.dumps({"success": True, "sha": sha, "state": state, "context": context})
+    except Exception as exc:
+        logger.error("set_commit_status unhandled exception", exc_info=True)
+        return json.dumps({"success": False, "error": f"{type(exc).__name__}: {exc}"})
 
 
 @mcp.tool(tags={"autoskillit", "kitchen", "ci"}, annotations={"readOnlyHint": True})
@@ -254,31 +266,37 @@ async def get_ci_status(
 
     Returns:
         JSON with runs list, each containing id, status, conclusion, failed_jobs.
+
+    Never raises.
     """
     if (gate := _require_enabled()) is not None:
         return gate
-    from autoskillit.server import _get_ctx
+    try:
+        from autoskillit.server import _get_ctx
 
-    tool_ctx = _get_ctx()
-    if tool_ctx.ci_watcher is None:
-        return json.dumps({"runs": [], "error": "CI watcher not configured"})
+        tool_ctx = _get_ctx()
+        if tool_ctx.ci_watcher is None:
+            return json.dumps({"runs": [], "error": "CI watcher not configured"})
 
-    if branch is None and run_id is None:
-        return json.dumps({"runs": [], "error": "Provide branch or run_id"})
+        if branch is None and run_id is None:
+            return json.dumps({"runs": [], "error": "Provide branch or run_id"})
 
-    scope = CIRunScope(
-        workflow=workflow or tool_ctx.default_ci_scope.workflow,
-        event=event or tool_ctx.default_ci_scope.event,
-    )
+        scope = CIRunScope(
+            workflow=workflow or tool_ctx.default_ci_scope.workflow,
+            event=event or tool_ctx.default_ci_scope.event,
+        )
 
-    result = await tool_ctx.ci_watcher.status(
-        branch or "",
-        repo=repo,
-        run_id=run_id,
-        scope=scope,
-        cwd=cwd,
-    )
-    return json.dumps(result)
+        result = await tool_ctx.ci_watcher.status(
+            branch or "",
+            repo=repo,
+            run_id=run_id,
+            scope=scope,
+            cwd=cwd,
+        )
+        return json.dumps(result)
+    except Exception as exc:
+        logger.error("get_ci_status unhandled exception", exc_info=True)
+        return json.dumps({"runs": [], "error": f"{type(exc).__name__}: {exc}"})
 
 
 @mcp.tool(tags={"autoskillit", "kitchen", "ci"}, annotations={"readOnlyHint": False})
@@ -307,43 +325,48 @@ async def toggle_auto_merge(
     Returns:
         JSON: {"success": bool, "pr_number": int} on success,
               {"success": false, "error": str} on failure.
+
+    Never raises.
     """
     if (gate := _require_enabled()) is not None:
         return gate
-
-    structlog.contextvars.clear_contextvars()
-    structlog.contextvars.bind_contextvars(
-        tool="toggle_auto_merge", pr_number=pr_number, target_branch=target_branch
-    )
-    await _notify(
-        ctx,
-        "info",
-        f"Toggling auto-merge for PR #{pr_number} on {target_branch!r}",
-        "autoskillit.toggle_auto_merge",
-        extra={"pr_number": pr_number, "target_branch": target_branch},
-    )
-
-    from autoskillit.server import _get_ctx
-
-    tool_ctx = _get_ctx()
-
-    if tool_ctx.merge_queue_watcher is None:
-        return json.dumps(
-            {
-                "success": False,
-                "error": "merge_queue_watcher not configured (missing GITHUB_TOKEN?)",
-            }
+    try:
+        structlog.contextvars.clear_contextvars()
+        structlog.contextvars.bind_contextvars(
+            tool="toggle_auto_merge", pr_number=pr_number, target_branch=target_branch
+        )
+        await _notify(
+            ctx,
+            "info",
+            f"Toggling auto-merge for PR #{pr_number} on {target_branch!r}",
+            "autoskillit.toggle_auto_merge",
+            extra={"pr_number": pr_number, "target_branch": target_branch},
         )
 
-    resolved_repo = await infer_repo_from_remote(cwd, hint=remote_url or repo or None)
+        from autoskillit.server import _get_ctx
 
-    result = await tool_ctx.merge_queue_watcher.toggle(
-        pr_number=pr_number,
-        target_branch=target_branch,
-        repo=resolved_repo or None,
-        cwd=cwd,
-    )
-    return json.dumps(result)
+        tool_ctx = _get_ctx()
+
+        if tool_ctx.merge_queue_watcher is None:
+            return json.dumps(
+                {
+                    "success": False,
+                    "error": "merge_queue_watcher not configured (missing GITHUB_TOKEN?)",
+                }
+            )
+
+        resolved_repo = await infer_repo_from_remote(cwd, hint=remote_url or repo or None)
+
+        result = await tool_ctx.merge_queue_watcher.toggle(
+            pr_number=pr_number,
+            target_branch=target_branch,
+            repo=resolved_repo or None,
+            cwd=cwd,
+        )
+        return json.dumps(result)
+    except Exception as exc:
+        logger.error("toggle_auto_merge unhandled exception", exc_info=True)
+        return json.dumps({"success": False, "error": f"{type(exc).__name__}: {exc}"})
 
 
 @mcp.tool(tags={"autoskillit", "kitchen", "ci"}, annotations={"readOnlyHint": False})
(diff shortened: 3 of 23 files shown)


src/autoskillit/execution/headless.py
```diff
@@ -1034,32 +1035,36 @@ async def run_headless_core(
...
+                    exception_text=_exc_text,
+                )
+            except Exception:
+                logger.debug("flush_session_log during crash failed", exc_info=True)
+            return SkillResult.crashed(
+                exception=exc,
+                skill_command=original_skill_command,
+                order_id=order_id,
+            )
         _elapsed = time.monotonic() - _start_mono
```
Trecek, line 1075: [warning] defense: Defensive guard removed — the old `if _result is None: raise RuntimeError('runner() did not return a result — cannot build SkillResult')` sentinel was dropped. If a runner ever exits normally (no exception) without assigning `_result`, the code falls through to `dataclasses.replace(None, …)` at the line below, producing an opaque `TypeError` at a semantically unrelated location instead of a clear `RuntimeError` at the point of invariant violation. The guard was cheap and provided explicit contract enforcement. Restore it as `if _result is None: return SkillResult.crashed(exception=RuntimeError('runner() did not return a result'), order_id=order_id)` after the except block and before `_elapsed`.

src/autoskillit/core/_type_results.py
```diff
@@ -179,6 +179,33 @@ def to_json(self) -> str:
             data["order_id"] = self.order_id
         return json.dumps(data, default=lambda o: o.value if isinstance(o, Enum) else str(o))
 
+    @classmethod
+    def crashed(
+        cls,
+        exception: BaseException,
+        skill_command: str = "",
```
Trecek, line 186: [warning] slop: `skill_command: str = ""` is accepted as a parameter of `SkillResult.crashed()` but is never forwarded to the `cls(...)` constructor — `SkillResult` has no `skill_command` field. Callers such as `headless.py` pass `skill_command=original_skill_command` and `run_skill` passes `skill_command=resolved_command`, but both values are silently dropped and do not appear in the serialised JSON envelope. Remove the parameter or — if preserving the skill_command in the crash result is desired — expose it through the existing `result` field or a dedicated field.

src/autoskillit/core/_type_results.py
```diff
@@ -179,6 +179,33 @@ def to_json(self) -> str:
...
+            success=False,
+            result=f"{type(exception).__name__}: {exception}",
+            session_id=session_id,
+            subtype="crashed",
+            is_error=True,
+            exit_code=-1,
+            needs_retry=False,
+            retry_reason=RetryReason.NONE,
+            stderr="",
+            kill_reason=KillReason.NATURAL_EXIT,
```
Trecek: [warning] arch: `kill_reason=KillReason.NATURAL_EXIT` is semantically wrong for a crash caused by an unhandled exception. `NATURAL_EXIT` implies the process ran to completion without external intervention; a runner that raised is not a natural exit. Downstream consumers that inspect `kill_reason` to distinguish clean exits from crashes will be misled. Use a neutral or crash-specific enum value — e.g. `KillReason.EXCEPTION` if one exists, or at minimum the current `kill_reason` sentinel that the rest of the SkillResult lifecycle uses for unknown/unresolved terminations.

src/autoskillit/server/tools_recipe.py
```diff
@@ -167,27 +173,33 @@ async def load_recipe(name: str, overrides: dict[str, str] | None = None) -> str
...
+            resolved_defaults=_defaults,
+            ingredient_overrides=overrides,
+            temp_dir=tool_ctx.temp_dir,
+            temp_dir_relpath=temp_dir_display_str(tool_ctx.config.workspace.temp_dir),
+        )
+        recipe_info = tool_ctx.recipes.find(name, Path.cwd())
+        return json.dumps(await _apply_triage_gate(result, name, recipe_info=recipe_info))
+    except Exception as exc:
+        logger.error("load_recipe unhandled exception", exc_info=True)
+        return json.dumps({"error": f"{type(exc).__name__}: {exc}"})
```
Trecek: [warning] defense: `load_recipe` outer exception handler returns `{"error": …}` without a `"success"` key, while every other `@mcp.tool()` exception handler added in this PR returns `{"success": False, "error": …}`. Callers checking `result.get("success")` receive `None` (falsy, but not `False`) on this code path; callers that do `result["success"]` receive `KeyError`. Change to `json.dumps({"success": False, "error": f"{type(exc).__name__}: {exc}"})` to match the uniform contract.

src/autoskillit/server/tools_recipe.py
```diff
@@ -41,16 +41,22 @@ async def list_recipes() -> str:
...
+        structlog.contextvars.clear_contextvars()
+        structlog.contextvars.bind_contextvars(tool="list_recipes")
+        tool_ctx = _get_ctx_or_none()
+        if tool_ctx is None or tool_ctx.recipes is None:
+            return json.dumps([])
+        result = tool_ctx.recipes.list_all(Path.cwd())
+        return json.dumps(result)
+    except Exception as exc:
+        logger.error("list_recipes unhandled exception", exc_info=True)
+        return json.dumps({"success": False, "error": f"{type(exc).__name__}: {exc}"})
```
Trecek: [warning] cohesion: `list_recipes` success path returns a JSON array, but the new exception handler at this line returns a JSON object `{"success": False, "error": …}`. Callers expecting an array will fail to handle the error case. The error response should either use the same shape as the empty-list case (`json.dumps([])`) or be a well-known error envelope — but not silently switch schema types on the same code path.

src/autoskillit/server/tools_ci.py
```diff
@@ -67,89 +67,95 @@ async def wait_for_ci(
...
+            )
+
+            return json.dumps(result)
+        except Exception as exc:
+            logger.error("autoskillit.wait_for_ci failed", exc_info=exc)
+            return json.dumps({"success": False, "error": f"{type(exc).__name__}: {exc}"})
+        finally:
+            if step_name:
+                tool_ctx.timing_log.record(step_name, time.monotonic() - _start)
     except Exception as exc:
```
Trecek, line 158: [warning] bugs: If `_get_ctx()` raises before `_start = time.monotonic()` is reached (setup phase failure), the outer `except` catches it and returns an error JSON — but `step_name` timing is silently lost because `timing_log.record` lives only in the inner `finally`, which never executes in this path. This is a new regression: in the original code, `_get_ctx()` failure would propagate as an unhandled exception (no timing issue); now it is silently swallowed without recording the step time. Consider recording a zero-duration timing entry in the outer except block when `step_name` is set.

src/autoskillit/server/tools_ci.py
```diff
@@ -401,57 +424,62 @@ async def wait_for_merge_queue(
...
+                cwd=cwd,
+                timeout_seconds=timeout_seconds,
+                poll_interval=poll_interval,
+                stall_grace_period=stall_grace_period,
+                max_stall_retries=max_stall_retries,
+                not_in_queue_confirmation_cycles=not_in_queue_confirmation_cycles,
+            )
+            return json.dumps(result)
+        except Exception as exc:
+            logger.error("autoskillit.wait_for_merge_queue failed", exc_info=exc)
```
Trecek: [warning] cohesion: Inner except handler logs `"autoskillit.wait_for_merge_queue failed"` (legacy format) while the outer handler at L481 logs `"wait_for_merge_queue unhandled exception"` (new format introduced by this PR). All outer handlers in this PR were normalised to the new format; the pre-existing inner handlers were left inconsistent. Update to `"wait_for_merge_queue ci_watcher error"` or another consistent naming scheme, and align `exc_info=exc` → `exc_info=True` to match the outer handler.

Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (12 blocking, 1 decision, 8 info findings)

src/autoskillit/server/tools_ci.py
```diff
@@ -483,35 +511,52 @@ async def check_repo_merge_state(
...
+            owner, repo = resolved_repo.split("/", 1)
+            state = await fetch_repo_merge_state(
+                owner=owner,
+                repo=repo,
+                branch=branch,
+                token=tool_ctx.config.github.token,
+            )
+            return json.dumps(state)
+        except Exception as exc:
+            logger.error("autoskillit.check_repo_merge_state failed", exc_info=exc)
```
Trecek: [warning] cohesion: Inner except handler logs `"autoskillit.check_repo_merge_state failed"` (legacy format) while the outer handler at L559 logs `"check_repo_merge_state unhandled exception"` (new format). Same inconsistency as wait_for_merge_queue above — the inner handlers were not normalised when the outer handlers were added.

src/autoskillit/server/tools_kitchen.py
```diff
@@ -245,168 +245,184 @@ async def open_kitchen(
...
+    """
+    try:
+        if (h := _require_not_headless("close_kitchen")) is not None:
+            return h
+        _close_kitchen_handler()
+        await ctx.reset_visibility()
+        return "Kitchen is closed."
+    except Exception as exc:
+        logger.error("close_kitchen unhandled exception", exc_info=True)
+        return json.dumps({"error": f"{type(exc).__name__}: {exc}"})
```
Trecek: [warning] cohesion: `close_kitchen` success path returns a plain string `"Kitchen is closed."` but the new exception handler returns JSON `{"error": …}`. The response type is asymmetric: callers or hooks that try to parse the error path as JSON will succeed, but callers that always expect plain text will break on the error path. Consider returning `json.dumps({"success": False, "error": …})` in the except block, or wrapping the success path as `json.dumps({"success": True, "message": "Kitchen is closed."})` — whichever matches the documented contract.

src/autoskillit/server/tools_git.py
```diff
@@ -41,49 +41,55 @@ async def merge_worktree(
...
+                    extra={"reason": result["error"]},
+                )
+
+            return json.dumps(result)
+        finally:
+            if step_name:
+                tool_ctx.timing_log.record(step_name, time.monotonic() - _start)
+    except Exception as exc:
+        logger.error("merge_worktree unhandled exception", exc_info=True)
+        return json.dumps({"error": f"{type(exc).__name__}: {exc}"})
```
Trecek: [warning] cohesion: `merge_worktree` outer exception handler returns `{"error": …}` without `"success": False`. All peer git tools added in this PR (`classify_fix` L234, `create_unique_branch` L381, `check_pr_mergeable` L446) return `{"success": False, "error": …}`. Add `"success": False` to make the error response consistent with its siblings.

tests/server/test_tools_load_recipe.py
```diff
@@ -319,8 +319,7 @@ async def test_unexpected_exception_returns_structured_error(
             side_effect=AttributeError("programming error"),
         ):
             result = json.loads(await load_recipe(name="test"))
-        assert result["success"] is False
-        assert result["subtype"] == "tool_exception"
+        assert "error" in result
```
Trecek, line 323: [warning] tests: `assert result["success"] is False` was removed because `load_recipe`'s outer except now returns `{"error": …}` without a `success` field. This weakening is a symptom of the bug in `tools_recipe.py:202` (missing `success=False`). Once that handler is fixed to return `{"success": False, "error": …}`, this test should restore the `assert result["success"] is False` assertion. As written, a response that accidentally returns `{"error": ""}` (empty message) would still pass.

tests/execution/test_headless.py
```diff
@@ -2148,13 +2150,63 @@ async def raising_runner(*args: object, **kwargs: object) -> None:
...
+        async def raising_runner(*args: object, **kwargs: object) -> None:
+            raise ValueError("bad input")
+
+        tool_ctx.runner = raising_runner  # type: ignore[assignment]
+
+        with patch("autoskillit.execution.headless.logger") as mock_logger:
+            await run_headless_core("/investigate test", cwd=str(tmp_path), ctx=tool_ctx)
+            mock_logger.error.assert_called_once()
+            call_kwargs = mock_logger.error.call_args
+            assert call_kwargs[1].get("exc_info") is not None
```
Trecek: [warning] tests: `assert call_kwargs[1].get('exc_info') is not None` only checks presence/non-None, not truthiness. A call like `logger.error('msg', exc_info=False)` (explicitly disabled) would pass this assertion since `False is not None`. Change to `assert call_kwargs[1].get('exc_info')` (truthy check) to verify that exc_info was actually enabled, which is the semantically correct assertion for this test.

src/autoskillit/execution/headless.py
```diff
@@ -1034,32 +1035,36 @@ async def run_headless_core(
...
-                except Exception:
-                    logger.debug("flush_session_log during crash failed", exc_info=True)
-        if _result is None:
-            raise RuntimeError("runner() did not return a result — cannot build SkillResult")
+        except Exception as exc:
+            logger.error("run_headless_core runner crashed", exc_info=True)
+            _exc_text = traceback.format_exc()
+            _log_dir = ctx.config.linux_tracing.log_dir
+            try:
+                from autoskillit.execution import flush_session_log
```
Trecek, line 1045: [warning] arch: `from autoskillit.execution import flush_session_log` is imported inside the `except` block rather than at the top of the file. Both modules are L1 (execution layer), so there is no layer violation — but lazy imports inside exception paths are unusual and can obscure real circular-import issues. If the deferred import exists to break a circular dependency, that dependency should be documented. If there is no circular dependency, move the import to the top of the file.

src/autoskillit/core/_type_results.py
```diff
@@ -179,6 +179,33 @@ def to_json(self) -> str:
             data["order_id"] = self.order_id
         return json.dumps(data, default=lambda o: o.value if isinstance(o, Enum) else str(o))
 
+    @classmethod
+    def crashed(
+        cls,
+        exception: BaseException,
```
Trecek: [info] defense: `exception: BaseException` is broader than necessary — all call sites (`headless.py`, `run_skill`) catch `Exception` and pass it here. Accepting `BaseException` permits accidental use with `KeyboardInterrupt` or `SystemExit`, which should propagate rather than be silenced into a `SkillResult`. Narrowing to `Exception` tightens the public API contract with no behavioural change for current callers.

src/autoskillit/core/_type_results.py
```diff
@@ -179,6 +179,33 @@ def to_json(self) -> str:
...
+        skill_command: str = "",
+        session_id: str = "",
+        order_id: str = "",
+    ) -> SkillResult:
+        """Construct a SkillResult for a runner crash (pre-launch or mid-flight exception).
+
+        Produces the same 13+ field envelope as _build_skill_result, ensuring
```
Trecek, line 192: [info] slop: `"13+ field envelope"` is a vague quantifier that reads as AI filler. `SkillResult` has a fixed, known number of fields — use the actual count or drop the quantifier entirely ("the same field envelope as `_build_skill_result`").

src/autoskillit/server/tools_ci.py
```diff
@@ -67,89 +67,95 @@ async def wait_for_ci(
...
+                ctx,
+                level,
+                f"CI result: {conclusion}",
+                "autoskillit.wait_for_ci",
+                extra={"run_id": result.get("run_id")},
+            )
+
+            return json.dumps(result)
+        except Exception as exc:
+            logger.error("autoskillit.wait_for_ci failed", exc_info=exc)
```
Trecek: [info] cohesion: Inner except uses `exc_info=exc` (passing exception object directly) while outer except at L157 uses `exc_info=True`. Both are valid in structlog, but within the same function the inconsistency is confusing and could lead to copy-paste errors in future handlers. Standardise to `exc_info=True` in both handlers.

src/autoskillit/server/tools_execution.py
```diff
@@ -212,119 +236,134 @@ async def run_skill(
...
+                skill_command=resolved_command,
+                order_id=order_id,
+            ).to_json()
+        finally:
+            if step_name:
+                tool_ctx.timing_log.record(step_name, time.monotonic() - _start, order_id=order_id)
+    except Exception as exc:
+        logger.error("run_skill unhandled exception", exc_info=True)
+        return SkillResult.crashed(
+            exception=exc,
```
Trecek, line 366: [info] arch: The `SkillResult.crashed(...)` call in the outer except block (L365–L369) has 4 extra spaces of indentation compared to the inner except block at L355–L359 and relative to the `except Exception as exc:` at L363. Pre-commit ruff-format will fix this, but it signals the block was not re-checked after writing.

tests/execution/test_headless.py
```diff
@@ -2148,13 +2150,63 @@ async def raising_runner(*args: object, **kwargs: object) -> None:
...
+            flushed.append(dict(kwargs))
+
+        monkeypatch.setattr("autoskillit.execution.flush_session_log", fake_flush)
+
+        async def raising_runner(*args: object, **kwargs: object) -> None:
+            raise OSError("disk full")
+
+        tool_ctx.runner = raising_runner  # type: ignore[assignment]
+
+        await run_headless_core("/investigate test", cwd=str(tmp_path), ctx=tool_ctx)
```
Trecek, line 2185: [info] tests: `test_crash_exception_text_passed_to_flush` does not capture or assert the return value of `run_headless_core`. Since the PR's primary behavioral change is that the crash path returns `SkillResult.crashed()` instead of raising, the test only validates a side effect (flush call) and does not verify the new return-value contract. Add `result = await run_headless_core(...)` with `assert isinstance(result, SkillResult)` and `assert result.success is False`.

tests/execution/test_headless.py
```diff
@@ -2148,13 +2150,63 @@ async def raising_runner(*args: object, **kwargs: object) -> None:
...
+        assert len(crash_calls) >= 1
+        assert "exception_text" in crash_calls[0]
+        assert "OSError: disk full" in crash_calls[0]["exception_text"]
+
+    @pytest.mark.anyio
+    async def test_crash_logs_exception_with_logger_error(self, monkeypatch, tool_ctx, tmp_path):
+        """Runner crash is logged at ERROR level with exc_info."""
+        from autoskillit.execution.headless import run_headless_core
+
+        monkeypatch.setattr("autoskillit.execution.flush_session_log", lambda **kw: None)
```
Trecek, line 2197: [info] tests: `test_crash_logs_exception_with_logger_error` similarly omits any assertion on the return value of `run_headless_core`. The crash path now returns `SkillResult.crashed()`, but the test only asserts on logger side effects. Capture the result and assert `result.success is False` to confirm the function returns correctly through the logger path.

tests/core/test_types.py
```diff
@@ -375,3 +375,31 @@ def test_pr_state_enum_members_are_locked():
...
+        assert result.subtype == "crashed"
+        assert result.is_error is True
+        assert result.exit_code == -1
+        assert result.needs_retry is False
+        assert result.retry_reason == RetryReason.NONE
+        assert "RuntimeError: boom" in result.result
+        assert result.session_id == ""
+        assert result.stderr == ""
+
+    def test_crashed_to_json_produces_valid_envelope(self):
```
Trecek, line 396: [info] tests: `test_crashed_to_json_produces_valid_envelope` checks for `needs_retry`, `session_id`, and `subtype` keys in the serialised JSON but does not assert `data['success'] is False` or `data['exit_code'] == -1`. If `to_json()` silently dropped those fields, the test would not catch it. Add `assert data['success'] is False` and `assert data['exit_code'] == -1` to fully validate the crash envelope.

tests/server/test_tools_execution.py
```diff
@@ -1498,3 +1498,27 @@ async def test_run_skill_idle_output_timeout_defaults_to_none(tool_ctx, monkeypa
...
+    class ExplodingExecutor:
+        async def run(self, *args, **kwargs) -> SkillResult:
+            raise RuntimeError("unexpected executor failure")
+
+    tool_ctx.executor = ExplodingExecutor()
+    monkeypatch.setattr("autoskillit.server._ctx", tool_ctx)
+
+    from autoskillit.server.tools_execution import run_skill
+
+    result_json = await run_skill("/test cmd", "/tmp")
```
Trecek: [info] tests: `cwd="/tmp"` is a hardcoded real filesystem path. Prefer `str(tmp_path)` (the `tmp_path` fixture is already in scope) to make the test fully hermetic and avoid any side effects on the real `/tmp` directory.

Trecek (commented):
AutoSkillit review found 12 blocking issues. See inline comments.

## Summary

| Category | Count |
|----------|-------|
| Blocking (warning) | 12 |
| Decision (requires human) | 1 |
| Info | 8 |

## Blocking Issues

**Defense (2)**
- `headless.py:1068` — Removed defensive guard for `_result is None` (implicit TypeError vs. explicit RuntimeError)
- `tools_recipe.py:202` — `load_recipe` error response missing `"success": False` (all other tools include it)

**Slop/dead code (2)**
- `_type_results.py:186` — `skill_command` param accepted by `SkillResult.crashed()` but never forwarded to the dataclass constructor (silently dropped)
- `_type_results.py:205` — `kill_reason=KillReason.NATURAL_EXIT` semantically wrong for an exception crash

**Cohesion (5)**
- `tools_recipe.py:59` — `list_recipes` error response switches schema type (array → object)
- `tools_ci.py:475,545` — Inner except handlers retain legacy log message format inconsistent with outer handlers added in this PR
- `tools_kitchen.py:428` — `close_kitchen` error path returns JSON but success path returns plain string
- `tools_git.py:92` — `merge_worktree` outer except missing `"success": False` (all sibling git tools include it)

**Bugs (1)**
- `tools_ci.py:156` — `step_name` timing silently lost when outer try fails before `_start` is set

**Tests (2)**
- `test_tools_load_recipe.py:322` — Test weakened: `assert result['success'] is False` removed (fix the handler at `tools_recipe.py:202` and restore)
- `test_headless.py:2208` — `is not None` should be truthy check for `exc_info`

## Decision Finding (1)
- `headless.py:1043` — Deferred `from autoskillit.execution import flush_session_log` inside the except block: no layer violation (both L1), but unclear whether this is intentional for circular-import avoidance or can be moved to module scope

## Deletion Regression
None found (no files or symbols deleted from `integration` since branch point).

src/autoskillit/core/_type_results.py
Trecek: Agreed — fixed in 854d0c7d. All call sites catch Exception, not BaseException; narrowed to prevent misuse with KeyboardInterrupt/SystemExit.

src/autoskillit/server/tools_ci.py
Trecek: Agreed — fixed in 6ae08ea9. Harmonized inner except to use `exc_info=True` and dropped legacy `autoskillit.` prefix, matching outer handler and sibling functions fixed in 14f0584d.

tests/execution/test_headless.py
Trecek: Agreed — fixed in 53947587. Added `assert result.success is False` to lock in the SkillResult.crashed() return contract.

tests/server/test_tools_execution.py
Trecek: Agreed — fixed in c934abe3. Added `tmp_path` fixture and replaced hardcoded `/tmp` with `str(tmp_path)` for hermetic test isolation.

src/autoskillit/execution/headless.py
Trecek: Investigated — this is already fixed. Commit 30af2834 restored the `_result is None` guard. The current code at L1070 has the five-line `if _result is None: return SkillResult.crashed(...)` sentinel.

src/autoskillit/execution/headless.py
Trecek: Investigated — this is intentional. Commit 30af2834 added an explanatory comment: `autoskillit.execution.__init__` imports `headless.py` at L39-42, so a top-level import of `autoskillit.execution` in headless.py would be circular.

src/autoskillit/core/_type_results.py
Trecek: Investigated — this is already fixed. Commit 0d21762f forwards `skill_command` by appending it to the result string (L196-197: `if skill_command: _result += f' | skill_command={skill_command!r}'`). The value is not silently dropped.

src/autoskillit/core/_type_results.py
Trecek: Investigated — this is already fixed. Commit 0d21762f added `KillReason.EXCEPTION` to `_type_enums.py` and updated `crashed()` to use it (L208). The `NATURAL_EXIT` concern no longer applies.

src/autoskillit/server/tools_recipe.py
Trecek: Investigated — this is already fixed. Commit 93b7f41c updated `load_recipe` outer except to include `"success": False` in the error response.

src/autoskillit/server/tools_recipe.py
Trecek: Investigated — this is already fixed. Commit 93b7f41c changed the `list_recipes` exception handler to return `json.dumps([])`, preserving the array contract on the error path.

src/autoskillit/server/tools_ci.py
Trecek: Investigated — this is already fixed. Commit 14f0584d moved `_start = time.monotonic()` before `_get_ctx()` and added timing recording in the outer except. The timing gap no longer exists.

src/autoskillit/server/tools_ci.py
Trecek: Investigated — this is already fixed. Commit 14f0584d harmonized `wait_for_merge_queue` inner/outer handlers to use consistent format and `exc_info=True`.

src/autoskillit/server/tools_kitchen.py
Trecek: Investigated — this is already fixed. Commit 0df48964 updated `close_kitchen` exception handler to return `{"success": False, "error": ...}`.

src/autoskillit/server/tools_git.py
Trecek: Investigated — this is already fixed. Commit 0df48964 updated `merge_worktree` outer except to include `"success": False`, matching peer git tools.

tests/server/test_tools_load_recipe.py
Trecek: Investigated — this is already fixed. Commit 1b2d4538 restored `assert result["success"] is False` after the production handler fix.

tests/execution/test_headless.py
Trecek: Investigated — this is already fixed. Commit 1b2d4538 changed the assertion to `assert call_kwargs[1].get("exc_info")` (truthy check).

PR #918, opened 2026-04-14
Rectify: Merge Queue Watcher Inconclusive Budget Short-Circuits Outer Timeout

### LINKED ISSUE #911: Merge queue watcher inconclusive budget short-circuits outer timeout
## Summary

`wait_for_merge_queue` returns `success: False, pr_state: timeout, reason: "Inconclusive after 5 retries: checks still running"` after ~75 seconds, even though the outer `timeout_seconds` deadline (600s default, 900s in recipes) has not been reached. The inconclusive retry budget (5 retries × 15s poll interval) short-circuits the outer timeout when CI legitimately takes multiple minutes.

Observed on PR #909 (`doctor-destroys-plugin-cache-startup-race-regression-dual-re/902` → `integration`). The `Test on ubuntu-latest` job takes ~4.5 minutes, but the watcher gives up after ~105 seconds total (30s confirmation window + 75s inconclusive polling).

## Root Cause

Commit `86bbbcbb` (PR #813, "Rectify: Merge Queue Classifier — Positive-Signal Immunity") correctly replaced dangerous negative-inference ejection with a positive-signal-only classifier. As a safety valve, it introduced `ClassifierInconclusive` with a bounded retry budget (`max_inconclusive_retries=5`). However, the budget conflates two fundamentally different conditions:

- **"checks still running"** (`checks_state in {"PENDING", "EXPECTED"}`) — a legitimate, expected transient state that should poll until the outer `timeout_seconds` deadline
- **"no positive signal matched"** — a genuinely ambiguous state that correctly needs a bounded retry budget

Both consume the same 5-retry budget, meaning "checks still running" can never survive longer than ~75 seconds regardless of `timeout_seconds`.

### Failure chain

1. PR enters the merge queue. Auto-merge is enabled.
2. The merge queue creates a readonly-queue branch and runs CI.
3. The watcher polls and sees `in_queue=True` — loops normally.
4. PR exits the queue (or watcher observes `in_queue=False` transiently).
5. Confirmation window: 2 cycles × 15s = 30s of not-in-queue before acting.
6. `_classify_pr_state` sees `checks_state="PENDING"`, raises `ClassifierInconclusive("checks still running")`.
7. `inconclusive_count` increments from 0→5 over 5 cycles (~75s).
8. At `inconclusive_count >= 5`: returns `{success: False, pr_state: "timeout", reason: "Inconclusive after 5 retries: checks still running"}`.
9. Total elapsed: ~105 seconds — well under the 600s outer deadline.

## Affected Components

| File | Role |
|------|------|
| `src/autoskillit/execution/merge_queue.py` | `DefaultMergeQueueWatcher.wait()` inconclusive budget loop (lines 366-380); `_classify_pr_state()` raise sites (lines 232-235) |
| `src/autoskillit/server/_factory.py:239` | Instantiates `DefaultMergeQueueWatcher(token_factory)` with no `max_inconclusive_retries` override |
| `src/autoskillit/server/tools_ci.py` | `wait_for_merge_queue` tool handler — does not expose `max_inconclusive_retries` as a parameter |
| `src/autoskillit/core/_type_protocols.py` | `MergeQueueWatcher.wait()` Protocol — does not include `max_inconclusive_retries` in its signature |

### Bug 1: Budget conflation (primary)

In `merge_queue.py`, the `except ClassifierInconclusive` handler (lines 366-380) increments `inconclusive_count` regardless of the exception's `reason`. When `reason == "checks still running"`, this is incorrect — CI is legitimately in progress and the watcher should continue polling until `timeout_seconds` expires.

```python
# Current behavior (merge_queue.py lines 366-380):
except ClassifierInconclusive as exc:
    if not_in_queue_cycles < not_in_queue_confirmation_cycles:
        await asyncio.sleep(poll_interval)
        continue
    inconclusive_count += 1  # <-- incremented for BOTH "checks still running" AND "no positive signal"
    if inconclusive_count >= self._max_inconclusive_retries:
        return _make_result(
            False, PRState.TIMEOUT,
            f"Inconclusive after {self._max_inconclusive_retries} retries: {exc.reason}",
        )
    await asyncio.sleep(poll_interval)
    continue
```

### Bug 2: Non-resetting `inconclusive_count`

`inconclusive_count` is strictly monotone — it never resets. When the PR re-enters the queue, `not_in_queue_cycles` resets to 0 (line 355) but `inconclusive_count` is left untouched. This means queue oscillation (in → out → in → out) bleeds earlier inconclusive counts into later phases, accelerating the timeout.

For example:
- Phase 1 out-of-queue: 2 inconclusive cycles → `inconclusive_count = 2`
- Re-enters queue → `not_in_queue_cycles = 0`, but `inconclusive_count` stays at 2
- Phase 2 out-of-queue: only 3 more inconclusive cycles needed to hit the budget

## Historical Context

This is a **recurring pattern** — the merge queue classifier has been modified 5+ times:

| Commit | PR | Fix |
|--------|-----|-----|
| `d0b75ab3` | #422 | False ejection from race condition between two API calls |
| `24aa2c58` | #519 | Negative-inference ejection when CI was PENDING — added first PENDING guard |
| `03ff0c78` | #628 | Queue ejection loop — CI failure vs conflict not distinguished |
| `86bbbcbb` | #813 | Replaced negative inference with positive-signal classifier + inconclusive budget (**introduced current bug**) |
| `c9f7e1ef` | #818 | Pre-queue routing block consolidation |

### Primary: Exempt "checks still running" from the inconclusive budget

When `ClassifierInconclusive.reason == "checks still running"`, do **not** increment `inconclusive_count`. Instead, just `await asyncio.sleep(poll_interval)` and continue — respecting only the outer `timeout_seconds` deadline. Reserve the bounded budget exclusively for the "no positive signal matched" catch-all path.

This can be done by checking `exc.reason` in the handler, or by splitting `ClassifierInconclusive` into two distinct exception types (e.g., `ChecksStillRunning` vs `NoPositiveSignal`).

### Secondary: Reset `inconclusive_count` on queue re-entry

Add `inconclusive_count = 0` alongside the existing `not_in_queue_cycles = 0` reset at line 355 when `in_queue` becomes True.

### Optional: Expose `max_inconclusive_retries` as a tool parameter

Currently hardcoded at construction in `_factory.py` and not exposed through the tool handler or Protocol. If the "no positive signal" path retains a budget, recipes should be able to tune it via the tool's `with:` block.

## Contrast with CI Watcher

The CI watcher (`execution/ci.py`) handles "still running" correctly — it polls until the run's `status == "completed"` or the outer deadline expires. There is no intermediate budget that short-circuits the timeout. The merge queue watcher should match this pattern for the "checks still running" case.

Trecek:
## Summary

`DefaultMergeQueueWatcher.wait()` used a single exception class (`ClassifierInconclusive`) as a control-flow signal for two semantically incompatible situations: (1) CI checks are still running — an *expected transient* state that should poll until the outer `timeout_seconds` deadline — and (2) no positive classification signal matched at all — a *genuinely ambiguous* state that warrants a bounded retry ceiling before giving up. Both raised the same exception and incremented the same `inconclusive_count` counter, so CI that took longer than `max_inconclusive_retries × poll_interval` (default: 75 seconds) received a `timeout` result before the outer deadline was reached.

The fix is a **type-level distinction**: `ClassifierInconclusive` is split into `CIStillRunning` (transient — no budget consumed) and `NoPositiveSignal` (ambiguous — bounded budget), caught separately in `wait()`. Additionally, `inconclusive_count` resets on queue re-entry so budget from a prior exit episode does not bleed into subsequent ones.

Closes #911

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260414-074614-472907/.autoskillit/temp/rectify/rectify_merge_queue_inconclusive_budget_2026-04-14_083000_part_a.md`

## Token Usage Summary

| investigate | 120 | 8.4k | 389.7k | 40.7k | 1 | 5m 44s |
| rectify | 5.1k | 24.2k | 1.0M | 70.4k | 1 | 9m 13s |
| review | 98 | 8.2k | 346.5k | 39.7k | 1 | 2m 38s |
| dry_walkthrough | 308 | 28.1k | 1.7M | 105.3k | 2 | 8m 31s |
| implement | 492 | 21.1k | 2.6M | 94.0k | 2 | 6m 37s |
| assess | 226 | 8.5k | 1.2M | 42.8k | 1 | 10m 40s |
| prepare_pr | 52 | 5.6k | 145.0k | 25.7k | 1 | 1m 23s |
| run_arch_lenses | 261 | 24.6k | 1.1M | 163.1k | 3 | 7m 6s |
| compose_pr | 51 | 3.4k | 134.0k | 16.2k | 1 | 1m 5s |
| **Total** | 6.7k | 132.1k | 8.7M | 598.0k | | 53m 0s |

### DIFF
--- a/src/autoskillit/execution/merge_queue.py
+++ b/src/autoskillit/execution/merge_queue.py
@@ -171,6 +171,21 @@ class ClassifierInconclusive(Exception):
         self.reason = reason
 
 
+class CIStillRunning(ClassifierInconclusive):
+    """CI checks are legitimately in-progress (PENDING or EXPECTED).
+
+    Expected transient state — outer timeout_seconds bounds the wait.
+    Must NOT consume the inconclusive budget.
+    """
+
+
+class NoPositiveSignal(ClassifierInconclusive):
+    """No positive classifier gate matched — state is genuinely ambiguous.
+
+    Bounded by max_inconclusive_retries. Counts against the inconclusive budget.
+    """
+
+
 def _is_positive_stall(state: PRFetchState) -> bool:
     """True when auto-merge is enabled and merge_state_status indicates the PR is
     stuck in a state where it should be in queue but is not."""
@@ -230,9 +245,9 @@ def _classify_pr_state(state: PRFetchState) -> ClassificationResult:
         return ClassificationResult(PRState.DROPPED_HEALTHY, "PR healthy but auto_merge cleared")
 
     if state["checks_state"] in {"PENDING", "EXPECTED"}:
-        raise ClassifierInconclusive(state, "checks still running")
+        raise CIStillRunning(state, "checks still running")
 
-    raise ClassifierInconclusive(state, "no positive signal matched")
+    raise NoPositiveSignal(state, "no positive signal matched")
 
 
 class DefaultMergeQueueWatcher:
@@ -353,6 +368,7 @@ class DefaultMergeQueueWatcher:
             # In queue: reset window and continue
             if state["in_queue"]:
                 not_in_queue_cycles = 0
+                inconclusive_count = 0
                 if state["queue_state"] == "UNMERGEABLE":
                     return _make_result(False, PRState.EJECTED, "PR is UNMERGEABLE in merge queue")
                 await asyncio.sleep(poll_interval)
@@ -363,8 +379,10 @@ class DefaultMergeQueueWatcher:
             # Classify state using positive-signal gates
             try:
                 classification = _classify_pr_state(state)
-            except ClassifierInconclusive as exc:
-                # Apply confirmation window before consuming inconclusive budget
+            except CIStillRunning:
+                await asyncio.sleep(poll_interval)
+                continue
+            except NoPositiveSignal as exc:
                 if not_in_queue_cycles < not_in_queue_confirmation_cycles:
                     await asyncio.sleep(poll_interval)
                     continue
--- a/src/autoskillit/execution/merge_queue.py
+++ b/src/autoskillit/execution/merge_queue.py
@@ -301,6 +301,7 @@ class DefaultMergeQueueWatcher:
         stall_grace_period: int = 60,
         max_stall_retries: int = 3,
         not_in_queue_confirmation_cycles: int = 2,
+        max_inconclusive_retries: int = 5,
     ) -> dict[str, Any]:
         """Poll until PR is merged, ejected, stalled, dropped, or timeout expires.
 
@@ -319,6 +320,8 @@ class DefaultMergeQueueWatcher:
             not_in_queue_confirmation_cycles: Consecutive "not in queue" cycles required
                 before acting on absence. Guards against race between queue exit and
                 merged=true propagation (default 2).
+            max_inconclusive_retries: Maximum NoPositiveSignal cycles (beyond the
+                confirmation window) before returning pr_state="timeout" (default 5).
 
         Returns:
             {
@@ -387,12 +390,11 @@ class DefaultMergeQueueWatcher:
                     await asyncio.sleep(poll_interval)
                     continue
                 inconclusive_count += 1
-                if inconclusive_count >= self._max_inconclusive_retries:
+                if inconclusive_count >= max_inconclusive_retries:
                     return _make_result(
                         False,
                         PRState.TIMEOUT,
-                        f"Inconclusive after {self._max_inconclusive_retries} retries:"
-                        f" {exc.reason}",
+                        f"Inconclusive after {max_inconclusive_retries} retries: {exc.reason}",
                     )
                 await asyncio.sleep(poll_interval)
                 continue
--- a/tests/execution/test_merge_queue.py
+++ b/tests/execution/test_merge_queue.py
@@ -977,6 +977,76 @@ class TestInconclusiveBudget:
         assert result["pr_state"] == "merged"
         assert call_count == 8
 
+    @pytest.mark.anyio
+    async def test_wait_accepts_max_inconclusive_retries_per_call_param(self):
+        """max_inconclusive_retries must be a wait() parameter, not constructor-only.
+        Passing it must not raise TypeError. Passing a small value (1) must cause
+        budget exhaustion after 1 NoPositiveSignal cycle beyond the confirmation window.
+        """
+        watcher = _make_watcher()
+        # State: unknown — no positive classifier fires → NoPositiveSignal
+        watcher._fetch_pr_and_queue_state = AsyncMock(  # type: ignore[method-assign]
+            return_value=_queue_state(
+                merged=False,
+                in_queue=False,
+                checks_state="SUCCESS",
+                mergeable="UNKNOWN",
+                merge_state_status="UNKNOWN",
+                auto_merge_enabled_at=None,
+            )
+        )
+        with patch("autoskillit.execution.merge_queue.asyncio.sleep", new_callable=AsyncMock):
+            result = await watcher.wait(
+                pr_number=42,
+                target_branch="main",
+                repo="owner/repo",
+                poll_interval=1,
+                timeout_seconds=99999,
+                max_inconclusive_retries=1,  # ← per-call override
+            )
+
+        assert result["success"] is False
+        assert result["pr_state"] == "timeout"
+        assert "Inconclusive after 1 retries" in result["reason"]
+
+    @pytest.mark.anyio
+    async def test_per_call_max_inconclusive_retries_overrides_constructor_default(self):
+        """Per-call value must take precedence over the constructor default of 5.
+        budget=2 per call, constructor default would be 5. Budget exhaustion must
+        occur after 2 ambiguous cycles (+ confirmation window), not 5.
+        """
+        watcher = _make_watcher()  # constructor default = 5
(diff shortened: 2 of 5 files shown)


src/autoskillit/execution/merge_queue.py
```diff
@@ -363,18 +382,19 @@ def _make_result(
             # Classify state using positive-signal gates
             try:
                 classification = _classify_pr_state(state)
-            except ClassifierInconclusive as exc:
-                # Apply confirmation window before consuming inconclusive budget
+            except CIStillRunning:
+                await asyncio.sleep(poll_interval)
+                continue
+            except NoPositiveSignal as exc:
                 if not_in_queue_cycles < not_in_queue_confirmation_cycles:
```
Trecek, line 389: [critical] bugs: The confirmation window guard `not_in_queue_cycles < not_in_queue_confirmation_cycles` inside the NoPositiveSignal handler reads `not_in_queue_cycles`, but this variable is not incremented anywhere in the visible diff. If `not_in_queue_cycles` is incremented after the `continue` statements (i.e., in code after the try/except block), any `continue` inside the except will bypass the increment — permanently freezing the window counter and preventing `inconclusive_count` from ever advancing. Verify that `not_in_queue_cycles` is incremented before the try/except block, not after, so confirmation window cycles correctly accumulate across all NoPositiveSignal iterations.

src/autoskillit/execution/merge_queue.py
```diff
@@ -353,6 +371,7 @@ def _make_result(
             # In queue: reset window and continue
             if state["in_queue"]:
                 not_in_queue_cycles = 0
+                inconclusive_count = 0
```
Trecek, line 374: [warning] cohesion: `inconclusive_count` is reset to 0 on queue re-entry (L374), but is NOT reset when `CIStillRunning` is caught (L385-L387). If CIStillRunning and NoPositiveSignal cycles interleave, the inconclusive budget could be partially consumed before any NoPositiveSignal handling, potentially causing premature timeout. Consider whether `CIStillRunning` should also reset `inconclusive_count`, or document why it should not.

src/autoskillit/execution/merge_queue.py
```diff
@@ -286,6 +301,7 @@ async def wait(
         stall_grace_period: int = 60,
         max_stall_retries: int = 3,
         not_in_queue_confirmation_cycles: int = 2,
+        max_inconclusive_retries: int = 5,
```
Trecek, line 304: [info] cohesion: The protocol adds `max_inconclusive_retries` to `wait()` as a per-call parameter, but tests construct `DefaultMergeQueueWatcher(token=None, max_inconclusive_retries=3)` — implying the constructor also accepts it. If this is now a per-call wait() parameter, the constructor-level parameter may be redundant. Verify the constructor signature is consistent with the intended design.

tests/execution/test_merge_queue.py
```diff
@@ -792,6 +792,262 @@ def test_implements_merge_queue_watcher_protocol(self):
...
+        assert result["pr_state"] == "merged"
+        assert call_count == 8
+
+    @pytest.mark.anyio
+    async def test_wait_accepts_max_inconclusive_retries_per_call_param(self):
+        """max_inconclusive_retries must be a wait() parameter, not constructor-only.
+        Passing it must not raise TypeError. Passing a small value (1) must cause
+        budget exhaustion after 1 NoPositiveSignal cycle beyond the confirmation window.
+        """
+        watcher = _make_watcher()
```
Trecek, line 986: [info] cohesion: Tests at L804-L949 construct `DefaultMergeQueueWatcher(token=None, max_inconclusive_retries=3)` directly, but L986/L1017 use `_make_watcher()` (constructor default=5). The difference is intentional (testing per-call override), but a brief comment at these `_make_watcher()` call sites would clarify why the constructor default is used here.

Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (1 actionable finding)

src/autoskillit/execution/merge_queue.py
Trecek: Investigated — this is intentional. `not_in_queue_cycles += 1` is at line 380, unconditionally executed after the `if state["in_queue"]: ... continue` guard and **before** the try/except block at line 383. The diff hunk visible in review started at the in-queue reset and omitted line 380, making the increment appear absent. Reading the full loop body confirms the increment is correctly positioned so no `continue` inside the except can bypass it. No code change is warranted.

src/autoskillit/execution/merge_queue.py
Trecek: Investigated — this is intentional. `inconclusive_count` is incremented **only** inside the `NoPositiveSignal` handler (line 392). The `CIStillRunning` handler (lines 385-387) never reads or writes `inconclusive_count`, so interleaved CIStillRunning cycles cannot partially consume the budget. Commit 01628cca documents the explicit design choice: 'CIStillRunning no longer consumes the inconclusive budget — outer timeout_seconds bounds CI wait time.' Adding a reset on CIStillRunning would enable indefinite looping if CI perpetually cycles between running and idle without ever reaching a positive signal, defeating the max_inconclusive_retries guarantee. No code change is warranted.

PR #923, opened 2026-04-14
Add MCP tool to disable quota guard in order sessions

### LINKED ISSUE #919: Add MCP tool to disable quota guard in order sessions
## Summary

Add a user-invocable MCP tool (or Claude command) that disables the quota guard for the current session. This tool should:

- Be **free-range** (visible without `open_kitchen`, alongside `open_kitchen`/`close_kitchen` in `FREE_RANGE_TOOLS`)
- Be **user-invocable only** — not callable by headless automation
- Only be **available in AutoSkillit order sessions** (the only context where quota guard blocking is encountered)

## Motivation

When the quota guard fires during an order session, the only current options are:
1. Wait 20+ minutes for the quota window to reset
2. Manually hack the cache file or config

Neither is ergonomic. A user should be able to consciously override the quota guard for a session when they decide the work is worth the quota spend.

## Behavior

- When invoked, disables the quota guard **for the remainder of the current session only**
- Does not persist across sessions (not a config change)
- Could work by either:
  - Setting a session-scoped flag that the `quota_guard.py` hook checks (e.g., a marker file in the session's temp dir)
  - Writing a "disabled" cache entry and suppressing cache refresh for the session
  - Directly mutating the in-memory `QuotaGuardConfig.enabled = False` on the `ToolContext`

## Context

The quota guard hook (`quota_guard.py`) is stdlib-only and reads from `~/.claude/autoskillit_quota_cache.json`. The config `enabled: false` flag only affects the server-side `check_and_sleep_if_needed` — not the hook. Any solution needs to account for both the hook read path and the server write path.

Trecek:
## Summary

Adds a new `disable_quota_guard` MCP tool that allows order sessions to bypass quota guard enforcement when needed. The quota guard hook (`quota_guard.py`) is updated to respect the new disabled state, and `_hook_settings.py` is extended to expose the disable flag. Kitchen tooling, CLI cook path, and type constants are updated to register and expose the new tool.

Closes #919

## Token Usage Summary

| review_approach | 158 | 12.4k | 1.1M | 98.1k | 1 | 4m 53s |
| make_plan | 195 | 14.2k | 1.4M | 83.3k | 1 | 4m 22s |
| dry_walkthrough | 2.3k | 13.0k | 935.8k | 77.9k | 1 | 3m 52s |
| implement | 570 | 25.7k | 4.6M | 81.3k | 1 | 8m 37s |
| resolve_failures | 226 | 10.6k | 1.4M | 52.7k | 1 | 11m 17s |
| prepare_pr | 101 | 5.3k | 174.6k | 28.2k | 1 | 1m 27s |
| run_arch_lenses | 157 | 13.7k | 597.9k | 43.2k | 1 | 9m 22s |
| compose_pr | 67 | 1.8k | 178.2k | 14.4k | 1 | 43s |
| **Total** | 3.8k | 96.7k | 10.3M | 479.1k | | 44m 35s |

### DIFF
--- a/src/autoskillit/hooks/pretty_output_hook.py
+++ b/src/autoskillit/hooks/pretty_output_hook.py
@@ -189,6 +189,7 @@ _UNFORMATTED_TOOLS: frozenset[str] = frozenset(
         "get_ci_status",  # ci status dict
         "get_quota_events",  # list of quota events, generic renders correctly
         "close_kitchen",  # simple ack
+        "disable_quota_guard",  # simple success/error result
         "register_clone_status",  # simple registered/error result
         "batch_cleanup_clones",  # bulk delete summary dict
     }
--- a/src/autoskillit/server/tools_kitchen.py
+++ b/src/autoskillit/server/tools_kitchen.py
@@ -114,7 +114,7 @@ _DISPLAY_CATEGORIES: tuple[tuple[str, tuple[str, ...]], ...] = (
             "get_quota_events",
         ),
     ),
-    ("Kitchen", ("open_kitchen", "close_kitchen")),
+    ("Kitchen", ("open_kitchen", "close_kitchen", "disable_quota_guard")),
 )
 
 
@@ -425,4 +425,50 @@ async def close_kitchen(ctx: Context = CurrentContext()) -> str:
         return "Kitchen is closed."
     except Exception as exc:
         logger.error("close_kitchen unhandled exception", exc_info=True)
+        return json.dumps({"success": False, "error": f"{type(exc).__name__}: {exc}"})
+
+
+@mcp.tool(tags={"autoskillit"})
+@track_response_size("disable_quota_guard")
+def disable_quota_guard() -> str:
+    """Disable the quota guard for the remainder of this kitchen session.
+
+    The quota guard blocks run_skill calls when API utilization exceeds a
+    threshold. Invoke this tool when you decide the work is worth the quota
+    spend and want to override the guard for the current session.
+
+    Session-scoped only: the guard re-activates when the kitchen is closed
+    and reopened. Does not modify persistent configuration.
+
+    Never raises.
+    """
+    try:
+        if (h := _require_not_headless("disable_quota_guard")) is not None:
+            return h
+        hook_cfg_path = _hook_config_path(Path.cwd())
+        if not hook_cfg_path.exists():
+            return json.dumps(
+                {
+                    "success": False,
+                    "error": "Kitchen is not open — hook config file absent. Open the kitchen first.",
+                }
+            )
+        try:
+            payload = json.loads(hook_cfg_path.read_text())
+        except (OSError, json.JSONDecodeError) as exc:
+            return json.dumps(
+                {"success": False, "error": f"Failed to read hook config: {type(exc).__name__}: {exc}"}
+            )
+        quota_section = payload.get("quota_guard", {})
+        quota_section["disabled"] = True
+        payload["quota_guard"] = quota_section
+        atomic_write(hook_cfg_path, json.dumps(payload))
+        return json.dumps(
+            {
+                "success": True,
+                "content": "Quota guard disabled for this session. run_skill calls will no longer be blocked by quota checks.",
+            }
+        )
+    except Exception as exc:
+        logger.error("disable_quota_guard unhandled exception", exc_info=True)
         return json.dumps({"success": False, "error": f"{type(exc).__name__}: {exc}"})
--- a/src/autoskillit/server/tools_kitchen.py
+++ b/src/autoskillit/server/tools_kitchen.py
@@ -430,7 +430,7 @@ async def close_kitchen(ctx: Context = CurrentContext()) -> str:
 
 @mcp.tool(tags={"autoskillit"})
 @track_response_size("disable_quota_guard")
-def disable_quota_guard() -> str:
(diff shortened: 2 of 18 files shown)


src/autoskillit/server/tools_kitchen.py
```diff
@@ -426,3 +426,55 @@ async def close_kitchen(ctx: Context = CurrentContext()) -> str:
...
+        try:
+            payload = json.loads(hook_cfg_path.read_text())
+        except (OSError, json.JSONDecodeError) as exc:
+            return json.dumps(
+                {
+                    "success": False,
+                    "error": f"Failed to read hook config: {type(exc).__name__}: {exc}",
+                }
+            )
+        quota_section = payload.get("quota_guard", {})
```
Trecek, line 465: [warning] bugs: `_quota_refresh_loop` (runs every ~240s) and post-run_skill `_refresh_quota_cache` are not modified in this PR. If either path overwrites the full `quota_guard` section in the hook config (rather than merging only a `quota_cache` key), `disabled=True` written here will be silently cleared on the next refresh cycle. Verify both refresh paths perform a read-modify-write that preserves existing keys, or document that they only update a disjoint key.

src/autoskillit/hooks/pretty_output_hook.py
```diff
@@ -189,6 +189,7 @@ def _fmt_generic(short_name: str, data: dict, _pipeline: bool) -> str:
         "get_ci_status",  # ci status dict
         "get_quota_events",  # list of quota events, generic renders correctly
         "close_kitchen",  # simple ack
+        "disable_quota_guard",  # simple success/error result
```
Trecek, line 192: [warning] bugs: `quota_post_hook.py` (PostToolUse hook that appends quota warnings to run_skill output) is not modified in this PR. After `disable_quota_guard` is called, the pre-tool hook (`quota_guard.py`) correctly exits 0, but the post-hook will still append quota warning banners on every run_skill call. Add the same `settings.disabled` early-exit to `quota_post_hook.py` to stop appending warnings once the guard is disabled.

src/autoskillit/server/tools_kitchen.py
```diff
@@ -426,3 +426,55 @@ async def close_kitchen(ctx: Context = CurrentContext()) -> str:
...
+        quota_section["disabled"] = True
+        payload["quota_guard"] = quota_section
+        atomic_write(hook_cfg_path, json.dumps(payload))
```
Trecek, line 468: [info] design: The feasibility note recommended mutating `ctx.config.quota_guard.enabled = False` and cancelling `ctx.quota_refresh_task` at the server level. The current implementation uses a hook_config.json file write instead — this only gates the PreToolUse hook; any server-side quota refresh actions (background token polling, cache writes) continue unaffected. Assess whether halting server-side quota refresh is also required, or confirm the hook-file approach is intentionally scope-limited to hook blocking only. (requires_decision: design trade-off)

Trecek (commented):

Findings from feasibility concern assessment:
- 2 actionable warnings (refresh loop overwrite gap, quota_post_hook.py gap)
- 2 design trade-off notes (server-side state mutation, session-scoped markers)

src/autoskillit/server/tools_kitchen.py
```diff
@@ -426,3 +426,55 @@ async def close_kitchen(ctx: Context = CurrentContext()) -> str:
...
+
+    Session-scoped only: the guard re-activates when the kitchen is closed
+    and reopened. Does not modify persistent configuration.
+
+    Never raises.
+    """
+    try:
+        if (h := _require_not_headless("disable_quota_guard")) is not None:
+            return h
+        hook_cfg_path = _hook_config_path(Path.cwd())
```
Trecek, line 448: [info] design: The implementation uses `.hook_config.json` (shared per-CWD) rather than session-scoped marker files (e.g., `.autoskillit/temp/quota_disabled/{session_id}.json` as suggested in the feasibility note). Trade-off: the shared file approach is simpler but may affect multiple concurrent sessions sharing the same CWD. Session-scoped files would provide cleaner isolation at the cost of session-ID discovery. Accept or document the chosen trade-off. (requires_decision: design choice)

src/autoskillit/hooks/pretty_output_hook.py
Trecek: Agreed — fixed in 7867591060978f2c13b7cb8975d65011bd907f2a. quota_guard.py lines 103-104 have `if settings.disabled: sys.exit(0)` immediately after resolve_quota_settings(). quota_post_hook.py line 113 called resolve_quota_settings() but never read settings.disabled, so warning banners would still append after disable_quota_guard set disabled=True. Added the matching early-exit guard.

src/autoskillit/server/tools_kitchen.py
Trecek: Investigated — this is intentional. _refresh_quota_cache (quota.py:278-303) calls _write_cache(config.cache_path, fetch_result) which writes only to the quota cache file (e.g. ~/.claude/autoskillit_quota_cache.json), not to the hook_config file (.autoskillit/temp/.hook_config.json). These are entirely separate files. The _quota_refresh_loop (helpers.py:507-524) never reads or writes hook_config. disable_quota_guard correctly performs a read-modify-write on hook_config preserving all existing keys. The disabled=True flag is safe from any refresh cycle overwrite.

PR #924, opened 2026-04-14
Rectify: Doctor-Install Disconnect

### LINKED ISSUE #912: Doctor-Install Disconnect: install does not resolve doctor warnings
## Problem

Running `autoskillit doctor` reports six warnings/errors. It suggests running `autoskillit install`. After running `autoskillit install` (which prints output that looks like success), re-running `autoskillit doctor` shows **the exact same issues**. Nothing changed.

### Reproduction

```
$ autoskillit install
Checking for updates...
Marketplace prepared: /home/talon/.autoskillit/marketplace
Marketplace registered.
Plugin installed: autoskillit@autoskillit-local (scope: user)
  Next steps:
  1. autoskillit init  create project config
  2. autoskillit doctor  verify your setup

$ autoskillit doctor
WARNING: autoskillit entry missing from installed_plugins.json. Run `autoskillit install` to fix.
ERROR: Hook scripts not found: echo '{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny","permissionDecisionReason":"Plan mode is permanently disabled."}}'
WARNING: [project] Hook registry has changed since last install. Run 'autoskillit install' to deploy 14 new/changed hook(s).
WARNING: [local] Hook registry has changed since last install. Run 'autoskillit install' to deploy 14 new/changed hook(s).
WARNING: Outdated recipes: full-audit, smoke-test. Will be auto-migrated on next load.
WARNING: Files in .autoskillit/ not covered by .gitignore: .onboarded, sync_manifest.json, .secrets.yaml. Add to _AUTOSKILLIT_GITIGNORE_ENTRIES or _COMMITTED_BY_DESIGN.
```

## Root Cause

Three independent bugs and two design gaps produce the six symptoms.

---

### Bug 1 — Wrong dict level in `_check_installed_plugins_entry` (false positive)

**File:** `src/autoskillit/cli/_doctor.py:629`

```python
if "autoskillit@autoskillit-local" in data:  # BUG
```

The actual JSON structure of `~/.claude/plugins/installed_plugins.json` is:

```json
{"version": 2, "plugins": {"autoskillit@autoskillit-local": [...]}}
```

The key is nested under `data["plugins"]`, not at the top level of `data`. The top-level keys are `version` and `plugins`. The check always fails because it looks in the wrong dict.

**Verified:** The entry IS present on disk at `data["plugins"]["autoskillit@autoskillit-local"]`. The doctor falsely reports it as missing.

**Fix:** Change to `if "autoskillit@autoskillit-local" in data.get("plugins", {}):`.

**Note:** Two other call sites in the same codebase already use the correct pattern:
- `src/autoskillit/cli/_mcp_names.py:39` — `data.get("plugins", {})`
- `src/autoskillit/cli/_init_helpers.py:344-346` — `data.get("plugins", {})`

---

### Bug 2 — Same wrong dict level in `_clear_plugin_cache` (silent no-op)

**File:** `src/autoskillit/cli/_marketplace.py:43-44`

```python
if plugin_ref in data:      # BUG: never matches top-level keys
    del data[plugin_ref]     # never reached
```

Same structural issue as Bug 1. The cache-clear function silently no-ops because `plugin_ref` is never a top-level key. Should be:

```python
if plugin_ref in data.get("plugins", {}):
    del data["plugins"][plugin_ref]
```

---

### Bug 3 — `CLAUDECODE=1` early-exit masks failure as success

**File:** `src/autoskillit/cli/_marketplace.py:122-128`

```python
if os.environ.get("CLAUDECODE"):
    print("\nRun these commands in a regular terminal to complete installation:")
    print(f"  claude plugin marketplace add {marketplace_dir}")
    print(f"  claude plugin install {plugin_ref} --scope {scope}")
    print("\nThen run: autoskillit init (in your project directory)")
    return
```

When running inside a Claude Code session (`CLAUDECODE=1`), the install function:
1. Prints "Marketplace prepared: ..." (looks like success)
2. Immediately returns **without executing**:
   - `claude plugin install` subprocess
   - `sync_hooks_to_settings()` (hook deployment)
   - `sweep_all_scopes_for_orphans()` (orphan cleanup)

The instructions to run commands manually in a regular terminal are printed, but the overall output format (especially after `_print_next_steps` runs) makes it look like a successful install. The user sees "Marketplace prepared" + "Plugin installed" style output and reasonably concludes it worked.

**This is the primary cause of hook drift warnings persisting.** Hooks are never synced to `settings.json` because the sync step is after the early return.

---

### Design Gap 4 — `install` never repairs `.autoskillit/.gitignore`

`autoskillit install` calls `_marketplace.install()` which has no call to `ensure_project_temp()`. Only `autoskillit init` triggers the gitignore auto-repair path at `src/autoskillit/core/io.py:170-176`.

The `.autoskillit/.gitignore` currently contains only `temp/`, but the canonical entries list (`_AUTOSKILLIT_GITIGNORE_ENTRIES` at `core/io.py:121`) requires four entries: `temp/`, `.secrets.yaml`, `.onboarded`, `sync_manifest.json`. The repair logic exists and works — it's just never called from the install path.

Doctor suggests "Run `autoskillit install` to fix" but install has no gitignore awareness at all.

---

### Design Gap 5 — `install` has no recipe migration step

`autoskillit install` has zero awareness of project recipes. The "Outdated recipes: full-audit, smoke-test" warning (both recipes lack an `autoskillit_version` field, treated as version `0.0.0`) can only be resolved by explicit `migrate_recipe` MCP tool calls or the `/autoskillit:migrate-recipes` skill.

The doctor message "Will be auto-migrated on next load" is misleading — `load_recipe` (`src/autoskillit/server/tools_recipe.py:170`) is explicitly read-only and does NOT trigger migration. Migration requires an explicit `migrate_recipe` tool call (`tools_recipe.py:250-308`).

---

### Side Note — "Plan mode is permanently disabled" ERROR

The `ERROR: Hook scripts not found: echo '{"hookSpecificOutput":...}'` is a separate issue. This is a manually-placed inline `echo` hook in `~/.claude/settings.json` that denies `EnterPlanMode`. It is NOT an autoskillit hook. Doctor's `find_broken_hook_scripts()` (`hook_registry.py:276-288`) processes all hooks (not just autoskillit-owned ones), splits on whitespace, and checks if `parts[-1]` is a file on disk. For the `echo '...'` command, `parts[-1]` is the JSON string — not a file path — so it's flagged as "not found." This is a false positive for non-autoskillit inline hooks.

---

## Affected Components

| File | Line(s) | Issue |
|---|---|---|
| `src/autoskillit/cli/_doctor.py` | 629 | Wrong dict level: `in data` should be `in data.get("plugins", {})` |
| `src/autoskillit/cli/_marketplace.py` | 43-44 | Wrong dict level: same as above, plus `del` targets wrong level |
| `src/autoskillit/cli/_marketplace.py` | 122-128 | CLAUDECODE early-exit looks like success |
| `src/autoskillit/cli/_marketplace.py` | 100-178 | `install()` missing `ensure_project_temp()` call |
| `src/autoskillit/core/io.py` | 138-177 | `ensure_project_temp()` works correctly, just not called from install |
| `src/autoskillit/hook_registry.py` | 276-288 | `find_broken_hook_scripts` false-positive on inline echo hooks |
| `.autoskillit/.gitignore` | — | Only contains `temp/`, missing 3 entries |

## Historical Context

This area has extensive prior fix history (5+ rectify PRs), suggesting a recurring pattern of narrowly-scoped fixes that don't address the cross-cutting lifecycle invariant: **install must resolve what doctor reports**.

Key prior fixes:
- `afc9f69` — "Rectify: Init Gitignore Completeness Immunity (#442)": Fixed gitignore coverage but only for `init`, not `install`.
- `4b390ed` — "Rectify: Doctor Destroys Plugin Cache + Startup Race Regression (#902)": Fixed doctor calling `_clear_plugin_cache()` but did NOT fix the wrong-dict-level bug in the same function.
- `8e9b3c7` — "Rectify: Ghost Hook Registrations Survive Git Revert (#670)": Fixed orphaned hook detection but not the install-inside-CLAUDECODE failure path.
- `f92d781` — "Rectify: Stale-Hooks Check Infinite Prompt Loop (#682)": Fixed hook-check UX, scoped only to the interaction loop.

## Recommendations

1. **Fix Bug 1**: `_doctor.py:629` — `in data` → `in data.get("plugins", {})`.
2. **Fix Bug 2**: `_marketplace.py:43-44` — same dict-level fix, and target `data["plugins"]` for deletion.
3. **Fix Bug 3**: Make the CLAUDECODE early-exit clearly communicate failure. Use WARNING/ERROR-level prefix. Explicitly state what was NOT done.
4. **Fix Gap 4**: Call `ensure_project_temp()` in the install flow (doesn't require the `claude` subprocess, so it can run even inside CLAUDECODE).
5. **Fix Gap 5**: Either trigger recipe migration in install, or fix the doctor message from "Will be auto-migrated on next load" to accurately describe how to resolve.
6. **Fix false positive**: `find_broken_hook_scripts` should skip hooks that aren't autoskillit-owned (already has `_is_own_hook()` available).
7. **Add roundtrip test**: Create a realistic `installed_plugins.json` with `{"version": 2, "plugins": {...}}` structure and verify `_check_installed_plugins_entry` returns OK.
8. **Add lifecycle invariant test**: For every doctor WARNING/ERROR that says "Run X to fix", verify that running X actually resolves the issue.

Trecek:
## Summary

Four discrete bugs share a single architectural root: there is no canonical owner for `installed_plugins.json`. Every caller that touches the file re-implements its own structural interpretation of the JSON format, and two of those implementations got the nesting wrong (`{"plugins": {...}}` vs flat `{}`). The test fixtures matched the wrong implementations, so the regression passed CI.

The architectural fix is a **typed repository class** — `InstalledPluginsFile` — that becomes the single authorized interface for all reads and writes of `installed_plugins.json`. Once all call sites are routed through this class, the correct nesting is in exactly one place. A wrong structure cannot be silently introduced by a new call site: the class's API enforces the contract.

Two companion fixes complete the immunity:
- `install()` must return a typed value distinguishing "complete" from "deferred (CLAUDECODE)" so callers cannot proceed with misleading success output.
- `find_broken_hook_scripts` must apply the same `_is_own_hook` ownership filter that `_extract_script_basenames` already uses, eliminating false positives on user-defined inline hooks.

Part B (separate task) will cover `ensure_project_temp` call site completeness across CLI entry points.

### Repository/Data Access Diagram

```mermaid
flowchart LR
    %% CLASS DEFINITIONS %%

    subgraph Callers ["CALLERS"]
        direction TB
        APP["● app.py<br/>━━━━━━━━━━<br/>CLI entry: install,<br/>init, doctor, upgrade"]
        MARKET["● _marketplace.py<br/>━━━━━━━━━━<br/>install() / upgrade()<br/>_clear_plugin_cache()"]
        DOCTOR["● _doctor.py<br/>━━━━━━━━━━<br/>_check_installed_plugins_entry()<br/>_check_hook_registry_drift()<br/>_check_hook_health()"]
    end

    subgraph PluginRepo ["★ PLUGIN REGISTRY (new)"]
        direction TB
        IPF["★ InstalledPluginsFile<br/>━━━━━━━━━━<br/>get_plugins() → read<br/>contains(ref) → read<br/>remove(ref) → read+write<br/>cli/_installed_plugins.py"]
    end

    subgraph HookRepo ["● HOOK REGISTRY (modified)"]
        direction TB
        HREG["● HOOK_REGISTRY<br/>━━━━━━━━━━<br/>List[HookDef] — canonical<br/>source of truth<br/>hook_registry.py"]
        GEN["● generate_hooks_json()<br/>━━━━━━━━━━<br/>HookDef → JSON structure<br/>hook_registry.py"]
        DRIFT["● _count_hook_registry_drift()<br/>━━━━━━━━━━<br/>canonical vs deployed<br/>hook_registry.py"]
        FIND["find_broken_hook_scripts()<br/>━━━━━━━━━━<br/>deployed scripts<br/>existence check"]
        EXTRACT["_extract_script_basenames()<br/>━━━━━━━━━━<br/>settings.json hooks<br/>→ basename set"]
    end

    subgraph IOLayer ["I/O PRIMITIVE"]
        AW["atomic_write()<br/>━━━━━━━━━━<br/>fsync + os.replace<br/>core/io.py"]
    end

    subgraph Storage ["STORAGE"]
        direction TB
        INSTJSON[("installed_plugins.json<br/>━━━━━━━━━━<br/>~/.claude/plugins/<br/>installed_plugins.json")]
        HOOKSJSON[("hooks.json<br/>━━━━━━━━━━<br/>hooks/hooks.json<br/>plugin manifest")]
        SETTINGS[("settings.json<br/>━━━━━━━━━━<br/>~/.claude/settings.json<br/>per-scope hook registration")]
        CLAUDEJSON[("~/.claude.json<br/>━━━━━━━━━━<br/>Legacy mcpServers<br/>config")]
        CACHE[("Plugin Cache Dir<br/>━━━━━━━━━━<br/>~/.claude/plugins/<br/>cache/autoskillit-local/")]
    end

    %% Caller → Repository relationships %%
    APP -->|"invokes"| MARKET
    APP -->|"invokes"| DOCTOR

    MARKET -->|"reads/writes"| IPF
    MARKET -->|"reads"| HREG
    MARKET -->|"writes hooks.json"| AW

    DOCTOR -->|"reads"| IPF
    DOCTOR -->|"reads"| DRIFT
    DOCTOR -->|"reads"| FIND
    DOCTOR -->|"reads"| SETTINGS

    %% Hook Registry internal %%
    HREG -->|"drives"| GEN
    HREG -->|"drives"| DRIFT
    GEN -->|"writes via"| AW
    DRIFT -->|"reads"| EXTRACT
    EXTRACT -->|"parses"| SETTINGS
    FIND -->|"parses"| SETTINGS

    %% Repository → Storage %%
    IPF -->|"reads"| INSTJSON
    IPF -->|"writes via atomic_write"| AW
    AW -->|"writes"| INSTJSON
    AW -->|"writes"| HOOKSJSON

    %% Cache eviction (marketplace) %%
    MARKET -->|"deletes"| CACHE

    %% Doctor reads legacy config %%
    DOCTOR -->|"reads"| CLAUDEJSON

    %% CLASS ASSIGNMENTS %%
    class APP,MARKET cli;
    class DOCTOR detector;
    class IPF newComponent;
    class HREG,GEN stateNode;
    class DRIFT,FIND,EXTRACT phase;
    class AW handler;
    class INSTJSON,HOOKSJSON,SETTINGS,CLAUDEJSON,CACHE integration;
```

### Process Flow Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINALS %%
    INSTALL_START([autoskillit install])
    INSTALL_DEFERRED([DEFERRED<br/>━━━━━━━━━━<br/>user must run manually])
    INSTALL_COMPLETE([INSTALL COMPLETE])
    DOCTOR_START([autoskillit doctor])
    DOCTOR_END([DOCTOR COMPLETE])

    %% ─────────────────────────────────────────────
    %%  INSTALL FLOW
    %% ─────────────────────────────────────────────

    subgraph InstallPreflight ["Install — Preflight"]
        direction TB
        ValidateScope{"● scope valid?<br/>━━━━━━━━━━<br/>user / project / local"}
        EnsureMarketplace["_ensure_marketplace()<br/>━━━━━━━━━━<br/>mkdir ~/.autoskillit/marketplace<br/>write marketplace.json<br/>symlink → pkg_root()"]
        GuardCLAUDECODE{"CLAUDECODE env set?<br/>━━━━━━━━━━<br/>running inside session?"}
        GuardClaude{"claude on PATH?<br/>━━━━━━━━━━<br/>shutil.which('claude')"}
        EnsureWorkspace["_ensure_workspace_ready()<br/>━━━━━━━━━━<br/>ensure_project_temp()<br/>upgrade() if scripts/ present"]
    end

    subgraph CacheClear ["Install — Cache Clearing (●)"]
        direction TB
        RmCacheDir["shutil.rmtree(cache_dir)<br/>━━━━━━━━━━<br/>~/.claude/plugins/cache/<br/>autoskillit-local/autoskillit"]
        InstalledPluginsRemove["★ InstalledPluginsFile.remove()<br/>━━━━━━━━━━<br/>reads installed_plugins.json<br/>deletes 'autoskillit@autoskillit-local'<br/>atomic_write() back"]
        HooksJsonRegen["● generate_hooks_json()<br/>━━━━━━━━━━<br/>iterates HOOK_REGISTRY<br/>writes hooks/hooks.json<br/>embeds HOOK_REGISTRY_HASH"]
    end

    subgraph PluginRegistration ["Install — Plugin Registration"]
        direction TB
        MarketplaceAdd["claude plugin marketplace add<br/>━━━━━━━━━━<br/>subprocess, timeout=30s"]
        PluginInstall["claude plugin install<br/>━━━━━━━━━━<br/>plugin_ref @ scope"]
        EvictDirect["evict_direct_mcp_entry()<br/>━━━━━━━━━━<br/>remove stale direct entry<br/>from ~/.claude.json"]
    end

    subgraph HookSync ["Install — Hook Sync (●)"]
        direction TB
        SweepAllScopes["sweep_all_scopes_for_orphans()<br/>━━━━━━━━━━<br/>iterate user + project + local<br/>remove orphaned hook entries"]
        SyncHooks["sync_hooks_to_settings()<br/>━━━━━━━━━━<br/>write canonical hooks<br/>to target scope settings.json"]
    end

    %% ─────────────────────────────────────────────
    %%  SHARED STATE
    %% ─────────────────────────────────────────────

    subgraph SharedState ["★ InstalledPluginsFile (_installed_plugins.py)"]
        direction LR
        IPF_Read["_read()<br/>━━━━━━━━━━<br/>json.loads(path)<br/>→ {} on missing/error"]
        IPF_Contains["contains(plugin_ref)<br/>━━━━━━━━━━<br/>plugin_ref in get_plugins()"]
        IPF_Remove["remove(plugin_ref)<br/>━━━━━━━━━━<br/>del plugins[ref]<br/>atomic_write() back"]
        InstalledPluginsJSON[("installed_plugins.json<br/>━━━━━━━━━━<br/>~/.claude/plugins/<br/>{version,plugins:{}}")]
    end

    %% ─────────────────────────────────────────────
    %%  DOCTOR FLOW
    %% ─────────────────────────────────────────────

    subgraph DoctorMCP ["Doctor — MCP Checks"]
        direction TB
        ChkStaleMCP["check: stale_mcp_servers<br/>━━━━━━━━━━<br/>dead binary paths in ~/.claude.json"]
        ChkMCPReg["check: mcp_server_registered<br/>━━━━━━━━━━<br/>mcpServers entry OR<br/>claude plugin list"]
        ChkDualReg["● check: dual_mcp_registration<br/>━━━━━━━━━━<br/>direct entry AND plugin both present?<br/>→ WARNING: split-brain"]
        ChkPluginCache["check: plugin_cache_exists<br/>━━━━━━━━━━<br/>~/.claude/plugins/cache/... dir?"]
        ChkInstalledPlugins["● check: installed_plugins_entry<br/>━━━━━━━━━━<br/>InstalledPluginsFile.contains()<br/>'autoskillit@autoskillit-local'?"]
    end

    subgraph DoctorHooks ["Doctor — Hook Checks (●)"]
        direction TB
        ChkHookHealth["check: hook_health<br/>━━━━━━━━━━<br/>find_broken_hook_scripts()<br/>script files exist on disk?"]
        ChkHookReg["check: hook_registration<br/>━━━━━━━━━━<br/>canonical_script_basenames()<br/>all scripts in settings.json?"]
        ChkHookDrift["● check: hook_registry_drift<br/>━━━━━━━━━━<br/>_count_hook_registry_drift()<br/>orphaned | missing hooks<br/>iterates all scopes"]
    end

    subgraph DoctorHealth ["Doctor — Health Checks"]
        direction TB
        ChkVersion["check: version_consistency<br/>━━━━━━━━━━<br/>pkg version == plugin.json version?"]
        ChkConfig["check: project_config<br/>━━━━━━━━━━<br/>.autoskillit/config.yaml exists?"]
        ChkGitignore["check: gitignore_completeness<br/>━━━━━━━━━━<br/>.autoskillit/ entries covered?"]
        ChkQuota["check: quota_cache_schema<br/>━━━━━━━━━━<br/>schema_version drift?"]
        ChkInstallClass["check: install_classification<br/>━━━━━━━━━━<br/>detect_install() → InstallType"]
        ChkSourceDrift["check: source_version_drift<br/>━━━━━━━━━━<br/>commit SHA vs reference SHA<br/>(network + disk-cache TTL)"]
        ChkEditableInst["check: editable_install_source_exists<br/>━━━━━━━━━━<br/>direct_url.json → source path exists?"]
        ChkDismissal["check: update_dismissal_state<br/>━━━━━━━━━━<br/>dismissed_at + window expiry"]
    end

    %% ─────────────────────────────────────────────
    %%  INSTALL EDGES
    %% ─────────────────────────────────────────────
    INSTALL_START --> ValidateScope
    ValidateScope -->|"invalid"| INSTALL_COMPLETE
    ValidateScope -->|"valid"| EnsureMarketplace
    EnsureMarketplace --> GuardCLAUDECODE
    GuardCLAUDECODE -->|"yes — inside session"| INSTALL_DEFERRED
    GuardCLAUDECODE -->|"no"| GuardClaude
    GuardClaude -->|"missing"| INSTALL_COMPLETE
    GuardClaude -->|"found"| EnsureWorkspace
    EnsureWorkspace --> RmCacheDir
    RmCacheDir --> InstalledPluginsRemove
    InstalledPluginsRemove -->|"reads/writes"| InstalledPluginsJSON
    InstalledPluginsRemove --> HooksJsonRegen
    HooksJsonRegen --> MarketplaceAdd
    MarketplaceAdd -->|"returncode != 0"| INSTALL_COMPLETE
    MarketplaceAdd -->|"ok"| PluginInstall
    PluginInstall -->|"returncode != 0"| INSTALL_COMPLETE
    PluginInstall -->|"ok"| EvictDirect
    EvictDirect --> SweepAllScopes
    SweepAllScopes --> SyncHooks
    SyncHooks --> INSTALL_COMPLETE

    %% ─────────────────────────────────────────────
    %%  DOCTOR EDGES
    %% ─────────────────────────────────────────────
    DOCTOR_START --> ChkStaleMCP
    ChkStaleMCP --> ChkMCPReg
    ChkMCPReg --> ChkDualReg
    ChkDualReg --> ChkPluginCache
    ChkPluginCache --> ChkInstalledPlugins
    ChkInstalledPlugins -->|"reads"| IPF_Contains
    IPF_Contains -->|"reads"| IPF_Read
    IPF_Read -->|"reads"| InstalledPluginsJSON
    ChkInstalledPlugins --> ChkHookHealth
    ChkHookHealth --> ChkHookReg
    ChkHookReg --> ChkHookDrift
    ChkHookDrift --> ChkVersion
    ChkVersion --> ChkConfig
    ChkConfig --> ChkGitignore
    ChkGitignore --> ChkQuota
    ChkQuota --> ChkInstallClass
    ChkInstallClass --> ChkSourceDrift
    ChkSourceDrift --> ChkEditableInst
    ChkEditableInst --> ChkDismissal
    ChkDismissal --> DOCTOR_END

    %% ─────────────────────────────────────────────
    %%  SHARED STATE INTERNAL EDGES
    %% ─────────────────────────────────────────────
    IPF_Remove -->|"atomic_write()"| InstalledPluginsJSON

    %% ─────────────────────────────────────────────
    %%  CLASS ASSIGNMENTS
    %% ─────────────────────────────────────────────
    class INSTALL_START,INSTALL_DEFERRED,INSTALL_COMPLETE,DOCTOR_START,DOCTOR_END terminal;
    class ValidateScope,GuardCLAUDECODE,GuardClaude stateNode;
    class EnsureMarketplace,EnsureWorkspace,HooksJsonRegen,MarketplaceAdd,PluginInstall,EvictDirect handler;
    class RmCacheDir,SweepAllScopes,SyncHooks phase;
    class InstalledPluginsRemove,IPF_Read,IPF_Contains,IPF_Remove newComponent;
    class InstalledPluginsJSON output;
    class ChkStaleMCP,ChkMCPReg,ChkDualReg,ChkPluginCache,ChkInstalledPlugins detector;
    class ChkHookHealth,ChkHookReg,ChkHookDrift detector;
    class ChkVersion,ChkConfig,ChkGitignore,ChkQuota,ChkInstallClass,ChkSourceDrift,ChkEditableInst,ChkDismissal phase;
```

Closes #912

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260414-075239-201688/.autoskillit/temp/rectify/rectify_doctor-install-disconnect_2026-04-14_120000_part_a.md`

## Token Usage Summary

| review_approach | 158 | 12.4k | 1.1M | 98.1k | 1 | 4m 53s |
| make_plan | 195 | 14.2k | 1.4M | 83.3k | 1 | 4m 22s |
| dry_walkthrough | 2.3k | 13.0k | 935.8k | 77.9k | 1 | 3m 52s |
| implement | 570 | 25.7k | 4.6M | 81.3k | 1 | 8m 37s |
| resolve_failures | 226 | 10.6k | 1.4M | 52.7k | 1 | 11m 17s |
| prepare_pr | 101 | 5.3k | 174.6k | 28.2k | 1 | 1m 27s |
| run_arch_lenses | 157 | 13.7k | 597.9k | 43.2k | 1 | 9m 22s |
| compose_pr | 67 | 1.8k | 178.2k | 14.4k | 1 | 43s |
| review_pr | 94 | 21.1k | 374.8k | 48.0k | 1 | 6m 6s |
| **Total** | 3.9k | 117.8k | 10.7M | 527.1k | | 50m 42s |

### DIFF
--- a/src/autoskillit/cli/_doctor.py
+++ b/src/autoskillit/cli/_doctor.py
@@ -609,24 +609,16 @@ def _check_installed_plugins_entry(
     plugins_json_path: Path | None = None,
 ) -> DoctorResult:
     """Check that installed_plugins.json contains the autoskillit entry."""
-    _plugins_json = plugins_json_path or (
-        Path.home() / ".claude" / "plugins" / "installed_plugins.json"
-    )
-    if not _plugins_json.exists():
+    from autoskillit.cli._installed_plugins import InstalledPluginsFile
+
+    store = InstalledPluginsFile(plugins_json_path)
+    if not store.path.exists():
         return DoctorResult(
             Severity.WARNING,
             "installed_plugins_entry",
             "installed_plugins.json not found. Run `autoskillit install`.",
         )
-    try:
-        data = json.loads(_plugins_json.read_text())
-    except (json.JSONDecodeError, OSError):
-        return DoctorResult(
-            Severity.WARNING,
-            "installed_plugins_entry",
-            "installed_plugins.json could not be parsed.",
-        )
-    if "autoskillit@autoskillit-local" in data:
+    if store.contains("autoskillit@autoskillit-local"):
         return DoctorResult(
             Severity.OK,
             "installed_plugins_entry",
--- a/src/autoskillit/cli/_installed_plugins.py
+++ b/src/autoskillit/cli/_installed_plugins.py
@@ -1,0 +1,65 @@
+"""Canonical accessor for ~/.claude/plugins/installed_plugins.json.
+
+The real file structure produced by `claude plugin install` is:
+    {"version": 2, "plugins": {"autoskillit@autoskillit-local": {...}}}
+
+All access to this file must go through InstalledPluginsFile so that the
+nesting contract is defined in exactly one place.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any
+
+from autoskillit.core.io import atomic_write
+
+
+def _default_path() -> Path:
+    return Path.home() / ".claude" / "plugins" / "installed_plugins.json"
+
+
+class InstalledPluginsFile:
+    """Repository over installed_plugins.json.
+
+    Exposes typed query and mutation methods; enforces the {'plugins': {}} nesting.
+    """
+
+    def __init__(self, path: Path | None = None) -> None:
+        self._path = path or _default_path()
+
+    @property
+    def path(self) -> Path:
+        return self._path
+
+    def _read(self) -> dict[str, Any]:
+        if not self._path.exists():
+            return {}
+        try:
+            return json.loads(self._path.read_text())
+        except (json.JSONDecodeError, OSError):
+            return {}
+
+    def get_plugins(self) -> dict[str, Any]:
+        """Return the nested plugins dict (never raises)."""
+        return self._read().get("plugins", {})
+
+    def contains(self, plugin_ref: str) -> bool:
+        """Return True iff plugin_ref is present in data['plugins']."""
+        return plugin_ref in self.get_plugins()
+
+    def remove(self, plugin_ref: str) -> None:
+        """Remove plugin_ref from data['plugins'] and atomically rewrite the file.
+
+        No-op if the file does not exist or the key is absent.
+        """
+        if not self._path.exists():
+            return
+        data = self._read()
+        plugins = data.get("plugins", {})
+        if plugin_ref not in plugins:
+            return
+        del plugins[plugin_ref]
+        data["plugins"] = plugins
+        atomic_write(self._path, json.dumps(data, indent=2))
--- a/src/autoskillit/cli/_installed_plugins.py
+++ b/src/autoskillit/cli/_installed_plugins.py
@@ -13,7 +13,7 @@ import json
 from pathlib import Path
 from typing import Any
 
-from autoskillit.core.io import atomic_write
+from autoskillit.core import atomic_write
 
 
 def _default_path() -> Path:
--- a/src/autoskillit/cli/_marketplace.py
+++ b/src/autoskillit/cli/_marketplace.py
@@ -93,6 +93,24 @@ def _ensure_marketplace() -> Path:
     return marketplace_dir
 
 
+def _ensure_workspace_ready() -> None:
+    """Repair project workspace state that install() is responsible for.
+
+    Called after the CLAUDECODE guard — only when the actual install proceeds.
+    Idempotent: safe to call on any project state.
+    """
+    from autoskillit.core import ensure_project_temp
+
+    project_dir = Path.cwd()
+    # Repair .autoskillit/.gitignore and ensure temp/ exists
+    if (project_dir / ".autoskillit").is_dir():
+        ensure_project_temp(project_dir)
+
+    # Migrate legacy .autoskillit/scripts/ to .autoskillit/recipes/ if present
+    if (project_dir / ".autoskillit" / "scripts").exists():
+        upgrade()
+
+
 def install(*, scope: str = "user") -> bool:
     """Install the plugin persistently for Claude Code.
 
@@ -131,6 +149,7 @@ def install(*, scope: str = "user") -> bool:
         print("\nThen run: autoskillit init (in your project directory)")
         sys.exit(1)
 
+    _ensure_workspace_ready()
     _clear_plugin_cache()
 
     # Regenerate hooks.json from the canonical registry with absolute paths
--- a/src/autoskillit/cli/_marketplace.py
+++ b/src/autoskillit/cli/_marketplace.py
@@ -17,6 +17,7 @@ from autoskillit.cli._hooks import (
 )
 from autoskillit.cli._init_helpers import _user_claude_json_path, evict_direct_mcp_entry
 from autoskillit.core import atomic_write, is_git_worktree, pkg_root
+from autoskillit.hooks import generate_hooks_json
 
 _VALID_SCOPES = {"user", "project", "local"}
 _MARKETPLACE_NAME = "autoskillit-local"
@@ -153,8 +154,6 @@ def install(*, scope: str = "user") -> bool:
     _clear_plugin_cache()
 
     # Regenerate hooks.json from the canonical registry with absolute paths
-    from autoskillit.hooks import generate_hooks_json
-
     hooks_json_path = pkg_root() / "hooks" / "hooks.json"
     atomic_write(hooks_json_path, json.dumps(generate_hooks_json(), indent=2) + "\n")
 
--- a/src/autoskillit/cli/_marketplace.py
+++ b/src/autoskillit/cli/_marketplace.py
@@ -35,16 +35,12 @@ def _clear_plugin_cache() -> None:
     if cache_dir.is_dir():
         shutil.rmtree(cache_dir)
 
-    installed_json = Path.home() / ".claude" / "plugins" / "installed_plugins.json"
-    if installed_json.exists():
-        try:
-            data = json.loads(installed_json.read_text())
-            plugin_ref = f"autoskillit@{_MARKETPLACE_NAME}"
-            if plugin_ref in data:
-                del data[plugin_ref]
-                atomic_write(installed_json, json.dumps(data, indent=2))
-        except (OSError, json.JSONDecodeError):
-            pass  # non-fatal — install will proceed regardless
+    from autoskillit.cli._installed_plugins import InstalledPluginsFile
+
+    try:
+        InstalledPluginsFile().remove(f"autoskillit@{_MARKETPLACE_NAME}")
+    except OSError:
+        pass  # non-fatal — install will proceed regardless
 
 
 def _ensure_marketplace() -> Path:
@@ -97,7 +93,7 @@ def _ensure_marketplace() -> Path:
     return marketplace_dir
 
 
-def install(*, scope: str = "user"):
+def install(*, scope: str = "user") -> bool:
     """Install the plugin persistently for Claude Code.
 
     Sets up a local marketplace and installs the plugin so it loads
@@ -125,7 +121,7 @@ def install(*, scope: str = "user"):
         print(f"  claude plugin marketplace add {marketplace_dir}")
         print(f"  claude plugin install {plugin_ref} --scope {scope}")
         print("\nThen run: autoskillit init (in your project directory)")
-        return
+        return False  # deferred: user must complete manually in a regular terminal
 
     if shutil.which("claude") is None:
         print("\nERROR: 'claude' command not found on PATH.")
@@ -176,6 +172,7 @@ def install(*, scope: str = "user"):
     sweep_all_scopes_for_orphans(Path.cwd())
     settings_path = _hooks_mod._claude_settings_path(scope)
     sync_hooks_to_settings(settings_path)
+    return True
 
 
 def upgrade():
--- a/src/autoskillit/cli/_marketplace.py
+++ b/src/autoskillit/cli/_marketplace.py
@@ -109,7 +109,10 @@ def _ensure_workspace_ready() -> None:
 
     # Migrate legacy .autoskillit/scripts/ to .autoskillit/recipes/ if present
     if (project_dir / ".autoskillit" / "scripts").exists():
-        upgrade()
+        try:
+            upgrade()
+        except OSError as exc:
+            print(f"Warning: migration upgrade() failed (non-fatal): {exc}")
 
 
 def install(*, scope: str = "user") -> bool:
--- a/src/autoskillit/hook_registry.py
+++ b/src/autoskillit/hook_registry.py
@@ -281,6 +281,8 @@ def find_broken_hook_scripts(settings_path: Path) -> list[str]:
         for entry in data.get("hooks", {}).get(event_type, []):
             for hook in entry.get("hooks", []):
                 cmd = hook.get("command", "")
+                if not _is_own_hook(cmd):  # skip non-autoskillit hooks
+                    continue
                 parts = cmd.split()
                 if len(parts) >= 2:
                     if not Path(parts[-1]).is_file():
--- a/tests/cli/test_doctor.py
+++ b/tests/cli/test_doctor.py
@@ -1784,6 +1784,7 @@ def test_doctor_no_dual_when_only_direct(tmp_path: Path, monkeypatch: pytest.Mon
 def test_check_installed_plugins_entry_real_structure_is_ok(tmp_path: Path) -> None:
     """With the real nested format, the check must report OK."""
     from autoskillit.cli._doctor import _check_installed_plugins_entry
+    from autoskillit.core import Severity
 
     p = tmp_path / "installed_plugins.json"
     p.write_text(
@@ -1795,14 +1796,15 @@ def test_check_installed_plugins_entry_real_structure_is_ok(tmp_path: Path) -> N
         )
     )
     result = _check_installed_plugins_entry(plugins_json_path=p)
-    assert result.severity.name == "OK"
+    assert result.severity == Severity.OK
 
 
 def test_check_installed_plugins_entry_flat_structure_is_warning(tmp_path: Path) -> None:
     """A flat structure (wrong format) must not be silently treated as OK."""
     from autoskillit.cli._doctor import _check_installed_plugins_entry
+    from autoskillit.core import Severity
 
     p = tmp_path / "installed_plugins.json"
     p.write_text(json.dumps({"autoskillit@autoskillit-local": {}}))
     result = _check_installed_plugins_entry(plugins_json_path=p)
-    assert result.severity.name == "WARNING"
+    assert result.severity == Severity.WARNING
--- a/tests/cli/test_doctor.py
+++ b/tests/cli/test_doctor.py
@@ -94,7 +94,7 @@ class TestCLIDoctor:
         )
         # Create installed_plugins.json for Check 2d
         (tmp_path / ".claude" / "plugins" / "installed_plugins.json").write_text(
-            json.dumps({"autoskillit@autoskillit-local": {}})
+            json.dumps({"version": 2, "plugins": {"autoskillit@autoskillit-local": {}}})
         )
         # Register hooks so hook_registration check passes
         # Use explicit path (tmp_path already monkeypatched as Path.home())
@@ -639,13 +639,17 @@ def test_doctor_does_not_modify_plugin_state(tmp_path, monkeypatch, capsys):
     (cache_dir / "0.3.0" / "hooks" / "pretty_output_hook.py").write_text("# cached")
 
     plugins_json = tmp_path / ".claude" / "plugins" / "installed_plugins.json"
-    plugins_json.write_text('{"autoskillit@autoskillit-local": {"version": "0.8.25"}}')
+    plugins_json.write_text(
+        json.dumps(
+            {"version": 2, "plugins": {"autoskillit@autoskillit-local": {"version": "0.8.25"}}}
+        )
+    )
 
     cli.doctor()
 
     assert cache_dir.exists(), "Doctor must not delete the plugin cache directory"
     data = json.loads(plugins_json.read_text())
-    assert "autoskillit@autoskillit-local" in data, (
+    assert "autoskillit@autoskillit-local" in data.get("plugins", {}), (
         "Doctor must not remove installed_plugins.json entries"
     )
 
@@ -1776,3 +1780,30 @@ def test_doctor_no_dual_when_only_direct(tmp_path: Path, monkeypatch: pytest.Mon
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
     result = _check_dual_mcp_registration()
     assert result.severity == Severity.OK
+
+
+def test_check_installed_plugins_entry_real_structure_is_ok(tmp_path: Path) -> None:
+    """With the real nested format, the check must report OK."""
+    from autoskillit.cli._doctor import _check_installed_plugins_entry
+
+    p = tmp_path / "installed_plugins.json"
+    p.write_text(
+        json.dumps(
+            {
+                "version": 2,
+                "plugins": {"autoskillit@autoskillit-local": {"name": "autoskillit"}},
+            }
+        )
+    )
+    result = _check_installed_plugins_entry(plugins_json_path=p)
+    assert result.severity.name == "OK"
+
+
+def test_check_installed_plugins_entry_flat_structure_is_warning(tmp_path: Path) -> None:
+    """A flat structure (wrong format) must not be silently treated as OK."""
+    from autoskillit.cli._doctor import _check_installed_plugins_entry
+
+    p = tmp_path / "installed_plugins.json"
+    p.write_text(json.dumps({"autoskillit@autoskillit-local": {}}))
+    result = _check_installed_plugins_entry(plugins_json_path=p)
+    assert result.severity.name == "WARNING"
--- a/tests/cli/test_doctor.py
+++ b/tests/cli/test_doctor.py
@@ -69,7 +69,6 @@ class TestCLIDoctor:
                 {
                     "mcpServers": {
                         "other-server": {"type": "stdio", "command": str(fake_bin)},
-                        "autoskillit": {"type": "stdio", "command": "autoskillit"},
                     }
                 }
             )
--- a/tests/cli/test_install.py
+++ b/tests/cli/test_install.py
@@ -678,3 +678,74 @@ def test_install_sweeps_all_scopes_for_orphans(
                 assert "pretty_output.py" not in hook["command"], (
                     "Orphaned hook was not evicted from project scope"
                 )
+
+
+def test_install_creates_autoskillit_gitignore(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """After install(), .autoskillit/.gitignore must exist (ensure_project_temp was called)."""
+    import importlib as _importlib
+
+    from autoskillit.cli._marketplace import install as _install
+
+    _app_mod = _importlib.import_module("autoskillit.cli._marketplace")
+    monkeypatch.setattr(_app_mod, "is_git_worktree", lambda path: False)
+    monkeypatch.setattr(Path, "home", lambda: tmp_path)
+    monkeypatch.setattr(Path, "cwd", lambda: tmp_path)
+    monkeypatch.delenv("CLAUDECODE", raising=False)
+    monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/claude")
+    monkeypatch.setattr(
+        "subprocess.run",
+        lambda *a, **kw: type("R", (), {"returncode": 0, "stdout": "", "stderr": ""})(),
+    )
+    monkeypatch.setattr("autoskillit.cli._marketplace.evict_direct_mcp_entry", lambda _: False)
+    monkeypatch.setattr(
+        "autoskillit.cli._marketplace.sweep_all_scopes_for_orphans", lambda _: None
+    )
+    monkeypatch.setattr("autoskillit.cli._marketplace.sync_hooks_to_settings", lambda _: None)
+    monkeypatch.setattr("autoskillit.cli._marketplace.generate_hooks_json", lambda: {})
+    monkeypatch.setattr("autoskillit.cli._marketplace.atomic_write", lambda *a, **kw: None)
+
+    (tmp_path / ".autoskillit").mkdir()
+    _install(scope="user")
+
+    assert (tmp_path / ".autoskillit" / ".gitignore").exists(), (
+        ".autoskillit/.gitignore must be created by install(), not just by init()"
+    )
+
+
+def test_install_calls_upgrade_when_scripts_dir_exists(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """install() must migrate .autoskillit/scripts/ → .autoskillit/recipes/ if scripts/ exists."""
+    import importlib as _importlib
+
+    from autoskillit.cli._marketplace import install as _install
+
+    _app_mod = _importlib.import_module("autoskillit.cli._marketplace")
+    monkeypatch.setattr(_app_mod, "is_git_worktree", lambda path: False)
+    monkeypatch.setattr(Path, "home", lambda: tmp_path)
+    monkeypatch.setattr(Path, "cwd", lambda: tmp_path)
+    monkeypatch.delenv("CLAUDECODE", raising=False)
+    monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/claude")
+    monkeypatch.setattr(
+        "subprocess.run",
+        lambda *a, **kw: type("R", (), {"returncode": 0, "stdout": "", "stderr": ""})(),
+    )
+    monkeypatch.setattr("autoskillit.cli._marketplace.evict_direct_mcp_entry", lambda _: False)
+    monkeypatch.setattr(
+        "autoskillit.cli._marketplace.sweep_all_scopes_for_orphans", lambda _: None
+    )
+    monkeypatch.setattr("autoskillit.cli._marketplace.sync_hooks_to_settings", lambda _: None)
+    monkeypatch.setattr("autoskillit.cli._marketplace.generate_hooks_json", lambda: {})
+    monkeypatch.setattr("autoskillit.cli._marketplace.atomic_write", lambda *a, **kw: None)
+
+    scripts_dir = tmp_path / ".autoskillit" / "scripts"
+    scripts_dir.mkdir(parents=True)
+
+    _install(scope="user")
+
+    assert (tmp_path / ".autoskillit" / "recipes").exists(), (
+        "install() must migrate scripts/ to recipes/ when scripts/ exists"
+    )
+    assert not scripts_dir.exists(), "scripts/ must be renamed away"
--- a/tests/cli/test_install.py
+++ b/tests/cli/test_install.py
@@ -555,6 +555,85 @@ class TestGroupFInstall:
         )
 
 
+def test_clear_plugin_cache_removes_nested_entry(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """_clear_plugin_cache must remove the entry from data['plugins'], not top-level data."""
+    from autoskillit.cli._marketplace import _clear_plugin_cache
+
+    plugins_dir = tmp_path / ".claude" / "plugins"
+    plugins_dir.mkdir(parents=True)
+    installed_json = plugins_dir / "installed_plugins.json"
+    installed_json.write_text(
+        json.dumps(
+            {
+                "version": 2,
+                "plugins": {"autoskillit@autoskillit-local": {"name": "autoskillit"}},
+            }
+        )
+    )
+    monkeypatch.setattr(Path, "home", lambda: tmp_path)
+    _clear_plugin_cache()
+    data = json.loads(installed_json.read_text())
+    assert "autoskillit@autoskillit-local" not in data.get("plugins", {})
+    assert data["version"] == 2  # other keys preserved
+
+
+def test_clear_plugin_cache_noop_when_entry_absent(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """_clear_plugin_cache must not raise when the entry is already absent."""
+    from autoskillit.cli._marketplace import _clear_plugin_cache
+
+    plugins_dir = tmp_path / ".claude" / "plugins"
+    plugins_dir.mkdir(parents=True)
+    installed_json = plugins_dir / "installed_plugins.json"
+    installed_json.write_text(json.dumps({"version": 2, "plugins": {}}))
+    monkeypatch.setattr(Path, "home", lambda: tmp_path)
+    _clear_plugin_cache()  # must not raise
+    data = json.loads(installed_json.read_text())
+    assert data == {"version": 2, "plugins": {}}
+
+
+def test_install_claudecode_guard_returns_false(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """install() must return False when CLAUDECODE guard fires, not None-as-success."""
+    import importlib as _importlib
+
+    from autoskillit.cli._marketplace import install as _install
+
+    _app_mod = _importlib.import_module("autoskillit.cli._marketplace")
+    monkeypatch.setattr(_app_mod, "is_git_worktree", lambda path: False)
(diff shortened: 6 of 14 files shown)


tests/cli/test_doctor.py
```diff
@@ -1776,3 +1779,30 @@ def test_doctor_no_dual_when_only_direct(tmp_path: Path, monkeypatch: pytest.Mon
...
+    p.write_text(
+        json.dumps(
+            {
+                "version": 2,
+                "plugins": {"autoskillit@autoskillit-local": {"name": "autoskillit"}},
+            }
+        )
+    )
+    result = _check_installed_plugins_entry(plugins_json_path=p)
+    assert result.severity.name == "OK"
```
Trecek: [warning] tests: Assertion uses `result.severity.name == "OK"` (string comparison on enum name) instead of `result.severity == Severity.OK`. This bypasses type safety — if the Severity enum is ever refactored the string comparison silently succeeds or fails for the wrong reason. Import and compare the enum directly.

Trecek (commented):

9 blocking findings (warning severity) require fixes before merge. See inline comments for details.

**Summary:**
- defense: 4 warnings (InstalledPluginsFile silent error swallowing, corrupt file overwrite, unguarded upgrade() exception, missing return type)
- tests: 4 warnings (enum name string comparisons x2, broad subprocess.run patches x2)
- cohesion: 1 warning (_ensure_workspace_ready mixes concerns, hides migration side-effect)

Additionally: 13 info-level findings (see inline comments)

tests/cli/test_doctor.py
```diff
@@ -1776,3 +1779,30 @@ def test_doctor_no_dual_when_only_direct(tmp_path: Path, monkeypatch: pytest.Mon
...
+
+
+def test_check_installed_plugins_entry_flat_structure_is_warning(tmp_path: Path) -> None:
+    """A flat structure (wrong format) must not be silently treated as OK."""
+    from autoskillit.cli._doctor import _check_installed_plugins_entry
+
+    p = tmp_path / "installed_plugins.json"
+    p.write_text(json.dumps({"autoskillit@autoskillit-local": {}}))
+    result = _check_installed_plugins_entry(plugins_json_path=p)
+    assert result.severity.name == "WARNING"
```
Trecek: [warning] tests: Assertion uses `result.severity.name == "WARNING"` (string comparison on enum name) instead of `result.severity == Severity.WARNING`. Same type-safety concern as the sibling test above. Use the enum directly.

tests/cli/test_install.py
```diff
@@ -599,3 +678,74 @@ def test_install_sweeps_all_scopes_for_orphans(
...
+    import importlib as _importlib
+
+    from autoskillit.cli._marketplace import install as _install
+
+    _app_mod = _importlib.import_module("autoskillit.cli._marketplace")
+    monkeypatch.setattr(_app_mod, "is_git_worktree", lambda path: False)
+    monkeypatch.setattr(Path, "home", lambda: tmp_path)
+    monkeypatch.setattr(Path, "cwd", lambda: tmp_path)
+    monkeypatch.delenv("CLAUDECODE", raising=False)
+    monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/claude")
```
Trecek, line 696: [warning] tests: `monkeypatch.setattr("subprocess.run", ...)` patches the global `subprocess.run` symbol. This is overly broad — any code in the process that calls `subprocess.run` during this test will silently get the stub. Prefer patching the module-level reference such as `autoskillit.cli._marketplace.subprocess.run` to limit blast radius.

tests/cli/test_install.py
```diff
@@ -599,3 +678,74 @@ def test_install_sweeps_all_scopes_for_orphans(
...
```
Trecek, line 730: [warning] tests: Same broad `subprocess.run` global patch as `test_install_creates_autoskillit_gitignore`. Patch the module-level reference instead to avoid an overly wide mock scope.

src/autoskillit/cli/_installed_plugins.py
```diff
@@ -0,0 +1,65 @@
...
+    @property
+    def path(self) -> Path:
+        return self._path
+
+    def _read(self) -> dict[str, Any]:
+        if not self._path.exists():
+            return {}
+        try:
+            return json.loads(self._path.read_text())
+        except (json.JSONDecodeError, OSError):
```
Trecek, line 41: [warning] defense: `_read()` silently swallows OSError and json.JSONDecodeError by returning an empty dict without any logging. Callers like `contains()` and `remove()` cannot distinguish "file was corrupt/unreadable" from "file legitimately has no data". The original `_doctor.py` code surfaced a WARNING DoctorResult on parse failure; that error-context is now permanently lost.

src/autoskillit/cli/_installed_plugins.py
```diff
@@ -0,0 +1,65 @@
...
+            return {}
+
+    def get_plugins(self) -> dict[str, Any]:
```
Trecek, line 44: [warning] defense: `get_plugins()` is a public method with no return type annotation (`-> dict[str, Any]` is missing). Public methods on a repository class should carry full type annotations at the boundary.

src/autoskillit/cli/_installed_plugins.py
```diff
@@ -0,0 +1,65 @@
...
+        """Return the nested plugins dict (never raises)."""
+        return self._read().get("plugins", {})
+
+    def contains(self, plugin_ref: str) -> bool:
+        """Return True iff plugin_ref is present in data['plugins']."""
+        return plugin_ref in self.get_plugins()
+
+    def remove(self, plugin_ref: str) -> None:
```
Trecek, line 52: [warning] defense: `remove()` calls `_read()` which silently returns `{}` on parse failure, then immediately overwrites the file with `json.dumps({...})`. If the file exists but is corrupt, this will silently replace a potentially-recoverable file with an empty-plugins structure, destroying partial data.

src/autoskillit/cli/_marketplace.py
```diff
@@ -97,7 +94,25 @@ def _ensure_marketplace() -> Path:
     return marketplace_dir
 
 
-def install(*, scope: str = "user"):
+def _ensure_workspace_ready() -> None:
```
Trecek, line 97: [warning] cohesion: `_ensure_workspace_ready()` mixes two concerns — filesystem bootstrap (`ensure_project_temp`) and recipe migration (`upgrade()`). Calling `upgrade()` silently inside `install()` whenever `scripts/` exists hides a side-effect migration. `install()`'s docstring does not mention this migration, so the behaviour is invisible to callers.

src/autoskillit/cli/_marketplace.py
```diff
@@ -97,7 +94,25 @@ def _ensure_marketplace() -> Path:
...
+    from autoskillit.core import ensure_project_temp
+
+    project_dir = Path.cwd()
+    # Repair .autoskillit/.gitignore and ensure temp/ exists
+    if (project_dir / ".autoskillit").is_dir():
+        ensure_project_temp(project_dir)
+
+    # Migrate legacy .autoskillit/scripts/ to .autoskillit/recipes/ if present
+    if (project_dir / ".autoskillit" / "scripts").exists():
+        upgrade()
```
Trecek: [warning] defense: `_ensure_workspace_ready()` calls `upgrade()` without guarding against exceptions. If the migration fails (e.g. `scripts/` rename raises PermissionError or `recipes/` already exists), `install()` will abort with an unhandled exception rather than surfacing a clean error message to the user.

tests/cli/test_doctor.py
```diff
@@ -1776,3 +1779,30 @@ def test_doctor_no_dual_when_only_direct(tmp_path: Path, monkeypatch: pytest.Mon
...
+                "version": 2,
+                "plugins": {"autoskillit@autoskillit-local": {"name": "autoskillit"}},
+            }
+        )
+    )
+    result = _check_installed_plugins_entry(plugins_json_path=p)
+    assert result.severity.name == "OK"
+
+
+def test_check_installed_plugins_entry_flat_structure_is_warning(tmp_path: Path) -> None:
```
Trecek, line 1802: [info] tests: `test_check_installed_plugins_entry_flat_structure_is_warning` only asserts severity is WARNING but not the check name or message. The WARNING could be triggered by the file-not-found path rather than the bad-format path, since the flat file has no `plugins` key. Consider asserting `result.check` to confirm the right code path fired.

tests/cli/test_installed_plugins_file.py
```diff
@@ -0,0 +1,81 @@
+"""Unit tests for the InstalledPluginsFile repository."""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+from autoskillit.cli._installed_plugins import InstalledPluginsFile
+
+REAL_STRUCTURE = {
```
Trecek, line 10: [info] tests: `REAL_STRUCTURE` is a module-level mutable dict. While no test mutates it directly, best practice for shared test fixtures is to use `dict.copy()` or a factory function to prevent accidental cross-test contamination if the structure is ever passed somewhere that mutates it.

tests/cli/test_install.py
```diff
@@ -599,3 +678,74 @@ def test_install_sweeps_all_scopes_for_orphans(
...
+        "autoskillit.cli._marketplace.sweep_all_scopes_for_orphans", lambda _: None
+    )
+    monkeypatch.setattr("autoskillit.cli._marketplace.sync_hooks_to_settings", lambda _: None)
+    monkeypatch.setattr("autoskillit.cli._marketplace.generate_hooks_json", lambda: {})
+    monkeypatch.setattr("autoskillit.cli._marketplace.atomic_write", lambda *a, **kw: None)
+
+    scripts_dir = tmp_path / ".autoskillit" / "scripts"
+    scripts_dir.mkdir(parents=True)
+
+    _install(scope="user")
```
Trecek, line 746: [info] tests: This test exercises the real `upgrade()` function (not mocked) alongside heavily-mocked `install()` helpers. If `upgrade()` has side-effects beyond renaming the directory, those will execute against `tmp_path`. This is intentional integration behavior but should be documented in the docstring.

tests/test_hook_registry.py
```diff
@@ -130,3 +133,55 @@ def test_hook_registry_uses_new_filenames() -> None:
...
+            }
+        )
+    )
+    broken = find_broken_hook_scripts(settings)
+    assert broken == []
+
+
+def test_find_broken_hook_scripts_flags_missing_autoskillit_script(tmp_path: Path) -> None:
+    """A python3 /abs/path/autoskillit/hooks/script.py that doesn't exist must be flagged."""
+    missing = str(HOOKS_DIR / "nonexistent_guard.py")
```
Trecek, line 156: [info] tests: `test_find_broken_hook_scripts_flags_missing_autoskillit_script` relies on `HOOKS_DIR` being inside the autoskillit package tree for `_is_own_hook` to return True. If `HOOKS_DIR` is ever relocated, the test may vacuously pass with an empty `broken` list. Consider asserting `_is_own_hook` returns True for the constructed path as a guard.

src/autoskillit/cli/_marketplace.py
```diff
@@ -35,16 +36,12 @@ def _clear_plugin_cache() -> None:
...
-            if plugin_ref in data:
-                del data[plugin_ref]
-                atomic_write(installed_json, json.dumps(data, indent=2))
-        except (OSError, json.JSONDecodeError):
-            pass  # non-fatal — install will proceed regardless
+    from autoskillit.cli._installed_plugins import InstalledPluginsFile
+
+    try:
+        InstalledPluginsFile().remove(f"autoskillit@{_MARKETPLACE_NAME}")
+    except OSError:
```
Trecek, line 43: [info] defense: The outer `except OSError` in `_clear_plugin_cache` is effectively dead for json errors — `InstalledPluginsFile.remove()` internally swallows both OSError and json.JSONDecodeError before they can propagate. The outer try/except is misleading about the actual error surface.

src/autoskillit/cli/_marketplace.py
```diff
@@ -35,16 +36,12 @@ def _clear_plugin_cache() -> None:
...
-    if installed_json.exists():
-        try:
-            data = json.loads(installed_json.read_text())
-            plugin_ref = f"autoskillit@{_MARKETPLACE_NAME}"
-            if plugin_ref in data:
-                del data[plugin_ref]
-                atomic_write(installed_json, json.dumps(data, indent=2))
-        except (OSError, json.JSONDecodeError):
-            pass  # non-fatal — install will proceed regardless
+    from autoskillit.cli._installed_plugins import InstalledPluginsFile
```
Trecek, line 39: [info] cohesion: `InstalledPluginsFile` is imported with a deferred `from` inside `_clear_plugin_cache()`. All other references also use deferred imports. For a class introduced to centralize access, always importing it lazily is inconsistent with how other cli/ collaborators are imported at module scope. Either promote to module level or add a comment explaining why deferred (e.g. circular import risk).

src/autoskillit/cli/_doctor.py
```diff
@@ -609,24 +609,16 @@ def _check_installed_plugins_entry(
     plugins_json_path: Path | None = None,
 ) -> DoctorResult:
     """Check that installed_plugins.json contains the autoskillit entry."""
-    _plugins_json = plugins_json_path or (
-        Path.home() / ".claude" / "plugins" / "installed_plugins.json"
-    )
-    if not _plugins_json.exists():
+    from autoskillit.cli._installed_plugins import InstalledPluginsFile
```
Trecek, line 612: [info] cohesion: `InstalledPluginsFile` is imported deferred inside `_check_installed_plugins_entry()`, inconsistent with module-level import style used for other doctor helpers in the same file. The pattern should be uniform across all call sites.

src/autoskillit/cli/_marketplace.py
```diff
@@ -97,7 +94,25 @@ def _ensure_marketplace() -> Path:
...
+    # Repair .autoskillit/.gitignore and ensure temp/ exists
+    if (project_dir / ".autoskillit").is_dir():
+        ensure_project_temp(project_dir)
+
+    # Migrate legacy .autoskillit/scripts/ to .autoskillit/recipes/ if present
+    if (project_dir / ".autoskillit" / "scripts").exists():
+        upgrade()
+
+
+def install(*, scope: str = "user") -> bool:
```
Trecek, line 118: [info] cohesion: `install()` now returns `bool` but `upgrade()` (sibling function in the same file) still implicitly returns `None`. If the return-value check pattern from `app.py` is ever needed for `upgrade`, this asymmetry will cause a silent bug.

src/autoskillit/cli/_marketplace.py
```diff
@@ -97,7 +94,25 @@ def _ensure_marketplace() -> Path:
...
+def _ensure_workspace_ready() -> None:
+    """Repair project workspace state that install() is responsible for.
+
+    Called after the CLAUDECODE guard — only when the actual install proceeds.
+    Idempotent: safe to call on any project state.
+    """
+    from autoskillit.core import ensure_project_temp
+
+    project_dir = Path.cwd()
+    # Repair .autoskillit/.gitignore and ensure temp/ exists
```
Trecek, line 106: [info] slop: Inline comment restates what `ensure_project_temp` already makes obvious. The function docstring covers the intent; this comment adds no new information.

src/autoskillit/cli/_marketplace.py
```diff
@@ -97,7 +94,25 @@ def _ensure_marketplace() -> Path:
...
+    if (project_dir / ".autoskillit").is_dir():
+        ensure_project_temp(project_dir)
+
+    # Migrate legacy .autoskillit/scripts/ to .autoskillit/recipes/ if present
```
Trecek, line 110: [info] slop: Inline comment restates the condition and call below it verbatim. The condition `(... / "scripts").exists()` and the `upgrade()` call are self-documenting.

src/autoskillit/cli/_installed_plugins.py
```diff
@@ -0,0 +1,65 @@
...
+            return json.loads(self._path.read_text())
+        except (json.JSONDecodeError, OSError):
+            return {}
+
+    def get_plugins(self) -> dict[str, Any]:
+        """Return the nested plugins dict (never raises)."""
+        return self._read().get("plugins", {})
+
+    def contains(self, plugin_ref: str) -> bool:
+        """Return True iff plugin_ref is present in data['plugins']."""
```
Trecek, line 49: [info] slop: Docstring on `contains` adds marginal value over the one-liner body `return plugin_ref in self.get_plugins()`. Could be dropped or reduced to a one-liner.

src/autoskillit/cli/_marketplace.py
```diff
@@ -125,7 +140,7 @@ def install(*, scope: str = "user"):
         print(f"  claude plugin marketplace add {marketplace_dir}")
         print(f"  claude plugin install {plugin_ref} --scope {scope}")
         print("\nThen run: autoskillit init (in your project directory)")
-        return
+        return False  # deferred: user must complete manually in a regular terminal
```
Trecek, line 146: [info] slop: Inline comment on `return False` restates the CLAUDECODE guard rationale already printed to stdout in the preceding print block.

src/autoskillit/hook_registry.py
```diff
@@ -281,6 +281,8 @@ def find_broken_hook_scripts(settings_path: Path) -> list[str]:
         for entry in data.get("hooks", {}).get(event_type, []):
             for hook in entry.get("hooks", []):
                 cmd = hook.get("command", "")
+                if not _is_own_hook(cmd):  # skip non-autoskillit hooks
```
Trecek, line 284: [info] slop: Inline comment `# skip non-autoskillit hooks` restates the function name `_is_own_hook` exactly. The comment adds no information beyond what the code already expresses.

tests/cli/test_doctor.py
Trecek: Agreed — fixed in b78355bf. Added `from autoskillit.core import Severity` import and switched to direct enum comparison `result.severity == Severity.OK`.

tests/cli/test_doctor.py
Trecek: Agreed — fixed in b78355bf. Added `from autoskillit.core import Severity` import and switched to direct enum comparison `result.severity == Severity.WARNING`.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in e89139e3. Wrapped `upgrade()` call in try/except OSError with a non-fatal warning print so install() continues even if migration fails (PermissionError, cross-device rename, etc.).

tests/cli/test_install.py
Trecek: Investigated — this is intentional. The broad `monkeypatch.setattr('subprocess.run', ...)` pattern is established across 5+ tests in this file (lines 474, 511, 538, 698, 732). monkeypatch restores state after each test automatically. The concern about over-broad scope is real in principle but doesn't apply here since all subprocess.run callers in the test path are already individually mocked.

tests/cli/test_install.py
Trecek: Investigated — this is intentional. Same as sibling comment 3081497440: the broad subprocess.run patch is consistent with 5+ prior tests in the same file (category: false_positive_intentional_pattern). monkeypatch restores state after each test automatically.

src/autoskillit/cli/_installed_plugins.py
Trecek: Investigated — this is intentional. The actual code at line 44 reads `def get_plugins(self) -> dict[str, Any]:`. The `-> dict[str, Any]` return type annotation is present. The reviewer's claim that the annotation is missing is factually incorrect (category: false_positive_intentional_pattern).

src/autoskillit/cli/_installed_plugins.py
Trecek: Investigated — this is intentional. The reviewer's concern is based on a misread of the control flow. When `_read()` returns `{}` due to corruption: `plugins = {}.get('plugins', {}) = {}`, then `if plugin_ref not in plugins: return` at line 61 always short-circuits — `atomic_write` is never reached. The corrupt-file case is handled correctly by the early-exit guard (category: false_positive_intentional_pattern).

src/autoskillit/cli/_installed_plugins.py
Trecek: Valid observation — flagged for design decision. The 'never raises' contract in `_read()` is intentional (see `get_plugins()` docstring). Both `contains()` and `remove()` handle `{}` gracefully. Whether to add a debug-level log for operational visibility is a policy question for the team.

src/autoskillit/cli/_marketplace.py
Trecek: Valid observation — flagged for design decision. The mixing of `ensure_project_temp` and `upgrade()` in `_ensure_workspace_ready()` was designed explicitly in commit 8f45bb18 to address 'Design Gap 5: install() never called upgrade() when scripts/ existed.' The `upgrade()` call is now also guarded with try/except OSError (see commit e89139e3). Whether to further separate the concerns is an architectural decision for the team.

PR #969, opened 2026-04-16
AST Import Walker Module

### LINKED ISSUE #931: Create AST Import Walker module
**Category:** Filter Infrastructure
**Depends on:** #01-#05 (structural refactoring complete)

## Problem

No mechanism exists to determine which source modules a test file depends on. This is the foundation for deciding which tests to skip based on git diff.

## Solution

Create `tests/_test_filter.py` — a standalone Python module containing all pure filtering logic. Importable and testable independently, no pytest dependencies in the logic layer.

### Components

**`FilterMode` enum**: `NONE`, `CONSERVATIVE`, `AGGRESSIVE`

**`ImportContext` enum**: `TOP_LEVEL`, `CONDITIONAL`, `TYPE_CHECKING`, `DEFERRED`, `IMPORTLIB`

**`ASTImportWalker(ast.NodeVisitor)`**: Extracts all import statements with context tracking.
- Visits `Import`, `ImportFrom`, `FunctionDef`, `AsyncFunctionDef`, `If`, `Try`, `Call`
- Detects `TYPE_CHECKING` guards via `ast.Name(id="TYPE_CHECKING")` or `ast.Attribute(attr="TYPE_CHECKING")`
- Detects `importlib.import_module("literal")` via `visit_Call` + `ast.Constant` check
- Returns list of `(module_path: str, context: ImportContext)` tuples
- All contexts count for triggering (TYPE_CHECKING and DEFERRED included)

**`load_manifest(path)`**: Loads `.autoskillit/test-filter-manifest.yaml` from the given path.

**`apply_manifest(changed_files, manifest)`**: Applies the manifest to a set of changed files, returning the union of test directories for all matched patterns.

**`git_changed_files(cwd, base_ref)`**: Subprocess wrapper.
- Handles `GITHUB_BASE_REF`, `AUTOSKILLIT_TEST_BASE_REF`, detached HEAD
- Uses three-dot notation: `base_ref...HEAD`
- Returns `set[str] | None` (None = unavailable, triggers fail-open)
- 10s timeout on subprocess

**`check_bucket_a(changed_files)`**: Full-run trigger check.
- Checks against `BUCKET_A_PATTERNS` frozenset and `BUCKET_A_GLOBS` tuple
- Patterns: `tests/conftest.py`, `tests/_helpers.py`, `tests/arch/_helpers.py`, `tests/arch/_rules.py`, `pyproject.toml`, `uv.lock`, `.pre-commit-config.yaml`, `src/autoskillit/server/_factory.py`
- Globs: `tests/*/conftest.py`
- Returns `True` if any file matches

**Why `_factory.py` is in Bucket A:** `tool_ctx` fixture (root conftest) calls `make_context()` from `server/_factory.py`, which imports all 8 production layers. Since `tool_ctx` is used by tests in `tests/execution/`, `tests/recipe/`, `tests/contracts/`, and `tests/infra/` (not just `tests/server/`), a change to `_factory.py` affects test files across many directories that the normal server cascade wouldn't reach. Treating it as Bucket A is the safe, correct choice.

**Layer cascade maps**: `LAYER_CASCADE_CONSERVATIVE` and `LAYER_CASCADE_AGGRESSIVE` dicts mapping source packages to test directories.

#### Critical: cascade maps must account for cross-layer test imports

The hardcoded cascade maps are initially based on the SOURCE architecture (L0-L3 layers), but test files have their own cross-layer imports that the source architecture doesn't capture. After tickets #01-#05 are implemented, the implementer MUST validate the cascade maps against actual test file imports using the AST walker.

Known cross-layer test imports that survive the refactoring:
- `test_headless.py` imports from `pipeline.gate`, `pipeline.audit`, `recipe.contracts` — so `pipeline` and `recipe` cascades must include `tests/execution/`
- `test_session.py` imports from `server.tools_execution` — so `server` cascade must include `tests/execution/` (it already does)
- `test_recording.py` imports from `server._factory` — same as above

In conservative mode (no per-file AST refinement), the cascade map is the ONLY mechanism for routing. If `pipeline` doesn't cascade to `tests/execution/`, a change to `pipeline/gate.py` would silently miss `test_headless.py`. The cascade map for conservative mode must be wider than the source architecture — it must include every test directory that has ANY test file importing from that package.

Recommended approach: after tickets #01-#05 land, run the AST walker over all test files to compute the actual test-level import graph, and use that to derive the final conservative cascade map. Hardcode the result.

#### Transitive re-export closure

When a changed file (e.g., `core/_type_subprocess.py`) is re-exported through `__init__.py` chains, the scope builder must expand the changed-modules set to include all packages that re-export from the changed file. Implementation: for each changed source file, walk up through `__init__.py` files that import from it and add those packages to the affected set.

**Always-run sets**: `ALWAYS_RUN_CONSERVATIVE` (arch, contracts, infra, docs) and `ALWAYS_RUN_AGGRESSIVE` (arch, contracts).

**`build_test_scope(changed_files, mode, manifest, tests_root, src_root)`**: Main scope computation function. Returns `set[Path] | None` (None = run everything).

### AST parsing performance

~400 files at ~200 lines average: 0.3-1.5s cold. No persistent cache needed.

## Acceptance Criteria

- [ ] `tests/_test_filter.py` exists with all components listed above
- [ ] ASTImportWalker correctly extracts: top-level, from, deferred, conditional, TYPE_CHECKING, importlib literal
- [ ] `git_changed_files` handles all edge cases (failure, timeout, detached HEAD)
- [ ] `check_bucket_a` catches all trigger patterns
- [ ] `build_test_scope` returns correct scope for all layer combinations
- [ ] Module has no pytest dependencies (pure Python + ast + subprocess)
- [ ] Comprehensive test suite in `tests/test_test_filter.py` (~30 test cases)
- [ ] Large changeset (>30 files) triggers fail-open

### Test cases to write

- Walker: 8 cases (all import types + context tracking)
- Bucket A: 8 cases (all trigger patterns + negative cases)
- build_test_scope: 10 cases (L0-L3 changes, modes, edge cases)
- Conservative vs aggressive: 4 cases
- Large changeset fallback: 2 cases
- Git diff edge cases: 5 cases

## Context from plan

See plan sections: 5 (Filter Design), 8 (Filter Infrastructure)

## Design details from research

The `build_test_scope` algorithm:
1. If changed_files is None → return None (run everything)
2. If len(changed_files) > 30 → return None (large changeset fallback)
3. Check Bucket A → if triggered, return None
4. Classify changed files: source Python → layer cascade; non-Python → manifest; test Python → include directly; unknown → fail-open
5. Compute always-run set for mode
6. If any src/**/*.py changed → add arch/ and contracts/ (Step 6 operates IN ADDITION to step 4's cascade — it unconditionally adds arch/ and contracts/ whenever any src Python file changed, regardless of which layers the cascade already included.)
7. Union all sets
8. For aggressive mode: AST-based file-level refinement (check each test file's imports against changed modules)
9. Resolve to concrete test file paths
10. Return the set

Trecek:
## Summary

Create `tests/_test_filter.py` — a standalone, pytest-free Python module that determines which test files to run based on `git diff` output. The module contains: `FilterMode` and `ImportContext` enums, `ASTImportWalker` (an `ast.NodeVisitor` subclass), `load_manifest`/`apply_manifest` for YAML-based overrides, `git_changed_files` for subprocess-based diff extraction, `check_bucket_a` for full-run triggers, hardcoded layer cascade maps derived from the actual cross-layer test import graph, transitive re-export closure logic, and `build_test_scope` as the top-level entry point. A companion test suite in `tests/test_test_filter.py` validates all components with ~30 test cases.

## Requirements

- [ ] `tests/_test_filter.py` exists with all components listed above
- [ ] ASTImportWalker correctly extracts: top-level, from, deferred, conditional, TYPE_CHECKING, importlib literal
- [ ] `git_changed_files` handles all edge cases (failure, timeout, detached HEAD)
- [ ] `check_bucket_a` catches all trigger patterns
- [ ] `build_test_scope` returns correct scope for all layer combinations
- [ ] Module has no pytest dependencies (pure Python + ast + subprocess)
- [ ] Comprehensive test suite in `tests/test_test_filter.py` (~30 test cases)
- [ ] Large changeset (>30 files) triggers fail-open

Closes #931

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-931-20260415-160044-646864/.autoskillit/temp/make-plan/ast_import_walker_plan_2026-04-15_160600.md`

## Token Usage Summary

| make_plan | 5.0k | 13.4k | 710.4k | 53.5k | 1 | 9m 54s |
| review_approach | 12 | 3.7k | 135.1k | 7.9k | 1 | 5m 55s |
| dry_walkthrough | 63 | 24.1k | 1.8M | 79.1k | 1 | 9m 54s |
| implement | 56 | 12.7k | 815.9k | 42.8k | 1 | 4m 34s |
| resolve_failures | 42 | 3.4k | 464.1k | 29.8k | 1 | 3m 46s |
| prepare_pr | 27 | 4.4k | 270.7k | 29.5k | 1 | 1m 40s |
| run_arch_lenses | 34 | 7.6k | 234.4k | 45.0k | 1 | 3m 53s |
| compose_pr | 23 | 1.9k | 130.8k | 14.7k | 1 | 51s |
| **Total** | 5.2k | 71.1k | 4.5M | 302.3k | | 40m 29s |

### DIFF
--- a/tests/_test_filter.py
+++ b/tests/_test_filter.py
@@ -6,6 +6,7 @@ import ast
 import enum
 import fnmatch
 import subprocess
+import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -228,8 +229,10 @@ class ASTImportWalker(ast.NodeVisitor):
             for child in node.body:
                 self.visit(child)
             self._context_stack.pop()
+            self._context_stack.append(ImportContext.CONDITIONAL)
             for child in node.orelse:
                 self.visit(child)
+            self._context_stack.pop()
         else:
             self._context_stack.append(ImportContext.CONDITIONAL)
             self.generic_visit(node)
@@ -292,7 +295,14 @@ def git_changed_files(
             timeout=10,
             check=True,
         )
-    except (subprocess.CalledProcessError, subprocess.TimeoutExpired, FileNotFoundError):
+    except subprocess.CalledProcessError as exc:
+        warnings.warn(f"git diff failed (exit {exc.returncode}): {exc.stderr or ''}", stacklevel=2)
+        return None
+    except subprocess.TimeoutExpired:
+        warnings.warn("git diff timed out after 10s", stacklevel=2)
+        return None
+    except FileNotFoundError:
+        warnings.warn("git binary not found on PATH", stacklevel=2)
         return None
 
     lines = result.stdout.strip().splitlines()
@@ -317,8 +327,15 @@ def load_manifest(path: str | Path) -> dict[str, Any] | None:
     manifest_path = Path(path) / ".autoskillit" / "test-filter-manifest.yaml"
     if not manifest_path.exists():
         return None
-    with manifest_path.open() as f:
-        return yaml.safe_load(f)
+    try:
+        with manifest_path.open() as f:
+            return yaml.safe_load(f)
+    except OSError as exc:
+        warnings.warn(f"Cannot read manifest {manifest_path}: {exc}", stacklevel=2)
+        return None
+    except yaml.YAMLError as exc:
+        warnings.warn(f"Malformed YAML in {manifest_path}: {exc}", stacklevel=2)
+        return None
 
 
 def apply_manifest(
@@ -374,13 +391,16 @@ def _expand_reexport_closure(
                     walker = ASTImportWalker()
                     walker.visit(tree)
                     for mod, _ctx in walker.imports:
-                        bare = mod.lstrip(".")
-                        if bare == module_name or bare.endswith(f".{module_name}"):
+                        bare = mod.split(".")[-1] if "." in mod else mod.lstrip(".")
+                        if bare == module_name:
                             rel_init = str(init_path.relative_to(src_root))
                             expanded.add(rel_init)
                             break
-                except SyntaxError:
-                    pass
+                except SyntaxError as exc:
+                    warnings.warn(
+                        f"Failed to parse {init_path}: {exc}",
+                        stacklevel=2,
+                    )
             parent = parent.parent
 
     return expanded
@@ -417,7 +437,6 @@ def build_test_scope(
     mode: FilterMode,
     manifest: dict[str, Any] | None = None,
     tests_root: str | Path = "tests",
-    src_root: str | Path = "src/autoskillit",
 ) -> set[Path] | None:
     """Compute the set of test paths to run, or None for a full run.
 
--- a/tests/_test_filter.py
+++ b/tests/_test_filter.py
@@ -1,0 +1,416 @@
+"""Standalone test-path filtering logic. No pytest dependencies."""
+from __future__ import annotations
+
+import ast
+import enum
+import fnmatch
+import subprocess
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Any
+
+
+class FilterMode(enum.StrEnum):
+    NONE = "none"
+    CONSERVATIVE = "conservative"
+    AGGRESSIVE = "aggressive"
+
+
+class ImportContext(enum.StrEnum):
+    TOP_LEVEL = "top_level"
+    CONDITIONAL = "conditional"
+    TYPE_CHECKING = "type_checking"
+    DEFERRED = "deferred"
+    IMPORTLIB = "importlib"
+
+
+BUCKET_A_PATTERNS: frozenset[str] = frozenset({
+    "tests/conftest.py",
+    "tests/_helpers.py",
+    "tests/arch/_helpers.py",
+    "tests/arch/_rules.py",
+    "pyproject.toml",
+    "uv.lock",
+    ".pre-commit-config.yaml",
+    "src/autoskillit/server/_factory.py",
+})
+
+BUCKET_A_GLOBS: tuple[str, ...] = ("tests/*/conftest.py",)
+
+ALWAYS_RUN_CONSERVATIVE: frozenset[str] = frozenset({
+    "arch", "contracts", "infra", "docs",
+})
+
+ALWAYS_RUN_AGGRESSIVE: frozenset[str] = frozenset({
+    "arch", "contracts",
+})
+
+_LARGE_CHANGESET_THRESHOLD: int = 30
+
+# ---------------------------------------------------------------------------
+# Layer cascade maps
+# ---------------------------------------------------------------------------
+
+LAYER_CASCADE_CONSERVATIVE: dict[str, frozenset[str]] = {
+    # L0 — imported by everything
+    "core": frozenset({
+        "core", "config", "execution", "pipeline", "workspace",
+        "recipe", "migration", "server", "cli",
+        "hooks", "skills",
+    }),
+    # L1
+    "config": frozenset({
+        "config", "pipeline", "workspace", "server", "cli",
+    }),
+    "execution": frozenset({
+        "execution", "core", "workspace", "migration",
+        "server", "cli", "infra", "skills",
+    }),
+    "pipeline": frozenset({
+        "pipeline", "execution", "server", "infra",
+    }),
+    "workspace": frozenset({
+        "workspace", "recipe", "server", "cli", "skills",
+    }),
+    # L2
+    "recipe": frozenset({
+        "recipe", "execution", "server", "cli", "infra", "skills",
+    }),
+    "migration": frozenset({
+        "migration", "server",
+    }),
+    # L3
+    "server": frozenset({
+        "server", "cli", "infra",
+    }),
+    "cli": frozenset({
+        "cli",
+    }),
+    # Infra (non-layered)
+    "hooks": frozenset({
+        "hooks", "infra", "cli",
+    }),
+    "hook_registry": frozenset({
+        "server", "infra", "cli", "docs",
+    }),
+}
+
+LAYER_CASCADE_AGGRESSIVE: dict[str, frozenset[str]] = {
+    "core": frozenset({"core"}),
+    "config": frozenset({"config"}),
+    "execution": frozenset({"execution"}),
+    "pipeline": frozenset({"pipeline"}),
+    "workspace": frozenset({"workspace"}),
+    "recipe": frozenset({"recipe"}),
+    "migration": frozenset({"migration"}),
+    "server": frozenset({"server"}),
+    "cli": frozenset({"cli"}),
+    "hooks": frozenset({"hooks"}),
+    "hook_registry": frozenset({"hooks"}),
+}
+
+# ---------------------------------------------------------------------------
+# ASTImportWalker
+# ---------------------------------------------------------------------------
+
+
+class ASTImportWalker(ast.NodeVisitor):
+    """Extract all import statements from a Python source file with context tracking."""
+
+    def __init__(self) -> None:
+        self.imports: list[tuple[str, ImportContext]] = []
+        self._context_stack: list[ImportContext] = []
+
+    @property
+    def _current_context(self) -> ImportContext:
+        return self._context_stack[-1] if self._context_stack else ImportContext.TOP_LEVEL
+
+    def visit_Import(self, node: ast.Import) -> None:
+        for alias in node.names:
+            self.imports.append((alias.name, self._current_context))
+        self.generic_visit(node)
+
+    def visit_ImportFrom(self, node: ast.ImportFrom) -> None:
+        if node.module:
+            module = ("." * (node.level or 0)) + node.module
+        else:
+            module = "." * (node.level or 0)
+        self.imports.append((module, self._current_context))
+        self.generic_visit(node)
+
+    def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
+        self._context_stack.append(ImportContext.DEFERRED)
+        self.generic_visit(node)
+        self._context_stack.pop()
+
+    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
+        self._context_stack.append(ImportContext.DEFERRED)
+        self.generic_visit(node)
+        self._context_stack.pop()
+
+    def visit_If(self, node: ast.If) -> None:
+        if self._is_type_checking_guard(node.test):
+            self._context_stack.append(ImportContext.TYPE_CHECKING)
+            for child in node.body:
+                self.visit(child)
+            self._context_stack.pop()
+            for child in node.orelse:
+                self.visit(child)
+        else:
+            self._context_stack.append(ImportContext.CONDITIONAL)
+            self.generic_visit(node)
+            self._context_stack.pop()
+
+    def visit_Try(self, node: ast.Try) -> None:
+        self._context_stack.append(ImportContext.CONDITIONAL)
+        self.generic_visit(node)
+        self._context_stack.pop()
+
+    def visit_Call(self, node: ast.Call) -> None:
+        if self._is_importlib_call(node) and node.args:
+            arg = node.args[0]
+            if isinstance(arg, ast.Constant) and isinstance(arg.value, str):
+                self.imports.append((arg.value, ImportContext.IMPORTLIB))
+        self.generic_visit(node)
+
+    @staticmethod
+    def _is_type_checking_guard(test: ast.expr) -> bool:
+        if isinstance(test, ast.Name) and test.id == "TYPE_CHECKING":
+            return True
+        if isinstance(test, ast.Attribute) and test.attr == "TYPE_CHECKING":
+            return True
+        return False
+
+    @staticmethod
+    def _is_importlib_call(node: ast.Call) -> bool:
+        func = node.func
+        if isinstance(func, ast.Attribute) and func.attr == "import_module":
+            return True
+        return False
+
+
+# ---------------------------------------------------------------------------
+# Utility functions
+# ---------------------------------------------------------------------------
+
+
+def git_changed_files(
+    cwd: str | Path,
+    base_ref: str | None = None,
+) -> set[str] | None:
+    """Return set of changed files relative to base_ref, or None on failure."""
+    import os
+
+    if base_ref is None:
+        base_ref = os.environ.get(
+            "AUTOSKILLIT_TEST_BASE_REF",
+            os.environ.get("GITHUB_BASE_REF"),
+        )
+    if base_ref is None:
+        return None
+
+    try:
+        result = subprocess.run(
+            ["git", "diff", "--name-only", f"{base_ref}...HEAD"],
+            cwd=str(cwd),
+            capture_output=True,
+            text=True,
+            timeout=10,
+            check=True,
+        )
+    except (subprocess.CalledProcessError, subprocess.TimeoutExpired, FileNotFoundError):
+        return None
+
+    lines = result.stdout.strip().splitlines()
+    return {line.strip() for line in lines if line.strip()}
+
+
+def check_bucket_a(changed_files: set[str]) -> bool:
+    """Return True if any changed file triggers a full test run."""
+    for f in changed_files:
+        if f in BUCKET_A_PATTERNS:
+            return True
+        for glob_pat in BUCKET_A_GLOBS:
+            if fnmatch.fnmatch(f, glob_pat):
+                return True
+    return False
+
+
+def load_manifest(path: str | Path) -> dict[str, Any] | None:
+    """Load .autoskillit/test-filter-manifest.yaml, or None if absent."""
+    import yaml
+
+    manifest_path = Path(path) / ".autoskillit" / "test-filter-manifest.yaml"
+    if not manifest_path.exists():
+        return None
+    with manifest_path.open() as f:
+        return yaml.safe_load(f)
+
+
+def apply_manifest(
+    changed_files: set[str],
+    manifest: dict[str, Any] | None,
+) -> set[str]:
+    """Return test directories matched by manifest patterns for the changed files."""
+    if manifest is None:
+        return set()
+    result: set[str] = set()
+    patterns = manifest.get("patterns", {})
+    for pattern, test_dirs in patterns.items():
+        for f in changed_files:
+            if fnmatch.fnmatch(f, pattern):
+                if isinstance(test_dirs, list):
+                    result.update(test_dirs)
+                elif isinstance(test_dirs, str):
+                    result.add(test_dirs)
+                break
+    return result
+
+
+# ---------------------------------------------------------------------------
+# Re-export closure
+# ---------------------------------------------------------------------------
+
+
+def _expand_reexport_closure(
+    changed_src_files: set[str],
+    src_root: str | Path,
+) -> set[str]:
+    """Expand changed files to include __init__.py files that directly re-export from them.
+
+    Walks up parent directories checking each __init__.py via AST parsing.
+    Only handles direct re-exports (from .module import X); transitive
+    chains through intermediate hubs are covered by cascade maps.
+    """
+    src_root = Path(src_root)
+    expanded = set(changed_src_files)
+
+    for changed in list(changed_src_files):
+        changed_path = src_root / changed
+        if not changed_path.exists():
+            continue
+        module_name = changed_path.stem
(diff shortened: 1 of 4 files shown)


Trecek (commented):

tests/test_test_filter.py
```diff
@@ -0,0 +1,457 @@
...
+
+from tests._test_filter import (
+    ALWAYS_RUN_AGGRESSIVE,
+    ALWAYS_RUN_CONSERVATIVE,
+    LAYER_CASCADE_AGGRESSIVE,
+    LAYER_CASCADE_CONSERVATIVE,
+    ASTImportWalker,
+    FilterMode,
+    ImportContext,
+    _expand_reexport_closure,
```
Trecek, line 19: [warning] cohesion: _expand_reexport_closure has an underscore prefix (private) but is explicitly imported and directly tested, while load_manifest and apply_manifest are public functions (no prefix) but are not imported or tested at all. This inverts the expected visibility/coverage pattern: private internals get direct tests, public API does not.

tests/test_test_filter.py
```diff
@@ -0,0 +1,457 @@
...
+
+from __future__ import annotations
+
+import ast
+import subprocess
+from pathlib import Path
+
+import pytest
+
+from tests._test_filter import (
```
Trecek, line 11: [warning] cohesion: load_manifest and apply_manifest are public functions defined in _test_filter.py but are absent from the test import list and have no test coverage. Both functions form part of the manifest-based non-Python file handling path exercised indirectly by build_test_scope tests.

tests/_test_filter.py
```diff
@@ -0,0 +1,484 @@
...
+        pass
+    return None
+
+
+def build_test_scope(
+    changed_files: set[str] | None,
+    mode: FilterMode,
+    manifest: dict[str, Any] | None = None,
+    tests_root: str | Path = "tests",
+    src_root: str | Path = "src/autoskillit",
```
Trecek: [warning] cohesion: build_test_scope accepts src_root as a parameter (default 'src/autoskillit') but never passes it to _expand_reexport_closure or any other call site inside the function body. The parameter is dead: it is accepted but silently ignored, creating a false impression that callers can control the source root.

tests/_test_filter.py
```diff
@@ -0,0 +1,484 @@
+"""Standalone test-path filtering logic. No pytest dependencies."""
```
Trecek, line 1: [warning] cohesion: Module docstring states 'No pytest dependencies' but load_manifest imports yaml (PyYAML) inside the function body at line 315. PyYAML is a third-party runtime dependency, not a stdlib module. The stated contract of the module (stdlib-only) is silently broken.

tests/test_test_filter.py
```diff
@@ -0,0 +1,457 @@
...
+
+    def test_bucket_a_subdir_conftest(self) -> None:
+        assert check_bucket_a({"tests/execution/conftest.py"}) is True
+
+    def test_bucket_a_negative(self) -> None:
+        assert check_bucket_a({"src/autoskillit/core/io.py"}) is False
+
+
+# ---------------------------------------------------------------------------
+# build_test_scope Tests (S1–S10)
```
Trecek, line 123: [info] cohesion: Comment labels the test group as 'S1-S10' (10 tests expected) but the class TestBuildTestScope only contains 9 test methods. The count in the section label does not match the actual test count.

tests/test_test_filter.py
```diff
@@ -0,0 +1,457 @@
...
+            mode=FilterMode.CONSERVATIVE,
+            tests_root=tests_root,
+        )
+        assert result is not None
+        dir_names = {p.name for p in result}
+        assert dir_names == {"arch", "contracts", "infra", "docs"}
+
+
+# ---------------------------------------------------------------------------
+# Conservative vs Aggressive Tests (M1–M4)
```
Trecek, line 326: [info] cohesion: FilterMode.NONE is a named enum variant in FilterMode but has no dedicated test. TestFilterModes covers conservative vs aggressive mode behaviour but the NONE mode short-circuit path (build_test_scope returns None immediately when mode==NONE) is never explicitly exercised by any test class.

Trecek (commented):
AutoSkillit review found 4 blocking cohesion issues. See inline comments.

tests/_test_filter.py
Trecek, line 1: [critical] arch: Production logic (FilterMode, ImportContext, ASTImportWalker, cascade maps, git_changed_files, build_test_scope) lives in tests/_test_filter.py. The tests/ directory is for test code and test helpers only. This module contains standalone utility/library code with no test dependencies. It belongs in src/autoskillit/ (e.g. core/ or a dedicated testing/ subpackage). Placing it here makes the module non-importable from production code and sets a precedent for leaking production logic into the test tree.

tests/_test_filter.py
```diff
@@ -0,0 +1,484 @@
...
+ALWAYS_RUN_AGGRESSIVE: frozenset[str] = frozenset(
+    {
+        "arch",
+        "contracts",
+    }
+)
+
+_LARGE_CHANGESET_THRESHOLD: int = 30
+
+# ---------------------------------------------------------------------------
```
Trecek, line 64: [warning] arch: LAYER_CASCADE_CONSERVATIVE and LAYER_CASCADE_AGGRESSIVE encode the entire project layer topology as data. This topology is already the authoritative subject of arch tests (tests/arch/). Duplicating it here creates two sources of truth: the arch rules and this cascade map can drift silently. The cascade maps should either be derived from the arch rule definitions or asserted to stay in sync with them via a test.

tests/_test_filter.py
```diff
@@ -0,0 +1,484 @@
...
+        self.generic_visit(node)
+        self._context_stack.pop()
+
+    def visit_If(self, node: ast.If) -> None:
+        if self._is_type_checking_guard(node.test):
+            self._context_stack.append(ImportContext.TYPE_CHECKING)
+            for child in node.body:
+                self.visit(child)
+            self._context_stack.pop()
+            for child in node.orelse:
```
Trecek, line 233: [warning] bugs: The `orelse` branch of an `if TYPE_CHECKING:` guard is visited without pushing any context, so imports in the else-branch inherit whatever context is current (e.g. TOP_LEVEL). An `if TYPE_CHECKING: ... else: ...` else-branch should be visited under CONDITIONAL context. Fix: push `ImportContext.CONDITIONAL` before visiting `node.orelse` and pop afterward.

tests/_test_filter.py
```diff
@@ -0,0 +1,484 @@
...
+    try:
+        result = subprocess.run(
+            ["git", "diff", "--name-only", f"{base_ref}...HEAD"],
+            cwd=str(cwd),
+            capture_output=True,
+            text=True,
+            timeout=10,
+            check=True,
+        )
+    except (subprocess.CalledProcessError, subprocess.TimeoutExpired, FileNotFoundError):
```
Trecek: [warning] defense: CalledProcessError, TimeoutExpired, and FileNotFoundError are all collapsed into a silent `return None` with no diagnostic. A timeout is indistinguishable from a missing git binary. The exception type/message should be surfaced via `warnings.warn()` or a debug log so callers can distinguish transient failures from permanent ones.

tests/_test_filter.py
```diff
@@ -0,0 +1,484 @@
...
+    for f in changed_files:
+        if f in BUCKET_A_PATTERNS:
+            return True
+        for glob_pat in BUCKET_A_GLOBS:
+            if fnmatch.fnmatch(f, glob_pat):
+                return True
+    return False
+
+
+def load_manifest(path: str | Path) -> dict[str, Any] | None:
```
Trecek, line 323: [warning] defense: `load_manifest` opens the manifest file but does not catch `OSError`/`yaml.YAMLError`. If the file exists but is unreadable or contains malformed YAML, an unhandled exception propagates with no contextual message indicating the file path. Also, deferred `import yaml` hides the dependency from import-time checks.

tests/_test_filter.py
```diff
@@ -0,0 +1,484 @@
...
+    src_root = Path(src_root)
+    expanded = set(changed_src_files)
+
+    for changed in list(changed_src_files):
+        changed_path = src_root / changed
+        if not changed_path.exists():
+            continue
+        module_name = changed_path.stem
+        parent = changed_path.parent
+        while parent >= src_root:
```
Trecek, line 385: [warning] bugs: `while parent >= src_root` uses Python Path comparison, which is lexicographic. On some systems or with unusual paths this can produce incorrect results (e.g. `/srv/foo` >= `/src/foo` is True lexicographically but not a parent relationship). Use `parent.is_relative_to(src_root)` (Python 3.9+) or `src_root in (parent, *parent.parents)` for correct containment checks.

tests/_test_filter.py
```diff
@@ -0,0 +1,484 @@
...
+            init_path = parent / "__init__.py"
+            if init_path.exists():
+                try:
+                    source = init_path.read_text(errors="replace")
+                    tree = ast.parse(source, filename=str(init_path))
+                    walker = ASTImportWalker()
+                    walker.visit(tree)
+                    for mod, _ctx in walker.imports:
+                        bare = mod.lstrip(".")
```
Trecek: [warning] bugs: `mod.lstrip('.')` strips ALL leading dots, so a relative import like `from ...sibling import foo` (`...sibling`) becomes `sibling`, losing depth information. This means a deeply-nested relative import could falsely match an unrelated module with the same leaf name. Consider `mod.lstrip('.').split('.')[0]` or matching against the full relative path instead.

tests/_test_filter.py
```diff
@@ -0,0 +1,484 @@
...
+                        if bare == module_name or bare.endswith(f".{module_name}"):
+                            rel_init = str(init_path.relative_to(src_root))
+                            expanded.add(rel_init)
+                            break
+                except SyntaxError:
```
Trecek: [warning] defense: Bare `except SyntaxError: pass` silently discards the error. An `__init__.py` that fails to parse is quietly skipped, meaning the reexport closure walk produces an incomplete result with no indication of why. At minimum surface via `warnings.warn()` so callers can diagnose unexpected filtering misses.

tests/_test_filter.py
```diff
@@ -0,0 +1,484 @@
...
+         'src/autoskillit/execution/headless.py' -> 'execution'
+         'src/autoskillit/server/_factory.py' -> 'server'
+         'src/autoskillit/hook_registry.py' -> 'hook_registry'
+    """
+    parts = Path(filepath).parts
+    try:
+        idx = parts.index("autoskillit")
+        if idx + 1 < len(parts):
+            candidate = parts[idx + 1]
+            if candidate.endswith(".py"):
```
Trecek, line 427: [warning] bugs: When a file is directly under `src/autoskillit/` (e.g. `version.py`), `_file_to_package` returns `"version"` which is not in `cascade_map`, causing `build_test_scope` to return `None` (full run). This may be intentional fail-safe behavior but is inconsistent with the stated cascade design — if top-level files are meant to cascade to all tests, that should be explicit (e.g. a catch-all entry in the cascade map) rather than an accidental full-run via `None`.

**Requires human decision:** Is the full-run fallback for top-level module files intentional?

tests/_test_filter.py
Trecek: [warning] cohesion: `src_root` parameter is accepted by `build_test_scope` but never passed to `_expand_reexport_closure` or any internal call — it is dead code. Either wire it through to the functions that need it or remove the parameter.

tests/_test_filter.py
```diff
@@ -0,0 +1,484 @@
...
+    """Compute the set of test paths to run, or None for a full run.
+
+    Algorithm:
+    1. None changed_files -> None (fail-open)
+    2. >30 files -> None (large changeset)
+    3. Bucket A triggered -> None (full run)
+    4. Classify: src Python -> cascade, test Python -> direct, non-Python -> manifest
+    5. Compute always-run set for mode (includes arch/contracts for both modes)
+    6. Union all sets
+    7. For aggressive mode -> AST-based file-level refinement (future)
```
Trecek, line 450: [warning] slop: Algorithm step 7 documents unimplemented future behavior: `For aggressive mode -> AST-based file-level refinement (future)`. This is a TODO embedded in a docstring with no issue reference. Either implement it, track it in an issue and reference it here, or remove the note.

tests/test_test_filter.py
Trecek, line 19: [warning] cohesion: `_expand_reexport_closure` (private, underscore prefix) is directly imported and tested, while `load_manifest` and `apply_manifest` (public, no prefix) have zero test coverage. This inverts the expected visibility/coverage pattern — public API should have coverage before private helpers.

tests/test_test_filter.py
```diff
@@ -0,0 +1,457 @@
...
+
+
+class TestGitChangedFiles:
+    def test_git_changed_files_success(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        fake_result = subprocess.CompletedProcess(
+            args=[],
+            returncode=0,
+            stdout="src/autoskillit/core/io.py\ntests/core/test_io.py\n",
+        )
+        monkeypatch.setattr(subprocess, "run", lambda *a, **kw: fake_result)
```
Trecek, line 380: [warning] tests: `monkeypatch.setattr(subprocess, "run", ...)` patches at the module level. If `_test_filter` ever binds `subprocess.run` locally (e.g. `from subprocess import run`), this patch would not intercept the call and the test would execute a real git subprocess. Safer to patch as `monkeypatch.setattr("tests._test_filter.subprocess.run", ...)`.

tests/test_test_filter.py
```diff
@@ -0,0 +1,457 @@
...
+        assert result == {"src/autoskillit/core/io.py", "tests/core/test_io.py"}
+
+    def test_git_changed_files_failure_returns_none(
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
+        def _raise(*a: object, **kw: object) -> None:
+            raise subprocess.CalledProcessError(1, "git")
+
+        monkeypatch.setattr(subprocess, "run", _raise)
```
Trecek, line 391: [warning] tests: Same subprocess.run patch target concern as L370 — patching `subprocess.run` globally rather than the symbol as imported in the module under test.

tests/test_test_filter.py
```diff
@@ -0,0 +1,457 @@
...
+        assert result is None
+
+    def test_git_changed_files_timeout_returns_none(
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
+        def _raise(*a: object, **kw: object) -> None:
+            raise subprocess.TimeoutExpired("git", 10)
+
+        monkeypatch.setattr(subprocess, "run", _raise)
```
Trecek, line 402: [warning] tests: Same subprocess.run patch target concern as L370/L381 for the TimeoutExpired test.

Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (posted as COMMENT — cannot REQUEST_CHANGES on own PR)

**17 actionable findings** (critical/warning) across 6 audit dimensions.
2 findings require human decision.

| Dimension | Critical | Warning | Info |
|-----------|----------|---------|------|
| arch | 1 | 3 | 0 |
| bugs | 0 | 4 | 1 |
| defense | 0 | 2 | 3 |
| cohesion | 0 | 3 | 0 |
| tests | 0 | 3 | 2 |
| slop | 0 | 1 | 10 |

Trecek (commented):
AutoSkillit review found 15 blocking issues (1 critical, 14 warning) across 6 audit dimensions. 2 findings require human decision. See inline comments.

tests/_test_filter.py
Trecek: Agreed — fixed in b1383d5f. Removed dead `src_root` parameter from `build_test_scope`.

Trecek: Agreed — fixed in b1383d5f. Duplicate of comment on line 420; removed dead `src_root` parameter.

tests/_test_filter.py
Trecek: Agreed — fixed in b1383d5f. Added `CONDITIONAL` context push for the `orelse` branch of TYPE_CHECKING guards.

tests/_test_filter.py
Trecek: Agreed — fixed in b1383d5f. Split the bare except into three separate handlers with `warnings.warn()` for CalledProcessError (with exit code+stderr), TimeoutExpired, and FileNotFoundError.

tests/_test_filter.py
Trecek: Agreed — fixed in b1383d5f. Wrapped `open()`/`yaml.safe_load()` in try/except for `OSError` and `yaml.YAMLError`, returning None with a `warnings.warn()` that includes the file path.

tests/_test_filter.py
Trecek: Agreed — fixed in b1383d5f. Replaced `mod.lstrip('.')` with `mod.split('.')[-1]` to extract only the leaf module name, preserving relative import depth.

tests/_test_filter.py
Trecek: Agreed — fixed in b1383d5f. Replaced bare `except SyntaxError: pass` with `warnings.warn()` that includes the init file path and parse error.

tests/test_test_filter.py
Trecek: Agreed — fixed in 03631c20. Added `test_scope_none_mode_returns_none` to TestBuildTestScope that exercises the FilterMode.NONE short-circuit path.

tests/test_test_filter.py
Trecek: Agreed — fixed in 03631c20. Added TestLoadManifest (3 tests: absent, valid, malformed YAML) and TestApplyManifest (4 tests: none, match, no-match, list-dirs). Both `load_manifest` and `apply_manifest` are now imported and tested.

tests/_test_filter.py
Trecek: Investigated — this is intentional. The module docstring (line 1) states 'No pytest dependencies', not 'stdlib-only'. PyYAML is a declared runtime dependency in pyproject.toml (pyyaml>=6.0), so it is always available at runtime. The deferred `import yaml` is a stylistic choice to defer loading, not a contract violation.

tests/_test_filter.py
Trecek: Investigated — this is intentional. The cascade maps in _test_filter.py represent test scope calculation (which test directories to run given a changed package), not import enforcement rules. A search of tests/arch/ finds zero references to LAYER_CASCADE or cascade_map data structures. The arch tests enforce import rules via AST analysis — a fundamentally different abstraction from test-scope cascade maps.

tests/_test_filter.py
Trecek: Investigated — this is intentional. The loop invariant makes false positives impossible: `parent` is initialized to `changed_path.parent` where `changed_path = src_root / changed`, so parent always starts at or below src_root. The loop walks strictly upward via `parent = parent.parent`, and terminates correctly when parent steps above src_root (e.g. 'src' < 'src/autoskillit'). The lexicographic comparison aligns with the parent-child relationship in this constrained traversal.

tests/_test_filter.py
Trecek: Valid observation — flagged for design decision. The module contains utility logic with zero pytest/test dependencies — it was placed in tests/ deliberately as test infrastructure used exclusively by the test suite (imported only from tests/test_test_filter.py). No production code consumes it. Whether to promote to src/autoskillit/ depends on whether the module's audience expands beyond test infrastructure.

tests/_test_filter.py
Trecek: Valid observation — flagged for design decision. `_file_to_package('src/autoskillit/version.py')` returns 'version' which is absent from cascade maps, triggering `return None` (full run). This is the intended fail-safe per the docstring ('else full run'). Whether to explicitly map top-level files like version.py to specific test directories (e.g. 'infra'/'contracts') is a design decision that affects the fail-open vs. targeted-run tradeoff.

PR #976, opened 2026-04-16
Build conftest filter plugin (#933)

### LINKED ISSUE #933: Build conftest filter plugin
**Category:** Filter Infrastructure
**Depends on:** #06, #07

## Problem

No mechanism exists to selectively run tests based on what changed. The conftest.py plugin is the integration point that connects the filter logic (from #06) and the manifest (from #07) to pytest's collection machinery.

## Solution

Add three pytest hooks to `tests/conftest.py`:

### `pytest_addoption`

Register CLI options for discoverability:
- `--filter-mode` (choices: `none`, `conservative`, `aggressive`, default: `none`): Filter mode selection
- `--filter-base-ref` (string): Git diff base ref override

### `pytest_configure`

1. Read `AUTOSKILLIT_TEST_FILTER` env var (or `--filter-mode` CLI flag)
2. If not set or `"none"` → store `FilterMode.NONE` on config, return
3. Base ref resolution priority: (1) `AUTOSKILLIT_TEST_BASE_REF` env var (set by DefaultTestRunner from sidecar), (2) `--filter-base-ref` CLI option, (3) `GITHUB_BASE_REF` env var. The conftest does NOT do sidecar lookup — that's DefaultTestRunner's responsibility (#09).
4. Call `git_changed_files(cwd, base_ref)` → store result on config
5. If `git_changed_files()` returns None (subprocess failure, timeout, not a git repo), store None on config. `pytest_collection_modifyitems` checks for None and returns immediately (fail-open, run everything).
6. Load manifest from `.autoskillit/test-filter-manifest.yaml` → store on config

### `pytest_collection_modifyitems`

1. Read mode and changed files from config
2. If mode is NONE → return (no filtering)
3. Call `build_test_scope()` → compute the set of test files to run
4. If scope is None → return (fail-open, run everything)
5. Partition items into selected/deselected based on scope
6. Call `config.hook.pytest_deselected(items=deselected)` BEFORE mutating `items[:] = selected`
7. Mutate `items[:] = selected`
8. Log summary via `config.pluginmanager.get_plugin('terminalreporter').write_line()` (pytest-idiomatic terminal output): `[smart-filter] mode=X, changed=N files, selected=M/T tests, deselected=D`

### Key design decisions

- **Opt-in via env var**: `AUTOSKILLIT_TEST_FILTER` must be set. CI never sees it.
- **`pytest_collection_modifyitems`** (not `pytest_ignore_collect`): Safe with parametrized arch tests that use `_SOURCE_FILES` at import time.
- **Deterministic across xdist workers**: Each worker independently computes the same result from the same env vars and git state.
- **Fail-open**: Any error in filter logic → run everything.
- **Deselection order**: call `config.hook.pytest_deselected(items=deselected)` BEFORE mutating `items[:] = selected`.

### Update `tests/CLAUDE.md` — path filtering rules for test authors

Add a concise section to `tests/CLAUDE.md`:

```
## Path Filtering

Tests support smart path filtering (`AUTOSKILLIT_TEST_FILTER` env var). Filter accuracy depends on test imports:

- **Import only from your layer + core.** A test in `tests/execution/` should import from `autoskillit.execution` and `autoskillit.core`. Cross-layer imports widen the filter's cascade.
- **Use `minimal_ctx` over `tool_ctx`** when you only need `gate`, `runner`, or `config`. `tool_ctx` pulls all 8 layers and triggers a wider filter scope.
- **Non-Python dependencies** (YAML, SKILL.md, config files) must be registered in `.autoskillit/test-filter-manifest.yaml` or the filter won't trigger on their changes.
- **`tests/arch/` and `tests/contracts/`** always run on any source change — no special handling needed.
- **Full-run triggers**: changes to `conftest.py`, `pyproject.toml`, `uv.lock`, `tests/_helpers.py`, or `server/_factory.py` bypass filtering entirely.
```

## Acceptance Criteria

- [ ] `pytest_addoption` registers `--filter-mode` (choices: none/conservative/aggressive) and `--filter-base-ref`
- [ ] Filter is inactive when `AUTOSKILLIT_TEST_FILTER` is unset
- [ ] Filter activates in conservative/aggressive modes when env var is set
- [ ] Summary line logged to terminal
- [ ] `pytest_deselected` hook called with correct items
- [ ] Works with xdist `-n 4` (deterministic collection)
- [ ] pytester-based hook tests go in `tests/test_test_filter.py` alongside the unit tests for filter logic: 6 cases (deselection, summary, fallback, modes)
- [ ] Verify `AUTOSKILLIT_TEST_FILTER` and `AUTOSKILLIT_TEST_BASE_REF` are NOT in `AUTOSKILLIT_PRIVATE_ENV_VARS` in `core/_type_constants.py`
- [ ] `tests/CLAUDE.md` updated with the path filtering rules section above

## Context from plan

See plan sections: 5 (Filter Design — bucket system), 8 (conftest.py additions)

Trecek:
## Summary

Wire the test filter engine (from `_test_filter.py`) into pytest's collection machinery via three conftest hooks: `pytest_addoption`, `pytest_configure`, and `pytest_collection_modifyitems`. The filter is opt-in via `AUTOSKILLIT_TEST_FILTER` env var or `--filter-mode` CLI flag, with fail-open semantics on any error. 8 pytester-based integration tests validate hook behavior. `tests/CLAUDE.md` updated with path filtering rules for test authors.

### Acceptance Criteria (from issue #933)

- `pytest_addoption` registers `--filter-mode` (choices: none/conservative/aggressive) and `--filter-base-ref`
- Filter is inactive when `AUTOSKILLIT_TEST_FILTER` is unset
- Filter activates in conservative/aggressive modes when env var is set
- Summary line logged to terminal
- `pytest_deselected` hook called with correct items
- Works with xdist `-n 4` (deterministic collection)
- pytester-based hook tests in `tests/test_test_filter.py`: 6 cases (deselection, summary, fallback, modes)
- Verify `AUTOSKILLIT_TEST_FILTER` and `AUTOSKILLIT_TEST_BASE_REF` are NOT in `AUTOSKILLIT_PRIVATE_ENV_VARS` in `core/_type_constants.py`
- `tests/CLAUDE.md` updated with path filtering rules section

Closes #933

## Token Usage Summary

| make_plan | 5.4k | 15.9k | 899.5k | 75.0k | 1 | 9m 37s |
| review_approach | 35 | 5.3k | 300.3k | 50.4k | 1 | 4m 29s |
| dry_walkthrough | 54 | 14.9k | 1.4M | 68.3k | 1 | 5m 30s |
| implement | 74 | 11.4k | 1.3M | 49.1k | 1 | 3m 56s |
| prepare_pr | 26 | 4.1k | 215.0k | 29.8k | 1 | 1m 31s |
| run_arch_lenses | 50 | 8.5k | 627.0k | 59.3k | 1 | 2m 39s |
| compose_pr | 24 | 1.9k | 156.6k | 14.7k | 1 | 53s |
| **Total** | 5.7k | 62.0k | 4.9M | 346.6k | | 28m 38s |

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,6 +13,8 @@ from autoskillit.core.types import (
 )
 from tests._helpers import _flush_structlog_proxy_caches
 
+_scope_key = pytest.StashKey[set[_Path] | None]()
+
 
 class StatefulMockTester:
     """Test double for TestRunner returning pre-configured results on successive calls.
@@ -292,3 +294,140 @@ def tool_ctx(monkeypatch, tmp_path):
     monkeypatch.setattr(_state, "_ctx", ctx)
     monkeypatch.setattr(_state, "_startup_ready", None)
     return ctx
+
+
+# ---------------------------------------------------------------------------
+# Test filter hooks (opt-in via AUTOSKILLIT_TEST_FILTER env var)
+# ---------------------------------------------------------------------------
+
+
+def pytest_addoption(parser: pytest.Parser) -> None:
+    parser.addoption(
+        "--filter-mode",
+        default=None,
+        choices=("none", "conservative", "aggressive"),
+        help="Test filter mode (overrides AUTOSKILLIT_TEST_FILTER env var).",
+    )
+    parser.addoption(
+        "--filter-base-ref",
+        default=None,
+        help="Git base ref for changed-file detection (overrides AUTOSKILLIT_TEST_BASE_REF).",
+    )
+
+
+def pytest_configure(config: pytest.Config) -> None:
+    """Compute test filter scope from env var + git diff + manifest.
+
+    Opt-in via AUTOSKILLIT_TEST_FILTER env var or --filter-mode CLI flag.
+    Fail-open: any error sets scope to None (full test run).
+    """
+    import os
+    import warnings
+
+    config.stash[_scope_key] = None
+
+    cli_mode = config.getoption("--filter-mode", default=None)
+    env_val = os.environ.get("AUTOSKILLIT_TEST_FILTER", "")
+
+    if not cli_mode and not env_val:
+        return
+    if not cli_mode and env_val.lower() in ("0", "false", "no"):
+        return
+
+    try:
+        from tests._test_filter import (
+            FilterMode,
+            build_test_scope,
+            git_changed_files,
+            load_manifest,
+        )
+
+        if cli_mode:
+            mode = FilterMode(cli_mode)
+        elif env_val.lower() in ("1", "true", "yes"):
+            mode = FilterMode.CONSERVATIVE
+        else:
+            mode = FilterMode(env_val)
+
+        if mode == FilterMode.NONE:
+            return
+
+        cli_base_ref = config.getoption("--filter-base-ref", default=None)
+        changed = git_changed_files(config.rootpath, base_ref=cli_base_ref)
+
+        raw_manifest = load_manifest(config.rootpath)
+        manifest = {"patterns": raw_manifest} if raw_manifest is not None else None
+
+        scope = build_test_scope(
+            changed_files=changed,
+            mode=mode,
+            manifest=manifest,
+            tests_root=config.rootpath / "tests",
+        )
+        config.stash[_scope_key] = scope
+
+    except Exception as exc:
+        warnings.warn(
+            f"Test filter setup failed, running all tests: {exc}",
+            stacklevel=1,
+        )
+
+
+def pytest_collection_modifyitems(
+    items: list[pytest.Item],
+    config: pytest.Config,
+) -> None:
+    """Deselect test items outside the computed filter scope.
+
+    Fail-open: any error leaves all items selected.
+    """
+    import warnings
+
+    scope: set[_Path] | None = config.stash.get(_scope_key, None)
+    if scope is None:
+        return
+
+    try:
+        root = config.rootpath
+        scope_abs: set[_Path] = set()
+        for p in scope:
+            scope_abs.add(p if p.is_absolute() else root / p)
+
+        selected: list[pytest.Item] = []
+        deselected: list[pytest.Item] = []
+
+        for item in items:
+            item_path = item.path
+            matched = False
+            for sp in scope_abs:
+                if sp.is_file():
+                    if item_path == sp:
+                        matched = True
+                        break
+                else:
+                    try:
+                        item_path.relative_to(sp)
+                        matched = True
+                        break
+                    except ValueError:
+                        continue
+            if matched:
+                selected.append(item)
+            else:
+                deselected.append(item)
+
+        if deselected:
+            config.hook.pytest_deselected(items=deselected)
+            items[:] = selected
+
+        warnings.warn(
+            f"Test filter: {len(selected)} selected, {len(deselected)} deselected "
+            f"({len(scope)} scope paths)",
+            stacklevel=1,
+        )
+
+    except Exception as exc:
+        warnings.warn(
+            f"Test filter deselection failed, running all tests: {exc}",
+            stacklevel=1,
+        )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -355,8 +355,7 @@ def pytest_configure(config: pytest.Config) -> None:
         cli_base_ref = config.getoption("--filter-base-ref", default=None)
         changed = git_changed_files(config.rootpath, base_ref=cli_base_ref)
 
-        raw_manifest = load_manifest(config.rootpath)
-        manifest = {"patterns": raw_manifest} if raw_manifest is not None else None
+        manifest = load_manifest(config.rootpath)
 
         scope = build_test_scope(
             changed_files=changed,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -419,12 +419,11 @@ def pytest_collection_modifyitems(
         if deselected:
             config.hook.pytest_deselected(items=deselected)
             items[:] = selected
-
-        warnings.warn(
-            f"Test filter: {len(selected)} selected, {len(deselected)} deselected "
-            f"({len(scope)} scope paths)",
-            stacklevel=1,
-        )
+            warnings.warn(
+                f"Test filter: {len(selected)} selected, {len(deselected)} deselected "
+                f"({len(scope)} scope paths)",
+                stacklevel=1,
+            )
 
     except Exception as exc:
         warnings.warn(
--- a/tests/test_test_filter.py
+++ b/tests/test_test_filter.py
@@ -26,6 +26,8 @@ from tests._test_filter import (
     load_manifest,
 )
 
+pytest_plugins = ["pytester"]
+
 PROJECT_ROOT = Path(__file__).resolve().parents[1]
 MANIFEST_PATH = PROJECT_ROOT / ".autoskillit" / "test-filter-manifest.yaml"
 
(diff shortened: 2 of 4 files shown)


Trecek (commented):

9 actionable findings (warning), 10 informational. See inline comments.

tests/test_test_filter.py
```diff
@@ -26,6 +26,8 @@
     load_manifest,
 )
 
+pytest_plugins = ["pytester"]
```
Trecek, line 29: [warning] tests: pytest_plugins = ["pytester"] at module level in a test file can cause issues with pytest-xdist parallel execution. Consider moving this declaration to a conftest.py.

tests/test_test_filter.py
```diff
@@ -576,3 +578,190 @@ def test_apply_manifest_multiple_files_union(self) -> None:
...
+    def test_deselection_reports_correctly(
+        self, pytester: pytest.Pytester, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        monkeypatch.setenv("AUTOSKILLIT_TEST_FILTER", "1")
+        pytester.makeconftest(_CONFTEST_HOOKS_SOURCE)
+        pytester.makepyfile(test_keep="def test_keep(): pass")
+        pytester.makepyfile(test_drop="def test_drop(): pass")
+        result = pytester.runpytest("-v")
+        # Both are at root level, not under subdir_a — both deselected
+        result.assert_outcomes(deselected=2)
```
Trecek, line 726: [warning] tests: test_deselection_reports_correctly asserts deselected=2 but does not verify the warning message content. The test name implies it verifies correct reporting but only confirms the count.

tests/conftest.py
```diff
@@ -292,3 +294,140 @@ def tool_ctx(monkeypatch, tmp_path):
...
+
+        scope = build_test_scope(
+            changed_files=changed,
+            mode=mode,
+            manifest=manifest,
+            tests_root=config.rootpath / "tests",
+        )
+        config.stash[_scope_key] = scope
+
+    except Exception as exc:
```
Trecek, line 368: [warning] defense: Bare except Exception swallows all errors including ValueError from FilterMode() with an invalid env value, hiding misconfiguration. Consider separating mode-parsing from I/O calls so bad AUTOSKILLIT_TEST_FILTER values produce clear errors.

tests/conftest.py
```diff
@@ -292,3 +294,140 @@ def tool_ctx(monkeypatch, tmp_path):
...
+            git_changed_files,
+            load_manifest,
+        )
+
+        if cli_mode:
+            mode = FilterMode(cli_mode)
+        elif env_val.lower() in ("1", "true", "yes"):
+            mode = FilterMode.CONSERVATIVE
+        else:
+            mode = FilterMode(env_val)
```
Trecek, line 350: [warning] defense: FilterMode(env_val) is called without stripping whitespace. An env var like 'conservative ' (trailing space) will raise ValueError caught by broad except, silently running all tests.

tests/conftest.py
```diff
@@ -292,3 +294,140 @@ def tool_ctx(monkeypatch, tmp_path):
...
+            if matched:
+                selected.append(item)
+            else:
+                deselected.append(item)
+
+        if deselected:
+            config.hook.pytest_deselected(items=deselected)
+            items[:] = selected
+
+        warnings.warn(
```
Trecek: [warning] bugs: Summary warning fires unconditionally even when deselected==0. The warnings.warn is outside the 'if deselected:' block, emitting '0 deselected' noise on every filtered run where all tests match scope. Also diverges from the pytester stub behavior.

tests/conftest.py
```diff
@@ -292,3 +294,140 @@ def tool_ctx(monkeypatch, tmp_path):
...
+            f"Test filter: {len(selected)} selected, {len(deselected)} deselected "
+            f"({len(scope)} scope paths)",
+            stacklevel=1,
+        )
+
+    except Exception as exc:
```
Trecek, line 427: [warning] defense: Bare except Exception in pytest_collection_modifyitems masks programmer errors (AttributeError, TypeError). Consider narrowing the caught exception types.

tests/conftest.py
```diff
@@ -292,3 +294,140 @@ def tool_ctx(monkeypatch, tmp_path):
...
+    items: list[pytest.Item],
+    config: pytest.Config,
+) -> None:
+    """Deselect test items outside the computed filter scope.
+
+    Fail-open: any error leaves all items selected.
+    """
+    import warnings
+
+    scope: set[_Path] | None = config.stash.get(_scope_key, None)
```
Trecek, line 385: [warning] bugs: config.stash.get(_scope_key, None) returns None both when key was never set AND when explicitly set to None. If pytest_collection_modifyitems runs before pytest_configure (unusual plugin ordering), filtering is silently skipped.

tests/conftest.py
```diff
@@ -292,3 +294,140 @@ def tool_ctx(monkeypatch, tmp_path):
...
+        else:
+            mode = FilterMode(env_val)
+
+        if mode == FilterMode.NONE:
+            return
+
+        cli_base_ref = config.getoption("--filter-base-ref", default=None)
+        changed = git_changed_files(config.rootpath, base_ref=cli_base_ref)
+
+        raw_manifest = load_manifest(config.rootpath)
```
Trecek: [warning] cohesion: load_manifest returns raw patterns dict, but conftest wraps it as {'patterns': raw_manifest} — an impedance mismatch creating an inconsistent calling convention vs direct callers of apply_manifest.

tests/test_test_filter.py
```diff
@@ -576,3 +578,190 @@ def test_apply_manifest_multiple_files_union(self) -> None:
             ["src/autoskillit/recipes/cook.yaml", "docs/guide.md"], manifest
         )
         assert result == {"recipe/", "docs/"}
+
+
+# ---------------------------------------------------------------------------
+# Conftest filter plugin – pytester integration tests (P1–P8)
+# ---------------------------------------------------------------------------
+
+_CONFTEST_HOOKS_SOURCE = """
```
Trecek, line 587: [warning] cohesion: _CONFTEST_HOOKS_SOURCE duplicates production conftest hook logic. Any change to production hooks must be manually mirrored in all three string constants.

tests/conftest.py
Trecek: Agreed — fixed in 79e7606. Moved the summary warning inside the `if deselected:` block so it only emits when tests are actually deselected. Updated both production conftest and pytester stub.

tests/conftest.py
Trecek: Agreed — fixed in d9c9e16. Removed the impedance mismatch: apply_manifest now iterates the raw manifest dict directly instead of requiring a {"patterns": ...} wrapper. Conftest no longer wraps, and tests updated to pass raw dicts.

tests/conftest.py
Trecek: Investigated — this is intentional. The except Exception at line 369 is the explicit fail-open mechanism (see docstring: 'Fail-open: any error sets scope to None'). The ValueError from FilterMode() is caught and surfaced via warnings.warn with the exc repr included — not silently swallowed. The fail-open-on-any-error contract is a deliberate design choice for a conftest hook that must never crash collection.

tests/conftest.py
Trecek: Investigated — this is intentional. A trailing-space env value like 'conservative ' raises ValueError from FilterMode(), which is caught by the fail-open except at line 369 and emits warnings.warn('Test filter setup failed, running all tests: ...') with the exception detail. The failure is not silent — the warning includes the exc repr. This is the intended fail-open behavior; adding .strip() would be a cosmetic improvement but the current code does not silently hide the misconfiguration.

tests/conftest.py
Trecek: Investigated — this is intentional. The except Exception at line 429 is the documented fail-open safety net (docstring: 'Fail-open: any error leaves all items selected'). AttributeError/TypeError caught here emit warnings.warn('Test filter deselection failed, running all tests: {exc}') — errors are surfaced, not masked. This broad catch is symmetric with the pytest_configure pattern and exists because a conftest collection hook must never crash the test run.

tests/conftest.py
Trecek: Investigated — this is intentional. pytest_configure sets config.stash[_scope_key] = None as its very first action (line 331), before any early-return paths. If pytest_collection_modifyitems hypothetically ran first (which pytest's plugin system does not permit for conftest hooks), stash.get returns None → all tests run, which is the correct fail-open outcome. There is no scenario where this produces incorrect filtering.

tests/test_test_filter.py
Trecek: Investigated — this is intentional. pytest_plugins = ["pytester"] at module level is the documented mechanism for enabling pytester in a specific test file (per pytest docs). Plugin declarations are processed by pytest's plugin collection machinery before worker distribution, so this does not cause xdist issues. No conftest.py currently declares pytester because only this file needs it.

tests/test_test_filter.py
Trecek: Investigated — this is intentional. _CONFTEST_HOOKS_SOURCE is an intentional test double, not a copy of production code. It omits FilterMode enum, load_manifest, build_test_scope, and git_changed_files entirely, instead hardcoding scope to {config.rootpath / 'subdir_a'}. This is structurally different from the production hooks — test doubles are expected to diverge from production implementations.

tests/test_test_filter.py
Trecek: Valid observation — flagged for design decision. The test name 'test_deselection_reports_correctly' implies verifying warning content but only asserts deselected=2. Either renaming to 'test_deselects_all_root_level_tests' or adding result.stdout.fnmatch_lines() to verify the warning text would resolve the ambiguity.

PR #978, opened 2026-04-16
Extend TestCheckConfig and DefaultTestRunner for Filtering

### LINKED ISSUE #934: Extend TestCheckConfig and DefaultTestRunner for filtering
**Category:** Integration
**Depends on:** #08 (conftest plugin exists to consume the env vars)

## Problem

The filter is activated by environment variables, but nothing sets them. `DefaultTestRunner.run()` currently passes a sanitized env to the test subprocess with no filter vars. The configuration has no fields for filter mode or base ref.

**Design constraint**: The `test_check` MCP tool is codebase-agnostic — users configure their own test command. The filter works by setting env vars that the conftest plugin reads, not by modifying the test command.

### Extend `TestCheckConfig`

In `config/settings.py`:
```python
@dataclass
class TestCheckConfig:
    command: list[str] = field(default_factory=lambda: ["task", "test-check"])
    timeout: int = 600
    filter_mode: Literal["none", "conservative", "aggressive"] = "none"
    base_ref: str = ""
```

Consider using `Literal['none', 'conservative', 'aggressive']` for `filter_mode` type annotation to enable mypy enforcement.

Update `config/defaults.yaml` with the new fields (defaulting to `none` / empty).

**Note:** The `from_dynaconf()` method in `settings.py` constructs `TestCheckConfig` with explicit named args — adding `filter_mode` and `base_ref` to the dataclass also requires extending the `from_dynaconf` block.

### Thread env vars through `DefaultTestRunner.run()`

In `execution/testing.py`, after `env = build_sanitized_env()`:
- If `config.test_check.filter_mode != "none"`, set `env["AUTOSKILLIT_TEST_FILTER"]`
- Resolve base ref via `_resolve_base_ref(cwd)`:
  1. Config `base_ref` field (explicit override)
  2. Worktree sidecar file at `{project_root}/.autoskillit/temp/worktrees/{name}/base-branch`
  3. `git rev-parse --abbrev-ref @{upstream}` in cwd
  4. `None` (filter auto-detects or runs everything)
- Set `env["AUTOSKILLIT_TEST_BASE_REF"]` if resolved

**Note:** `_resolve_base_ref()` must use async subprocess calls (not bare `subprocess.run`) since `DefaultTestRunner.run()` is an async method.

**Note:** `_resolve_base_ref(cwd)` needs the project root (not the worktree cwd) to find sidecar files. Derive via `git rev-parse --show-toplevel` or pass from config.

### Ensure filter vars pass through

Verify `AUTOSKILLIT_TEST_FILTER` and `AUTOSKILLIT_TEST_BASE_REF` are NOT in `AUTOSKILLIT_PRIVATE_ENV_VARS` in `core/_type_constants.py`. They must not be stripped by `build_sanitized_env()`.

## Acceptance Criteria

- [ ] `TestCheckConfig` has `filter_mode` and `base_ref` fields
- [ ] `defaults.yaml` updated with defaults
- [ ] `DefaultTestRunner.run()` sets filter env vars when configured
- [ ] Filter vars not in `AUTOSKILLIT_PRIVATE_ENV_VARS`
- [ ] Base ref resolution: sidecar → upstream → None fallback chain
- [ ] Tests: `test_runner_sets_filter_env_when_configured`, `test_runner_omits_filter_env_when_none`, `test_build_sanitized_env_preserves_filter_vars` (set `AUTOSKILLIT_TEST_FILTER` in env, call `build_sanitized_env()`, assert it's present in the result dict), `test_runner_resolves_base_ref_from_sidecar`

## Context from plan

See plan sections: 9 (Integration & Wiring)

Trecek:
## Summary

Add `filter_mode` and `base_ref` fields to `TestCheckConfig`. Thread `AUTOSKILLIT_TEST_FILTER` and `AUTOSKILLIT_TEST_BASE_REF` env vars through `DefaultTestRunner.run()` into the subprocess environment. Implement a base ref resolution chain: config override -> worktree sidecar -> git upstream -> None. Verify that the filter env vars are NOT in `AUTOSKILLIT_PRIVATE_ENV_VARS` (they must pass through to subprocesses, not be stripped). Update `defaults.yaml` and `from_dynaconf()`.

## Requirements

- `TestCheckConfig` has `filter_mode` and `base_ref` fields
- `defaults.yaml` updated with defaults
- `DefaultTestRunner.run()` sets filter env vars when configured
- Filter vars not in `AUTOSKILLIT_PRIVATE_ENV_VARS`
- Base ref resolution: sidecar → upstream → None fallback chain
- Tests: `test_runner_sets_filter_env_when_configured`, `test_runner_omits_filter_env_when_none`, `test_build_sanitized_env_preserves_filter_vars`, `test_runner_resolves_base_ref_from_sidecar`

Closes #934

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-934-20260415-200306-865354/.autoskillit/temp/make-plan/extend_testcheckconfig_filtering_plan_2026-04-15_201300.md`

## Token Usage Summary

| make_plan | 88 | 17.4k | 893.1k | 74.9k | 1 | 8m 49s |
| review_approach | 2.6k | 4.1k | 304.2k | 52.2k | 1 | 3m 4s |
| dry_walkthrough | 45 | 10.1k | 1.0M | 63.4k | 1 | 5m 50s |
| implement | 64 | 10.0k | 1.2M | 47.5k | 1 | 3m 47s |
| resolve_failures | 53 | 8.4k | 925.0k | 38.7k | 1 | 5m 44s |
| prepare_pr | 26 | 4.5k | 217.8k | 25.1k | 1 | 1m 42s |
| run_arch_lenses | 59 | 10.0k | 438.8k | 43.0k | 1 | 3m 36s |
| compose_pr | 22 | 2.0k | 106.4k | 14.9k | 1 | 48s |
| **Total** | 3.0k | 66.7k | 5.1M | 359.7k | | 33m 22s |

### DIFF
--- a/src/autoskillit/execution/testing.py
+++ b/src/autoskillit/execution/testing.py
@@ -90,8 +90,11 @@ async def _resolve_base_ref(config_base_ref: str | None, cwd: Path) -> str | Non
         try:
             await asyncio.wait_for(proc.wait(), timeout=15.0)
         except TimeoutError:
-            proc.kill()
-            await proc.wait()
+            try:
+                proc.kill()
+                await proc.wait()
+            except OSError:
+                pass
             return None
         if proc.returncode == 0:
             assert proc.stdout is not None
--- a/src/autoskillit/execution/testing.py
+++ b/src/autoskillit/execution/testing.py
@@ -7,6 +7,7 @@ _logging.
 
 from __future__ import annotations
 
+import asyncio
 import os
 import re
 from pathlib import Path
@@ -30,6 +31,78 @@ def build_sanitized_env() -> dict[str, str]:
     subprocess runner get full env inheritance minus the internal vars.
     """
     return {k: v for k, v in os.environ.items() if k not in AUTOSKILLIT_PRIVATE_ENV_VARS}
+
+
+def _read_sidecar_base_branch(cwd: Path) -> str | None:
+    """Read base-branch from worktree sidecar if cwd is a linked worktree.
+
+    Worktree sidecars are written by implement-worktree skills at
+    ``<project_root>/.autoskillit/temp/worktrees/<wt_name>/base-branch``.
+    Returns None if cwd is not a worktree or no sidecar exists.
+    """
+    git_path = cwd / ".git"
+    if not git_path.is_file():
+        return None
+    try:
+        content = git_path.read_text().strip()
+        if not content.startswith("gitdir:"):
+            return None
+        gitdir = Path(content.split(":", 1)[1].strip())
+        if not gitdir.is_absolute():
+            gitdir = (cwd / gitdir).resolve()
+        main_git = gitdir.parent.parent
+        project_root = main_git.parent
+        wt_name = cwd.name
+        sidecar = project_root / ".autoskillit" / "temp" / "worktrees" / wt_name / "base-branch"
+        if sidecar.is_file():
+            return sidecar.read_text().strip() or None
+    except (OSError, ValueError):
+        pass
+    return None
+
+
+async def _resolve_base_ref(config_base_ref: str | None, cwd: Path) -> str | None:
+    """Resolve the base ref for test filtering.
+
+    Resolution chain (first non-None wins):
+    1. Config override (explicit base_ref in TestCheckConfig)
+    2. Worktree sidecar (base-branch file written by implement-worktree skills)
+    3. Git upstream tracking ref (``@{upstream}`` of current branch)
+    4. None (no base ref available)
+    """
+    if config_base_ref:
+        return config_base_ref
+
+    sidecar_ref = _read_sidecar_base_branch(cwd)
+    if sidecar_ref:
+        return sidecar_ref
+
+    try:
+        proc = await asyncio.create_subprocess_exec(
+            "git",
+            "rev-parse",
+            "--abbrev-ref",
+            "@{upstream}",
+            cwd=cwd,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.DEVNULL,
+        )
+        try:
+            await asyncio.wait_for(proc.wait(), timeout=15.0)
+        except TimeoutError:
+            proc.kill()
+            await proc.wait()
+            return None
+        if proc.returncode == 0:
+            assert proc.stdout is not None
+            raw = await proc.stdout.read()
+            ref = raw.decode().strip()
+            if ref:
+                return ref
+    except OSError:
+        pass
+
+    return None
 
 
 _OUTCOME_PATTERN = re.compile(
@@ -116,6 +189,15 @@ class DefaultTestRunner:
         command = self._config.test_check.command
         timeout = float(self._config.test_check.timeout)
         env = build_sanitized_env()
+
+        filter_mode = self._config.test_check.filter_mode
+        if filter_mode:
+            env["AUTOSKILLIT_TEST_FILTER"] = filter_mode
+
+        base_ref = await _resolve_base_ref(self._config.test_check.base_ref, cwd)
+        if base_ref:
+            env["AUTOSKILLIT_TEST_BASE_REF"] = base_ref
+
         result = await self._runner(command, cwd=cwd, timeout=timeout, env=env)
         passed = check_test_passed(result.returncode, result.stdout, result.stderr)
         return TestResult(passed=passed, stdout=result.stdout, stderr=result.stderr)
--- a/tests/execution/test_testing.py
+++ b/tests/execution/test_testing.py
@@ -434,6 +434,16 @@ async def test_resolve_base_ref_git_upstream_fallback(tmp_path):
     repo.mkdir()
     subprocess.run(["git", "init", str(repo)], check=True, capture_output=True)
     subprocess.run(
+        ["git", "-C", str(repo), "config", "user.email", "test@test.com"],
+        check=True,
+        capture_output=True,
+    )
+    subprocess.run(
+        ["git", "-C", str(repo), "config", "user.name", "test"],
+        check=True,
+        capture_output=True,
+    )
+    subprocess.run(
         ["git", "-C", str(repo), "commit", "--allow-empty", "-m", "init"],
         check=True,
         capture_output=True,
--- a/tests/execution/test_testing.py
+++ b/tests/execution/test_testing.py
@@ -6,6 +6,7 @@ from pathlib import Path
 
 import pytest
 
+from autoskillit.config.settings import TestCheckConfig
 from autoskillit.core.types import (
     AUTOSKILLIT_PRIVATE_ENV_VARS,
     SubprocessResult,
@@ -13,6 +14,8 @@ from autoskillit.core.types import (
 )
 from autoskillit.execution.testing import (
     DefaultTestRunner,
+    _read_sidecar_base_branch,
+    _resolve_base_ref,
     build_sanitized_env,
 )
 from autoskillit.execution.testing import (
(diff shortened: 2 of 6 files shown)


src/autoskillit/execution/testing.py
```diff
@@ -32,6 +33,78 @@ def build_sanitized_env() -> dict[str, str]:
...
+        )
+        try:
+            await asyncio.wait_for(proc.wait(), timeout=15.0)
+        except TimeoutError:
+            proc.kill()
+            await proc.wait()
+            return None
+        if proc.returncode == 0:
+            assert proc.stdout is not None
+            raw = await proc.stdout.read()
```
Trecek, line 101: [warning] bugs: `proc.stdout.read()` is called after `proc.wait()` completes. If the subprocess wrote enough output to fill the OS pipe buffer, `wait()` would deadlock. For `git rev-parse --abbrev-ref` the output is always tiny so this is safe in practice, but the canonical pattern is `proc.communicate()` or reading stdout concurrently with waiting.

src/autoskillit/execution/testing.py
```diff
@@ -32,6 +33,78 @@ def build_sanitized_env() -> dict[str, str]:
...
+            stderr=asyncio.subprocess.DEVNULL,
+        )
+        try:
+            await asyncio.wait_for(proc.wait(), timeout=15.0)
+        except TimeoutError:
+            proc.kill()
+            await proc.wait()
+            return None
+        if proc.returncode == 0:
+            assert proc.stdout is not None
```
Trecek, line 100: [warning] defense: `assert proc.stdout is not None` is used for type narrowing, but asserts can be stripped with `-O`. Since `stdout=PIPE` is unconditionally set above, this is always true — but an explicit `if proc.stdout is None: return None` guard would be more robust than an assert.

src/autoskillit/execution/testing.py
```diff
@@ -32,6 +33,78 @@ def build_sanitized_env() -> dict[str, str]:
...
+            "--abbrev-ref",
+            "@{upstream}",
+            cwd=cwd,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.DEVNULL,
+        )
+        try:
+            await asyncio.wait_for(proc.wait(), timeout=15.0)
+        except TimeoutError:
+            proc.kill()
```
Trecek: [warning] defense: After `proc.kill()` on timeout, the subsequent `await proc.wait()` can itself raise `OSError` if the process already exited. This is caught by the outer `except OSError`, which swallows the timeout context — a timeout event becomes indistinguishable from a spawn failure in logs.

src/autoskillit/execution/testing.py
```diff
@@ -32,6 +33,78 @@ def build_sanitized_env() -> dict[str, str]:
...
+    if not git_path.is_file():
+        return None
+    try:
+        content = git_path.read_text().strip()
+        if not content.startswith("gitdir:"):
+            return None
+        gitdir = Path(content.split(":", 1)[1].strip())
+        if not gitdir.is_absolute():
+            gitdir = (cwd / gitdir).resolve()
+        main_git = gitdir.parent.parent
```
Trecek, line 53: [warning] defense: `gitdir.parent.parent` assumes the worktree gitdir is always exactly `<repo>/.git/worktrees/<name>` (two levels deep). If the gitdir layout differs (submodule, custom GIT_DIR), `project_root` will be wrong and the sidecar lookup silently reads the wrong path or misses entirely. Consider validating that `main_git.name == '.git'` before proceeding.

src/autoskillit/execution/testing.py
```diff
@@ -116,6 +189,15 @@ async def run(self, cwd: Path) -> TestResult:
         command = self._config.test_check.command
         timeout = float(self._config.test_check.timeout)
         env = build_sanitized_env()
+
+        filter_mode = self._config.test_check.filter_mode
+        if filter_mode:
+            env["AUTOSKILLIT_TEST_FILTER"] = filter_mode
```
Trecek, line 198: [warning] cohesion: The env var `AUTOSKILLIT_TEST_FILTER` drops `_MODE` from the config field name `filter_mode`, while `base_ref` maps symmetrically to `AUTOSKILLIT_TEST_BASE_REF`. A symmetric mapping would be `AUTOSKILLIT_TEST_FILTER_MODE`. (requires human decision — the env var name may be intentionally shortened)

tests/execution/test_testing.py
```diff
@@ -326,3 +328,170 @@ def test_test_result_dataclass_fields() -> None:
...
+
+
+@pytest.mark.anyio
+async def test_resolve_base_ref_git_upstream_fallback(tmp_path):
+    import subprocess
+
+    repo = tmp_path / "repo"
+    repo.mkdir()
+    subprocess.run(["git", "init", str(repo)], check=True, capture_output=True)
+    subprocess.run(
```
Trecek, line 446: [warning] tests: The git commit in this test does not configure `user.email`/`user.name`, which can cause the commit to fail in CI environments where git identity is not globally configured. Add `git -C <repo> config user.email test@test.com` and `git -C <repo> config user.name test` before the commit.

tests/execution/test_testing.py
```diff
@@ -326,3 +328,170 @@ def test_test_result_dataclass_fields() -> None:
...
+
+
+@pytest.mark.anyio
+async def test_resolve_base_ref_returns_none_when_no_source(tmp_path):
+    result = await _resolve_base_ref(None, tmp_path)
+    assert result is None
+
+
+@pytest.mark.anyio
+async def test_resolve_base_ref_git_upstream_fallback(tmp_path):
```
Trecek, line 430: [warning] tests: `git branch --set-upstream-to=upstream-branch` sets the tracking branch for the current local branch to another local branch, not a remote tracking ref. If this command fails silently (e.g., because no upstream-branch exists at that point in the git init sequence), `_resolve_base_ref` returns None and the test passes trivially — a false green.

Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (7 actionable warnings found across src/autoskillit/execution/testing.py and tests/execution/test_testing.py)

Trecek (commented):
AutoSkillit review found 7 blocking warnings. See inline comments. (Note: could not submit as REQUEST_CHANGES on own PR — using COMMENT instead.)

src/autoskillit/execution/testing.py
Trecek: Valid observation — flagged for design decision. The reviewer correctly identifies the read-after-wait pattern. For `git rev-parse --abbrev-ref @{upstream}` output is always a single short line so deadlock is impossible in practice, but `proc.communicate()` would be the canonical pattern. Deferring to human decision on whether to refactor.

src/autoskillit/execution/testing.py
Trecek: Investigated — this is intentional. `stdout=asyncio.subprocess.PIPE` is unconditionally set at subprocess creation (line 87), so `proc.stdout` is structurally never `None`. The `assert` is a valid type-narrowing idiom used throughout the codebase. The `-O` concern is theoretical since this project never runs with optimizations stripped.

src/autoskillit/execution/testing.py
Trecek: Agreed — fixed in bb5ce788. Wrapped `proc.kill()` + `await proc.wait()` in an inner `try/except OSError` so a race (process already exited before kill) doesn't propagate into the outer handler, which would silently swallow the timeout as a spawn failure.

src/autoskillit/execution/testing.py
Trecek: Investigated — this is intentional. The function guards with `if not git_path.is_file(): return None` (line 44), ensuring only linked worktrees reach this code. For linked worktrees, git always writes gitdir as `<repo>/.git/worktrees/<name>`, making `parent.parent` correct. The function is fail-open by design — if the sidecar path is wrong (e.g., submodule), the file simply doesn't exist and `None` is returned.

src/autoskillit/execution/testing.py
Trecek: Investigated — this is intentional. `AUTOSKILLIT_TEST_FILTER` is a pre-existing canonical env var established before this commit, used throughout `tests/conftest.py`, `tests/CLAUDE.md`, and `tests/_test_filter.py`. The config field `filter_mode` maps to this existing env var by design. Renaming to `AUTOSKILLIT_TEST_FILTER_MODE` would break the conftest plugin and all documentation.

tests/execution/test_testing.py
Trecek: Agreed — fixed in 0abbe624. Added `git -C <repo> config user.email test@test.com` and `git -C <repo> config user.name test` before the commit to ensure the test works in CI environments without global git identity.

tests/execution/test_testing.py
Trecek: Investigated — this is intentional. The branch `upstream-branch` is explicitly created at lines 441-444 before the `--set-upstream-to` at line 446, so the command cannot fail. All `subprocess.run` calls use `check=True`, raising `CalledProcessError` on any failure — no silent pass is possible. `git rev-parse --abbrev-ref @{upstream}` correctly returns the tracking branch name when set to a local branch; this is valid git behavior and the test correctly exercises the local-branch upstream fallback.

PR #980, opened 2026-04-16
Anti-Drift Verification System for Test Path Filter

### LINKED ISSUE #936: Anti-drift verification system
**Category:** Verification
**Depends on:** #08, #09, #10

## Problem

The filter could develop false negatives over time as the codebase evolves — tests that should run but don't. Without a verification mechanism, regressions silently ship.

### Shadow-mode comparison script

Create `scripts/verify-test-filter.sh`:
- The script takes a `base_ref` argument that sets `AUTOSKILLIT_TEST_BASE_REF` for the filtered collection run. `HEAD~5` means "compare against 5 commits ago on main" — this is the git diff base, not a number of commits to check.
- CI workflow must include `task install-worktree` step before running the script, and set `TMPDIR` per the Taskfile pattern.

**Note:** This script uses `.venv/bin/pytest` directly (not `task test-check`) because it needs raw `--collect-only` output without the Taskfile harness. This is an intentional exception to the "no bare pytest" convention — the script is a standalone audit tool under `scripts/`, not a test runner.

1. Run `.venv/bin/pytest tests/ --collect-only -q --no-header -o "addopts=" -m "not smoke"` without filter → `full_selected.txt`
2. Run `.venv/bin/pytest tests/ --collect-only -q --no-header -o "addopts=" -m "not smoke"` with conservative filter (set `AUTOSKILLIT_TEST_BASE_REF=$1 AUTOSKILLIT_TEST_FILTER=conservative`) → `filter_selected.txt`
3. `comm -23` to find missed tests
4. Run missed tests — if any fail, alert "false negative detected"
5. Report reduction percentage

### CI workflow

Create `.github/workflows/test-filter-audit.yml`:
- Schedule: Weekly (Monday 3am UTC)
- Also triggered via `workflow_dispatch`
- Runs `scripts/verify-test-filter.sh HEAD~5` against recent main commits
- Fails the job if a false negative is detected

### Canary test pattern documentation

Document the canary pattern in `tests/test_test_filter.py`:
- Canary tests must be gated behind a pytest marker (e.g., `@pytest.mark.canary`) and excluded from default runs via `-m 'not canary'` in the Taskfile. They only run when explicitly invoked during filter development.
- Development-time technique, not a permanent test

## Acceptance Criteria

- [ ] `scripts/verify-test-filter.sh` exists and runs correctly
- [ ] `.github/workflows/test-filter-audit.yml` exists with weekly schedule
- [ ] Script reports: total tests, filtered tests, reduction %, missed tests
- [ ] Script runs missed tests and detects false negatives
- [ ] Canary pattern documented
- [ ] Unit test: `test_shadow_diff_detects_missed_tests` — given two lists of test IDs (full and filtered), verify the `comm` logic correctly identifies IDs present in full but absent from filtered

## Context from plan

See plan sections: 10 (Anti-Drift Verification)

Trecek:
## Summary

Create a shadow-mode verification system that detects false negatives in the test path filter by comparing full vs filtered pytest collection, running any missed tests, and reporting reduction metrics. Three deliverables: a shell script (`scripts/verify-test-filter.sh`), a GitHub Actions workflow (`.github/workflows/test-filter-audit.yml`), and unit tests with canary pattern documentation in `tests/test_test_filter.py`.

## Requirements

- `scripts/verify-test-filter.sh` exists and runs correctly
- `.github/workflows/test-filter-audit.yml` exists with weekly schedule
- Script reports: total tests, filtered tests, reduction %, missed tests
- Script runs missed tests and detects false negatives
- Canary pattern documented
- Unit test: `test_shadow_diff_detects_missed_tests` — given two lists of test IDs (full and filtered), verify the `comm` logic correctly identifies IDs present in full but absent from filtered

Closes #936

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-936-20260415-215002-555319/.autoskillit/temp/make-plan/anti_drift_verification_plan_2026-04-15_215500.md`

## Token Usage Summary

| make_plan | 1.6k | 9.7k | 817.4k | 61.8k | 2 | 4m 49s |
| review_approach | 12 | 4.7k | 125.4k | 6.6k | 1 | 4m 9s |
| dry_walkthrough | 42 | 12.0k | 886.3k | 74.8k | 1 | 5m 47s |
| implement | 70 | 10.0k | 968.7k | 38.9k | 1 | 3m 28s |
| prepare_pr | 28 | 3.1k | 287.6k | 27.1k | 1 | 1m 37s |
| run_arch_lenses | 1.2k | 4.6k | 193.4k | 23.7k | 1 | 2m 58s |
| compose_pr | 22 | 1.7k | 105.9k | 14.4k | 1 | 36s |
| **Total** | 2.9k | 45.7k | 3.4M | 247.3k | | 23m 27s |

### DIFF
--- a/.github/workflows/test-filter-audit.yml
+++ b/.github/workflows/test-filter-audit.yml
@@ -7,7 +7,7 @@ on:
     branches: [main]
     paths:
       - "tests/**"
-      - "conftest.py"
+      - "tests/conftest.py"
       - "tests/_test_filter.py"
       - "src/autoskillit/_test_filter.py"
       - ".autoskillit/test-filter-manifest.yaml"
--- a/.github/workflows/test-filter-audit.yml
+++ b/.github/workflows/test-filter-audit.yml
@@ -1,0 +1,56 @@
+name: Test Filter Audit
+
+on:
+  schedule:
+    - cron: "17 3 * * 1"  # Weekly Monday 3:17am UTC (offset avoids peak contention)
+  push:
+    branches: [main]
+    paths:
+      - "tests/**"
+      - "conftest.py"
+      - "tests/_test_filter.py"
+      - "src/autoskillit/_test_filter.py"
+      - ".autoskillit/test-filter-manifest.yaml"
+      - "Taskfile.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  audit:
+    name: Shadow-mode filter verification
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    env:
+      TMPDIR: /dev/shm/pytest-audit
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 10  # Need history for HEAD~5
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          uv-version: "0.9.21"
+
+      - name: Install go-task
+        uses: arduino/setup-task@v2
+        with:
+          version: "3.43.3"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Configure git auth for private deps
+        run: git config --global url."https://${{ secrets.GH_USER }}:${{ secrets.GH_PAT }}@github.com/TalonT-Org/".insteadOf "https://github.com/TalonT-Org/"
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Set up environment
+        run: |
+          mkdir -p "$TMPDIR"
+          task install-worktree
+
+      - name: Run test filter audit
+        run: scripts/verify-test-filter.sh HEAD~5
--- a/scripts/verify-test-filter.sh
+++ b/scripts/verify-test-filter.sh
@@ -80,11 +80,11 @@ echo "Missed tests (in full but not in filtered): $MISSED"
 echo ""
 echo "--- Report ---"
 if [[ "$TOTAL" -gt 0 ]]; then
-    REDUCTION=$(( (TOTAL - FILTERED) * 100 / TOTAL ))
+    REDUCTION=$(awk "BEGIN {printf \"%.1f\", ($TOTAL - $FILTERED) * 100 / $TOTAL}")
     echo "Reduction: ${REDUCTION}% ($TOTAL → $FILTERED)"
 else
-    REDUCTION=0
-    echo "Reduction: 0% (no tests collected)"
+    REDUCTION="0.0"
+    echo "Reduction: 0.0% (no tests collected)"
 fi
 
 echo "Total tests:    $TOTAL"
--- a/scripts/verify-test-filter.sh
+++ b/scripts/verify-test-filter.sh
@@ -1,0 +1,122 @@
+#!/usr/bin/env bash
+# Anti-drift verification for test path filtering.
+#
+# Shadow-mode comparison: collects tests with and without the filter,
+# identifies missed tests, runs them to detect false negatives, and
+# reports reduction percentage.
+#
+# Usage: scripts/verify-test-filter.sh <base_ref>
+#   base_ref: git ref for AUTOSKILLIT_TEST_BASE_REF (e.g. HEAD~5)
+#
+# NOTE: This script uses .venv/bin/python -m pytest directly (not task
+# test-check) because it needs raw --collect-only output without the
+# Taskfile harness. This is an intentional exception to the "no bare
+# pytest" convention.
+
+set -euo pipefail
+
+# Deterministic byte-order for sort and comm — prevents locale-dependent
+# collation surprises with parametrized test IDs (brackets, hyphens, etc.)
+export LC_ALL=C
+
+BASE_REF="${1:?Usage: $0 <base_ref>}"
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+cd "$PROJECT_ROOT"
+
+PYTEST=".venv/bin/python -m pytest"
+if [[ ! -x ".venv/bin/python" ]]; then
+    echo "ERROR: .venv/bin/python not found. Run 'task install-worktree' first." >&2
+    exit 1
+fi
+
+WORK_DIR="$(mktemp -d "${TMPDIR:-/tmp}/test-filter-audit.XXXXXX")"
+trap 'rm -rf "$WORK_DIR"' EXIT
+
+FULL_FILE="$WORK_DIR/full_selected.txt"
+FILTER_FILE="$WORK_DIR/filter_selected.txt"
+MISSED_FILE="$WORK_DIR/missed_tests.txt"
+
+echo "=== Test Filter Anti-Drift Verification ==="
+echo "Base ref: $BASE_REF"
+echo ""
+
+# --- Step 1: Full collection (no filter) ---
+echo "--- Step 1: Collecting all tests (no filter) ---"
+$PYTEST tests/ --collect-only -q --no-header --disable-warnings -o "addopts=" -m "not smoke and not canary" 2>/dev/null \
+    | grep '::' \
+    | sort > "$FULL_FILE"
+
+TOTAL=$(wc -l < "$FULL_FILE")
+echo "Total tests collected: $TOTAL"
+
+if [[ "$TOTAL" -eq 0 ]]; then
+    echo "ERROR: No tests collected in full run. Something is wrong." >&2
+    exit 1
+fi
+
+# --- Step 2: Filtered collection (conservative) ---
+echo ""
+echo "--- Step 2: Collecting tests with conservative filter (base_ref=$BASE_REF) ---"
+AUTOSKILLIT_TEST_BASE_REF="$BASE_REF" \
+AUTOSKILLIT_TEST_FILTER=conservative \
+$PYTEST tests/ --collect-only -q --no-header --disable-warnings -o "addopts=" -m "not smoke and not canary" 2>/dev/null \
+    | grep '::' \
+    | sort > "$FILTER_FILE"
+
+FILTERED=$(wc -l < "$FILTER_FILE")
+echo "Filtered tests collected: $FILTERED"
+
+# --- Step 3: Find missed tests ---
+echo ""
+echo "--- Step 3: Computing shadow diff ---"
+comm -23 "$FULL_FILE" "$FILTER_FILE" > "$MISSED_FILE"
+
+MISSED=$(wc -l < "$MISSED_FILE")
+echo "Missed tests (in full but not in filtered): $MISSED"
+
+# --- Step 4: Report reduction ---
+echo ""
+echo "--- Report ---"
+if [[ "$TOTAL" -gt 0 ]]; then
+    REDUCTION=$(( (TOTAL - FILTERED) * 100 / TOTAL ))
+    echo "Reduction: ${REDUCTION}% ($TOTAL → $FILTERED)"
+else
+    REDUCTION=0
+    echo "Reduction: 0% (no tests collected)"
+fi
+
+echo "Total tests:    $TOTAL"
+echo "Filtered tests: $FILTERED"
+echo "Missed tests:   $MISSED"
+echo "Reduction:      ${REDUCTION}%"
+
+# --- Step 5: Run missed tests if any ---
+if [[ "$MISSED" -gt 0 ]]; then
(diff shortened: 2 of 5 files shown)


scripts/verify-test-filter.sh
```diff
@@ -0,0 +1,122 @@
...
+echo "Missed tests:   $MISSED"
+echo "Reduction:      ${REDUCTION}%"
+
+# --- Step 5: Run missed tests if any ---
+if [[ "$MISSED" -gt 0 ]]; then
+    echo ""
+    echo "--- Step 5: Running $MISSED missed tests to detect false negatives ---"
+
+    set +e
+    $PYTEST $(cat "$MISSED_FILE" | tr '\n' ' ') \
```
Trecek, line 101: [critical] bugs: Unquoted `$(cat "$MISSED_FILE" | tr '\n' ' ')` is unsafe: word-splitting on spaces in test IDs, ARG_MAX limits for large lists. If comm writes content without trailing newline, `wc -l` returns 0 and the L96 guard is bypassed, silently skipping missed test execution. Use xargs or a pytest nodeids file.

scripts/verify-test-filter.sh
```diff
@@ -0,0 +1,122 @@
...
+comm -23 "$FULL_FILE" "$FILTER_FILE" > "$MISSED_FILE"
+
+MISSED=$(wc -l < "$MISSED_FILE")
+echo "Missed tests (in full but not in filtered): $MISSED"
+
+# --- Step 4: Report reduction ---
+echo ""
+echo "--- Report ---"
+if [[ "$TOTAL" -gt 0 ]]; then
+    REDUCTION=$(( (TOTAL - FILTERED) * 100 / TOTAL ))
```
Trecek: [critical] defense: Integer division `(TOTAL - FILTERED) * 100 / TOTAL` truncates toward zero — a sub-1% reduction reports as 0%. The L54 guard prevents actual division-by-zero, but the truncation is misleading in audit output.

scripts/verify-test-filter.sh
```diff
@@ -0,0 +1,122 @@
...
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+cd "$PROJECT_ROOT"
+
+PYTEST=".venv/bin/python -m pytest"
+if [[ ! -x ".venv/bin/python" ]]; then
+    echo "ERROR: .venv/bin/python not found. Run 'task install-worktree' first." >&2
+    exit 1
+fi
+
+WORK_DIR="$(mktemp -d "${TMPDIR:-/tmp}/test-filter-audit.XXXXXX")"
```
Trecek, line 34: [warning] defense: WORK_DIR is created in TMPDIR or /tmp, not in the project's .autoskillit/temp/ directory. CLAUDE.md section 3.2 requires all temp files go in .autoskillit/temp/. In CI this may be acceptable but local runs will litter system tmp.

scripts/verify-test-filter.sh
```diff
@@ -0,0 +1,122 @@
...
+    | grep '::' \
+    | sort > "$FILTER_FILE"
+
+FILTERED=$(wc -l < "$FILTER_FILE")
+echo "Filtered tests collected: $FILTERED"
+
+# --- Step 3: Find missed tests ---
+echo ""
+echo "--- Step 3: Computing shadow diff ---"
+comm -23 "$FULL_FILE" "$FILTER_FILE" > "$MISSED_FILE"
```
Trecek, line 74: [warning] bugs: `comm -23` requires both inputs sorted with identical locale collation. The script exports LC_ALL=C and pipes through sort, but if a pytest plugin overrides locale between sort and comm steps, output could be incorrect. The dependency is implicit.

.github/workflows/test-filter-audit.yml
```diff
@@ -0,0 +1,56 @@
...
+
+on:
+  schedule:
+    - cron: "17 3 * * 1"  # Weekly Monday 3:17am UTC (offset avoids peak contention)
+  push:
+    branches: [main]
+    paths:
+      - "tests/**"
+      - "conftest.py"
+      - "tests/_test_filter.py"
```
Trecek, line 11: [warning] arch: `tests/_test_filter.py` is listed as a trigger path but the canonical module location is `src/autoskillit/_test_filter.py` (already listed on L12). If the file does not exist at `tests/_test_filter.py`, this trigger path never fires.

.github/workflows/test-filter-audit.yml
```diff
@@ -0,0 +1,56 @@
...
+    name: Shadow-mode filter verification
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    env:
+      TMPDIR: /dev/shm/pytest-audit
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 10  # Need history for HEAD~5
```
Trecek, line 31: [warning] defense: `fetch-depth: 10` with `HEAD~5` leaves only a 5-commit safety margin. If the workflow triggers after a long squash chain, HEAD~5 may not resolve. Consider fetch-depth: 0 or validating that HEAD~5 exists before invoking the script.

.github/workflows/test-filter-audit.yml
```diff
@@ -0,0 +1,56 @@
+name: Test Filter Audit
+
+on:
+  schedule:
+    - cron: "17 3 * * 1"  # Weekly Monday 3:17am UTC (offset avoids peak contention)
+  push:
+    branches: [main]
+    paths:
+      - "tests/**"
+      - "conftest.py"
```
Trecek: [warning] cohesion: Path trigger `conftest.py` refers to root-level file — the actual conftest lives at `tests/conftest.py`. This path will never match a push event, silently breaking automatic trigger for conftest changes.

tests/test_test_filter.py
```diff
@@ -765,3 +765,109 @@ def test_summary_warning_emitted(
...
+        ]
+
+        full_file = tmp_path / "full_selected.txt"
+        filtered_file = tmp_path / "filter_selected.txt"
+        missed_file = tmp_path / "missed_tests.txt"
+
+        full_file.write_text("\n".join(full_ids) + "\n")
+        filtered_file.write_text("\n".join(filtered_ids) + "\n")
+
+        result = subprocess.run(
```
Trecek, line 829: [warning] tests: `comm -23` requires sorted input. Test data is hand-sorted but not enforced — future edits reordering the list would silently produce wrong output. Also, subprocess.run is called without `env={'LC_ALL': 'C'}` unlike the script itself, creating a locale asymmetry.

tests/test_test_filter.py
```diff
@@ -765,3 +765,109 @@ def test_summary_warning_emitted(
...
+            ["comm", "-23", str(full_file), str(filtered_file)],
```
Trecek, line 830: [warning] tests: `comm` is a POSIX utility not available on Windows. If CI ever runs on Windows, this test will fail with FileNotFoundError. Consider a platform guard or reimplementing the set-difference in pure Python.

tests/test_test_filter.py
```diff
@@ -765,3 +765,109 @@ def test_summary_warning_emitted(
...
+
+    def test_shadow_diff_empty_filtered(self, tmp_path: Path) -> None:
+        """When filter selects nothing, all full IDs are missed."""
+        full_ids = ["tests/core/test_core.py::test_a", "tests/core/test_core.py::test_b"]
+
+        full_file = tmp_path / "full_selected.txt"
+        filtered_file = tmp_path / "filter_selected.txt"
+
+        full_file.write_text("\n".join(full_ids) + "\n")
+        filtered_file.write_text("\n")
```
Trecek, line 875: [warning] tests: `filtered_file` is written as `'\n'` (single newline) rather than `''` (empty). comm on some platforms may treat a file with a blank line differently from truly empty, potentially causing unexpected behavior. Use `write_text('')` for an empty selection.

tests/test_test_filter.py
```diff
@@ -765,3 +765,109 @@ def test_summary_warning_emitted(
         pytester.makepyfile(test_drop="def test_drop(): pass")
         result = pytester.runpytest("-v", "-W", "always")
         result.stdout.fnmatch_lines(["*Test filter:*selected*deselected*"])
+
+
+# ---------------------------------------------------------------------------
```
Trecek, line 770: [warning] slop: Large block comment (L770-L790) documents a canary pattern that 'should be removed once filter validation is complete'. This is a stale planning note in test code — no actual canary-marked tests are added in this PR. Either add the tests or remove the comment.

Trecek (commented):

Trecek (commented):
AutoSkillit review found 11 blocking issues (2 critical, 9 warning). See inline comments. verdict=changes_requested

scripts/verify-test-filter.sh
Trecek: Valid observation — flagged for design decision. The word-splitting concern for parametrized test IDs is real, but the wc-l bypass claim is factually incorrect: comm always writes newline-terminated output per POSIX, so the L96 guard cannot misfire. Flagged for human review on whether to adopt xargs/nodeid-file approach.

scripts/verify-test-filter.sh
Trecek: Agreed — fixed in 86a627c8555ee56cd64da95963e1655465064ca7. Switched to awk for floating-point percentage display so sub-1% reductions are accurately reported.

scripts/verify-test-filter.sh
Trecek: Investigated — this is intentional. The script uses mktemp with an EXIT trap (line 35) that unconditionally cleans up. Using /tmp for shell script working directories is idiomatic POSIX practice; the CLAUDE.md 3.2 convention targets application-layer temp files that may persist, not shell scripts with guaranteed cleanup.

scripts/verify-test-filter.sh
Trecek: Investigated — this is intentional. LC_ALL=C is exported at line 20 of the script, inherited by all child processes (sort on lines 49/66 and comm on line 74). A pytest subprocess cannot retroactively modify the parent shell's exported environment. The dependency is documented explicitly in the comment at lines 18-20.

.github/workflows/test-filter-audit.yml
Trecek: Investigated — this is intentional. tests/_test_filter.py exists as a real standalone module (conftest plugin for test filtering). It is distinct from src/autoskillit/_test_filter.py on line 12 — both files exist and both paths are legitimate workflow triggers.

.github/workflows/test-filter-audit.yml
Trecek: Valid observation — flagged for design decision. fetch-depth: 10 provides a 4-commit margin beyond HEAD~5, which is sufficient for typical single-commit pushes but tight for weekly schedule triggers where many commits may have accumulated. Whether to use fetch-depth: 0 (slower but bulletproof) vs. adding a pre-check guard is a cost/robustness trade-off.

.github/workflows/test-filter-audit.yml
Trecek: Agreed — fixed in 130abfe981e648af6fddbc0b40373b44a0855005. Root-level conftest.py does not exist; corrected to tests/conftest.py so the trigger path actually fires on conftest changes.

tests/test_test_filter.py
Trecek: Agreed — fixed in fc9204b13428d2db3fa2fadd039775b969ae5ac2. Added explicit sorted() calls to enforce ordering and passed env={"LC_ALL": "C"} to all subprocess.run calls to match the production script's locale handling.

tests/test_test_filter.py
Trecek: Investigated — this is intentional. CI matrix (.github/workflows/tests.yml) runs only ubuntu-latest and macos-15 — no Windows runners exist or are planned. comm is a standard POSIX utility available on both Linux and macOS. The concern is purely hypothetical given the project's CI targets.

tests/test_test_filter.py
Trecek: Investigated — this is intentional. comm -23 produces identical output for a file with a single newline vs. truly empty. The assertion uses strip().splitlines() which discards blank lines in both cases. The '\n' write is consistent with the '\n'.join(...) + '\n' pattern used for non-empty files in the same test class.

tests/test_test_filter.py
Trecek: Investigated — this is intentional. The canary block comment was introduced in the same commit (4cfcc9d5) that registers the canary marker in pyproject.toml and excludes it via -m 'not canary' in Taskfile.yml. It documents a development-time technique for filter developers — intentional scaffolding documentation, not a stale planning note.

PR #983, opened 2026-04-16
Protocol-Based Test Fakes

### LINKED ISSUE #937: [Tier 3] Protocol-based test fakes
**Category:** Tier 3 — Enhancements
**Depends on:** #01-#11 (core system operational)

## Problem

Tests for higher layers (server, cli) import concrete implementations from lower layers to set up test dependencies. This creates cross-layer import chains that reduce filter accuracy. The fix in #04 addresses the worst cases, but a systemic solution would eliminate the pattern entirely.

Note: `tests/conftest.py` already contains `StatefulMockTester` (TestRunner fake) and `MockSubprocessRunner` (SubprocessRunner fake). These should be migrated to `tests/fakes.py` as part of this work.

## Solution

For each protocol in `core/_type_protocols.py`, create a corresponding in-memory fake. Prioritized by test isolation impact:

Priority 1 (highest test isolation impact): `InMemoryRecipeRepository`, `InMemoryHeadlessExecutor`, `InMemoryTestRunner` (consolidating `StatefulMockTester`). Priority 2: `InMemoryCIWatcher`, `InMemoryMergeQueueWatcher`, `InMemoryDatabaseReader`. Priority 3 (defer): remaining 15 protocols (`GateState`, `AuditLog`, `TokenLog`, etc.) — these are mostly used only within their own layer's tests.

Collect them in `tests/fakes.py`. Tests for higher layers inject fakes exclusively via fixtures. mypy verifies conformance via structural subtyping.

## Value

**Highest-value Tier 3 item.** Directly improves filter accuracy and test isolation simultaneously — every test that switches from concrete imports to fakes removes one cross-layer dependency the filter must account for.

## Acceptance Criteria

- [ ] `tests/fakes.py` exists with all protocol fakes
- [ ] Each fake satisfies its protocol (`isinstance` check)
- [ ] At least server/ tests migrated to use fakes for non-server dependencies
- [ ] At least 5 server/ test files no longer import from execution/ or recipe/ at module level after switching to fakes
- [ ] All existing tests pass

## Context from plan

See plan section: 11 (Tier 3 — 11.4 Protocol-Based Test Fakes)

Trecek:
## Summary

Create `tests/fakes.py` as the single authoritative module for in-memory test doubles that satisfy protocols defined in `core/_type_protocols.py`. This consolidates 10+ inline `MockExecutor` classes scattered across server tests, replaces untyped `AsyncMock`/`MagicMock` usage for CI/merge-queue watchers, and migrates `MockSubprocessRunner` and `StatefulMockTester` out of `conftest.py`. Each fake carries `isinstance` conformance tests via a new `tests/test_fakes_conformance.py`. Six server/ test files are migrated to import from `tests.fakes`.

## Requirements

- [ ] `tests/fakes.py` exists with all protocol fakes
- [ ] Each fake satisfies its protocol (`isinstance` check)
- [ ] At least server/ tests migrated to use fakes for non-server dependencies
- [ ] At least 5 server/ test files no longer import from execution/ or recipe/ at module level after switching to fakes
- [ ] All existing tests pass

Closes #937

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-937-20260415-225324-347495/.autoskillit/temp/make-plan/protocol_test_fakes_plan_2026-04-15_225500.md`

## Token Usage Summary

| make_plan | 1.1k | 16.9k | 1.0M | 70.4k | 1 | 8m 13s |
| review_approach | 24 | 3.2k | 149.4k | 23.8k | 1 | 4m 32s |
| dry_walkthrough | 63 | 12.3k | 822.7k | 47.1k | 1 | 6m 32s |
| implement | 107 | 34.8k | 6.2M | 132.5k | 1 | 16m 9s |
| resolve_failures | 57 | 7.3k | 1.2M | 40.4k | 1 | 8m 8s |
| prepare_pr | 26 | 4.1k | 234.7k | 31.0k | 1 | 1m 41s |
| compose_pr | 25 | 2.2k | 186.3k | 18.5k | 1 | 47s |
| **Total** | 1.4k | 80.8k | 9.8M | 363.8k | | 46m 6s |

### DIFF
--- a/tests/fakes.py
+++ b/tests/fakes.py
@@ -140,7 +140,7 @@ class InMemoryHeadlessExecutor(HeadlessExecutor):
             )
         )
         if self._queue:
-            return self._queue.popleft()
+            return dataclasses.replace(self._queue.popleft())
         # Return a defensive copy so callers mutating fields (e.g. run_skill
         # setting order_id) don't pollute the shared default across tests.
         return dataclasses.replace(self._default)
--- a/tests/fakes.py
+++ b/tests/fakes.py
@@ -442,7 +442,7 @@ class MockSubprocessRunner(SubprocessRunner):
     """
 
     def __init__(self) -> None:
-        self._queue: list[SubprocessResult] = []
+        self._queue: deque[SubprocessResult] = deque()
         self._default = SubprocessResult(
             returncode=0,
             stdout="",
@@ -470,5 +470,5 @@ class MockSubprocessRunner(SubprocessRunner):
     ) -> SubprocessResult:
         self.call_args_list.append((cmd, cwd, timeout, kwargs))
         if self._queue:
-            return self._queue.pop(0)
+            return self._queue.popleft()
         return self._default
--- a/tests/fakes.py
+++ b/tests/fakes.py
@@ -118,7 +118,9 @@ class InMemoryHeadlessExecutor(HeadlessExecutor):
         )
         if self._queue:
             return self._queue.popleft()
-        return self._default
+        # Return a defensive copy so callers mutating fields (e.g. run_skill
+        # setting order_id) don't pollute the shared default across tests.
+        return dataclasses.replace(self._default)
 
 
 # ---------------------------------------------------------------------------
--- a/tests/fakes.py
+++ b/tests/fakes.py
@@ -1,0 +1,459 @@
+"""Protocol-based test fakes for autoskillit.
+
+Single authoritative module for in-memory test doubles that satisfy protocols
+defined in ``core/_type_protocols.py``. Imports only from L0 (``autoskillit.core``).
+"""
+
+from __future__ import annotations
+
+import dataclasses
+from collections import deque
+from collections.abc import Awaitable, Callable, Sequence
+from pathlib import Path
+from typing import Any
+
+from autoskillit.core.types import (
+    CIRunScope,
+    CIWatcher,
+    DatabaseReader,
+    HeadlessExecutor,
+    MergeQueueWatcher,
+    RecipeRepository,
+    SkillResult,
+    SubprocessResult,
+    SubprocessRunner,
+    TerminationReason,
+    TestResult,
+    TestRunner,
+    WriteBehaviorSpec,
+)
+
+# ---------------------------------------------------------------------------
+# HeadlessExecutor fake
+# ---------------------------------------------------------------------------
+
+
+@dataclasses.dataclass
+class ExecutorCall:
+    """Record of a single ``HeadlessExecutor.run()`` invocation."""
+
+    skill_command: str
+    cwd: str
+    model: str = ""
+    step_name: str = ""
+    kitchen_id: str = ""
+    order_id: str = ""
+    add_dirs: tuple[Any, ...] = ()
+    timeout: float | None = None
+    stale_threshold: float | None = None
+    idle_output_timeout: float | None = None
+    expected_output_patterns: tuple[str, ...] = ()
+    write_behavior: Any | None = None
+    completion_marker: str = ""
+
+
+_DEFAULT_SKILL_RESULT = SkillResult(
+    success=True,
+    result="ok",
+    session_id="",
+    subtype="success",
+    is_error=False,
+    exit_code=0,
+    needs_retry=False,
+    retry_reason="none",
+    stderr="",
+    token_usage=None,
+)
+
+
+class InMemoryHeadlessExecutor(HeadlessExecutor):
+    """In-memory test double for :class:`HeadlessExecutor`.
+
+    Supports a FIFO queue via :meth:`push` and records every call in
+    :attr:`calls`.
+    """
+
+    def __init__(self, default_result: SkillResult | None = None) -> None:
+        self._default = default_result or _DEFAULT_SKILL_RESULT
+        self._queue: deque[SkillResult] = deque()
+        self.calls: list[ExecutorCall] = []
+
+    def push(self, result: SkillResult) -> None:
+        """Enqueue a result to be returned by the next :meth:`run` call."""
+        self._queue.append(result)
+
+    async def run(
+        self,
+        skill_command: str,
+        cwd: str,
+        *,
+        model: str = "",
+        step_name: str = "",
+        kitchen_id: str = "",
+        order_id: str = "",
+        add_dirs: Sequence[Any] = (),
+        timeout: float | None = None,
+        stale_threshold: float | None = None,
+        idle_output_timeout: float | None = None,
+        expected_output_patterns: Sequence[str] = (),
+        write_behavior: WriteBehaviorSpec | None = None,
+        completion_marker: str = "",
+    ) -> SkillResult:
+        self.calls.append(
+            ExecutorCall(
+                skill_command=skill_command,
+                cwd=cwd,
+                model=model,
+                step_name=step_name,
+                kitchen_id=kitchen_id,
+                order_id=order_id,
+                add_dirs=tuple(add_dirs),
+                timeout=timeout,
+                stale_threshold=stale_threshold,
+                idle_output_timeout=idle_output_timeout,
+                expected_output_patterns=tuple(expected_output_patterns),
+                write_behavior=write_behavior,
+                completion_marker=completion_marker,
+            )
+        )
+        if self._queue:
+            return self._queue.popleft()
+        return self._default
+
+
+# ---------------------------------------------------------------------------
+# TestRunner fake
+# ---------------------------------------------------------------------------
+
+
+class InMemoryTestRunner(TestRunner):
+    """In-memory test double for :class:`TestRunner`.
+
+    Pops pre-configured results from a deque; falls back to a passing
+    ``TestResult`` when the deque is exhausted.
+    """
+
+    def __init__(self, results: list[TestResult] | None = None) -> None:
+        self._results: deque[TestResult] = deque(results or [])
+        self._call_count = 0
+        self.calls: list[Path] = []
+
+    async def run(self, cwd: Path) -> TestResult:
+        self._call_count += 1
+        self.calls.append(cwd)
+        if self._results:
+            return self._results.popleft()
+        return TestResult(passed=True, stdout="", stderr="")
+
+    @property
+    def call_count(self) -> int:
+        return self._call_count
+
+
+# ---------------------------------------------------------------------------
+# RecipeRepository fake
+# ---------------------------------------------------------------------------
+
+
+class InMemoryRecipeRepository(RecipeRepository):
+    """In-memory test double for :class:`RecipeRepository`."""
+
+    def __init__(self) -> None:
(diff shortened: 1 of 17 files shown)


tests/fakes.py
```diff
@@ -0,0 +1,461 @@
...
+                "pr_number": pr_number,
+                "target_branch": target_branch,
+                "repo": repo,
+                "cwd": cwd,
+                "timeout_seconds": timeout_seconds,
+                "poll_interval": poll_interval,
+            }
+        )
+        if self.wait_side_effect is not None:
+            return InMemoryCIWatcher._resolve_side_effect(self.wait_side_effect)
```
Trecek: [critical] cohesion: `InMemoryMergeQueueWatcher.wait()` delegates side-effect resolution to `InMemoryCIWatcher._resolve_side_effect()` — a private static method of an unrelated sibling class. Extract `_resolve_side_effect` as a module-level helper function so all three fakes reference it independently without cross-class coupling.

tests/fakes.py
```diff
@@ -0,0 +1,461 @@
...
+            {
+                "db_path": db_path,
+                "sql": sql,
+                "params": params,
+                "timeout_sec": timeout_sec,
+                "max_rows": max_rows,
+            }
+        )
+        if self.side_effect is not None:
+            return InMemoryCIWatcher._resolve_side_effect(self.side_effect)
```
Trecek: [critical] cohesion: `InMemoryDatabaseReader.query()` also delegates to `InMemoryCIWatcher._resolve_side_effect()`. Three unrelated fakes now depend on a private method of CIWatcher. Extract to a module-level `_resolve_side_effect()` helper.

tests/fakes.py
```diff
@@ -0,0 +1,461 @@
...
+
+
+_DEFAULT_SKILL_RESULT = SkillResult(
+    success=True,
+    result="ok",
+    session_id="",
+    subtype="success",
+    is_error=False,
+    exit_code=0,
+    needs_retry=False,
```
Trecek, line 85: [warning] defense: `retry_reason="none"` passes a plain `str`, but `SkillResult.retry_reason` is typed as `RetryReason` (a StrEnum from `_type_enums.py`). Use `RetryReason.NONE` for type-correct construction.

tests/fakes.py
```diff
@@ -0,0 +1,461 @@
...
+                timeout=timeout,
+                stale_threshold=stale_threshold,
+                idle_output_timeout=idle_output_timeout,
+                expected_output_patterns=tuple(expected_output_patterns),
+                write_behavior=write_behavior,
+                completion_marker=completion_marker,
+            )
+        )
+        if self._queue:
+            return self._queue.popleft()
```
Trecek: [warning] defense: Results popped from `_queue` are returned without a defensive copy. `SkillResult` has mutable fields (e.g. `write_path_warnings: list[str]`). Production code mutating the returned result would corrupt the pushed object. Apply `dataclasses.replace()` before returning, matching the treatment on line 123.

tests/fakes.py
```diff
@@ -0,0 +1,461 @@
...
+                "scope": scope,
+                "cwd": cwd,
+            }
+        )
+        if self.status_side_effect is not None:
+            return self._resolve_side_effect(self.status_side_effect)
+        return self._status_result
+
+    @staticmethod
+    def _resolve_side_effect(effect: Any) -> Any:
```
Trecek: [warning] defense: `_resolve_side_effect` calls `effect()` without awaiting when `callable(effect)` is true. If a caller sets `side_effect` to an async function, this returns an unawaited coroutine object instead of the expected result, producing a `RuntimeWarning` and wrong return type. The `Any` return annotation masks this from static analysis.

tests/fakes.py
```diff
@@ -0,0 +1,461 @@
...
+            return InMemoryCIWatcher._resolve_side_effect(self.side_effect)
+        return self._query_result
+
+
+# ---------------------------------------------------------------------------
+# SubprocessRunner fake (moved from conftest.py)
+# ---------------------------------------------------------------------------
+
+
+class MockSubprocessRunner(SubprocessRunner):
```
Trecek, line 434: [warning] cohesion: `MockSubprocessRunner` does not follow the `InMemory*` naming convention used by the other six fakes in this module. Consider renaming to `InMemorySubprocessRunner` for consistency, or document the intentional distinction.

tests/fakes.py
```diff
@@ -0,0 +1,461 @@
...
+
+    Inherits from SubprocessRunner (Protocol) so mypy verifies the __call__
+    signature matches the protocol at class definition, not just at call sites.
+
+    call_args_list stores (cmd, cwd, timeout, kwargs) tuples.
+    IMPORTANT: Assert [N][1] (cwd) when testing cwd propagation.
+    """
+
+    def __init__(self) -> None:
+        self._queue: list[SubprocessResult] = []
```
Trecek: [warning] cohesion: `_queue` is `list[SubprocessResult]` drained with `list.pop(0)` (O(n)), while `InMemoryHeadlessExecutor` and `InMemoryTestRunner` use `deque` + `popleft()` (O(1)). All three serve identical FIFO semantics — the implementation should be consistent.

Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (posted as COMMENT because reviewer is PR author)

tests/fakes.py
```diff
@@ -0,0 +1,461 @@
...
+        self._default = SubprocessResult(
```
Trecek, line 446: [warning] slop: Docstring line `IMPORTANT: Assert [N][1] (cwd) when testing cwd propagation.` is a usage instruction embedded in the class docstring with an `IMPORTANT:` escalation pattern. Usage instructions belong in the conformance test or a comment at the call site, not the implementation docstring.

tests/test_fakes_conformance.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+@pytest.mark.anyio
+async def test_runner_fallback_after_exhaustion():
+    runner = InMemoryTestRunner(results=[])
+    r = await runner.run(Path("/a"))
+    assert r.passed is True
+
+
+def test_runner_tracks_call_count():
```
Trecek: [warning] tests: `test_runner_tracks_call_count` only asserts `call_count == 0` on a freshly-constructed runner without invoking `.run()`. This is trivially true and does not exercise the tracking behavior. The test should call `runner.run()` and assert the count increments.

tests/fakes.py
```diff
@@ -0,0 +1,461 @@
...
+    """Test double for SubprocessRunner. Queues predetermined results.
+
+    Inherits from SubprocessRunner (Protocol) so mypy verifies the __call__
+    signature matches the protocol at class definition, not just at call sites.
+
+    call_args_list stores (cmd, cwd, timeout, kwargs) tuples.
+    IMPORTANT: Assert [N][1] (cwd) when testing cwd propagation.
+    """
+
+    def __init__(self) -> None:
```
Trecek, line 444: [warning] cohesion: `MockSubprocessRunner` exposes `set_default()` to override the fallback result post-construction, but `InMemoryHeadlessExecutor` accepts `default_result` only in `__init__` with no setter. The three queue-based fakes have inconsistent strategies for configuring their fallback.

tests/fakes.py
Trecek: Agreed — fixed in 772c6f95. Extracted `_resolve_side_effect` to a module-level helper function. All three fakes (`InMemoryCIWatcher`, `InMemoryMergeQueueWatcher`, `InMemoryDatabaseReader`) now reference the independent helper without cross-class coupling.

tests/fakes.py
Trecek: Agreed — fixed in 772c6f95. Same commit as above: `_resolve_side_effect` is now a module-level helper. The three-way cross-class coupling through `InMemoryCIWatcher._resolve_side_effect` is eliminated.

tests/fakes.py
Trecek: Agreed — fixed in 25c13767. Queue popleft now applies `dataclasses.replace()` before returning, matching the defensive copy treatment on the default path (line 123).

tests/fakes.py
Trecek: Agreed — fixed in 772c6f95. The module-level `_resolve_side_effect` now checks `asyncio.iscoroutine(result)` after calling `effect()` and runs the coroutine via `asyncio.get_event_loop().run_until_complete(result)` when detected.

tests/fakes.py
Trecek: Agreed — fixed in fe9204c5. `MockSubprocessRunner._queue` now uses `deque[SubprocessResult]` with `popleft()`, matching the O(1) FIFO pattern in `InMemoryHeadlessExecutor` and `InMemoryTestRunner`.

tests/test_fakes_conformance.py
Trecek: Agreed — fixed in 4629efe1. Test now calls `await runner.run()` twice and asserts `call_count` increments to 1 and then 2, exercising the actual tracking behavior.

tests/fakes.py
Trecek: Investigated — this is intentional. The `MockSubprocessRunner` name was preserved from the original conftest.py migration (commit 7c71bc6e). `tests/CLAUDE.md` explicitly documents both naming conventions: `InMemory*, MockSubprocessRunner`. There are 15 call sites importing this name. The naming reflects the migration history rather than an oversight.

tests/fakes.py
Trecek: Investigated — this is intentional. Usage instructions in test double docstrings are appropriate: test doubles exist to guide test authors, and the `call_args_list` tuple indexing convention (`[N][1]` for cwd) is non-obvious enough that inline documentation prevents errors at call sites. The `IMPORTANT:` pattern is standard in this codebase's documentation conventions.

tests/fakes.py
Trecek: Valid observation — flagged for design decision. The annotation mismatch is real (`str` vs `RetryReason`), but `RetryReason` is a `StrEnum` (subclass of `str`) making it transparent at runtime. Additionally, the pre-commit mypy hook only runs on `src/`, not `tests/`. The same pattern appears in `test_fakes_conformance.py` lines 78, 106, 117. Whether to add the import for type correctness in test code is a design preference.

tests/fakes.py
Trecek: Valid observation — flagged for design decision. The inconsistency is confirmed: `MockSubprocessRunner` has `set_default()`, `InMemoryHeadlessExecutor` accepts `default_result` only in `__init__`, and `InMemoryTestRunner` has no configurable fallback. However, each reflects its actual usage pattern — `MockSubprocessRunner` is used across 15 test files with varying setup sequences, while `InMemoryHeadlessExecutor` is typically constructed fresh per test. Unifying the API would be a larger design change.

PR #989, opened 2026-04-16
pytestmark Layer Markers

### LINKED ISSUE #938: [Tier 3] pytestmark layer markers
**Category:** Tier 3 — Enhancements
**Depends on:** #01-#11 (core system operational)

## Problem

The filter currently derives layer affinity from AST import analysis and directory structure. This is implicit and can be wrong if a test file imports across boundaries. An explicit declaration would be more auditable and allow the filter to cross-check against AST results.

## Solution

Add `pytestmark = [pytest.mark.layer("execution")]` to each test file. A conftest plugin validates at collection time that the marker matches the file's directory.

The filter can optionally use the marker instead of (or in addition to) AST analysis. AST analysis can verify the marker is correct by comparing declared layer against actual imports.

## Value

More maintenance (every new test file needs the marker) but more explicit, auditable, and grep-friendly. Enables future tooling like "show me all tests that declare layer=execution but import from server."

## Marker Registration

Register in pyproject.toml using the existing syntax: `layer(name): marks which source layer this test file exercises`. Parameterized markers work with plain registration.

## Scope

In scope: tests/core/, tests/config/, tests/pipeline/, tests/execution/, tests/workspace/, tests/recipe/, tests/migration/, tests/server/, tests/cli/. Out of scope: tests/arch/, tests/contracts/, tests/infra/, tests/docs/, tests/skills/ — these are cross-cutting and would carry `layer('cross-cutting')` or no layer marker.

## Relationship to Existing Arch Tests

This is additive. `test_layer_enforcement.py` enforces SOURCE import contracts. The layer marker validates that TEST files declare their layer affinity. Different concerns, no overlap.

## Implementation Notes

Consider splitting into two sub-tickets: (a) Add markers to all test files (mechanical, large diff), (b) Implement conftest validation plugin (design decisions required). The annotation pass can land first and provide value independently.

The filter does NOT consume markers in its first version — AST analysis remains the source of truth. Initial implementation: markers are documentation-only and used for cross-checking against AST results. Future: markers could replace AST analysis if proven reliable.

## Acceptance Criteria

- [ ] `pytest.mark.layer` marker registered in `pyproject.toml`
- [ ] All test files in L0-L3 directories have the marker
- [ ] Validation fixture in conftest warns on mismatch (marker vs directory)
- [ ] All existing tests pass with markers applied
- [ ] Update `tests/CLAUDE.md` with marker convention

## Context from plan

See plan section: 11 (Tier 3 — 11.1 pytestmark Source Module Markers)

Trecek:
## Summary

Add `pytestmark = [pytest.mark.layer("<subpackage>")]` to every `test_*.py` file in the 9 in-scope test directories (core, config, pipeline, execution, workspace, recipe, migration, server, cli). Register the `layer` marker in `pyproject.toml`. Add conftest-level collection validation that warns when a marker value mismatches the directory. Add an architectural enforcement test. Update `tests/CLAUDE.md`.

Out of scope: `tests/arch/`, `tests/contracts/`, `tests/infra/`, `tests/docs/`, `tests/skills/`, `tests/hooks/`, `tests/skills_extended/`, `tests/assets/`.

Closes #938

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-938-20260416-073648-433860/.autoskillit/temp/make-plan/pytestmark_layer_markers_plan_2026-04-16_074300.md`

## Token Usage Summary

| make_plan | 1.9k | 10.6k | 656.1k | 48.4k | 1 | 7m 10s |
| review_approach | 24 | 3.4k | 146.5k | 22.9k | 1 | 4m 23s |
| dry_walkthrough | 51 | 8.8k | 577.6k | 42.7k | 1 | 4m 49s |
| implement | 82 | 14.4k | 1.6M | 46.9k | 1 | 5m 31s |
| prepare_pr | 39 | 12.0k | 428.0k | 49.6k | 1 | 3m 30s |
| compose_pr | 24 | 1.9k | 171.3k | 18.5k | 1 | 55s |
| **Total** | 2.1k | 51.1k | 3.6M | 229.0k | | 26m 20s |

### DIFF
--- a/tests/arch/test_layer_markers.py
+++ b/tests/arch/test_layer_markers.py
@@ -90,6 +90,16 @@ def test_all_test_files_have_correct_layer_marker(directory: str) -> None:
     assert not errors, "\n".join(errors)
 
 
+def test_layer_directories_matches_conftest() -> None:
+    """LAYER_DIRECTORIES keys must match _LAYER_DIRS in conftest.py."""
+    from tests.conftest import _LAYER_DIRS
+
+    assert set(LAYER_DIRECTORIES.keys()) == _LAYER_DIRS, (
+        f"LAYER_DIRECTORIES keys {set(LAYER_DIRECTORIES.keys())} != "
+        f"conftest _LAYER_DIRS {_LAYER_DIRS}"
+    )
+
+
 def test_layer_marker_registered_in_pyproject() -> None:
     """The 'layer' marker is registered in pyproject.toml to avoid warnings."""
     import tomllib
--- a/tests/arch/test_layer_markers.py
+++ b/tests/arch/test_layer_markers.py
@@ -1,0 +1,102 @@
+"""Enforce pytestmark layer markers on all in-scope test files."""
+
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+import pytest
+
+TESTS_ROOT = Path(__file__).resolve().parent.parent
+
+LAYER_DIRECTORIES: dict[str, str] = {
+    "core": "core",
+    "config": "config",
+    "pipeline": "pipeline",
+    "execution": "execution",
+    "workspace": "workspace",
+    "recipe": "recipe",
+    "migration": "migration",
+    "server": "server",
+    "cli": "cli",
+}
+
+
+def _extract_layer_marker(path: Path) -> str | None:
+    """Parse a test file's AST and return the layer marker value, or None."""
+    tree = ast.parse(path.read_text())
+    for node in ast.iter_child_nodes(tree):
+        if not isinstance(node, ast.Assign):
+            continue
+        for target in node.targets:
+            if isinstance(target, ast.Name) and target.id == "pytestmark":
+                # pytestmark = [pytest.mark.layer("x"), ...]
+                if isinstance(node.value, ast.List):
+                    for elt in node.value.elts:
+                        val = _marker_layer_arg(elt)
+                        if val is not None:
+                            return val
+                # pytestmark = pytest.mark.layer("x")  (bare, no list)
+                val = _marker_layer_arg(node.value)
+                if val is not None:
+                    return val
+    return None
+
+
+def _marker_layer_arg(node: ast.expr) -> str | None:
+    """If node is pytest.mark.layer("x"), return "x"."""
+    if not isinstance(node, ast.Call):
+        return None
+    func = node.func
+    # pytest.mark.layer(...)
+    if (
+        isinstance(func, ast.Attribute)
+        and func.attr == "layer"
+        and isinstance(func.value, ast.Attribute)
+        and func.value.attr == "mark"
+    ):
+        if node.args and isinstance(node.args[0], ast.Constant):
+            return node.args[0].value
+    return None
+
+
+@pytest.mark.parametrize(
+    "directory",
+    sorted(LAYER_DIRECTORIES),
+    ids=sorted(LAYER_DIRECTORIES),
+)
+def test_all_test_files_have_correct_layer_marker(directory: str) -> None:
+    """Every test_*.py in an in-scope directory has pytestmark with correct layer."""
+    expected = LAYER_DIRECTORIES[directory]
+    dir_path = TESTS_ROOT / directory
+    test_files = sorted(dir_path.glob("test_*.py"))
+    assert test_files, f"No test files found in {dir_path}"
+
+    missing: list[str] = []
+    wrong: list[tuple[str, str]] = []
+
+    for tf in test_files:
+        layer = _extract_layer_marker(tf)
+        if layer is None:
+            missing.append(tf.name)
+        elif layer != expected:
+            wrong.append((tf.name, layer))
+
+    errors: list[str] = []
+    if missing:
+        errors.append(f"Missing layer marker: {missing}")
+    if wrong:
+        errors.append(f"Wrong layer marker: {wrong}")
+    assert not errors, "\n".join(errors)
+
+
+def test_layer_marker_registered_in_pyproject() -> None:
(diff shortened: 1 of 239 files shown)


tests/conftest.py
```diff
@@ -360,3 +374,26 @@ def pytest_collection_modifyitems(
             f"Test filter deselection failed, running all tests: {exc}",
             stacklevel=1,
         )
+
+    # Layer marker mismatch validation (controller-only under xdist)
```
Trecek: [critical] defense: The layer marker mismatch validation block (lines 378-399) is unreachable in normal test runs. `pytest_collection_modifyitems` returns early at line 331-332 when `scope is None` (i.e., `AUTOSKILLIT_TEST_FILTER` is not set), which is the default for `task test-all` and `task test-check`. The layer validation only executes when filtering is active (`task test-filtered`). Fix: move the xdist guard and layer validation loop before the `if scope is None: return` early exit so it always runs.

tests/execution/test_headless.py
```diff
@@ -3510,6 +3510,8 @@ def test_synthesis_uses_last_write_not_first(self, make_session):
 
 from autoskillit.pipeline.audit import DefaultAuditLog, FailureRecord  # noqa: E402
 
+pytestmark = [pytest.mark.layer("execution")]
```
Trecek: [warning] cohesion: `pytestmark` is placed at line 3513, deep inside the file after a `noqa: E402` late-import block. While this IS a module-level statement and pytest reads it from `module.__dict__` after full import (so all tests receive the marker), placing it this late violates the established convention of placing `pytestmark` immediately after imports at the top. It also defeats quick visual scanning and could mislead readers into thinking only the T3 tests below carry the marker. Move it to after the initial top-level import block near line 25.

tests/execution/test_session_adjudication.py
```diff
@@ -1672,6 +1672,8 @@ def test_compute_outcome_bold_wrapped_token_is_success_not_violation(
 
 import json  # noqa: E402 — imported here to keep T1 tests self-contained
 
+pytestmark = [pytest.mark.layer("execution")]
```
Trecek: [warning] cohesion: Same issue as `test_headless.py`: `pytestmark` is placed at line 1675 after a `noqa: E402` late import, rather than near the top of the file after the initial import block. This is functionally correct but inconsistent with every other modified file in this PR. Move it to after the initial import block near line 27.

tests/arch/test_layer_markers.py
```diff
@@ -0,0 +1,102 @@
...
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+import pytest
+
+TESTS_ROOT = Path(__file__).resolve().parent.parent
+
+LAYER_DIRECTORIES: dict[str, str] = {
```
Trecek, line 12: [warning] cohesion: `LAYER_DIRECTORIES` here and `_LAYER_DIRS` in `conftest.py` both enumerate the same 9 directories independently. Neither imports from the other. When a new layer is added, both must be updated manually — a silent divergence would cause the conftest runtime check to cover a different set than this hard enforcement test. Consider deriving one from the other or adding a parity assertion.

tests/arch/test_layer_markers.py
```diff
@@ -0,0 +1,102 @@
...
+                        if val is not None:
+                            return val
+                # pytestmark = pytest.mark.layer("x")  (bare, no list)
+                val = _marker_layer_arg(node.value)
+                if val is not None:
+                    return val
+    return None
+
+
+def _marker_layer_arg(node: ast.expr) -> str | None:
```
Trecek, line 46: [warning] defense: `_marker_layer_arg` returns `node.args[0].value` (type `Any` from `ast.Constant.value`) without an `isinstance(..., str)` guard. A non-string marker like `pytest.mark.layer(42)` would be silently returned and compared against the expected string, producing a confusing mismatch report. Add: `if not isinstance(node.args[0].value, str): return None`.

Trecek (commented):

5 actionable findings, 2 info findings.

tests/conftest.py
Trecek: Agreed — fixed in b92f8f29. Hoisted the xdist guard and layer validation loop before the `if scope is None: return` early exit so it runs unconditionally on every test collection.

tests/execution/test_headless.py
Trecek: Agreed — fixed in b92f8f29. Moved pytestmark to after the initial import block (line 26), consistent with all other test files.

tests/execution/test_session_adjudication.py
Trecek: Agreed — fixed in b92f8f29. Moved pytestmark to after the initial import block (line 28), consistent with convention.

tests/arch/test_layer_markers.py
Trecek: Agreed — fixed in b92f8f29. Added test_layer_directories_matches_conftest() that asserts LAYER_DIRECTORIES keys == _LAYER_DIRS from conftest.py. A new layer addition that misses either map will now fail this test.

tests/arch/test_layer_markers.py
Trecek: Investigated — this is intentional. _marker_layer_arg already guards with isinstance(node.args[0], ast.Constant) before returning node.args[0].value (line 58). For a non-string constant like pytest.mark.layer(42), the int value 42 is returned and compared against the expected string in test_all_test_files_have_correct_layer_marker — the comparison 42 != 'core' lands the file in the 'wrong marker' list with a clear diagnostic. Adding isinstance(value, str) would reclassify such files as 'missing marker' instead of 'wrong marker', losing the actual value found.

PR #1014, opened 2026-04-17
Rectify: Fail-Silent Manifest Handling and fnmatch/pathspec Mismatch in Test Filter

### LINKED ISSUE #997: Fix fail-silent manifest handling and fnmatch/pathspec mismatch in test filter
# Fix fail-silent manifest handling and fnmatch/pathspec mismatch in test filter

## Summary

Two correctness bugs in `tests/_test_filter.py::apply_manifest` cause the test path filter to silently skip tests for unmapped non-Python files instead of falling back to a full run.

## Bug 1: Fail-Silent on Unmatched Non-Python Files

The conftest-side `apply_manifest()` (lines 341-357) returns an empty `set()` when a non-Python file matches no manifest pattern. The call site in `build_test_scope()` (lines 487-489) treats this as "no tests needed" and silently continues:

```python
else:
    manifest_dirs = apply_manifest({f}, manifest)
    if manifest_dirs:
        test_dirs.update(manifest_dirs)
```

The production module `src/autoskillit/_test_filter.py::apply_manifest` correctly returns `None` on unmatched files, signaling fail-open (run the full suite). The conftest-side version diverges — it drops unmatched files silently, making manifest gaps dangerous rather than safe.

## Bug 2: fnmatch vs pathspec Glob Semantics

`apply_manifest()` uses `fnmatch.fnmatch()` for pattern matching, but the manifest was authored with gitwildmatch semantics (`**` = recursive). With fnmatch, `docs/**/*.md` does NOT match `docs/README.md`:

```python
>>> fnmatch.fnmatch('docs/README.md', 'docs/**/*.md')
False
>>> pathspec.PathSpec.from_lines('gitwildmatch', ['docs/**/*.md']).match_file('docs/README.md')
True
```

This causes 8 top-level `docs/*.md` files to be silently unmatched despite appearing covered by the manifest.

## Acceptance Criteria

- [ ] `apply_manifest` in `tests/_test_filter.py` returns a sentinel (e.g., `None`) when any non-Python file matches no manifest pattern
- [ ] `build_test_scope` at lines 487-489 checks for the sentinel and returns `None` (full run) when encountered
- [ ] Pattern matching uses `pathspec.PathSpec.from_lines("gitwildmatch", ...)` instead of `fnmatch.fnmatch`, aligning with `src/autoskillit/_test_filter.py`
- [ ] New test: `docs/README.md` matches against `docs/**/*.md` pattern
- [ ] New test: unmatched non-Python file triggers full run (returns `None`)
- [ ] Existing tests updated for new return type
- [ ] Shadow-diff tests (SD1-SD3) still pass

## Files to Modify

- `tests/_test_filter.py` — `apply_manifest()` function + `build_test_scope()` call site
- `tests/test_test_filter.py` — new/updated tests for fail-open and pathspec behavior

## Context

Discovered during milestone #4 evaluation. The production module (`src/autoskillit/_test_filter.py`) already handles this correctly with pathspec and fail-open returns. The conftest-side version needs to be aligned.

---

**Batch 0 (P0)** — Ship first. No dependencies.

**Blocks**: #998, #999 (manifest entries and completeness test depend on the pathspec fix)

---

## Sequencing

**Batch 0 (P0)** — Ship first. No dependencies.

**Blocks**: #998, #999

Trecek:
## Summary

The conftest-side test filter module (`tests/_test_filter.py`) independently re-implemented `apply_manifest` with two semantic divergences from the production module (`src/autoskillit/_test_filter.py`): it uses `fnmatch` instead of `pathspec` gitwildmatch (silently failing to match `**` patterns like `docs/**/*.md`), and it returns `set()` instead of `None` for the fail-open sentinel (silently narrowing the test scope when a non-Python file matches no manifest entry, instead of triggering a full run).

The architectural weakness is not the individual bugs but the structural absence of any contract between the two parallel implementations. There is no shared interface, no behavioral equivalence gate, and no type-checking coverage of the conftest module — so the divergence was invisible until runtime symptoms appeared.

The immunity plan replaces the shared semantic contract with an enforced one: `pathspec` (already a production dependency and importable in the conftest module) replaces `fnmatch` in the conftest module, eliminating the pattern-matching divergence at the source. Behavioral equivalence tests cross-validate both implementations on the same inputs. An AST-based contract test in `tests/arch/` enforces signature compatibility. Mypy coverage is extended to `tests/_test_filter.py`.

### Error/Resilience Diagram — Test Filter Pipeline

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    FULL_RUN([FULL SUITE RUN])
    FILTERED_RUN([FILTERED RUN])

    subgraph GitLayer ["GIT DIFF — Error Boundary"]
        GIT_CALL["● git_changed_files()<br/>━━━━━━━━━━<br/>git diff --name-only<br/>base_ref...HEAD"]
        GIT_ERR{"subprocess error?<br/>timeout? no git?"}
        GIT_WARN["warnings.warn()<br/>━━━━━━━━━━<br/>non-fatal, logged"]
    end

    subgraph ValidationGates ["VALIDATION GATES — Fail-Fast"]
        GATE_NONE{"FilterMode<br/>== NONE?"}
        GATE_NULL{"changed_files<br/>is None?"}
        GATE_SIZE{"len > 30<br/>files?"}
        GATE_BUCKET{"● check_bucket_a()<br/>━━━━━━━━━━<br/>conftest / pyproject /<br/>.pre-commit-config.yaml<br/>/ uv.lock / _factory.py"}
    end

    subgraph Classification ["FILE CLASSIFICATION"]
        CLASS_SRC["src/*.py<br/>━━━━━━━━━━<br/>Layer cascade map<br/>→ test directories"]
        CLASS_TEST["tests/*.py<br/>━━━━━━━━━━<br/>Direct inclusion"]
        CLASS_OTHER["non-.py files<br/>━━━━━━━━━━<br/>→ manifest lookup"]
        CLASS_UNKNOWN["unknown .py<br/>outside src/ tests/<br/>━━━━━━━━━━<br/>→ fail-open"]
    end

    subgraph ManifestLayer ["MANIFEST LAYER — Pathspec Guard"]
        LOAD_MAN["● load_manifest()<br/>━━━━━━━━━━<br/>reads .autoskillit/<br/>test-filter-manifest.yaml"]
        MAN_ERR{"absent /<br/>OSError /<br/>YAMLError?"}
        MAN_WARN["warnings.warn()<br/>━━━━━━━━━━<br/>returns None"]
        APPLY_MAN["● apply_manifest()<br/>━━━━━━━━━━<br/>pathspec gitwildmatch<br/>(fixed from fnmatch)"]
        MATCH_CHECK{"any file<br/>unmatched?"}
    end

    subgraph ContractLayer ["★ CONTRACT ENFORCEMENT (NEW)"]
        CONTRACT["★ TestApplyManifestSignatureContract<br/>━━━━━━━━━━<br/>tests/arch/test_filter_contracts.py<br/>AST inspection of both impls"]
        C_ANN["test_both_return_optional_set<br/>━━━━━━━━━━<br/>enforces None in return annotation<br/>(fail-open guarantee)"]
        C_PARAM["test_both_accept_manifest_parameter<br/>━━━━━━━━━━<br/>enforces manifest param present"]
    end

    subgraph ConftestPlugin ["CONFTEST PLUGIN — Fail-Open Guards"]
        P_CFG{"pytest_configure<br/>raises?"}
        P_CFG_WARN["warnings.warn()<br/>━━━━━━━━━━<br/>scope = None"]
        P_MOD{"pytest_collection_<br/>modifyitems raises?"}
        P_MOD_WARN["warnings.warn()<br/>━━━━━━━━━━<br/>deselection skipped"]
    end

    %% ── Git error path ──
    GIT_CALL --> GIT_ERR
    GIT_ERR -->|"CalledProcessError<br/>TimeoutExpired<br/>FileNotFoundError"| GIT_WARN
    GIT_WARN -->|"None propagates"| GATE_NULL
    GIT_ERR -->|"success"| GATE_NONE

    %% ── Validation gates ──
    GATE_NONE -->|"NONE"| FULL_RUN
    GATE_NONE -->|"conservative / aggressive"| GATE_NULL
    GATE_NULL -->|"None"| FULL_RUN
    GATE_NULL -->|"set[str]"| GATE_SIZE
    GATE_SIZE -->|"> 30"| FULL_RUN
    GATE_SIZE -->|"≤ 30"| GATE_BUCKET
    GATE_BUCKET -->|"Bucket A hit"| FULL_RUN
    GATE_BUCKET -->|"no match"| Classification

    %% ── Classification ──
    Classification --> CLASS_SRC
    Classification --> CLASS_TEST
    Classification --> CLASS_OTHER
    Classification --> CLASS_UNKNOWN
    CLASS_UNKNOWN -->|"fail-open"| FULL_RUN
    CLASS_SRC -->|"cascade"| FILTERED_RUN
    CLASS_TEST -->|"direct"| FILTERED_RUN
    CLASS_OTHER --> APPLY_MAN

    %% ── Manifest layer ──
    LOAD_MAN --> MAN_ERR
    MAN_ERR -->|"absent / OSError / YAMLError"| MAN_WARN
    MAN_WARN -->|"None"| APPLY_MAN
    MAN_ERR -->|"ok"| APPLY_MAN
    APPLY_MAN --> MATCH_CHECK
    MATCH_CHECK -->|"unmatched file"| FULL_RUN
    MATCH_CHECK -->|"all matched"| FILTERED_RUN

    %% ── Contract enforcement (build-time guard) ──
    CONTRACT --> C_ANN
    CONTRACT --> C_PARAM
    C_ANN -->|"assert None in annotation<br/>fails → CI breaks"| APPLY_MAN
    C_PARAM -->|"assert manifest param exists<br/>fails → CI breaks"| APPLY_MAN

    %% ── Conftest plugin fail-open ──
    P_CFG -->|"exception"| P_CFG_WARN
    P_CFG_WARN -->|"scope=None → all tests run"| FULL_RUN
    P_CFG -->|"ok"| P_MOD
    P_MOD -->|"exception"| P_MOD_WARN
    P_MOD_WARN -->|"deselection skipped → all tests run"| FULL_RUN
    P_MOD -->|"ok"| FILTERED_RUN

    %% CLASS ASSIGNMENTS %%
    class FULL_RUN,FILTERED_RUN terminal;
    class GIT_CALL,APPLY_MAN,LOAD_MAN handler;
    class GIT_ERR,MAN_ERR,MATCH_CHECK,GATE_NONE,GATE_NULL,GATE_SIZE,GATE_BUCKET,P_CFG,P_MOD stateNode;
    class GIT_WARN,MAN_WARN,P_CFG_WARN,P_MOD_WARN gap;
    class GATE_BUCKET detector;
    class CLASS_SRC,CLASS_TEST,CLASS_OTHER,CLASS_UNKNOWN phase;
    class CONTRACT,C_ANN,C_PARAM newComponent;
    class Classification output;
```

### Module Dependency Diagram — autoskillit PR #997 Scope

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 50, 'rankSpacing': 70, 'curve': 'basis'}}}%%
graph TB
    %% CLASS DEFINITIONS %%

    subgraph SrcLayer ["LAYER 0 — Foundation (unchanged)"]
        CoreIO["autoskillit.core<br/>━━━━━━━━━━<br/>load_yaml · get_logger<br/>Fan-in: 10 (highest)"]
    end

    subgraph AuxLayer ["AUX — Root Utility Module"]
        TestFilter["● autoskillit._test_filter<br/>━━━━━━━━━━<br/>Glob → test-dir YAML manifest<br/>Imports: core.load_yaml only<br/>Fan-in: 0 (leaf consumer)"]
    end

    subgraph TestInfra ["TEST INFRASTRUCTURE"]
        TestHelper["● tests/_test_filter.py<br/>━━━━━━━━━━<br/>apply_manifest(manifest, paths)<br/>Fixed: pathspec gitwildmatch<br/>replaces fnmatch"]
        TestSuite["● tests/test_test_filter.py<br/>━━━━━━━━━━<br/>Sentinel handling tests<br/>fnmatch/pathspec divergence<br/>coverage added"]
    end

    subgraph ArchContracts ["ARCH CONTRACTS (new)"]
        ContractTest["★ tests/arch/test_filter_contracts.py<br/>━━━━━━━━━━<br/>AST-based contract test<br/>Enforces apply_manifest<br/>signature compatibility"]
    end

    subgraph QualityGates ["QUALITY GATES"]
        PreCommit["● .pre-commit-config.yaml<br/>━━━━━━━━━━<br/>mypy hook: extended to cover<br/>tests/_test_filter.py<br/>Type safety enforcement"]
    end

    subgraph ExtLayer ["EXTERNAL DEPENDENCIES"]
        Pathspec["pathspec (gitwildmatch)<br/>━━━━━━━━━━<br/>Replaces stdlib fnmatch<br/>Git-compatible pattern matching"]
        ASTLib["ast (stdlib)<br/>━━━━━━━━━━<br/>Source introspection<br/>Signature contract checking"]
    end

    %% SOURCE → CORE %%
    TestFilter -->|"imports load_yaml"| CoreIO

    %% TEST HELPER → SOURCE + EXTERNAL %%
    TestHelper -->|"applies manifest data from"| TestFilter
    TestHelper -->|"uses for glob matching"| Pathspec

    %% TEST SUITE → HELPERS %%
    TestSuite -->|"tests apply_manifest"| TestHelper
    TestSuite -->|"tests YAML loading via"| TestFilter

    %% CONTRACT TEST %%
    ContractTest -->|"★ AST-inspects signature of"| TestHelper
    ContractTest -->|"parses source with"| ASTLib

    %% QUALITY GATE %%
    PreCommit -->|"mypy type-checks"| TestHelper

    %% CLASS ASSIGNMENTS %%
    class CoreIO handler;
    class TestFilter stateNode;
    class TestHelper phase;
    class TestSuite output;
    class ContractTest newComponent;
    class PreCommit detector;
    class Pathspec,ASTLib integration;
```

Closes #997

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260416-163632-205313/.autoskillit/temp/rectify/rectify_fail-silent-manifest-fnmatch-pathspec_2026-04-16_164500.md`

## Token Usage Summary

| investigate | 140 | 10.3k | 615.0k | 58.2k | 1 | 3m 15s |
| rectify | 97 | 13.8k | 410.7k | 48.0k | 1 | 6m 41s |
| review | 3.0k | 6.4k | 155.3k | 29.1k | 1 | 2m 47s |
| dry_walkthrough | 100 | 13.5k | 516.5k | 53.3k | 1 | 4m 11s |
| implement | 456 | 20.0k | 3.7M | 83.2k | 1 | 6m 14s |
| prepare_pr | 60 | 5.5k | 195.1k | 27.2k | 1 | 1m 30s |
| run_arch_lenses | 3.7k | 18.9k | 462.7k | 130.8k | 2 | 6m 19s |
| compose_pr | 75 | 6.8k | 242.4k | 26.3k | 1 | 2m 0s |
| **Total** | 7.6k | 95.2k | 6.3M | 456.1k | | 33m 1s |

### DIFF
--- a/tests/_test_filter.py
+++ b/tests/_test_filter.py
@@ -10,6 +10,8 @@ import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING
 
+import pathspec
+
 if TYPE_CHECKING:
     from typing import Any
 
@@ -341,20 +343,29 @@ def load_manifest(path: str | Path) -> dict[str, Any] | None:
 def apply_manifest(
     changed_files: set[str],
     manifest: dict[str, Any] | None,
-) -> set[str]:
-    """Return test directories matched by manifest patterns for the changed files."""
+) -> set[str] | None:
+    """Return test directories matched by manifest patterns for the changed files.
+
+    Returns None when manifest is None (fail-open) or when any changed file matches
+    no manifest pattern (fail-open: caller should run the full suite).
+    """
     if manifest is None:
-        return set()
-    result: set[str] = set()
-    for pattern, test_dirs in manifest.items():
-        for f in changed_files:
-            if fnmatch.fnmatch(f, pattern):
+        return None
+    compiled = {pat: pathspec.PathSpec.from_lines("gitwildmatch", [pat]) for pat in manifest}
+    matched_dirs: set[str] = set()
+    for f in changed_files:
+        file_matched = False
+        for pattern, spec in compiled.items():
+            if spec.match_file(f):
+                test_dirs = manifest[pattern]
                 if isinstance(test_dirs, list):
-                    result.update(test_dirs)
+                    matched_dirs.update(test_dirs)
                 elif isinstance(test_dirs, str):
-                    result.add(test_dirs)
-                break
-    return result
+                    matched_dirs.add(test_dirs)
+                file_matched = True
+        if not file_matched:
+            return None
+    return matched_dirs
 
 
 # ---------------------------------------------------------------------------
@@ -485,8 +496,9 @@ def build_test_scope(
             return None
         else:
             manifest_dirs = apply_manifest({f}, manifest)
-            if manifest_dirs:
-                test_dirs.update(manifest_dirs)
+            if manifest_dirs is None:
+                return None
+            test_dirs.update(manifest_dirs)
 
     test_dirs.update(always_run)
 
--- a/tests/arch/test_filter_contracts.py
+++ b/tests/arch/test_filter_contracts.py
@@ -1,0 +1,48 @@
+"""AST-based contract test enforcing signature compatibility between the two apply_manifest
+implementations: tests/_test_filter.py and src/autoskillit/_test_filter.py.
+"""
+
+from __future__ import annotations
+
+import ast
+import inspect
+from pathlib import Path
+
+import autoskillit._test_filter as src_filter
+import tests._test_filter as conftest_filter
+
+
+def _get_return_annotation(module: object, func_name: str) -> str:
+    """Return the string representation of func_name's return annotation in module."""
+    source = Path(inspect.getfile(module)).read_text()  # type: ignore[arg-type]
+    tree = ast.parse(source)
+    for node in ast.walk(tree):
+        if isinstance(node, ast.FunctionDef) and node.name == func_name:
+            if node.returns is not None:
+                return ast.unparse(node.returns)
+    return ""
+
+
+class TestApplyManifestSignatureContract:
+    """Enforces structural compatibility between the two apply_manifest implementations.
+
+    When the production module's apply_manifest signature changes, this test fails
+    immediately, forcing the conftest-side implementation to be updated in sync.
+    """
+
+    def test_both_return_optional_set(self) -> None:
+        """Both apply_manifest implementations must return set[str] | None."""
+        src_ann = _get_return_annotation(src_filter, "apply_manifest")
+        conftest_ann = _get_return_annotation(conftest_filter, "apply_manifest")
+        assert "None" in src_ann, f"src apply_manifest missing None return: {src_ann!r}"
+        assert "None" in conftest_ann, (
+            f"conftest apply_manifest missing None return: {conftest_ann!r}\n"
+            "The conftest module must return None (not empty set) to signal fail-open."
+        )
+
+    def test_both_accept_manifest_parameter(self) -> None:
+        """Both apply_manifest implementations must accept a manifest parameter."""
+        src_sig = inspect.signature(src_filter.apply_manifest)
+        conftest_sig = inspect.signature(conftest_filter.apply_manifest)
+        assert "manifest" in src_sig.parameters
+        assert "manifest" in conftest_sig.parameters
--- a/tests/arch/test_filter_contracts.py
+++ b/tests/arch/test_filter_contracts.py
@@ -34,9 +34,11 @@ class TestApplyManifestSignatureContract:
         """Both apply_manifest implementations must return set[str] | None."""
         src_ann = _get_return_annotation(src_filter, "apply_manifest")
         conftest_ann = _get_return_annotation(conftest_filter, "apply_manifest")
-        assert "None" in src_ann, f"src apply_manifest missing None return: {src_ann!r}"
-        assert "None" in conftest_ann, (
-            f"conftest apply_manifest missing None return: {conftest_ann!r}\n"
+        assert src_ann in ("set[str] | None", "Optional[set[str]]"), (
+            f"src apply_manifest unexpected return annotation: {src_ann!r}"
+        )
+        assert conftest_ann in ("set[str] | None", "Optional[set[str]]"), (
+            f"conftest apply_manifest unexpected return annotation: {conftest_ann!r}\n"
             "The conftest module must return None (not empty set) to signal fail-open."
         )
 
--- a/tests/test_test_filter.py
+++ b/tests/test_test_filter.py
@@ -293,6 +293,7 @@ class TestBuildTestScope:
         assert Path("tests/core/test_io.py") in result
 
     def test_scope_nonpython_no_manifest_only_alwaysrun(self, tmp_path: Path) -> None:
+        """Non-Python file with manifest=None → fail-open → result is None."""
         tests_root = tmp_path / "tests"
         for d in ["arch", "contracts", "infra", "docs"]:
             (tests_root / d).mkdir(parents=True, exist_ok=True)
@@ -302,9 +303,37 @@ class TestBuildTestScope:
             mode=FilterMode.CONSERVATIVE,
             tests_root=tests_root,
         )
+        assert result is None
+
+    def test_scope_nonpython_unmatched_manifest_returns_none(self, tmp_path: Path) -> None:
+        """A non-Python changed file with no manifest match must return None (full run)."""
+        tests_root = tmp_path / "tests"
+        for d in ["arch", "contracts", "infra", "docs"]:
+            (tests_root / d).mkdir(parents=True, exist_ok=True)
+        manifest = {"docs/**/*.md": ["docs"]}
+        result = build_test_scope(
+            changed_files={"some/unknown/file.txt"},
+            mode=FilterMode.CONSERVATIVE,
+            manifest=manifest,
+            tests_root=tests_root,
+        )
+        assert result is None
+
+    def test_scope_nonpython_matched_manifest_adds_dirs(self, tmp_path: Path) -> None:
+        """A non-Python file that DOES match a manifest pattern contributes its test dirs."""
+        tests_root = tmp_path / "tests"
+        for d in ["arch", "contracts", "infra", "docs"]:
+            (tests_root / d).mkdir(parents=True, exist_ok=True)
+        manifest = {"docs/**/*.md": ["docs"]}
+        result = build_test_scope(
+            changed_files={"docs/README.md"},
+            mode=FilterMode.CONSERVATIVE,
+            manifest=manifest,
+            tests_root=tests_root,
+        )
         assert result is not None
         dir_names = {p.name for p in result}
-        assert dir_names == {"arch", "contracts", "infra", "docs"}
+        assert "docs" in dir_names
 
     def test_scope_none_mode_returns_none(self, tmp_path: Path) -> None:
         result = build_test_scope(
@@ -511,7 +540,7 @@ class TestLoadManifest:
 class TestApplyManifest:
     def test_apply_manifest_none(self) -> None:
         result = apply_manifest({"README.md"}, None)
-        assert result == set()
+        assert result is None
 
     def test_apply_manifest_match(self) -> None:
         manifest = {"docs/*.md": ["docs"]}
@@ -519,14 +548,88 @@ class TestApplyManifest:
         assert result == {"docs"}
 
     def test_apply_manifest_no_match(self) -> None:
-        manifest = {"docs/*.md": ["docs"]}
+        manifest = {"tests/*.py": ["unit"]}
         result = apply_manifest({"src/foo.py"}, manifest)
-        assert result == set()
+        assert result is None
 
     def test_apply_manifest_list_dirs(self) -> None:
         manifest = {"*.yaml": ["config", "infra"]}
         result = apply_manifest({"defaults.yaml"}, manifest)
         assert result == {"config", "infra"}
+
+    def test_apply_manifest_doublestar_zero_segments(self) -> None:
+        """docs/**/*.md must match docs/README.md (zero intermediate path segments).
+
+        This pattern is in the production manifest. fnmatch fails this; pathspec passes it.
+        """
+        manifest = {"docs/**/*.md": ["docs"]}
+        result = apply_manifest({"docs/README.md"}, manifest)
+        assert result == {"docs"}
+
+    def test_apply_manifest_doublestar_nested(self) -> None:
+        """docs/**/*.md must match docs/developer/SETUP.md (non-zero intermediate segments)."""
+        manifest = {"docs/**/*.md": ["docs"]}
+        result = apply_manifest({"docs/developer/SETUP.md"}, manifest)
+        assert result == {"docs"}
+
+
+# ---------------------------------------------------------------------------
+# Behavioral equivalence cross-validation (EQ1–EQ2)
+# ---------------------------------------------------------------------------
+
+
+class TestApplyManifestEquivalence:
+    """Cross-validates conftest apply_manifest against production manifest_apply_manifest.
+
+    Both modules must agree on matched outputs. When a file matches patterns in one,
+    it must match the same patterns in the other. The two implementations are allowed
+    to differ only in how they handle the unmatched case (both must return None there too).
+    """
+
+    MANIFEST = {
+        "src/autoskillit/recipes/*.yaml": ["recipe", "contracts"],
+        "docs/**/*.md": ["docs"],
+        "Taskfile.yml": ["infra"],
+        "tests/recipe/fixtures/*": ["recipe"],
+        ".pre-commit-config.yaml": ["infra"],
+    }
+
+    @pytest.mark.parametrize(
+        "file_path,expected_dirs",
+        [
+            ("src/autoskillit/recipes/my.yaml", {"recipe", "contracts"}),
+            ("docs/README.md", {"docs"}),
+            ("docs/sub/dir/deep.md", {"docs"}),
+            ("Taskfile.yml", {"infra"}),
+            ("tests/recipe/fixtures/test.yaml", {"recipe"}),
+            (".pre-commit-config.yaml", {"infra"}),
+        ],
+    )
+    def test_matching_cases_agree(self, file_path: str, expected_dirs: set[str]) -> None:
+        """For files that match manifest patterns, both implementations return the same dirs."""
+        conftest_result = apply_manifest({file_path}, self.MANIFEST)
+        production_result = manifest_apply_manifest([file_path], self.MANIFEST)
+        assert conftest_result == production_result == expected_dirs, (
+            f"Implementations disagree for {file_path!r}: "
+            f"conftest={conftest_result!r}, production={production_result!r}"
+        )
+
+    @pytest.mark.parametrize(
+        "file_path",
+        [
+            "some/unknown/file.txt",
+            "README.md",
+            "pyproject.toml",
+        ],
+    )
+    def test_unmatched_cases_both_return_none(self, file_path: str) -> None:
+        """For files matching no manifest pattern, both implementations return None."""
+        conftest_result = apply_manifest({file_path}, self.MANIFEST)
+        production_result = manifest_apply_manifest([file_path], self.MANIFEST)
+        assert conftest_result is None, f"conftest returned {conftest_result!r} for {file_path!r}"
+        assert production_result is None, (
+            f"production returned {production_result!r} for {file_path!r}"
+        )
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_test_filter.py
+++ b/tests/test_test_filter.py
@@ -333,7 +333,7 @@ class TestBuildTestScope:
         )
         assert result is not None
         dir_names = {p.name for p in result}
-        assert "docs" in dir_names
+        assert dir_names == {"arch", "contracts", "infra", "docs"}
 
     def test_scope_none_mode_returns_none(self, tmp_path: Path) -> None:
         result = build_test_scope(
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
 
       - id: mypy
         name: mypy
-        entry: mypy src/ --ignore-missing-imports
+        entry: mypy src/ tests/_test_filter.py --ignore-missing-imports
         language: system
         pass_filenames: false
 


tests/test_test_filter.py
```diff
@@ -302,9 +303,37 @@ def test_scope_nonpython_no_manifest_only_alwaysrun(self, tmp_path: Path) -> Non
...
+            (tests_root / d).mkdir(parents=True, exist_ok=True)
+        manifest = {"docs/**/*.md": ["docs"]}
+        result = build_test_scope(
+            changed_files={"docs/README.md"},
+            mode=FilterMode.CONSERVATIVE,
+            manifest=manifest,
+            tests_root=tests_root,
+        )
         assert result is not None
         dir_names = {p.name for p in result}
```
Trecek, line 335: [warning] tests: Assertion weakened from `assert dir_names == {"arch", "contracts", "infra", "docs"}` to `assert "docs" in dir_names`. The old assertion verified the complete set; the new one only checks membership. If the intent is to verify only manifest-matched dirs (not always_run), use `assert dir_names == {"docs"}` or add a comment explaining why loose membership is intentional.

tests/test_test_filter.py
```diff
@@ -511,23 +540,97 @@ def test_load_manifest_malformed_yaml(self, tmp_path: Path) -> None:
...
         result = apply_manifest({"defaults.yaml"}, manifest)
         assert result == {"config", "infra"}
 
+    def test_apply_manifest_doublestar_zero_segments(self) -> None:
+        """docs/**/*.md must match docs/README.md (zero intermediate path segments).
+
+        This pattern is in the production manifest. fnmatch fails this; pathspec passes it.
+        """
+        manifest = {"docs/**/*.md": ["docs"]}
+        result = apply_manifest({"docs/README.md"}, manifest)
```
Trecek, line 566: [warning] slop: Duplicate assignment — `manifest = {"docs/**/*.md": ["docs"]}` assigned twice (L565–L566). The second line is dead code. Remove it.

tests/test_test_filter.py
```diff
@@ -511,23 +540,97 @@ def test_load_manifest_malformed_yaml(self, tmp_path: Path) -> None:
...
+        assert result == {"docs"}
+
```
Trecek, line 568: [warning] slop: Duplicate call — `result = apply_manifest({"docs/README.md"}, manifest)` called twice (L567–L568). First result is immediately overwritten. Remove the duplicate line.

tests/_test_filter.py
```diff
@@ -341,20 +343,29 @@ def load_manifest(path: str | Path) -> dict[str, Any] | None:
...
+                test_dirs = manifest[pattern]
                 if isinstance(test_dirs, list):
-                    result.update(test_dirs)
+                    matched_dirs.update(test_dirs)
                 elif isinstance(test_dirs, str):
-                    result.add(test_dirs)
-                break
-    return result
+                    matched_dirs.add(test_dirs)
+                file_matched = True
```
Trecek, line 365: [warning] bugs: `file_matched = True` is inside the isinstance branches, so it is only set when `test_dirs` is list or str. If a pattern matches but `test_dirs` is any other type, `file_matched` stays False and the function returns None incorrectly. Move `file_matched = True` immediately after `spec.match_file(f)` returns True, outside the isinstance check.

tests/arch/test_filter_contracts.py
```diff
@@ -0,0 +1,48 @@
...
+
+    When the production module's apply_manifest signature changes, this test fails
+    immediately, forcing the conftest-side implementation to be updated in sync.
+    """
+
+    def test_both_return_optional_set(self) -> None:
+        """Both apply_manifest implementations must return set[str] | None."""
+        src_ann = _get_return_annotation(src_filter, "apply_manifest")
+        conftest_ann = _get_return_annotation(conftest_filter, "apply_manifest")
+        assert "None" in src_ann, f"src apply_manifest missing None return: {src_ann!r}"
```
Trecek: [warning] tests: `assert "None" in src_ann` is a fragile substring match. Any annotation string containing the word None passes. Prefer `assert src_ann in ("set[str] | None", "Optional[set[str]]")` for an exact check.

tests/arch/test_filter_contracts.py
```diff
@@ -0,0 +1,48 @@
...
+import inspect
+from pathlib import Path
+
+import autoskillit._test_filter as src_filter
+import tests._test_filter as conftest_filter
+
+
+def _get_return_annotation(module: object, func_name: str) -> str:
+    """Return the string representation of func_name's return annotation in module."""
+    source = Path(inspect.getfile(module)).read_text()  # type: ignore[arg-type]
```
Trecek, line 17: [info] defense: `inspect.getfile(module)` can raise TypeError for built-in/frozen modules. The # type: ignore suppresses the type error but not the runtime exception. Wrap with try/except TypeError and call pytest.fail() with a descriptive message.

tests/_test_filter.py
```diff
@@ -341,20 +343,29 @@ def load_manifest(path: str | Path) -> dict[str, Any] | None:
...
+    no manifest pattern (fail-open: caller should run the full suite).
+    """
     if manifest is None:
-        return set()
-    result: set[str] = set()
-    for pattern, test_dirs in manifest.items():
-        for f in changed_files:
-            if fnmatch.fnmatch(f, pattern):
+        return None
+    compiled = {pat: pathspec.PathSpec.from_lines("gitwildmatch", [pat]) for pat in manifest}
```
Trecek, line 354: [info] defense: manifest is typed as dict[str, Any], so values at L360 are Any. The isinstance guards are the only runtime defence. Narrowing the type to dict[str, list[str] | str] at the public boundary would make misuse detectable at type-check time.

tests/test_test_filter.py
```diff
@@ -511,23 +540,97 @@ def test_load_manifest_malformed_yaml(self, tmp_path: Path) -> None:
...
+            ("tests/recipe/fixtures/test.yaml", {"recipe"}),
+            (".pre-commit-config.yaml", {"infra"}),
+        ],
+    )
+    def test_matching_cases_agree(self, file_path: str, expected_dirs: set[str]) -> None:
+        """For files that match manifest patterns, both implementations return the same dirs."""
+        conftest_result = apply_manifest({file_path}, self.MANIFEST)
+        production_result = manifest_apply_manifest([file_path], self.MANIFEST)
+        assert conftest_result == production_result == expected_dirs, (
+            f"Implementations disagree for {file_path!r}: "
```
Trecek, line 613: [warning] cohesion: The production function is called manifest_apply_manifest here but is named apply_manifest in the arch contract test. Clarify whether this is an import alias to avoid name collision, or an actual naming difference in src/autoskillit/_test_filter.py.

Trecek (commented):

Trecek (commented):
AutoSkillit review found 4 blocking issues (duplicated test lines, file_matched placement bug, fragile None assertion). See inline comments.

tests/test_test_filter.py
Trecek: Agreed — fixed in 9cf8a897. Original assertion `assert dir_names == {"arch", "contracts", "infra", "docs"}` was weakened in c9a7e3f1 without explanation; restored to verify the complete set.

tests/test_test_filter.py
Trecek: Investigated — this is intentional. The two `manifest = {"docs/**/*.md": ["docs"]}` assignments are in separate test methods (`test_apply_manifest_doublestar_zero_segments` at L560 and `test_apply_manifest_doublestar_nested` at L569). Each test method defines its own local variable independently — standard Python test isolation, not a duplicate within a single function.

tests/test_test_filter.py
Trecek: Investigated — this is intentional. Same reasoning as the previous comment: the two `result = apply_manifest(...)` calls belong to separate test methods (`test_apply_manifest_doublestar_zero_segments` and `test_apply_manifest_doublestar_nested`). No overwrite occurs within a single test.

tests/_test_filter.py
Trecek: Investigated — this is intentional. `file_matched = True` at line 365 is at the same indentation level as `test_dirs = manifest[pattern]` (line 360) and the `if isinstance(test_dirs, list):` block (line 361) — it is inside `if spec.match_file(f):` but outside the isinstance branches. It executes whenever any pattern matches the file, regardless of `test_dirs` type. No placement bug exists.

tests/arch/test_filter_contracts.py
Trecek: Agreed — fixed in 38ce5694. Tightened to `assert src_ann in ("set[str] | None", "Optional[set[str]]")` so the contract test validates the exact expected annotation rather than any string containing the word "None".

tests/test_test_filter.py
Trecek: Investigated — this is intentional. Line 11 of `test_test_filter.py`: `from autoskillit._test_filter import apply_manifest as manifest_apply_manifest`. The alias disambiguates between `tests._test_filter.apply_manifest` (conftest version, imported as `apply_manifest` on line 22) and `autoskillit._test_filter.apply_manifest` (production version). Both functions are named `apply_manifest` in their respective modules; the import alias is the standard resolution pattern.

tests/_test_filter.py
```diff
@@ -341,20 +343,29 @@ def load_manifest(path: str | Path) -> dict[str, Any] | None:
...
+    matched_dirs: set[str] = set()
```
Trecek, line 355: [warning] bugs: When `changed_files` is empty, the for-loop at L356 is never entered so `file_matched` is never evaluated and the function returns `matched_dirs` (an empty set) instead of `None`. The fail-open guard at L366-367 fires only per-file inside the loop, so an empty input bypasses it entirely. Callers that pass an empty set will receive `set()` rather than `None`, which may suppress full-suite fallback unexpectedly. Add a guard: `if not changed_files: return None`.

tests/_test_filter.py
Trecek, line 365: [warning] tests: `file_matched = True` fires whenever any pattern matches `f`, regardless of whether `test_dirs` is list or str. If `manifest[pattern]` is neither (e.g. `None` or an integer due to a malformed manifest), no dirs are added to `matched_dirs` but `file_matched` becomes `True`, so the function returns an empty set instead of `None` — silently suppressing a misconfiguration rather than failing open. Consider only setting `file_matched = True` inside the `isinstance` branches, or add an else-branch that logs a warning and leaves `file_matched = False`.

tests/arch/test_filter_contracts.py
```diff
@@ -0,0 +1,50 @@
...
+implementations: tests/_test_filter.py and src/autoskillit/_test_filter.py.
+"""
+
+from __future__ import annotations
+
+import ast
+import inspect
+from pathlib import Path
+
+import autoskillit._test_filter as src_filter
```
Trecek, line 11: [warning] defense: `inspect.getfile(module)` raises `TypeError` for built-in or frozen modules. The `# type: ignore[arg-type]` suppresses only mypy, not the runtime exception. If either `src_filter` or `conftest_filter` is ever imported from a `.pyc`-only or frozen distribution, this will raise an unhandled `TypeError` and produce a confusing test error. Wrap with `try/except TypeError` and call `pytest.fail()` with a descriptive message.

Trecek (commented):

Trecek (commented):
AutoSkillit review found 3 blocking issues. See inline comments.

Findings:
- `tests/_test_filter.py:355` [warning/bugs] — empty `changed_files` returns `set()` instead of `None`, bypassing fail-open guard
- `tests/_test_filter.py:365` [warning/tests] — `file_matched = True` fires for unhandled manifest value types (neither list nor str), silently suppressing misconfiguration
- `tests/arch/test_filter_contracts.py:11` [warning/defense] — `inspect.getfile()` can raise `TypeError` at runtime; unhandled

Info (non-blocking):
- `tests/_test_filter.py:354` — `pathspec.PathSpec.from_lines` raises `ValueError` for malformed patterns; no try/except
- `tests/test_test_filter.py:560` — `test_apply_manifest_doublestar_*` tests could be collapsed into one parametrized test

tests/_test_filter.py
Trecek: Investigated — this is intentional. The function's only production call site is `build_test_scope` at tests/_test_filter.py:498, which always passes a singleton set `{f}` — one file at a time. The empty-set path is unreachable in practice. Even if reached, `test_dirs.update(set())` is a no-op and `always_run` dirs are still appended — no full-suite suppression occurs.

tests/_test_filter.py
Trecek: Valid observation — flagged for design decision. Technically correct that `file_matched = True` fires regardless of isinstance branch match, so a malformed manifest value (None or int) would return `set()` instead of `None`. However: (a) `load_manifest` already catches YAML errors at load time and returns `None`, short-circuiting at L352-353 before reaching this line; (b) this is internal test-infrastructure, not production code. Moving `file_matched = True` inside the isinstance branches would improve defensive correctness. Flagged for human review.

tests/arch/test_filter_contracts.py
Trecek: Investigated — this is intentional. Both `autoskillit._test_filter` and `tests._test_filter` are plain Python source files loaded via `SourceFileLoader` under an editable hatchling install. The project uses `PYTHONDONTWRITEBYTECODE=1` and has no frozen-distribution pipeline (no PyInstaller/cx_Freeze/Nuitka/mypyc). The risk described applies only to built-in or frozen modules, which these are not. Note: the comment references line 11 (a bare import statement) — the actual `inspect.getfile()` call is on line 17; the `# type: ignore[arg-type]` correctly suppresses a mypy type-narrowing gap, not a runtime risk.

Trecek, line 11: [warning] arch: Import of `autoskillit._test_filter` is correct, but importing `tests._test_filter` by package path on L12 creates a lateral test-to-test dependency. This only works if `tests/` is a proper Python package (has `__init__.py` at every level). If the `tests/` layout changes, this import silently breaks the contract guard it was designed to enforce. Verify `tests/__init__.py` exists and document this coupling in the module docstring.

tests/arch/test_filter_contracts.py
```diff
@@ -0,0 +1,50 @@
...
+import tests._test_filter as conftest_filter
```
Trecek, line 12: [warning] arch: `import tests._test_filter as conftest_filter` treats a conftest-layer utility as an importable peer module from `tests/arch/`. This bypasses pytest's conftest discovery mechanism. If `tests/_test_filter.py` is ever moved or renamed, this import fails at collection time without a clear error. Consider using importlib to dynamically load by path, or document why direct import is the right approach here.

tests/test_test_filter.py
```diff
@@ -511,23 +540,97 @@ def test_load_manifest_malformed_yaml(self, tmp_path: Path) -> None:
...
         result = apply_manifest({"src/foo.py"}, manifest)
-        assert result == set()
+        assert result is None
 
     def test_apply_manifest_list_dirs(self) -> None:
         manifest = {"*.yaml": ["config", "infra"]}
         result = apply_manifest({"defaults.yaml"}, manifest)
         assert result == {"config", "infra"}
 
+    def test_apply_manifest_doublestar_zero_segments(self) -> None:
```
Trecek, line 560: [warning] tests: test_apply_manifest_doublestar_zero_segments (L560) and test_apply_manifest_doublestar_nested (L569) use identical manifest {'docs/**/*.md': ['docs']} and differ only in the input path. Merge into a single @pytest.mark.parametrize test to avoid duplication and make it trivial to add new glob coverage cases.

tests/test_test_filter.py
```diff
@@ -511,23 +540,97 @@ def test_load_manifest_malformed_yaml(self, tmp_path: Path) -> None:
 class TestApplyManifest:
     def test_apply_manifest_none(self) -> None:
```
Trecek, line 541: [warning] tests: test_apply_manifest_none now only checks `assert result is None` — a one-liner with no setup. Add a brief comment documenting the old behaviour (returned empty set()) to make the regression intent clear; otherwise future readers will not understand what bug this test guards against.

tests/test_test_filter.py
```diff
@@ -511,23 +540,97 @@ def test_load_manifest_malformed_yaml(self, tmp_path: Path) -> None:
...
+        [
+            ("src/autoskillit/recipes/my.yaml", {"recipe", "contracts"}),
+            ("docs/README.md", {"docs"}),
+            ("docs/sub/dir/deep.md", {"docs"}),
+            ("Taskfile.yml", {"infra"}),
+            ("tests/recipe/fixtures/test.yaml", {"recipe"}),
+            (".pre-commit-config.yaml", {"infra"}),
+        ],
+    )
+    def test_matching_cases_agree(self, file_path: str, expected_dirs: set[str]) -> None:
```
Trecek, line 608: [warning] tests: test_matching_cases_agree calls manifest_apply_manifest (L611), which is not defined in the diff. It is unclear from the diff whether this is an alias for the production apply_manifest or a different function. If they are the same function under a different import alias, the equivalence assertion is tautological. Document in the class docstring exactly what manifest_apply_manifest refers to and where it is imported from.

tests/test_test_filter.py
```diff
@@ -511,23 +540,97 @@ def test_load_manifest_malformed_yaml(self, tmp_path: Path) -> None:
...
         result = apply_manifest({"README.md"}, None)
-        assert result == set()
+        assert result is None
 
     def test_apply_manifest_match(self) -> None:
         manifest = {"docs/*.md": ["docs"]}
         result = apply_manifest({"docs/README.md"}, manifest)
         assert result == {"docs"}
 
     def test_apply_manifest_no_match(self) -> None:
```
Trecek, line 550: [warning] tests: test_apply_manifest_no_match now adds manifest = {'tests/*.py': ['unit']} at L551. A prior finding noted a duplicate manifest assignment in this area. Confirm the old duplicate pattern was removed and only the new manifest definition at L551 remains; otherwise the test retains dead code from the original no-match manifest.

tests/arch/test_filter_contracts.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class TestApplyManifestSignatureContract:
+    """Enforces structural compatibility between the two apply_manifest implementations.
+
+    When the production module's apply_manifest signature changes, this test fails
+    immediately, forcing the conftest-side implementation to be updated in sync.
+    """
+
+    def test_both_return_optional_set(self) -> None:
```
Trecek, line 33: [warning] tests: test_both_return_optional_set checks only the string annotation literal. An annotation can be correct while the runtime logic returns an empty set() instead of None. The TestApplyManifestEquivalence tests cover runtime behaviour, but this class docstring claims it 'enforces structural compatibility'. Either rename the class to TestApplyManifestAnnotationContract or add a note that runtime equivalence is separately tested to avoid misleading future reviewers.

tests/test_test_filter.py
```diff
@@ -302,6 +303,34 @@ def test_scope_nonpython_no_manifest_only_alwaysrun(self, tmp_path: Path) -> Non
...
+        manifest = {"docs/**/*.md": ["docs"]}
+        result = build_test_scope(
+            changed_files={"some/unknown/file.txt"},
+            mode=FilterMode.CONSERVATIVE,
+            manifest=manifest,
+            tests_root=tests_root,
+        )
+        assert result is None
+
+    def test_scope_nonpython_matched_manifest_adds_dirs(self, tmp_path: Path) -> None:
```
Trecek, line 322: [info] tests: test_scope_nonpython_matched_manifest_adds_dirs asserts dir_names == {'arch','contracts','infra','docs'} (L336), mixing manifest-matched dirs ('docs') with always_run defaults. If always_run defaults change, this test breaks for a reason unrelated to manifest logic. Consider asserting the manifest contribution separately to isolate the test's intent.

Trecek (commented):

Trecek (commented):
AutoSkillit review found 7 blocking issues (6 warnings + 1 decision finding). See inline comments for details. @Trecek Please review the decision finding at tests/test_test_filter.py:608 — clarification needed on manifest_apply_manifest import alias.

tests/arch/test_filter_contracts.py
Trecek: Investigated — this is intentional. `tests/` has `__init__.py` making it an importable package by design. The contract test (eb0385a1) specifically requires importing both `apply_manifest` implementations to compare AST-extracted return annotations. Importing via package path is the canonical pattern here.

tests/arch/test_filter_contracts.py
Trecek: Investigated — this is intentional. `tests/_test_filter.py` is not conftest-layer; it is a standalone utility module in an importable `tests` package (`tests/__init__.py` exists). The alias `conftest_filter` names the conftest-side `apply_manifest` implementation — it is accurate terminology for the contract test's purpose.

tests/test_test_filter.py
Trecek: Investigated — this is intentional. The two tests validate distinct behavioral cases: zero intermediate path segments (`docs/README.md`) vs. non-zero (`docs/developer/SETUP.md`). The docstrings call out the pathspec vs. fnmatch motivation. Sharing a fixture for two self-contained one-liners would obscure the specific regression each guards against.

tests/test_test_filter.py
Trecek: Investigated — this is intentional. The method name `test_apply_manifest_none` is self-documenting: `manifest=None` → `result=None` (fail-open contract). No other tests in `TestApplyManifest` carry inline comments either. Adding a redundant comment here would be slop.

tests/test_test_filter.py
Trecek: Investigated — this is intentional. `manifest_apply_manifest` is imported at line 11: `from autoskillit._test_filter import apply_manifest as manifest_apply_manifest`. The alias distinguishes the production implementation from the conftest-side `apply_manifest` imported at L22. Both are used in `TestApplyManifestEquivalence` to cross-validate behavior.

tests/test_test_filter.py
Trecek: Investigated — this is intentional. `test_apply_manifest_no_match` requires a non-empty manifest (with `tests/*.py: [unit]`) to distinguish the no-match case from the no-manifest case (`test_apply_manifest_none`). Without a manifest containing patterns, passing `src/foo.py` would be testing a different branch of the function.

tests/arch/test_filter_contracts.py
Trecek: Investigated — this is intentional. The check was deliberately tightened in commit 38ce5694 from a fragile substring match to exact membership in `("set[str] | None", "Optional[set[str]]")`. Since we own both implementations, enforcing canonical annotation forms is appropriate. The reviewer's concern about theoretically valid alternative forms (e.g. `Union[Set[str], None]`) does not apply — no such forms exist here.

PR #1066, opened 2026-04-19
Resilient Plugin Cache Lifecycle — Grace Period, Install Locking, Kitchen-Open Guard

### LINKED ISSUE #1065: Resilient plugin cache lifecycle: grace period, install locking, kitchen-open guard
## Problem Statement

When autoskillit auto-updates during an active pipeline, `_clear_plugin_cache()` calls `shutil.rmtree` on the entire `~/.claude/plugins/cache/autoskillit-local/autoskillit/` directory. Running Claude Code sessions hold in-memory references to the old version's `installPath`. Claude Code checks this path on **every hook dispatch** (not cached — confirmed via binary analysis of the `mH8` hook executor), producing "Plugin directory does not exist" errors. All hooks fail open — safety guards (branch protection, quota enforcement, orchestration depth) are silently bypassed for the remainder of the session.

This is the **7th investigation** in the hook/plugin lifecycle area. Prior fixes (PR #764, #902, #909, #924) all addressed install-time staleness. This is the first instance of **runtime** staleness caused by a concurrent update.

## Root Cause

`_clear_plugin_cache()` in `src/autoskillit/cli/_marketplace.py:26-44` has no session-awareness. It unconditionally deletes all version directories, breaking any session that loaded hooks from the old path. There is no grace period, no lock, and no signal to defer updates while a pipeline is running.

**Current code:**
```python
def _clear_plugin_cache() -> None:
    cache_dir = Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
    if cache_dir.is_dir():
        shutil.rmtree(cache_dir)

    from autoskillit.cli._installed_plugins import InstalledPluginsFile
    try:
        InstalledPluginsFile().remove(f"autoskillit@{_MARKETPLACE_NAME}")
    except OSError:
        pass
```

### Hook dispatch mechanism (Claude Code internals)

The `installPath` from `installed_plugins.json` is resolved **once at session start** (memoized via `bw()`), stored in memory, and the filesystem existence check (`c4(z)`) happens on **every hook dispatch** via `mH8`. Settings.json hooks (pointing to uv site-packages) have `pluginRoot: undefined` and are unaffected — only plugin hooks from the cached `hooks/hooks.json` fail.

### Impact when hooks fail

| Guard | Risk When Bypassed |
|---|---|
| `branch_protection_guard.py` | `merge_worktree`/`push_to_remote` to main/integration not blocked |
| `headless_orchestration_guard.py` | Leaf sessions can spawn recursive sessions (in-handler secondary check mitigates) |
| `quota_guard.py` | Pipeline can exceed API quota limits silently |
| `remove_clone_guard.py` | Clones with unpushed commits could be deleted |
| `pretty_output_hook.py` | Raw JSON output (cosmetic, higher token usage) |

## Solution Overview

Four components, ordered by priority and independence:

| # | Component | Risk | Scope |
|---|-----------|------|-------|
| 1 | Cache rotation with grace period | Low | `_marketplace.py`, new `_plugin_cache.py` |
| 2 | Retiring cache sweep | Low | `_plugin_cache.py`, `_lifespan.py` |
| 3 | Install-path locking | Low | `_marketplace.py` |
| 4 | Kitchen-open update guard | Medium | `tools_kitchen.py`, `_lifespan.py`, `_update_checks.py`, `_update.py` |

Components 1-3 can be implemented together. Component 4 is independent and can be done separately.

---

### What Changes

**File:** `src/autoskillit/cli/_marketplace.py` — `_clear_plugin_cache()`

Replace lines 35-37:
```python
cache_dir = Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
if cache_dir.is_dir():
    shutil.rmtree(cache_dir)
```

With:
```python
cache_dir = Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
if cache_dir.is_dir():
    from autoskillit import __version__ as new_version
    _retire_old_versions(cache_dir, new_version)
```

### `_retire_old_versions(cache_dir, new_version)` Logic

```
for each subdirectory in cache_dir:
    if subdir.name == new_version:
        # Same-version reinstall: delete immediately (no rollback value)
        shutil.rmtree(subdir)
    else:
        # Different version: register for deferred cleanup
        append_retiring_entry(version=subdir.name, path=str(subdir))

# Sweep any entries past the grace period
sweep_retiring_cache()
```

### Why Same-Version Gets Immediate Delete

When reinstalling the same version (e.g., `0.9.50` over `0.9.50`), the old and new directories would have the same path. Keeping the old one alive while `claude plugin install` tries to write the new one creates a name collision. Since the cache contents are identical (same source package), there is no rollback value — the collision is impossible by construction because you never create a retiring record for same-version.

### Key Invariant

`InstalledPluginsFile().remove()` stays at line 42, same ordering (before `claude plugin install`). This is the correctness invariant that forces `claude plugin install` to write a clean entry. It MUST stay, and it MUST execute before `claude plugin install`. The **only** change is to the `shutil.rmtree` call — directory lifecycle only.

### Verified Safe

Claude Code creates versioned subdirectories independently and ignores siblings. Multiple version directories can coexist — Claude Code only uses whichever `installed_plugins.json` points to. Confirmed via live disk inspection: the `arch-lens-mcp-local` plugin cache also has a single versioned subdir with the same structure, and no cleanup behavior was observed.

---

#### `append_retiring_entry(version: str, path: str) -> None`

Atomically appends an entry to `~/.autoskillit/retiring_cache.json`:

```json
{
  "schema_version": 1,
  "retiring": [
    {
      "version": "0.9.50",
      "path": "/home/talon/.claude/plugins/cache/autoskillit-local/autoskillit/0.9.50",
      "retired_at": "2026-04-19T10:00:00+00:00"
    }
  ]
}
```

- Uses `write_versioned_json()` from `core/io.py` (project convention for new JSON artifacts)
- Registry location: `~/.autoskillit/retiring_cache.json` (outside Claude Code's domain — no risk of confusing Claude's cache scanner)
- Locking: `fcntl.LOCK_EX` on `~/.autoskillit/retiring_cache.lock` (sidecar pattern from `clone_registry.py`)

#### `sweep_retiring_cache(grace_hours: int = 2) -> int`

Deletes version directories whose grace period has expired. Returns count deleted.

```
1. If retiring_cache.json does not exist → return 0
2. Acquire LOCK_EX on retiring_cache.lock
3. Read entries
4. For each entry:
   a. Parse retired_at, compute age
   b. If age >= grace_hours:
      - If path.is_dir(): shutil.rmtree(path), increment count
      - Drop entry from list (whether dir existed or not)
   c. If age < grace_hours: keep entry
5. Write surviving entries back (atomic_write)
6. Release lock
7. Return count
```

Error handling:
- Malformed entries (missing `retired_at`, bad JSON): treat as stale, drop silently (matches `sweep_stale_markers` pattern from `core/kitchen_state.py`)
- `shutil.rmtree` failure: keep entry, retry on next sweep
- Entire function is fail-open: wrapped in try/except at call sites, never aborts install

### Grace Period: 2 Hours

- `claude plugin install` completes in <30 seconds (hardcoded timeout in `_marketplace.py`)
- 2 hours is 240x that upper bound
- Long enough for any pipeline run to complete its current ticket
- Short enough to avoid meaningful disk accumulation (8.2 MB per version, ~390 files)

### Sweep Call Sites

1. **At install time** — inside `_retire_old_versions()` after registering entries. Cleans up previous installs' leftovers.

2. **At server startup** — as a background task in `_autoskillit_lifespan` (alongside `_run_drift_check_async`). Offloaded to `loop.run_in_executor(None, sweep_retiring_cache)`, fail-open. Catches orphans from installs that happened while no server was running. Insert at `_lifespan.py:121`:
```python
bg_tasks.append(create_background_task(_run_drift_check_async(), label="drift_check"))
bg_tasks.append(create_background_task(_run_retiring_sweep_async(), label="cache_sweep"))  # NEW
bg_tasks.append(create_background_task(_run_deferred_init(event), label="deferred_init"))
```

3. **NOT at server shutdown** — deletion during teardown risks incomplete I/O if the process is being killed.

---

### Context Manager (in `_plugin_cache.py`)

```python
class _InstallLock:
    """Exclusive fcntl lock for the autoskillit install critical section."""

    def __init__(self) -> None:
        self._lock_path = Path.home() / ".autoskillit" / "install.lock"
        self._lock_file: IO[str] | None = None

    def __enter__(self) -> _InstallLock:
        self._lock_path.parent.mkdir(parents=True, exist_ok=True)
        fh = open(self._lock_path, "w")
        try:
            fcntl.flock(fh, fcntl.LOCK_EX)
        except BaseException:
            fh.close()
            raise
        self._lock_file = fh
        return self

    def __exit__(self, *_: object) -> None:
        if self._lock_file is not None:
            self._lock_file.close()  # releases advisory lock
```

### Lock Scope in `install()` (`_marketplace.py:156-186`)

```python
_ensure_workspace_ready()

with _InstallLock():                                                    # NEW
    _clear_plugin_cache()                                                # existing (modified)
    atomic_write(hooks_json_path, ...)                                   # existing
    subprocess.run(["claude", "plugin", "marketplace", "add", ...])      # existing
    subprocess.run(["claude", "plugin", "install", ...])                 # existing
                                                                         # END lock scope
# Post-install cleanup (outside lock — doesn't touch cache or installed_plugins)
evict_direct_mcp_entry(...)
sweep_all_scopes_for_orphans(...)
sync_hooks_to_settings(...)
```

Design decisions:
- **Blocking, indefinite** (`LOCK_EX` without `LOCK_NB`) — matches `clone_registry.py` pattern. Max lock hold time bounded by subprocess timeouts (~60s).
- **Lock file at `~/.autoskillit/install.lock`** — user-global (plugin cache is user-global). Not CWD-scoped.
- **Lock covers `_clear_plugin_cache` through `claude plugin install`** — the full critical section. Post-install work (hook sync, orphan sweep) is outside the lock.

---

### Goal

Prevent auto-updates from running while any kitchen is open, regardless of the updater's CWD.

### Why a Global Signal is Required

The hook config file (`.autoskillit/temp/.autoskillit_hook_config.json`) and kitchen state markers are CWD-scoped to the MCP server's project directory. An `autoskillit update` invoked from a different terminal/CWD cannot find them. The readiness sentinel is also CWD-relative. There is no existing global "kitchen is open" signal.

### New Registry: `~/.autoskillit/active_kitchens.json`

```json
{
  "schema_version": 1,
  "kitchens": [
    {
      "kitchen_id": "uuid4-string",
      "pid": 12345,
      "project_path": "/abs/path/to/project",
      "opened_at": "2026-04-19T10:00:00+00:00"
    }
  ]
}
```

### Why JSON Registry Over fcntl LOCK_SH

Both work. JSON registry chosen because:
- Produces a human-readable diagnostic artifact (`cat ~/.autoskillit/active_kitchens.json` tells which project has a kitchen open)
- Matches existing patterns (`update_check.json`, `github_fetch_cache.json`)
- PID liveness sweep is ~10 lines, well-understood
- fcntl LOCK_SH would require storing an open FD on `ctx` across async tool calls — messier lifecycle

### Registration Points

**`open_kitchen` succeeds** — `src/autoskillit/server/tools_kitchen.py`, end of `_open_kitchen_handler()` (line 209, after all early-exit failure paths return):
```python
register_active_kitchen(ctx.kitchen_id, os.getpid(), str(Path.cwd()))
return None
```

**`close_kitchen`** — `src/autoskillit/server/tools_kitchen.py`, in `_close_kitchen_handler()` after `ctx.gate.disable()` (line 231), before hook config unlink (line 238):
```python
unregister_active_kitchen(ctx.kitchen_id)
```

**Server shutdown** — `src/autoskillit/server/_lifespan.py`, `finally:` block after `cleanup_readiness_sentinel()` (line 131):
```python
try:
    clear_kitchens_for_pid(os.getpid())
except Exception:
    logger.exception("lifespan kitchen registry cleanup error")
```

### Liveness Check (Stale Entry Recovery)

```python
def _pid_alive(pid: int) -> bool:
    try:
        os.kill(pid, 0)
        return True
    except ProcessLookupError:
        return False
    except PermissionError:
        return True  # process exists, not ours
```

Every read of the registry sweeps dead PIDs before returning results. A crashed server's entry is automatically cleaned on the next check — no daemon needed. Locking via `fcntl.LOCK_EX` on `~/.autoskillit/active_kitchens.lock`.

### Guard Insertion Points

**`src/autoskillit/cli/_update_checks.py` — `run_update_checks()`:**
Insert after existing env-var guards (CLAUDECODE, CI, TTY checks), before `detect_install()`:
```python
from autoskillit.cli._plugin_cache import any_kitchen_open
if any_kitchen_open():
    return  # silently skip — kitchen is active
```

**`src/autoskillit/cli/_update.py` — `run_update_command()`:**
Insert early in the function, after `_home` resolution, before `detect_install()`. Note: `run_update_command()` currently has **zero guards** — no TTY check, no CI check, nothing. This is the first guard added here:
```python
from autoskillit.cli._plugin_cache import any_kitchen_open
if any_kitchen_open():
    print("A kitchen is currently open. Close it or wait for the pipeline to finish.")
    raise SystemExit(1)
```

---

## New Module: `src/autoskillit/cli/_plugin_cache.py`

Consolidates all new functions:

```
_plugin_cache.py
├── _retire_old_versions(cache_dir, new_version)   # Called from _clear_plugin_cache
├── append_retiring_entry(version, path)            # Retiring registry write
├── sweep_retiring_cache(grace_hours=2) -> int      # TTL-based sweep
├── _InstallLock                                    # fcntl context manager
├── register_active_kitchen(kitchen_id, pid, path)  # Kitchen registry write
├── unregister_active_kitchen(kitchen_id)            # Kitchen registry remove
├── clear_kitchens_for_pid(pid)                      # Lifespan teardown cleanup
├── any_kitchen_open() -> bool                       # Update guard check
└── _pid_alive(pid) -> bool                          # Internal helper
```

All functions use `~/.autoskillit/` as the state directory root.

Lock files:
- `~/.autoskillit/install.lock` — install critical section
- `~/.autoskillit/retiring_cache.lock` — retiring registry read-modify-write
- `~/.autoskillit/active_kitchens.lock` — kitchen registry read-modify-write

---

## Files Modified

| File | Change |
|------|--------|
| `src/autoskillit/cli/_marketplace.py` | Replace `shutil.rmtree(cache_dir)` with `_retire_old_versions()`. Wrap critical section in `_InstallLock`. |
| `src/autoskillit/cli/_plugin_cache.py` | **NEW** — all cache lifecycle and kitchen registry functions |
| `src/autoskillit/server/_lifespan.py` | Add `sweep_retiring_cache` as background task at startup (line 121). Add `clear_kitchens_for_pid` in `finally:` block (after line 131). |
| `src/autoskillit/server/tools_kitchen.py` | Add `register_active_kitchen` at end of `_open_kitchen_handler` (line 209). Add `unregister_active_kitchen` in `_close_kitchen_handler` (between lines 231-238). |
| `src/autoskillit/cli/_update_checks.py` | Add `any_kitchen_open()` guard early in `run_update_checks()`. |
| `src/autoskillit/cli/_update.py` | Add `any_kitchen_open()` guard early in `run_update_command()`. |
| `tests/cli/test_install.py` | Update `test_clear_plugin_cache_removes_nested_entry` (line 560). Add 15+ new tests. |
| `tests/cli/test_plugin_cache.py` | **NEW** — sweep and registry tests |
| `tests/cli/test_doctor.py` | Add `test_doctor_plugin_cache_check_ok_when_old_version_dirs_present`. Extend `test_doctor_does_not_modify_plugin_state` (line 633). |

---

### Existing Tests That Need Modification

| Test | File | Change |
|------|------|--------|
| `test_clear_plugin_cache_removes_nested_entry` | `test_install.py:560` | Add version subdir to setup; assert old subdir survives; assert retiring_cache.json entry created |
| `test_install_calls_claude_cli` | `test_install.py:100` | Verify subprocess call count still matches (lock and retire are not subprocesses, so count should be unchanged) |

### New Tests Required

**Category A — Grace period on version change:**
- `test_clear_plugin_cache_retires_old_version_dir_on_version_change` — Setup: create `cache_dir/0.8.0/` and installed_plugins entry for 0.8.0. Monkeypatch `__version__` to 0.9.0. Call `_clear_plugin_cache()`. Assert: `0.8.0/` still exists, `retiring_cache.json` has entry with timestamp.
- `test_clear_plugin_cache_retiring_registry_includes_timestamp` — Assert `retired_at` is ISO-format, parseable, close to now.

**Category B — Same-version reinstall:**
- `test_clear_plugin_cache_deletes_existing_dir_on_same_version_reinstall` — Setup: `cache_dir/0.9.0/`, `__version__` = 0.9.0. Assert: `0.9.0/` deleted, no retiring entry created.
- `test_clear_plugin_cache_noop_when_no_cache_dir_exists` — No cache dir on disk. Assert: no exception, installed_plugins entry removed, no retiring_cache.json.

**Category C — Sweep deletes expired:**
- `test_sweep_retiring_cache_deletes_expired_entries` — Setup: version dir + registry entry with `retired_at` = 25 hours ago. Call `sweep_retiring_cache(grace_hours=24)`. Assert: dir deleted, entry removed.
- `test_sweep_retiring_cache_deletes_multiple_expired_entries` — Two expired dirs. Assert: both deleted.

**Category D — Sweep preserves fresh:**
- `test_sweep_retiring_cache_preserves_fresh_entries` — Setup: entry with `retired_at` = 1 hour ago. Assert: dir and entry survive.
- `test_sweep_retiring_cache_mixed_fresh_and_expired` — One expired (30h), one fresh (1h). Assert: expired deleted, fresh survives.

**Category E — Sweep error handling:**
- `test_sweep_retiring_cache_handles_already_deleted_dir` — Expired entry pointing to nonexistent path. Assert: no exception, entry pruned.
- `test_sweep_retiring_cache_noop_when_file_missing` — No `retiring_cache.json`. Assert: returns 0, no exception.
- `test_sweep_retiring_cache_handles_malformed_json` — Invalid JSON in file. Assert: no exception, fail-open.
- `test_sweep_retiring_cache_handles_missing_retired_at_field` — Valid JSON but missing `retired_at`. Assert: treated as stale, dropped.

**Category F — Install locking:**
- `test_install_lock_file_created_during_call` — Assert lock file appears at `~/.autoskillit/install.lock` during call, released after.

**Category G — Retiring registry integrity:**
- `test_retiring_cache_json_valid_after_multiple_retirements` — Three sequential retires. Assert: valid JSON with 3 entries.
- `test_retiring_cache_json_append_preserves_existing_entries` — Seed with one entry, retire another. Assert: both entries present.
- `test_retiring_cache_json_path_is_absolute` — Assert `Path(entry["path"]).is_absolute()`.

**Category H — Doctor compatibility:**
- `test_doctor_plugin_cache_check_ok_when_old_version_dirs_present` — Old + current version dirs coexist. Assert: `_check_plugin_cache_exists` returns OK.
- `test_doctor_does_not_modify_retiring_cache_json` — Extend existing test at line 633 with a `retiring_cache.json`. Assert: file unchanged after `cli.doctor()`.

**Category K — Kitchen registry:**
- `test_register_active_kitchen_creates_entry` — Call register, read JSON, assert entry exists.
- `test_unregister_active_kitchen_removes_entry` — Register then unregister. Assert: entry gone.
- `test_any_kitchen_open_returns_false_when_pid_dead` — Register with a dead PID (99999999). Assert: `any_kitchen_open()` returns False.
- `test_any_kitchen_open_sweeps_stale_entries` — Register with dead PID. Call `any_kitchen_open()`. Assert: stale entry removed from JSON.
- `test_clear_kitchens_for_pid_removes_all_entries_for_pid` — Register two entries for same PID. Call `clear_kitchens_for_pid`. Assert: both gone.

### xdist Safety

All tests use `monkeypatch.setattr(Path, "home", lambda: tmp_path)` for filesystem isolation. No test touches real `~/.autoskillit/` or `~/.claude/` paths. Lock tests use `tmp_path`-scoped lock files. Safe for `-n 4`.

---

## Edge Cases Addressed

| Edge Case | Handling |
|-----------|----------|
| Same-version reinstall | Immediate rmtree (no grace period, no collision — collision impossible by construction) |
| `claude plugin install` fails | Old version dir survives (grace period). `InstalledPluginsFile` entry is removed (current behavior). Re-run `autoskillit install` to retry. |
| Server crashes without `close_kitchen` | PID liveness check in `any_kitchen_open()` detects dead process, sweeps stale entry |
| Multiple concurrent `autoskillit install` | `_InstallLock` (fcntl LOCK_EX) serializes the critical section |
| Sweep runs during install | Install lock is separate from retiring registry lock. Grace period (2h) ensures sweep never touches a directory that install is actively working on. |
| Old directory already deleted externally | Sweep silently drops the registry entry (path.is_dir() returns False) |
| Malformed retiring_cache.json | Treated as stale, dropped silently (fail-open) |
| Disk accumulation | 8.2 MB per version, max 2 hours retention. At most 2-3 old directories at any time. |
| Update from different CWD | Kitchen registry is global (`~/.autoskillit/`), not CWD-scoped |
| `autoskillit update` (explicit command) | Guard added to `_update.py` in addition to `_update_checks.py` (update command currently has zero guards) |

---

## Implementation Order

1. **Create `src/autoskillit/cli/_plugin_cache.py`** with `_retire_old_versions`, `append_retiring_entry`, `sweep_retiring_cache`, `_InstallLock`
2. **Modify `_clear_plugin_cache()`** in `_marketplace.py` — replace rmtree with `_retire_old_versions()`, wrap critical section in `_InstallLock`
3. **Add sweep to `_lifespan.py`** — background task at startup
4. **Write tests for components 1-3** — categories A through H
5. **Add kitchen registry** to `_plugin_cache.py` — `register_active_kitchen`, `unregister_active_kitchen`, `clear_kitchens_for_pid`, `any_kitchen_open`
6. **Modify `tools_kitchen.py`** — registration/unregistration calls
7. **Modify `_lifespan.py`** — add `clear_kitchens_for_pid` to teardown
8. **Modify `_update_checks.py` and `_update.py`** — add guards
9. **Write tests for component 4** — category K
10. **Run `task test-all`** — full suite including doctor and install tests

---

## What This Does NOT Change

- Hook script content or registration format
- `hooks.json` generation or drift detection
- `settings.json` hook entries (they point to uv site-packages, not the cache)
- `InstalledPluginsFile` class interface (no new methods needed)
- `installed_plugins.json` ordering of `remove()` call (stays before `claude plugin install`)
- Doctor checks (they don't scan cache subdirectories — confirmed via `_check_plugin_cache_exists` which only checks `cache_dir.is_dir()`)
- The marketplace symlink mechanism
- Any MCP tool behavior
- Any recipe or skill behavior

## Validation Criteria

- [ ] `task test-all` passes
- [ ] `autoskillit install` succeeds on a clean system
- [ ] `autoskillit install` succeeds when an old version directory exists
- [ ] `autoskillit update` succeeds and old version directory survives for grace period
- [ ] A pipeline running during an update no longer shows "Plugin directory does not exist" errors
- [ ] `autoskillit doctor` reports no new warnings with old version directories present
- [ ] `autoskillit update` is blocked while a kitchen is open (prints message, exits 1)
- [ ] After server crash, `autoskillit update` is not permanently blocked (PID liveness clears stale entry)

## Key Patterns to Follow

- **Retiring cache sweep**: Follow `sweep_stale_markers()` in `core/kitchen_state.py` — glob-parse-unlink with fail-open semantics
- **fcntl locking**: Follow `CloneRegistry` in `workspace/clone_registry.py` — sidecar `.lock` file, `LOCK_EX`, short hold
- **Atomic writes**: Use `atomic_write()` from `core/io.py` for all JSON mutations
- **Versioned JSON**: Use `write_versioned_json()` from `core/io.py` for new artifacts (project convention)
- **PID liveness**: `os.kill(pid, 0)` with `ProcessLookupError`/`PermissionError` handling
- **Test isolation**: `monkeypatch.setattr(Path, "home", lambda: tmp_path)` — mandatory for all tests touching `~/.autoskillit/` or `~/.claude/`
- **Timestamp helpers**: `datetime.now(UTC) - timedelta(hours=N)` for expired entries in tests (mirrors `test_session_start_reminder.py`)

## Requirements

REQ-CACHE-001: Old plugin cache version directories MUST survive for a configurable grace period (default 2 hours) after a new version is installed.

REQ-CACHE-002: Same-version reinstalls MUST delete the existing version directory immediately (no grace period, no registry entry).

REQ-CACHE-003: The `InstalledPluginsFile().remove()` call MUST remain before `claude plugin install` — this ordering is the correctness invariant.

REQ-CACHE-004: A retiring cache sweep MUST run at install time and server startup, deleting directories whose grace period has expired.

REQ-CACHE-005: The sweep MUST be fail-open — malformed entries, missing directories, and rmtree failures must not abort the install or server startup.

REQ-LOCK-001: The critical section from `_clear_plugin_cache()` through `claude plugin install` MUST be serialized via fcntl advisory lock.

REQ-GUARD-001: `autoskillit update` and the auto-update check MUST be blocked while any kitchen is open, using a global (home-scoped) registry with PID liveness verification.

REQ-GUARD-002: A crashed server's kitchen entry MUST be automatically cleared via PID liveness check — never permanently blocking updates.

REQ-TEST-001: All new tests MUST be xdist-safe (`-n 4`) using `tmp_path` filesystem isolation.

Trecek:
## Summary

When autoskillit auto-updates during an active pipeline, `_clear_plugin_cache()` unconditionally deletes the entire plugin cache directory via `shutil.rmtree`. Running Claude Code sessions hold in-memory references to the old version's `installPath` and check this path on every hook dispatch. When the directory disappears, all hooks fail open — safety guards (branch protection, quota enforcement, orchestration depth) are silently bypassed.

This PR introduces four components:
1. **Cache rotation with grace period** — retire old version directories instead of deleting them immediately
2. **Retiring cache sweep** — TTL-based cleanup of retired directories after a 2-hour grace period
3. **Install-path locking** — fcntl advisory lock serializing the install critical section
4. **Kitchen-open update guard** — global registry preventing updates while any kitchen is active

All new logic is consolidated in a new module `src/autoskillit/core/_plugin_cache.py`.

## Requirements

REQ-CACHE-001: Old plugin cache version directories MUST survive for a configurable grace period (default 2 hours) after a new version is installed.

REQ-CACHE-002: Same-version reinstalls MUST delete the existing version directory immediately (no grace period, no registry entry).

REQ-CACHE-003: The `InstalledPluginsFile().remove()` call MUST remain before `claude plugin install` — this ordering is the correctness invariant.

REQ-CACHE-004: A retiring cache sweep MUST run at install time and server startup, deleting directories whose grace period has expired.

REQ-CACHE-005: The sweep MUST be fail-open — malformed entries, missing directories, and rmtree failures must not abort the install or server startup.

REQ-LOCK-001: The critical section from `_clear_plugin_cache()` through `claude plugin install` MUST be serialized via fcntl advisory lock.

REQ-GUARD-001: `autoskillit update` and the auto-update check MUST be blocked while any kitchen is open, using a global (home-scoped) registry with PID liveness verification.

REQ-GUARD-002: A crashed server's kitchen entry MUST be automatically cleared via PID liveness check — never permanently blocking updates.

REQ-TEST-001: All new tests MUST be xdist-safe (`-n 4`) using `tmp_path` filesystem isolation.

### Error/Resilience Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% ─── VALIDATION GATES (top row) ─── %%
    subgraph Gates ["VALIDATION GATES (Fail-Fast)"]
        direction LR
        WorktreeGate["WorktreeGuard<br/>━━━━━━━━━━<br/>is_git_worktree() → SystemExit<br/>Symlink target must outlive process"]
        KitchenOpenGate["● any_kitchen_open() guard<br/>━━━━━━━━━━<br/>Kitchen open → SystemExit/return<br/>Blocks install & update commands"]
        HeadlessGate["● headless_guard<br/>━━━━━━━━━━<br/>CLAUDECODE session check<br/>open_kitchen stage=headless_guard"]
    end

    %% ─── INSTALL LOCK (new) ─── %%
    subgraph InstallLock ["★ INSTALL LOCK (_plugin_cache.py: _InstallLock)"]
        direction TB
        Lock["★ _InstallLock context manager<br/>━━━━━━━━━━<br/>fcntl LOCK_EX on install.lock<br/>BaseException → fh.close() in __exit__"]
        OpenLockFn["★ _open_lock()<br/>━━━━━━━━━━<br/>Creates ~/.autoskillit/ if absent<br/>BaseException → fh.close() + reraise"]
    end

    %% ─── RETIRING CACHE (new) ─── %%
    subgraph RetiringCache ["★ RETIRING CACHE (_plugin_cache.py)"]
        direction TB
        RetireOld["★ _retire_old_versions()<br/>━━━━━━━━━━<br/>Current version dir → rmtree immediately<br/>Older dirs → append_retiring_entry()"]
        AppendRetiring["★ append_retiring_entry()<br/>━━━━━━━━━━<br/>JSON corrupt → entries=[] (reset)<br/>Writes retiring_cache.json atomically"]
        SweepCache["★ sweep_retiring_cache()<br/>━━━━━━━━━━<br/>Grace: ≥2h since retired_at<br/>OSError on rmtree → survivors list<br/>JSON corrupt → return 0"]
    end

    %% ─── KITCHEN REGISTRY (new) ─── %%
    subgraph KitchenRegistry ["★ KITCHEN REGISTRY (_plugin_cache.py)"]
        direction TB
        RegisterKitchen["★ register_active_kitchen()<br/>━━━━━━━━━━<br/>Appends PID + create_time<br/>JSON corrupt → entries=[]"]
        UnregisterKitchen["★ unregister_active_kitchen()<br/>━━━━━━━━━━<br/>Filters by kitchen_id<br/>JSON corrupt → entries=[]"]
        AnyOpen["★ any_kitchen_open()<br/>━━━━━━━━━━<br/>Prunes dead PIDs in-place<br/>JSON corrupt → return False<br/>Rewrites survivors atomically"]
        PidAlive["★ _pid_alive()<br/>━━━━━━━━━━<br/>ProcessLookupError → False<br/>PermissionError → psutil fallback<br/>create_time delta guard"]
    end

    %% ─── OPEN KITCHEN (modified) ─── %%
    subgraph OpenKitchenFlow ["● OPEN_KITCHEN STAGE PIPELINE (tools_kitchen.py)"]
        direction TB
        OKStage1["● Critical Stages<br/>━━━━━━━━━━<br/>write_hook_config<br/>prime_quota_cache<br/>start_quota_refresh"]
        OKRollback["● gate.disable() rollback<br/>━━━━━━━━━━<br/>Re-closes gate on critical fail<br/>Prevents partial-open state"]
        OKRegistry["● register_active_kitchen call<br/>━━━━━━━━━━<br/>Exception → WARNING only<br/>No rollback — non-fatal"]
        OKRecipe["● Recipe Stages<br/>━━━━━━━━━━<br/>load_and_validate<br/>recipe_find<br/>apply_triage_gate<br/>hook_diagnostic"]
        OKEnvelope["● _kitchen_failure_envelope()<br/>━━━━━━━━━━<br/>JSON {success:false, stage, error}<br/>Catch-all at tool boundary"]
    end

    %% ─── CLOSE KITCHEN + LIFESPAN (modified) ─── %%
    subgraph CloseFlow ["● CLOSE KITCHEN + LIFESPAN TEARDOWN (_lifespan.py, tools_kitchen.py)"]
        direction TB
        CloseKitchen["● close_kitchen()<br/>━━━━━━━━━━<br/>unregister → WARNING only<br/>Never raises"]
        LifespanTeardown["● _autoskillit_lifespan() finally:<br/>━━━━━━━━━━<br/>3 independent try/except blocks<br/>Each isolated — one fail ≠ skip rest"]
        ClearPidBlock["● clear_kitchens_for_pid()<br/>━━━━━━━━━━<br/>Removes all kitchens for this PID<br/>Called on SIGTERM / process exit"]
        RetireSweepAsync["● _run_retiring_sweep_async()<br/>━━━━━━━━━━<br/>Thread-pool offload at startup<br/>sweep_retiring_cache() non-blocking"]
        DriftCheck["run_startup_drift_check()<br/>━━━━━━━━━━<br/>Exception → swallowed + logged<br/>Never prevents server startup"]
    end

    %% ─── TERMINAL STATES ─── %%
    TERROR(["ERROR ENVELOPE\n{success:false}"])
    TEXIT(["SystemExit"])
    TCOMPLETE(["COMPLETE / OPEN"])

    %% ─── CONNECTIONS ─── %%

    %% Gates → fail paths %%
    WorktreeGate -->|"worktree detected"| TEXIT
    KitchenOpenGate -->|"kitchen open"| TEXIT
    HeadlessGate -->|"headless session"| OKEnvelope

    %% Install flow %%
    KitchenOpenGate -->|"no kitchen open"| Lock
    Lock --> OpenLockFn
    OpenLockFn -->|"locked"| RetireOld
    RetireOld --> AppendRetiring
    RetireOld --> SweepCache
    AppendRetiring -->|"JSON corrupt"| AppendRetiring
    SweepCache -->|"OSError → stay in survivors"| SweepCache
    SweepCache -->|"≥ grace period"| TCOMPLETE

    %% Kitchen registry %%
    RegisterKitchen --> AnyOpen
    AnyOpen --> PidAlive
    PidAlive -->|"dead PID pruned"| AnyOpen
    PidAlive -->|"alive"| AnyOpen
    AnyOpen -->|"JSON corrupt → false"| KitchenOpenGate

    %% open_kitchen pipeline %%
    HeadlessGate -->|"interactive"| OKStage1
    OKStage1 -->|"any stage throws"| OKRollback
    OKRollback --> OKEnvelope
    OKStage1 -->|"all stages pass"| OKRegistry
    OKRegistry -->|"exception → WARNING"| OKRegistry
    OKRegistry --> OKRecipe
    OKRecipe -->|"any stage throws"| OKEnvelope
    OKRecipe -->|"all pass"| TCOMPLETE
    OKEnvelope --> TERROR

    %% Close / Lifespan %%
    CloseKitchen --> UnregisterKitchen
    UnregisterKitchen -->|"exception → WARNING"| TCOMPLETE
    LifespanTeardown --> ClearPidBlock
    LifespanTeardown --> RetireSweepAsync
    LifespanTeardown --> DriftCheck
    ClearPidBlock -->|"exception → logged"| TCOMPLETE
    RetireSweepAsync --> SweepCache
    DriftCheck -->|"exception swallowed"| TCOMPLETE

    %% CLASS ASSIGNMENTS %%
    class WorktreeGate,HeadlessGate,KitchenOpenGate detector;
    class Lock,OpenLockFn newComponent;
    class RetireOld,AppendRetiring,SweepCache newComponent;
    class RegisterKitchen,UnregisterKitchen,AnyOpen,PidAlive newComponent;
    class OKStage1,OKRecipe handler;
    class OKRollback gap;
    class OKRegistry output;
    class OKEnvelope gap;
    class CloseKitchen,LifespanTeardown phase;
    class ClearPidBlock,RetireSweepAsync,DriftCheck phase;
    class TERROR,TEXIT,TCOMPLETE terminal;
```

### State Lifecycle Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% ── ENTRY POINTS ─────────────────────────────────────────────────── %%
    INSTALL(["● autoskillit install"])
    UPDATE(["● autoskillit update"])
    OPEN(["● open_kitchen MCP tool"])
    CLOSE(["● close_kitchen MCP tool"])
    LIFESPAN(["● _autoskillit_lifespan"])

    %% ── GATE 1: Install Lock ──────────────────────────────────────────── %%
    subgraph InstallLock ["★ GATE 1 — Install Critical Section (_InstallLock)"]
        direction TB
        ILOCK["★ _InstallLock<br/>━━━━━━━━━━<br/>fcntl.LOCK_EX<br/>install.lock"]
        KITCHEN_GUARD["★ any_kitchen_open()<br/>━━━━━━━━━━<br/>Refuse install if<br/>kitchen live"]
    end

    %% ── GATE 2: Cache Mutation Lock ──────────────────────────────────── %%
    subgraph CacheMutex ["★ GATE 2 — Per-File Mutex (_open_lock / fcntl.LOCK_EX)"]
        direction LR
        RLOCK["★ retiring_cache.lock<br/>━━━━━━━━━━<br/>Exclusive per write"]
        AKLOCK["★ active_kitchens.lock<br/>━━━━━━━━━━<br/>Exclusive per write"]
    end

    %% ── APPEND-ONLY STATE: Retiring Cache ────────────────────────────── %%
    subgraph RetiringState ["★ APPEND-ONLY — retiring_cache.json (schema_version=1)"]
        direction TB
        RETIRE_APPEND["★ append_retiring_entry()<br/>━━━━━━━━━━<br/>APPEND_ONLY<br/>{version, path, retired_at}"]
        GRACE["★ sweep_retiring_cache()<br/>━━━━━━━━━━<br/>grace_hours=2<br/>survivors-list pattern"]
        RETIRE_FS["★ shutil.rmtree(path)<br/>━━━━━━━━━━<br/>Physical deletion<br/>after grace"]
    end

    %% ── MUTABLE STATE: Active Kitchens Registry ──────────────────────── %%
    subgraph KitchenRegistry ["★ MUTABLE — active_kitchens.json (schema_version=1)"]
        direction TB
        REG["★ register_active_kitchen()<br/>━━━━━━━━━━<br/>MUTABLE: append entry<br/>{kitchen_id, pid, create_time,<br/>project_path, opened_at}"]
        UNREG["★ unregister_active_kitchen()<br/>━━━━━━━━━━<br/>MUTABLE: filter by kitchen_id"]
        PID_CLEAR["★ clear_kitchens_for_pid()<br/>━━━━━━━━━━<br/>MUTABLE: filter by pid<br/>lifespan teardown"]
    end

    %% ── GATE 3: PID Liveness (Ghost Eviction) ────────────────────────── %%
    subgraph LivenessGate ["★ GATE 3 — Ghost Eviction (_pid_alive + create_time)"]
        direction TB
        PID_CHECK["★ _pid_alive(pid, stored_create_time)<br/>━━━━━━━━━━<br/>os.kill(pid, 0) + psutil.create_time()<br/>±1.0 s tolerance"]
        DERIVED_READ["★ any_kitchen_open()<br/>━━━━━━━━━━<br/>DERIVED: reads kitchens[],<br/>evicts ghosts, rewrites survivors"]
    end

    %% ── MUTABLE: Kitchen Session Context ─────────────────────────────── %%
    subgraph KitchenCtx ["● MUTABLE — ToolContext kitchen session fields"]
        direction LR
        CTX_OPEN["● ctx.gate.enable()<br/>ctx.kitchen_id = uuid4()<br/>ctx.active_recipe_packs = frozenset()"]
        CTX_RECIPE["● ctx.recipe_name<br/>ctx.recipe_content_hash<br/>ctx.recipe_composite_hash<br/>ctx.recipe_version"]
        CTX_CLOSE["● ctx.gate.disable()<br/>ctx.quota_refresh_task = None<br/>ctx.active_recipe_packs = None<br/>ctx.recipe_* = ''"]
    end

    %% ── INIT_ONLY: Lifespan Startup ──────────────────────────────────── %%
    subgraph LifespanInit ["● INIT_ONLY — _lifespan startup (write-once per process)"]
        direction LR
        SENTINEL["● write_readiness_sentinel()<br/>━━━━━━━━━━<br/>INIT_ONLY<br/>written once; cleaned in finally"]
        STARTUP_EVENT["● _state._startup_ready = Event()<br/>━━━━━━━━━━<br/>INIT_ONLY per startup<br/>signalled by deferred_init"]
        BG_TASKS["● Background tasks (INIT_ONLY)<br/>━━━━━━━━━━<br/>drift_check<br/>cache_sweep → sweep_retiring_cache<br/>deferred_init"]
    end

    %% ── FLOWS ─────────────────────────────────────────────────────────── %%

    INSTALL --> KITCHEN_GUARD
    UPDATE --> KITCHEN_GUARD
    KITCHEN_GUARD -->|"kitchen OPEN → reject"| ILOCK
    KITCHEN_GUARD -->|"no live kitchen → proceed"| ILOCK
    ILOCK --> RLOCK
    RLOCK --> RETIRE_APPEND
    RETIRE_APPEND --> GRACE
    GRACE -->|"age ≥ grace_hours"| RETIRE_FS
    GRACE -->|"age < grace_hours → survivor"| RETIRE_APPEND

    OPEN --> CTX_OPEN
    CTX_OPEN --> AKLOCK
    AKLOCK --> REG
    OPEN --> CTX_RECIPE

    CLOSE --> CTX_CLOSE
    CTX_CLOSE --> AKLOCK
    AKLOCK --> UNREG

    LIFESPAN --> SENTINEL
    LIFESPAN --> STARTUP_EVENT
    LIFESPAN --> BG_TASKS
    BG_TASKS -->|"startup sweep"| RLOCK
    RLOCK -->|"background sweep"| GRACE

    LIFESPAN -->|"finally: teardown"| PID_CLEAR
    PID_CLEAR --> AKLOCK

    DERIVED_READ --> PID_CHECK
    PID_CHECK -->|"ghost detected"| AKLOCK
    PID_CHECK -->|"live entry"| DERIVED_READ
    KITCHEN_GUARD -->|"read"| DERIVED_READ

    %% ── CLASS ASSIGNMENTS ─────────────────────────────────────────────── %%
    class INSTALL,UPDATE,OPEN,CLOSE,LIFESPAN cli;
    class ILOCK,RLOCK,AKLOCK detector;
    class KITCHEN_GUARD,PID_CHECK,DERIVED_READ phase;
    class RETIRE_APPEND,GRACE handler;
    class RETIRE_FS gap;
    class REG,UNREG,PID_CLEAR newComponent;
    class CTX_OPEN,CTX_RECIPE,CTX_CLOSE stateNode;
    class SENTINEL,STARTUP_EVENT,BG_TASKS output;
```

### Process Flow Diagram

```mermaid
flowchart TB

    START([autoskillit install])

    subgraph PreGuards ["Pre-Install Guards"]
        direction TB
        CLAUDECODE{"CLAUDECODE env set?"}
        NO_CLAUDE{"claude on PATH?"}
        WORKTREE{"is_git_worktree?"}
    end

    subgraph CriticalSection ["★ Install Critical Section (_InstallLock)"]
        direction TB
        ACQUIRE["★ _InstallLock.__enter__<br/>━━━━━━━━━━<br/>fcntl.LOCK_EX on install.lock<br/>~/.autoskillit/install.lock"]
        CLEAR_CACHE["● _clear_plugin_cache()<br/>━━━━━━━━━━<br/>locate cache_dir<br/>~/.claude/plugins/cache/autoskillit-local/autoskillit"]
        RETIRE["★ _retire_old_versions(cache_dir, new_ver)<br/>━━━━━━━━━━<br/>iterdir() → classify subdirs"]
        REGEN_HOOKS["● regen hooks.json<br/>━━━━━━━━━━<br/>generate_hooks_json() → atomic_write"]
        MARKETPLACE_CMD["claude plugin marketplace add<br/>━━━━━━━━━━<br/>subprocess.run(timeout=30)"]
        INSTALL_CMD["claude plugin install autoskillit@autoskillit-local<br/>━━━━━━━━━━<br/>subprocess.run(timeout=30)"]
        RELEASE["★ _InstallLock.__exit__<br/>━━━━━━━━━━<br/>fh.close() → fcntl released"]
    end

    subgraph PostInstall ["Post-Install Housekeeping"]
        direction TB
        EVICT["evict_direct_mcp_entry()<br/>━━━━━━━━━━<br/>remove stale MCP from ~/.claude.json"]
        SWEEP_HOOKS["sweep_all_scopes_for_orphans()<br/>━━━━━━━━━━<br/>cross-scope hook cleanup"]
        SYNC_HOOKS["sync_hooks_to_settings()<br/>━━━━━━━━━━<br/>write canonical hooks to scope"]
    end

    DONE([DONE — plugin installed])
    DEFERRED([DEFERRED — manual steps printed])
    ERR_WORKTREE([ERROR — worktree detected])
    ERR_MARKETPLACE([ERROR — marketplace add failed])
    ERR_INSTALL([ERROR — plugin install failed])

    START --> WORKTREE
    WORKTREE -->|"yes"| ERR_WORKTREE
    WORKTREE -->|"no"| CLAUDECODE
    CLAUDECODE -->|"yes"| DEFERRED
    CLAUDECODE -->|"no"| NO_CLAUDE
    NO_CLAUDE -->|"not found"| ERR_WORKTREE
    NO_CLAUDE -->|"found"| ACQUIRE
    ACQUIRE --> CLEAR_CACHE
    CLEAR_CACHE --> RETIRE
    RETIRE --> REGEN_HOOKS
    REGEN_HOOKS --> MARKETPLACE_CMD
    MARKETPLACE_CMD -->|"returncode != 0"| ERR_MARKETPLACE
    MARKETPLACE_CMD -->|"ok"| INSTALL_CMD
    INSTALL_CMD -->|"returncode != 0"| ERR_INSTALL
    INSTALL_CMD -->|"ok"| RELEASE
    RELEASE --> EVICT
    EVICT --> SWEEP_HOOKS
    SWEEP_HOOKS --> SYNC_HOOKS
    SYNC_HOOKS --> DONE

    class START,DONE,DEFERRED terminal;
    class ERR_WORKTREE,ERR_MARKETPLACE,ERR_INSTALL detector;
    class CLAUDECODE,NO_CLAUDE,WORKTREE stateNode;
    class ACQUIRE,RELEASE newComponent;
    class CLEAR_CACHE,RETIRE,REGEN_HOOKS handler;
    class MARKETPLACE_CMD,INSTALL_CMD phase;
    class EVICT,SWEEP_HOOKS,SYNC_HOOKS output;
```

Closes #1065

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260419-133348-895289/.autoskillit/temp/make-plan/resilient_plugin_cache_lifecycle_plan_2026-04-19_134500.md`

## Token Usage Summary

| plan | 3.2k | 15.5k | 598.2k | 69.6k | 1 | 6m 43s |
| review | 18 | 5.5k | 103.5k | 9.8k | 1 | 6m 37s |
| verify | 994 | 12.6k | 1.2M | 62.8k | 1 | 6m 20s |
| implement | 390 | 30.4k | 3.4M | 103.1k | 1 | 6m 39s |
| fix | 392 | 43.8k | 3.7M | 106.3k | 1 | 16m 33s |
| prepare_pr | 60 | 6.9k | 225.8k | 36.7k | 1 | 1m 57s |
| run_arch_lenses | 199 | 27.5k | 743.5k | 132.2k | 3 | 7m 42s |
| compose_pr | 67 | 9.3k | 269.6k | 40.5k | 1 | 2m 15s |
| **Total** | 5.3k | 151.4k | 10.2M | 560.9k | | 54m 48s |

### DIFF
--- a/src/autoskillit/cli/_marketplace.py
+++ b/src/autoskillit/cli/_marketplace.py
@@ -38,6 +38,10 @@ def _clear_plugin_cache() -> None:
         from autoskillit.core import _retire_old_versions
 
         _retire_old_versions(cache_dir, _new_version)
+    else:
+        from autoskillit.core import sweep_retiring_cache
+
+        sweep_retiring_cache()
 
     from autoskillit.cli._installed_plugins import InstalledPluginsFile
 
--- a/src/autoskillit/cli/_marketplace.py
+++ b/src/autoskillit/cli/_marketplace.py
@@ -35,7 +35,7 @@ def _clear_plugin_cache() -> None:
     cache_dir = Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
     if cache_dir.is_dir():
         from autoskillit import __version__ as _new_version
-        from autoskillit.cli._plugin_cache import _retire_old_versions
+        from autoskillit.core import _retire_old_versions
 
         _retire_old_versions(cache_dir, _new_version)
 
@@ -158,7 +158,7 @@ def install(*, scope: str = "user") -> bool:
 
     _ensure_workspace_ready()
 
-    from autoskillit.cli._plugin_cache import _InstallLock
+    from autoskillit.core import _InstallLock
 
     with _InstallLock():
         _clear_plugin_cache()
--- a/src/autoskillit/cli/_marketplace.py
+++ b/src/autoskillit/cli/_marketplace.py
@@ -34,7 +34,10 @@ def _clear_plugin_cache() -> None:
     """
     cache_dir = Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
     if cache_dir.is_dir():
-        shutil.rmtree(cache_dir)
+        from autoskillit import __version__ as _new_version
+        from autoskillit.cli._plugin_cache import _retire_old_versions
+
+        _retire_old_versions(cache_dir, _new_version)
 
     from autoskillit.cli._installed_plugins import InstalledPluginsFile
 
@@ -154,36 +157,40 @@ def install(*, scope: str = "user") -> bool:
         sys.exit(1)
 
     _ensure_workspace_ready()
-    _clear_plugin_cache()
 
-    # Regenerate hooks.json from the canonical registry with absolute paths
-    hooks_json_path = pkg_root() / "hooks" / "hooks.json"
-    atomic_write(hooks_json_path, json.dumps(generate_hooks_json(), indent=2) + "\n")
+    from autoskillit.cli._plugin_cache import _InstallLock
 
-    # Register the marketplace (idempotent)
-    result = subprocess.run(
-        ["claude", "plugin", "marketplace", "add", str(marketplace_dir)],
-        capture_output=True,
-        text=True,
-        stdin=subprocess.DEVNULL,
-        timeout=30,
-    )
-    if result.returncode != 0:
-        print(f"Failed to register marketplace: {result.stderr.strip()}")
-        sys.exit(1)
-    print("Marketplace registered.")
+    with _InstallLock():
+        _clear_plugin_cache()
 
-    # Install the plugin
-    result = subprocess.run(
-        ["claude", "plugin", "install", plugin_ref, "--scope", scope],
-        capture_output=True,
-        text=True,
-        stdin=subprocess.DEVNULL,
-        timeout=30,
-    )
-    if result.returncode != 0:
-        print(f"Failed to install plugin: {result.stderr.strip()}")
-        sys.exit(1)
+        # Regenerate hooks.json from the canonical registry with absolute paths
+        hooks_json_path = pkg_root() / "hooks" / "hooks.json"
+        atomic_write(hooks_json_path, json.dumps(generate_hooks_json(), indent=2) + "\n")
+
+        # Register the marketplace (idempotent)
+        result = subprocess.run(
+            ["claude", "plugin", "marketplace", "add", str(marketplace_dir)],
+            capture_output=True,
+            text=True,
+            stdin=subprocess.DEVNULL,
+            timeout=30,
+        )
+        if result.returncode != 0:
+            print(f"Failed to register marketplace: {result.stderr.strip()}")
+            sys.exit(1)
+        print("Marketplace registered.")
+
+        # Install the plugin
+        result = subprocess.run(
+            ["claude", "plugin", "install", plugin_ref, "--scope", scope],
+            capture_output=True,
+            text=True,
+            stdin=subprocess.DEVNULL,
+            timeout=30,
+        )
+        if result.returncode != 0:
+            print(f"Failed to install plugin: {result.stderr.strip()}")
+            sys.exit(1)
 
     print(f"Plugin installed: {plugin_ref} (scope: {scope})")
     if evict_direct_mcp_entry(_user_claude_json_path()):
--- a/src/autoskillit/core/__init__.py
+++ b/src/autoskillit/core/__init__.py
@@ -6,6 +6,30 @@ explicit `from autoskillit.core.logging import get_logger`.
 """
 
 from ._claude_env import build_claude_env
+from ._plugin_cache import (
+    _InstallLock as _InstallLock,
+)
+from ._plugin_cache import (
+    _retire_old_versions as _retire_old_versions,
+)
+from ._plugin_cache import (
+    any_kitchen_open as any_kitchen_open,
+)
+from ._plugin_cache import (
+    append_retiring_entry as append_retiring_entry,
+)
+from ._plugin_cache import (
+    clear_kitchens_for_pid as clear_kitchens_for_pid,
+)
+from ._plugin_cache import (
+    register_active_kitchen as register_active_kitchen,
+)
+from ._plugin_cache import (
+    sweep_retiring_cache as sweep_retiring_cache,
+)
+from ._plugin_cache import (
+    unregister_active_kitchen as unregister_active_kitchen,
+)
 from ._terminal_table import TerminalColumn as TerminalColumn
 from ._terminal_table import _render_gfm_table as _render_gfm_table
 from ._terminal_table import _render_terminal_table as _render_terminal_table
@@ -273,4 +297,11 @@ __all__ = [
     "truncate_text",
     # _version_snapshot
     "collect_version_snapshot",
+    # _plugin_cache
+    "any_kitchen_open",
+    "append_retiring_entry",
+    "clear_kitchens_for_pid",
+    "register_active_kitchen",
+    "sweep_retiring_cache",
+    "unregister_active_kitchen",
 ]
--- a/src/autoskillit/core/_plugin_cache.py
+++ b/src/autoskillit/core/_plugin_cache.py
@@ -12,8 +12,8 @@ from typing import IO
 
 import psutil
 
-from autoskillit.core import atomic_write, write_versioned_json
-from autoskillit.core.logging import get_logger
+from .io import atomic_write, write_versioned_json
+from .logging import get_logger
 
 logger = get_logger(__name__)
 
--- a/src/autoskillit/core/_plugin_cache.py
+++ b/src/autoskillit/core/_plugin_cache.py
@@ -12,7 +12,7 @@ from typing import IO
 
 import psutil
 
-from .io import atomic_write, write_versioned_json
+from .io import write_versioned_json
 from .logging import get_logger
 
 logger = get_logger(__name__)
@@ -49,7 +49,7 @@ def _open_lock(lock_path: Path) -> IO[str]:
     fh = open(lock_path, "w")
     try:
         fcntl.flock(fh, fcntl.LOCK_EX)
-    except BaseException:
+    except Exception:
         fh.close()
         raise
     return fh
@@ -66,7 +66,9 @@ def append_retiring_entry(version: str, path: str) -> None:
                 entries = json.loads(cache.read_text()).get("retiring", [])
             except (json.JSONDecodeError, AttributeError):
                 entries = []
-        entries.append({"version": version, "path": path, "retired_at": datetime.now(UTC).isoformat()})
+        entries.append(
+            {"version": version, "path": path, "retired_at": datetime.now(UTC).isoformat()}
+        )
         write_versioned_json(cache, {"retiring": entries}, schema_version=_SCHEMA_VERSION)
     finally:
         fh.close()
@@ -91,18 +93,21 @@ def sweep_retiring_cache(grace_hours: int = 2) -> int:
         for entry in entries:
             retired_at_str = entry.get("retired_at")
             if not retired_at_str:
+                survivors.append(entry)
                 continue
             try:
                 retired_at = datetime.fromisoformat(retired_at_str)
                 age = datetime.now(UTC) - retired_at
             except (ValueError, TypeError):
+                survivors.append(entry)
                 continue
             if age >= cutoff:
                 path = entry.get("path", "")
                 if path and Path(path).is_dir():
                     try:
                         shutil.rmtree(path)
-                    except OSError:
+                    except OSError as exc:
+                        logger.warning("sweep_retiring_cache: failed to remove %s: %s", path, exc)
                         survivors.append(entry)
                         continue
                 count += 1
@@ -120,7 +125,12 @@ def _retire_old_versions(cache_dir: Path, new_version: str) -> None:
         if not subdir.is_dir():
             continue
         if subdir.name == new_version:
-            shutil.rmtree(subdir)
+            try:
+                shutil.rmtree(subdir)
+            except OSError as exc:
+                logger.warning(
+                    "_retire_old_versions: failed to remove same-version dir %s: %s", subdir, exc
+                )
         else:
             append_retiring_entry(version=subdir.name, path=str(subdir))
     sweep_retiring_cache()
@@ -132,7 +142,7 @@ class _InstallLock:
     def __init__(self) -> None:
         self._lock_file: IO[str] | None = None
 
-    def __enter__(self) -> "_InstallLock":
+    def __enter__(self) -> _InstallLock:
         self._lock_file = _open_lock(_install_lock_path())
         return self
 
@@ -179,13 +189,15 @@ def register_active_kitchen(kitchen_id: str, pid: int, project_path: str) -> Non
             create_time: float | None = psutil.Process(pid).create_time()
         except psutil.NoSuchProcess:
             create_time = None
-        entries.append({
-            "kitchen_id": kitchen_id,
-            "pid": pid,
-            "create_time": create_time,
-            "project_path": project_path,
-            "opened_at": datetime.now(UTC).isoformat(),
-        })
+        entries.append(
+            {
+                "kitchen_id": kitchen_id,
+                "pid": pid,
+                "create_time": create_time,
+                "project_path": project_path,
+                "opened_at": datetime.now(UTC).isoformat(),
+            }
+        )
         write_versioned_json(akp, {"kitchens": entries}, schema_version=_SCHEMA_VERSION)
     finally:
         fh.close()
@@ -242,10 +254,16 @@ def any_kitchen_open() -> bool:
             if not isinstance(pid, int):
                 continue
             create_time = entry.get("create_time")
-            stored: float | None = float(create_time) if isinstance(create_time, (int, float)) else None
+            stored: float | None = (
+                float(create_time) if isinstance(create_time, (int, float)) else None
+            )
             if _pid_alive(pid, stored_create_time=stored):
                 survivors.append(entry)
-        write_versioned_json(akp, {"kitchens": survivors}, schema_version=_SCHEMA_VERSION)
+        if len(survivors) < len(entries):
+            try:
+                write_versioned_json(akp, {"kitchens": survivors}, schema_version=_SCHEMA_VERSION)
+            except OSError as exc:
+                logger.warning("any_kitchen_open: failed to persist pruned kitchens: %s", exc)
         return len(survivors) > 0
     finally:
         fh.close()
--- a/tests/cli/test_plugin_cache.py
+++ b/tests/cli/test_plugin_cache.py
@@ -2,10 +2,10 @@
 
 from __future__ import annotations
 
+import fcntl
 import json
 import os
 import threading
-import fcntl
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
 
@@ -23,7 +23,7 @@ def test_retire_old_versions_registers_different_version(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import _retire_old_versions
+    from autoskillit.core._plugin_cache import _retire_old_versions
 
     cache_dir = tmp_path / "cache"
     old_dir = cache_dir / "0.8.0"
@@ -44,7 +44,7 @@ def test_retire_old_versions_multiple_old_dirs(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import _retire_old_versions
+    from autoskillit.core._plugin_cache import _retire_old_versions
 
     cache_dir = tmp_path / "cache"
     (cache_dir / "0.7.0").mkdir(parents=True)
@@ -69,7 +69,7 @@ def test_retire_old_versions_deletes_same_version(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import _retire_old_versions
+    from autoskillit.core._plugin_cache import _retire_old_versions
 
     cache_dir = tmp_path / "cache"
     same_dir = cache_dir / "0.9.0"
@@ -88,7 +88,7 @@ def test_retire_old_versions_noop_empty_cache_dir(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import _retire_old_versions
+    from autoskillit.core._plugin_cache import _retire_old_versions
 
     cache_dir = tmp_path / "cache"
     cache_dir.mkdir(parents=True)
@@ -106,11 +106,9 @@ def test_retire_old_versions_noop_empty_cache_dir(
 # ---------------------------------------------------------------------------
 
 
-def test_sweep_deletes_expired_entries(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_sweep_deletes_expired_entries(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import sweep_retiring_cache
+    from autoskillit.core._plugin_cache import sweep_retiring_cache
 
     old_dir = tmp_path / "cache" / "0.8.0"
     old_dir.mkdir(parents=True)
@@ -118,7 +116,12 @@ def test_sweep_deletes_expired_entries(
     retiring_json = tmp_path / ".autoskillit" / "retiring_cache.json"
     retiring_json.parent.mkdir(parents=True, exist_ok=True)
     retiring_json.write_text(
-        json.dumps({"retiring": [{"version": "0.8.0", "path": str(old_dir), "retired_at": retired_at}], "schema_version": 1})
+        json.dumps(
+            {
+                "retiring": [{"version": "0.8.0", "path": str(old_dir), "retired_at": retired_at}],
+                "schema_version": 1,
+            }
+        )
     )
 
     count = sweep_retiring_cache(grace_hours=24)
@@ -129,11 +132,9 @@ def test_sweep_deletes_expired_entries(
     assert data["retiring"] == []
 
 
-def test_sweep_deletes_multiple_expired(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_sweep_deletes_multiple_expired(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import sweep_retiring_cache
+    from autoskillit.core._plugin_cache import sweep_retiring_cache
 
     dir1 = tmp_path / "cache" / "0.7.0"
     dir2 = tmp_path / "cache" / "0.8.0"
@@ -143,13 +144,15 @@ def test_sweep_deletes_multiple_expired(
     retiring_json = tmp_path / ".autoskillit" / "retiring_cache.json"
     retiring_json.parent.mkdir(parents=True, exist_ok=True)
     retiring_json.write_text(
-        json.dumps({
-            "retiring": [
-                {"version": "0.7.0", "path": str(dir1), "retired_at": old_ts},
-                {"version": "0.8.0", "path": str(dir2), "retired_at": old_ts},
-            ],
-            "schema_version": 1,
-        })
+        json.dumps(
+            {
+                "retiring": [
+                    {"version": "0.7.0", "path": str(dir1), "retired_at": old_ts},
+                    {"version": "0.8.0", "path": str(dir2), "retired_at": old_ts},
+                ],
+                "schema_version": 1,
+            }
+        )
     )
 
     count = sweep_retiring_cache(grace_hours=24)
@@ -164,11 +167,9 @@ def test_sweep_deletes_multiple_expired(
 # ---------------------------------------------------------------------------
 
 
-def test_sweep_preserves_fresh_entries(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_sweep_preserves_fresh_entries(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import sweep_retiring_cache
+    from autoskillit.core._plugin_cache import sweep_retiring_cache
 
     fresh_dir = tmp_path / "cache" / "0.8.0"
     fresh_dir.mkdir(parents=True)
@@ -176,7 +177,14 @@ def test_sweep_preserves_fresh_entries(
     retiring_json = tmp_path / ".autoskillit" / "retiring_cache.json"
     retiring_json.parent.mkdir(parents=True, exist_ok=True)
     retiring_json.write_text(
-        json.dumps({"retiring": [{"version": "0.8.0", "path": str(fresh_dir), "retired_at": retired_at}], "schema_version": 1})
+        json.dumps(
+            {
+                "retiring": [
+                    {"version": "0.8.0", "path": str(fresh_dir), "retired_at": retired_at}
+                ],
+                "schema_version": 1,
+            }
+        )
     )
 
     count = sweep_retiring_cache(grace_hours=24)
@@ -187,11 +195,9 @@ def test_sweep_preserves_fresh_entries(
     assert len(data["retiring"]) == 1
 
 
-def test_sweep_mixed_fresh_and_expired(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_sweep_mixed_fresh_and_expired(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import sweep_retiring_cache
+    from autoskillit.core._plugin_cache import sweep_retiring_cache
 
     expired_dir = tmp_path / "cache" / "0.7.0"
     fresh_dir = tmp_path / "cache" / "0.8.0"
@@ -202,13 +208,15 @@ def test_sweep_mixed_fresh_and_expired(
     retiring_json = tmp_path / ".autoskillit" / "retiring_cache.json"
     retiring_json.parent.mkdir(parents=True, exist_ok=True)
     retiring_json.write_text(
-        json.dumps({
-            "retiring": [
-                {"version": "0.7.0", "path": str(expired_dir), "retired_at": expired_ts},
-                {"version": "0.8.0", "path": str(fresh_dir), "retired_at": fresh_ts},
-            ],
-            "schema_version": 1,
-        })
+        json.dumps(
+            {
+                "retiring": [
+                    {"version": "0.7.0", "path": str(expired_dir), "retired_at": expired_ts},
+                    {"version": "0.8.0", "path": str(fresh_dir), "retired_at": fresh_ts},
+                ],
+                "schema_version": 1,
+            }
+        )
     )
 
     count = sweep_retiring_cache(grace_hours=24)
@@ -230,7 +238,7 @@ def test_sweep_handles_already_deleted_dir(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import sweep_retiring_cache
+    from autoskillit.core._plugin_cache import sweep_retiring_cache
 
     missing_dir = tmp_path / "cache" / "0.8.0"
     # Deliberately do NOT create the directory
@@ -238,7 +246,12 @@ def test_sweep_handles_already_deleted_dir(
     retiring_json = tmp_path / ".autoskillit" / "retiring_cache.json"
     retiring_json.parent.mkdir(parents=True, exist_ok=True)
     retiring_json.write_text(
-        json.dumps({"retiring": [{"version": "0.8.0", "path": str(missing_dir), "retired_at": old_ts}], "schema_version": 1})
+        json.dumps(
+            {
+                "retiring": [{"version": "0.8.0", "path": str(missing_dir), "retired_at": old_ts}],
+                "schema_version": 1,
+            }
+        )
     )
 
     count = sweep_retiring_cache(grace_hours=24)  # must not raise
@@ -248,21 +261,17 @@ def test_sweep_handles_already_deleted_dir(
     assert data["retiring"] == []
 
 
-def test_sweep_noop_when_file_missing(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_sweep_noop_when_file_missing(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import sweep_retiring_cache
+    from autoskillit.core._plugin_cache import sweep_retiring_cache
 
     count = sweep_retiring_cache()
     assert count == 0
 
 
-def test_sweep_handles_malformed_json(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_sweep_handles_malformed_json(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import sweep_retiring_cache
+    from autoskillit.core._plugin_cache import sweep_retiring_cache
 
     retiring_json = tmp_path / ".autoskillit" / "retiring_cache.json"
     retiring_json.parent.mkdir(parents=True, exist_ok=True)
@@ -273,11 +282,9 @@ def test_sweep_handles_malformed_json(
     assert count == 0
 
 
-def test_sweep_handles_missing_retired_at(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_sweep_handles_missing_retired_at(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import sweep_retiring_cache
+    from autoskillit.core._plugin_cache import sweep_retiring_cache
 
     old_dir = tmp_path / "cache" / "0.8.0"
     old_dir.mkdir(parents=True)
@@ -298,11 +305,9 @@ def test_sweep_handles_missing_retired_at(
 # ---------------------------------------------------------------------------
 
 
-def test_install_lock_creates_lock_file(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_install_lock_creates_lock_file(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import _InstallLock
+    from autoskillit.core._plugin_cache import _InstallLock
 
     lock_path = tmp_path / ".autoskillit" / "install.lock"
     with _InstallLock():
@@ -315,7 +320,7 @@ def test_install_lock_blocks_concurrent_access(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import _install_lock_path
+    from autoskillit.core._plugin_cache import _install_lock_path
 
     lock_file_path = _install_lock_path()
     lock_file_path.parent.mkdir(parents=True, exist_ok=True)
@@ -357,7 +362,7 @@ def test_append_preserves_existing_entries(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import append_retiring_entry
+    from autoskillit.core._plugin_cache import append_retiring_entry
 
     append_retiring_entry("0.7.0", "/some/path/0.7.0")
     append_retiring_entry("0.8.0", "/some/path/0.8.0")
@@ -369,11 +374,9 @@ def test_append_preserves_existing_entries(
     assert "0.8.0" in versions
 
 
-def test_registry_path_is_absolute(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_registry_path_is_absolute(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import append_retiring_entry
+    from autoskillit.core._plugin_cache import append_retiring_entry
 
     abs_path = str(tmp_path / "cache" / "0.8.0")
     append_retiring_entry("0.8.0", abs_path)
@@ -389,11 +392,9 @@ def test_registry_path_is_absolute(
 # ---------------------------------------------------------------------------
 
 
-def test_register_creates_entry(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_register_creates_entry(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import register_active_kitchen
+    from autoskillit.core._plugin_cache import register_active_kitchen
 
     kitchen_id = "test-kitchen-001"
     pid = os.getpid()
@@ -411,11 +412,9 @@ def test_register_creates_entry(
     assert kitchens[0]["create_time"] is not None
 
 
-def test_unregister_removes_entry(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_unregister_removes_entry(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import register_active_kitchen, unregister_active_kitchen
+    from autoskillit.core._plugin_cache import register_active_kitchen, unregister_active_kitchen
 
     kitchen_id = "test-kitchen-002"
     register_active_kitchen(kitchen_id, os.getpid(), str(tmp_path))
@@ -430,7 +429,7 @@ def test_any_kitchen_open_false_when_pid_dead(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import any_kitchen_open, register_active_kitchen
+    from autoskillit.core._plugin_cache import any_kitchen_open, register_active_kitchen
 
     dead_pid = 99999999
     register_active_kitchen("test-kitchen-003", dead_pid, str(tmp_path))
@@ -439,11 +438,9 @@ def test_any_kitchen_open_false_when_pid_dead(
     assert result is False
 
 
-def test_any_kitchen_open_sweeps_stale(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
-) -> None:
+def test_any_kitchen_open_sweeps_stale(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(Path, "home", lambda: tmp_path)
-    from autoskillit.cli._plugin_cache import any_kitchen_open, register_active_kitchen
+    from autoskillit.core._plugin_cache import any_kitchen_open, register_active_kitchen
 
     dead_pid = 99999999
     register_active_kitchen("test-kitchen-004", dead_pid, str(tmp_path))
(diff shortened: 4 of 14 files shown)


src/autoskillit/core/__init__.py
```diff
@@ -6,6 +6,30 @@
 """
 
 from ._claude_env import build_claude_env
+from ._plugin_cache import (
```
Trecek, line 9: [warning] cohesion: Each symbol from _plugin_cache is exported in its own single-symbol import block (8 separate from/import statements). All other modules in this file use a single consolidated block per source module. Consolidate into one or two import blocks.

src/autoskillit/core/_plugin_cache.py
```diff
@@ -0,0 +1,251 @@
...
+
+def _active_kitchens_lock() -> Path:
+    return _autoskillit_home() / "active_kitchens.lock"
+
+
+def _install_lock_path() -> Path:
+    return _autoskillit_home() / "install.lock"
+
+
+def _open_lock(lock_path: Path) -> IO[str]:
```
Trecek, line 47: [warning] defense: _open_lock catches BaseException (L52) which is broader than necessary — it also catches SystemExit and KeyboardInterrupt. Use `except Exception` instead to avoid masking intentional program termination.

src/autoskillit/core/_plugin_cache.py
```diff
@@ -0,0 +1,251 @@
...
+            entries: list[dict[str, str]] = data.get("retiring", [])
+        except (json.JSONDecodeError, AttributeError, OSError):
+            return 0
+
+        survivors: list[dict[str, str]] = []
+        count = 0
+        cutoff = timedelta(hours=grace_hours)
+        for entry in entries:
+            retired_at_str = entry.get("retired_at")
+            if not retired_at_str:
```
Trecek, line 95: [warning] bugs: Entries whose `retired_at` key is missing are silently dropped with `continue` instead of being added to survivors. This means a malformed entry (e.g. from an interrupted write) is permanently lost from the registry on the next sweep call, rather than preserved for inspection or repair.

src/autoskillit/core/_plugin_cache.py
```diff
@@ -0,0 +1,251 @@
...
+                continue
+            try:
+                retired_at = datetime.fromisoformat(retired_at_str)
```
Trecek, line 99: [warning] bugs: Entries whose `retired_at` timestamp cannot be parsed (ValueError/TypeError) are also silently dropped rather than added to survivors. This permanently removes them from the registry on the first sweep instead of preserving them.

src/autoskillit/core/_plugin_cache.py
```diff
@@ -0,0 +1,251 @@
...
+                age = datetime.now(UTC) - retired_at
+            except (ValueError, TypeError):
+                continue
+            if age >= cutoff:
+                path = entry.get("path", "")
+                if path and Path(path).is_dir():
+                    try:
+                        shutil.rmtree(path)
```
Trecek, line 108: [warning] defense: When shutil.rmtree raises OSError, the entry is silently added back to survivors with no log call. The operator cannot distinguish a permission-denied failure from a transient I/O error, and the cache will grow indefinitely if the path is permanently unremovable. Add `logger.warning(..., exc_info=True)`.

src/autoskillit/core/_plugin_cache.py
```diff
@@ -0,0 +1,251 @@
...
+            else:
+                survivors.append(entry)
+
+        write_versioned_json(cache, {"retiring": survivors}, schema_version=_SCHEMA_VERSION)
+        return count
+    finally:
+        fh.close()
+
+
+def _retire_old_versions(cache_dir: Path, new_version: str) -> None:
```
Trecek, line 123: [warning] defense: `_retire_old_versions` calls `shutil.rmtree(subdir)` directly for the same-version case with no exception handling. If rmtree fails (permissions, in-use directory), the exception propagates uncaught and aborts the install flow. All other deletion operations in this module guard rmtree with try/except OSError — this is inconsistent.

src/autoskillit/core/_plugin_cache.py
```diff
@@ -0,0 +1,251 @@
...
+                entries = json.loads(akp.read_text()).get("kitchens", [])
+            except (json.JSONDecodeError, AttributeError):
+                entries = []
+        survivors = [e for e in entries if e.get("pid") != pid]
+        write_versioned_json(akp, {"kitchens": survivors}, schema_version=_SCHEMA_VERSION)
+    finally:
+        fh.close()
+
+
+def any_kitchen_open() -> bool:
```
Trecek, line 240: [warning] bugs: `any_kitchen_open` calls `write_versioned_json` on every invocation even when no dead-PID entries were pruned (survivors == entries). This causes unnecessary disk I/O and lock contention on every update-check call. Add `if len(survivors) != len(entries):` guard before the write.

Trecek (commented):

src/autoskillit/core/_plugin_cache.py
```diff
@@ -0,0 +1,251 @@
...
+        survivors = []
+        for entry in entries:
+            pid = entry.get("pid")
+            if not isinstance(pid, int):
+                continue
+            create_time = entry.get("create_time")
+            stored: float | None = float(create_time) if isinstance(create_time, (int, float)) else None
+            if _pid_alive(pid, stored_create_time=stored):
+                survivors.append(entry)
+        write_versioned_json(akp, {"kitchens": survivors}, schema_version=_SCHEMA_VERSION)
```
Trecek: [warning] defense: If `write_versioned_json` raises (disk full, atomic-write failure) inside `any_kitchen_open`, the exception propagates to callers (`run_update_command`, `run_update_checks`) that treat this as a simple boolean predicate with no try/except guard. The exception surfaces as an unhandled error at the CLI level. Wrap the write in try/except with logger.warning so the function can still return the computed boolean.

src/autoskillit/cli/_marketplace.py
```diff
@@ -34,7 +34,10 @@ def _clear_plugin_cache() -> None:
     """
     cache_dir = Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
     if cache_dir.is_dir():
-        shutil.rmtree(cache_dir)
+        from autoskillit import __version__ as _new_version
+        from autoskillit.core import _retire_old_versions
+
+        _retire_old_versions(cache_dir, _new_version)
```
Trecek, line 40: [warning] bugs: `_clear_plugin_cache` only calls `_retire_old_versions` (which includes `sweep_retiring_cache`) when `cache_dir.is_dir()`. If the cache directory does not exist, `sweep_retiring_cache` is never called, so stale entries from previous install attempts accumulate in retiring_cache.json indefinitely. Add a bare `sweep_retiring_cache()` call in the else branch.

tests/cli/test_plugin_cache.py
```diff
@@ -0,0 +1,479 @@
...
+    from autoskillit.core._plugin_cache import _retire_old_versions
+
+    cache_dir = tmp_path / "cache"
+    same_dir = cache_dir / "0.9.0"
+    same_dir.mkdir(parents=True)
+
+    _retire_old_versions(cache_dir, "0.9.0")
+
+    assert not same_dir.exists(), "Same-version dir must be deleted immediately"
+    retiring_json = tmp_path / ".autoskillit" / "retiring_cache.json"
```
Trecek, line 82: [warning] tests: `test_retire_old_versions_deletes_same_version` conditionally checks `retiring_json` only if the file exists (`if retiring_json.exists(): ...`). If a bug causes the file to never be created, the test passes vacuously. Assert unconditionally: either the file does not exist, or it contains no same-version entries.

tests/cli/test_plugin_cache.py
```diff
@@ -0,0 +1,479 @@
...
+) -> None:
+    monkeypatch.setattr(Path, "home", lambda: tmp_path)
+    from autoskillit.core._plugin_cache import _retire_old_versions
+
+    cache_dir = tmp_path / "cache"
+    cache_dir.mkdir(parents=True)
+
+    _retire_old_versions(cache_dir, "0.9.0")  # must not raise
+
+    retiring_json = tmp_path / ".autoskillit" / "retiring_cache.json"
```
Trecek, line 99: [warning] tests: `test_retire_old_versions_noop_empty_cache_dir` has the same conditional assertion pattern `if retiring_json.exists(): ...`. If the file is not created the test passes without asserting anything. Assert unconditionally that the retiring list is empty or the file does not exist.

tests/cli/test_plugin_cache.py
```diff
@@ -0,0 +1,479 @@
...
+    from autoskillit.core._plugin_cache import _InstallLock
+
+    lock_path = tmp_path / ".autoskillit" / "install.lock"
+    with _InstallLock():
+        assert lock_path.exists()
+    # File still exists after release (fcntl lock is released, not deleted)
+    assert lock_path.exists()
+
+
+def test_install_lock_blocks_concurrent_access(
```
Trecek, line 323: [warning] tests: `test_install_lock_blocks_concurrent_access` relies on `acquired.wait(timeout=2)` but never asserts the return value. If the background thread is slow to start (heavy CI load), the non-blocking acquire attempt runs before anyone holds the lock, the acquire succeeds, and the test reports a false failure. Assert the return value: `assert acquired.wait(timeout=2), 'lock holder thread did not acquire in time'`.

tests/cli/test_plugin_cache.py
```diff
@@ -0,0 +1,479 @@
...
+    release = threading.Event()
+
+    def hold_lock() -> None:
+        with open(lock_file_path, "w") as fh:
+            fcntl.flock(fh, fcntl.LOCK_EX)
+            acquired.set()
+            release.wait(timeout=5)
+
+    t = threading.Thread(target=hold_lock, daemon=True)
+    t.start()
```
Trecek, line 342: [warning] tests: No assertion that `acquired.wait(timeout=2)` returned True before proceeding to the non-blocking lock attempt — a timing failure and a real concurrency bug become indistinguishable from the test output.

tests/cli/test_plugin_cache.py
```diff
@@ -0,0 +1,479 @@
...
+    with open(lock_file_path, "w") as fh2:
+        try:
+            fcntl.flock(fh2, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            blocked = False
+            fcntl.flock(fh2, fcntl.LOCK_UN)
+        except OSError:
+            blocked = True
+
+    release.set()
+    t.join(timeout=2)
```
Trecek, line 355: [warning] tests: `t.join(timeout=2)` is called but `t.is_alive()` is not checked afterward. A hung daemon thread during teardown can produce confusing CI logs. Add `assert not t.is_alive()` after the join.

tests/cli/test_plugin_cache.py
```diff
@@ -0,0 +1,479 @@
...
+    assert data["kitchens"] == []
+
+
+def test_any_kitchen_open_false_when_pid_dead(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    monkeypatch.setattr(Path, "home", lambda: tmp_path)
+    from autoskillit.core._plugin_cache import any_kitchen_open, register_active_kitchen
+
+    dead_pid = 99999999
```
Trecek: [warning] tests: Hardcoded dead_pid=99999999 is fragile. On Linux the max PID defaults to 4194304 in modern kernels; 99999999 exceeds the kernel limit and is always dead in practice, but is brittle across configurations. Use a pid from a just-exited subprocess or read /proc/sys/kernel/pid_max to stay within the valid range.

tests/cli/test_plugin_cache.py
```diff
@@ -0,0 +1,479 @@
...
+    register_active_kitchen("test-kitchen-003", dead_pid, str(tmp_path))
+
+    result = any_kitchen_open()
+    assert result is False
+
+
+def test_any_kitchen_open_sweeps_stale(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setattr(Path, "home", lambda: tmp_path)
+    from autoskillit.core._plugin_cache import any_kitchen_open, register_active_kitchen
+
```
Trecek, line 451: [warning] tests: Same fragile dead_pid=99999999 issue as L434 — see comment there.

Trecek (commented):
AutoSkillit review found 16 blocking issues. See inline comments. (Note: REQUEST_CHANGES blocked on own PR — posting as comment.)

src/autoskillit/core/__init__.py
Trecek: Investigated — the ruff formatter enforces single-symbol-per-block style for explicit `name as name` re-exports in this file. Consolidating to a single block is reverted on every `pre-commit` run (magic-trailing-comma semantics). The split style is project-enforced and intentional.

src/autoskillit/core/_plugin_cache.py
Trecek: Agreed — fixed in c6c81b23. Changed `except BaseException` to `except Exception` in `_open_lock`.

src/autoskillit/core/_plugin_cache.py
Trecek: Agreed — fixed in c6c81b23. Missing `retired_at` entries are now added to `survivors` instead of being silently dropped. Test `test_sweep_handles_missing_retired_at` updated accordingly.

src/autoskillit/core/_plugin_cache.py
Trecek: Agreed — fixed in c6c81b23. Unparseable `retired_at` timestamps are now preserved in survivors to prevent permanent data loss.

src/autoskillit/core/_plugin_cache.py
Trecek: Agreed — fixed in c6c81b23. Added `logger.warning` when `shutil.rmtree` raises `OSError` in `sweep_retiring_cache` so failures are operator-visible.

src/autoskillit/core/_plugin_cache.py
Trecek: Agreed — fixed in c6c81b23. Wrapped `shutil.rmtree(subdir)` in `_retire_old_versions` with `try/except OSError` and a warning log.

src/autoskillit/core/_plugin_cache.py
Trecek: Agreed — fixed in c6c81b23. `write_versioned_json` is now guarded by `if len(survivors) < len(entries)` to skip unnecessary disk I/O when no dead PIDs were found.

src/autoskillit/core/_plugin_cache.py
Trecek: Agreed — fixed in c6c81b23. Wrapped `write_versioned_json` in `any_kitchen_open` with `try/except OSError` and a warning log so callers always receive a bool.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 9e845c20. Added an `else` branch that calls `sweep_retiring_cache()` when `cache_dir.is_dir()` is False, ensuring stale retiring entries are swept regardless of whether the cache directory exists.

tests/cli/test_plugin_cache.py
Trecek: Agreed — fixed in 13ffcfbd. Replaced the `if retiring_json.exists()` guard with `assert not retiring_json.exists()` — the file should never be created for a same-version reinstall.

tests/cli/test_plugin_cache.py
Trecek: Agreed — fixed in 13ffcfbd. Same fix — `assert not retiring_json.exists()` for the noop empty-cache-dir case.

tests/cli/test_plugin_cache.py
Trecek: Agreed — fixed in 13ffcfbd. Added `assert acquired.wait(timeout=2), 'background lock thread did not acquire within 2s'` so a timing failure is distinguishable from a real concurrency bug.

tests/cli/test_plugin_cache.py
Trecek: Agreed — fixed in 13ffcfbd (same commit as previous comment — both parts of the wait-assertion issue addressed together).

tests/cli/test_plugin_cache.py
Trecek: Agreed — fixed in 13ffcfbd. Added `assert not t.is_alive()` after `t.join(timeout=2)` to catch hung daemon threads.

tests/cli/test_plugin_cache.py
Trecek: Agreed — fixed in 13ffcfbd. Replaced `dead_pid = 99999999` with a real reaped subprocess: `proc = subprocess.Popen(['true']); proc.wait(); dead_pid = proc.pid`.

tests/cli/test_plugin_cache.py
Trecek: Agreed — fixed in 13ffcfbd. Same fix applied to `test_any_kitchen_open_sweeps_stale`.

PR #1254, opened 2026-04-25
fix: switch quota guard sleep 
  from run_cmd to asyncio.sleep (#1222)

### LINKED ISSUE #1222: fix: switch quota guard sleep from run_cmd to Bash to prevent MCP disconnect on ESC
## Summary

When the quota guard instructs the model to sleep via `run_cmd` with `python3 -c "import time; time.sleep(8606)"`, the sleep runs as a subprocess spawned through the MCP server. If the user presses ESC, the in-flight MCP tool call is cancelled. The subprocess is killed, but the cancellation path through `run_cmd` (which only has `except Exception`, not `except BaseException`) lets `CancelledError` propagate in a way that can tear down the MCP server connection.

Investigation report: `.autoskillit/temp/investigate/investigation_mcp-disconnect-account-switch_2026-04-24_230000.md`

## Root Cause

1. Quota guard blocks `run_skill` and instructs the LLM to call `run_cmd` with `python3 -c "import time; time.sleep(8606)"` — a subprocess spawned through the MCP server
2. User presses ESC → Claude Code sends `notifications/cancelled` → CancelledError propagates through the tool handler chain
3. The MCP SDK v1.27.0 `_handle_request` correctly catches CancelledError when `message.cancelled=True` (from notifications/cancelled) and returns cleanly — **the server SHOULD survive**
4. However, the subprocess-based sleep path through `run_managed_async` → `run_cmd` creates a complex cancellation chain where the subprocess cleanup and CancelledError propagation can interact with the transport layer, leading to server death in practice (confirmed by Token Reserve session 227cd5d1)

## Fix: Detect sleep commands in `run_cmd` and use `asyncio.sleep` internally

Instead of spawning a subprocess for `sleep N`, `run_cmd` detects pure sleep commands and converts them to `asyncio.sleep()` internally. This keeps the sleep inside the MCP server's asyncio event loop where cancellation is handled cleanly by the SDK.

### The change (~5 lines, single file)

**`src/autoskillit/server/tools_execution.py`** — In the `run_cmd` function, before calling `_run_subprocess`, add:

```python
import re
_PURE_SLEEP_RE = re.compile(r'^(?:python3?\s+-c\s+["\']import time;\s*time\.sleep\((\d+(?:\.\d+)?)\)["\']|sleep\s+(\d+(?:\.\d+)?))$')

m = _PURE_SLEEP_RE.match(cmd.strip())
if m:
    seconds = float(m.group(1) or m.group(2))
    await asyncio.sleep(seconds)
    return json.dumps({"success": True, "exit_code": 0, "stdout": "", "stderr": ""})
```

### Why this is the lowest blast radius approach

- **Single file, ~5 lines** — only `tools_execution.py` changes
- **Zero prompt changes** — the quota guard keeps telling the model to call `run_cmd` with the same syntax. All 14+ Bash prohibition sites remain untouched and self-consistent
- **Zero new tools** — no schema/registration changes
- **Zero SKILL.md changes** — the QUOTA WAIT PROTOCOL still says "via run_cmd" and is correct
- **Cancellation works correctly** — `asyncio.sleep` is directly cancellable by the SDK's cancel scope. When `notifications/cancelled` arrives, the sleep is cancelled, `_handle_request` sees `message.cancelled=True`, returns cleanly, server survives
- **The `python3 -c "import time; time.sleep(N)"` pattern is quota-guard-specific** — no other code path uses this exact subprocess command, so the regex won't accidentally match normal run_cmd usage

### Why not other approaches

| Approach | Why not |
|---|---|
| Switch to Bash tool | Conflicts with 14+ "NEVER use Bash" prohibition sites (system prompts, tool docstrings, open_kitchen, SKILL.md, recipe YAMLs, fleet prompts). Model faces contradictory instructions. Unreliable. |
| Dual-path "Bash if available, else run_cmd" | Same prohibition conflict. Also: hook can't detect cook vs order sessions (no distinguishing env vars). |
| Cap sleep at 270s + retry loop | 32 retries over 5 hours, each paying full conversation context — prohibitively expensive |
| Dedicated `quota_wait` MCP tool | Clean but higher blast radius: new tool registration, quota guard message change, SKILL.md update, _prompts.py update. The run_cmd detection approach achieves the same result with zero caller changes. |
| Resilient stdin wrapper | `to_thread.run_sync` without `abandon_on_cancel=True` hangs SIGTERM shutdown |
| HTTP transport mode | Kitchen tag gating doesn't persist across stateless HTTP requests |

## Upstream Issues (for reference)

- CC #49479 / #47724: ESC behavior with MCP tool calls (v2.1.105+ regression, both closed)
- CC #43177: stdio MCPs never auto-reconnect

Trecek:
Intercepts pure-sleep commands in run_cmd and converts to asyncio.sleep(). Prevents MCP disconnect on ESC during quota waits. Closes #1222

### DIFF
--- a/src/autoskillit/server/tools_execution.py
+++ b/src/autoskillit/server/tools_execution.py
@@ -2,8 +2,10 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
 import os
+import re
 import time
 from pathlib import Path
 
@@ -33,6 +35,11 @@ from autoskillit.server.helpers import (
 )
 
 logger = get_logger(__name__)
+
+_PURE_SLEEP_RE = re.compile(
+    r'^(?:python3?\s+-c\s+["\']import time;\s*time\.sleep\((\d+(?:\.\d+)?)\)["\']'
+    r"|sleep\s+(\d+(?:\.\d+)?))$"
+)
 
 
 def _is_absolute_path(path: str) -> bool:
@@ -76,6 +83,11 @@ async def run_cmd(
         tool_ctx = _get_ctx()
         _start = time.monotonic()
         try:
+            m = _PURE_SLEEP_RE.match(cmd.strip())
+            if m:
+                seconds = float(m.group(1) or m.group(2))
+                await asyncio.sleep(seconds)
+                return json.dumps({"success": True, "exit_code": 0, "stdout": "", "stderr": ""})
             _env: dict[str, str] | None = (
                 {**os.environ, SCENARIO_STEP_NAME_ENV: step_name} if step_name else None
             )
--- a/src/autoskillit/server/tools_execution.py
+++ b/src/autoskillit/server/tools_execution.py
@@ -37,8 +37,8 @@ from autoskillit.server.helpers import (
 logger = get_logger(__name__)
 
 _PURE_SLEEP_RE = re.compile(
-    r'^(?:python3?\s+-c\s+["\']import time;\s*time\.sleep\((\d+(?:\.\d+)?)\)["\']'
-    r"|sleep\s+(\d+(?:\.\d+)?))$"
+    r'^(?:python3?\s+-c\s+["\']import time;\s*time\.sleep\((?P<py_secs>\d+(?:\.\d+)?)\)["\']'
+    r"|sleep\s+(?P<sh_secs>\d+(?:\.\d+)?))$"
 )
 
 
@@ -85,7 +85,7 @@ async def run_cmd(
         try:
             m = _PURE_SLEEP_RE.match(cmd.strip())
             if m:
-                seconds = float(m.group(1) or m.group(2))
+                seconds = float(m.group("py_secs") or m.group("sh_secs"))
                 await asyncio.sleep(seconds)
                 return json.dumps({"success": True, "exit_code": 0, "stdout": "", "stderr": ""})
             _env: dict[str, str] | None = (
--- a/tests/server/test_tools_run_cmd.py
+++ b/tests/server/test_tools_run_cmd.py
@@ -252,4 +252,5 @@ class TestRunCmdSleepInterception:
     @pytest.mark.anyio
     async def test_step_name_timing_recorded(self, tool_ctx):
         await run_cmd(cmd="sleep 0", cwd="/tmp", step_name="quota_wait")
+        assert len(tool_ctx.runner.call_args_list) == 0
         assert any(e["step_name"] == "quota_wait" for e in tool_ctx.timing_log.get_report())
--- a/tests/server/test_tools_run_cmd.py
+++ b/tests/server/test_tools_run_cmd.py
@@ -40,7 +40,7 @@ class TestRunCmd:
     @pytest.mark.anyio
     async def test_custom_timeout(self, tool_ctx):
         tool_ctx.runner.push(_make_result(0, "", ""))
-        await run_cmd(cmd="sleep 1", cwd="/tmp", timeout=30)
+        await run_cmd(cmd="echo timeout_test", cwd="/tmp", timeout=30)
 
         assert tool_ctx.runner.call_args_list[-1][2] == 30.0
 
@@ -201,3 +201,55 @@ class TestRunPython:
         assert any("timed out" in log.get("event", "").lower() for log in logs), (
             f"Expected 'timed out' in warning event, got: {logs}"
         )
+
+
+class TestRunCmdSleepInterception:
+    """Sleep commands are intercepted and converted to asyncio.sleep."""
+
+    @pytest.mark.anyio
+    async def test_python_sleep_intercepted(self, tool_ctx):
+        result = json.loads(
+            await run_cmd(cmd='python3 -c "import time; time.sleep(60)"', cwd="/tmp")
+        )
+        assert result == {"success": True, "exit_code": 0, "stdout": "", "stderr": ""}
+        assert len(tool_ctx.runner.call_args_list) == 0
+
+    @pytest.mark.anyio
+    async def test_bare_sleep_intercepted(self, tool_ctx):
+        result = json.loads(await run_cmd(cmd="sleep 300", cwd="/tmp"))
+        assert result["success"] is True
+        assert len(tool_ctx.runner.call_args_list) == 0
+
+    @pytest.mark.anyio
+    async def test_python3_single_quotes_intercepted(self, tool_ctx):
+        result = json.loads(
+            await run_cmd(cmd="python3 -c 'import time; time.sleep(120)'", cwd="/tmp")
+        )
+        assert result["success"] is True
+        assert len(tool_ctx.runner.call_args_list) == 0
+
+    @pytest.mark.anyio
+    async def test_non_sleep_uses_subprocess(self, tool_ctx):
+        tool_ctx.runner.push(_make_result(0, "hello", ""))
+        await run_cmd(cmd="echo hello", cwd="/tmp")
+        assert len(tool_ctx.runner.call_args_list) == 1
+        assert tool_ctx.runner.call_args_list[0][0] == ["bash", "-c", "echo hello"]
+
+    @pytest.mark.anyio
+    async def test_decimal_seconds_intercepted(self, tool_ctx):
+        result = json.loads(
+            await run_cmd(cmd='python3 -c "import time; time.sleep(60.5)"', cwd="/tmp")
+        )
+        assert result["success"] is True
+        assert len(tool_ctx.runner.call_args_list) == 0
+
+    @pytest.mark.anyio
+    async def test_compound_sleep_not_intercepted(self, tool_ctx):
+        tool_ctx.runner.push(_make_result(0, "", ""))
+        await run_cmd(cmd="echo before && sleep 10 && echo after", cwd="/tmp")
+        assert len(tool_ctx.runner.call_args_list) == 1
+
+    @pytest.mark.anyio
+    async def test_step_name_timing_recorded(self, tool_ctx):
+        await run_cmd(cmd="sleep 0", cwd="/tmp", step_name="quota_wait")
+        assert any(e["step_name"] == "quota_wait" for e in tool_ctx.timing_log.get_report())
--- a/tests/server/test_tools_run_cmd.py
+++ b/tests/server/test_tools_run_cmd.py
@@ -209,21 +209,21 @@ class TestRunCmdSleepInterception:
     @pytest.mark.anyio
     async def test_python_sleep_intercepted(self, tool_ctx):
         result = json.loads(
-            await run_cmd(cmd='python3 -c "import time; time.sleep(60)"', cwd="/tmp")
+            await run_cmd(cmd='python3 -c "import time; time.sleep(0)"', cwd="/tmp")
         )
         assert result == {"success": True, "exit_code": 0, "stdout": "", "stderr": ""}
         assert len(tool_ctx.runner.call_args_list) == 0
 
     @pytest.mark.anyio
     async def test_bare_sleep_intercepted(self, tool_ctx):
-        result = json.loads(await run_cmd(cmd="sleep 300", cwd="/tmp"))
+        result = json.loads(await run_cmd(cmd="sleep 0", cwd="/tmp"))
         assert result["success"] is True
         assert len(tool_ctx.runner.call_args_list) == 0
 
     @pytest.mark.anyio
     async def test_python3_single_quotes_intercepted(self, tool_ctx):
         result = json.loads(
-            await run_cmd(cmd="python3 -c 'import time; time.sleep(120)'", cwd="/tmp")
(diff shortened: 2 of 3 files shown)


src/autoskillit/server/tools_execution.py
```diff
@@ -76,6 +83,11 @@ async def run_cmd(
         tool_ctx = _get_ctx()
         _start = time.monotonic()
         try:
+            m = _PURE_SLEEP_RE.match(cmd.strip())
+            if m:
+                seconds = float(m.group(1) or m.group(2))
```
Trecek: [critical] defense: `m.group(1) or m.group(2)` uses fragile positional group indices. If the regex is extended and group indices shift, this silently evaluates to `None` → `TypeError` at runtime. Use named groups (`(?P<py_secs>...)` and `(?P<sh_secs>...)`) to make intent explicit and eliminate positional coupling.

src/autoskillit/server/tools_execution.py
```diff
@@ -76,6 +83,11 @@ async def run_cmd(
...
+                await asyncio.sleep(seconds)
+                return json.dumps({"success": True, "exit_code": 0, "stdout": "", "stderr": ""})
```
Trecek, line 90: [warning] arch: The early `return` here exits before the `finally` block that records step_name timing. Intercepted sleep commands with a `step_name` will never appear in `timing_log`, so `test_step_name_timing_recorded` (L253-L255) should fail unless the `finally` block fires on return-inside-try (it does in Python). Verify the `finally` block at L115-L117 actually executes on this early return path — if it does, this is fine; if not, quota-wait steps vanish from diagnostics.

src/autoskillit/server/tools_execution.py
```diff
@@ -34,6 +36,11 @@
 
 logger = get_logger(__name__)
 
+_PURE_SLEEP_RE = re.compile(
+    r'^(?:python3?\s+-c\s+["\']import time;\s*time\.sleep\((\d+(?:\.\d+)?)\)["\']'
```
Trecek: [warning] bugs: The regex uses `["']` independently for both opening and closing quote positions, so mismatched quotes like `python3 -c "import time; time.sleep(5)'` are accepted. Use a backreference (`(?P<q>["'])...(?P=q)`) or separate alternatives per quote type to enforce matching pairs.

tests/server/test_tools_run_cmd.py
```diff
@@ -201,3 +201,55 @@ async def _hang(**_kw: object) -> None:
...
+    async def test_non_sleep_uses_subprocess(self, tool_ctx):
+        tool_ctx.runner.push(_make_result(0, "hello", ""))
+        await run_cmd(cmd="echo hello", cwd="/tmp")
+        assert len(tool_ctx.runner.call_args_list) == 1
+        assert tool_ctx.runner.call_args_list[0][0] == ["bash", "-c", "echo hello"]
+
+    @pytest.mark.anyio
+    async def test_decimal_seconds_intercepted(self, tool_ctx):
+        result = json.loads(
+            await run_cmd(cmd='python3 -c "import time; time.sleep(0.5)"', cwd="/tmp")
```
Trecek: [warning] tests: `test_decimal_seconds_intercepted` passes `time.sleep(0.5)` which causes a real 500ms wall-clock delay via `asyncio.sleep(0.5)`. Use `sleep(0)` to verify decimal parsing without burning wall-clock time — the interception path is already confirmed by zero-duration tests.

src/autoskillit/server/tools_execution.py
```diff
@@ -76,6 +83,11 @@ async def run_cmd(
         tool_ctx = _get_ctx()
         _start = time.monotonic()
         try:
+            m = _PURE_SLEEP_RE.match(cmd.strip())
```
Trecek, line 86: [warning] defense: The interception path bypasses all existing run_cmd scaffolding — audit logging, telemetry, token/timing records, and the tool_ctx pipeline are all skipped. A sleep that is intercepted leaves no trace in session diagnostics. Is this intentional, or should the intercepted path still record timing/audit entries?

Trecek (commented):

src/autoskillit/server/tools_execution.py
Trecek: Valid observation — flagged for design decision. The regex has exactly two mutually exclusive capturing groups; the `or` pattern works correctly because the alternation guarantees exactly one group is non-None. Named groups would improve robustness against future changes but are not required for current correctness.

src/autoskillit/server/tools_execution.py
Trecek: Investigated — this is intentional. Python always executes `finally` blocks when a `return` is reached inside the corresponding `try` block (PEP 341, Language Reference §8.4). The return at L90 is inside the inner `try:` at L85, so the `finally:` at L115-117 fires and records timing. The reviewer correctly identifies this behavior in the comment itself.

src/autoskillit/server/tools_execution.py
Trecek: Investigated — this is intentional. The regex intercepts machine-generated quota-guard sleep commands (src/autoskillit/hooks/quota_guard.py), not arbitrary user input. Mismatched quotes like `python3 -c "import time; time.sleep(5)'` aren't valid shell syntax and will never be produced by the generator. The backreference suggestion is technically correct but the risk scenario cannot occur in practice.

tests/server/test_tools_run_cmd.py
Trecek: Agreed — fixed in 27b52ffbcc03c546a29de047c071dbeb180293c0. Changed `time.sleep(0.5)` to `time.sleep(0.0)` to verify decimal parsing without burning 500ms wall-clock time.

src/autoskillit/server/tools_execution.py
Trecek: Valid observation — flagged for design decision. The intercepted path does record wall-clock timing via the `finally:` block at L115-117 (Python guarantees finally fires on return-inside-try), and the info-level `_notify` at L77-79 fires before the interception check. However, the intercepted path intentionally skips `_run_subprocess` telemetry — this is by design for ESC-resilience with minimal scaffolding per commit fce5d574.

Trecek (commented):

src/autoskillit/server/tools_execution.py
Trecek: [critical] defense: `m.group(1) or m.group(2)` uses fragile positional group indices. If the regex alternation order changes or a new capturing group is inserted, group numbering silently shifts and the wrong value is extracted at runtime. Use named groups (`(?P<py_secs>...)` and `(?P<sh_secs>...)`) and access via `m.group('py_secs') or m.group('sh_secs')` to make the mapping explicit and index-robust.

src/autoskillit/server/tools_execution.py
Trecek, line 86: [warning] defense: The interception path bypasses all existing run_cmd scaffolding — env injection (`SCENARIO_STEP_NAME_ENV`), audit logging, and subprocess telemetry are all skipped. The `finally` block fires correctly (Python guarantees finally on return-inside-try), so step_name timing is recorded. However, if future run_cmd logic adds security checks or rate limiting before subprocess dispatch, the sleep fast-path will silently skip them.

tests/server/test_tools_run_cmd.py
```diff
@@ -40,7 +40,7 @@ async def test_failing_command(self, tool_ctx):
     @pytest.mark.anyio
     async def test_custom_timeout(self, tool_ctx):
         tool_ctx.runner.push(_make_result(0, "", ""))
-        await run_cmd(cmd="sleep 1", cwd="/tmp", timeout=30)
+        await run_cmd(cmd="echo timeout_test", cwd="/tmp", timeout=30)
```
Trecek, line 43: [warning] arch: `test_custom_timeout` was changed from `sleep 1` to `echo timeout_test` because sleep commands are now intercepted. The test still correctly validates timeout parameter forwarding, but the substitution reason is non-obvious — a future reader may wonder why a timeout test uses `echo` instead of a sleep-like command.

tests/server/test_tools_run_cmd.py
```diff
@@ -201,3 +201,55 @@ async def _hang(**_kw: object) -> None:
...
+
+    @pytest.mark.anyio
+    async def test_compound_sleep_not_intercepted(self, tool_ctx):
+        tool_ctx.runner.push(_make_result(0, "", ""))
+        await run_cmd(cmd="echo before && sleep 10 && echo after", cwd="/tmp")
+        assert len(tool_ctx.runner.call_args_list) == 1
+
+    @pytest.mark.anyio
+    async def test_step_name_timing_recorded(self, tool_ctx):
+        await run_cmd(cmd="sleep 0", cwd="/tmp", step_name="quota_wait")
```
Trecek, line 254: [warning] defense: `test_step_name_timing_recorded` only asserts that `step_name` appears in the timing log but does not verify the intercepted path returns the expected success payload. A regression that accidentally removes the early return (falling through to the subprocess runner) would still pass this test if the runner happens to record timing. Consider also asserting `len(tool_ctx.runner.call_args_list) == 0` to confirm interception occurred.

Trecek (commented):
AutoSkillit PR Review — Verdict: **changes_requested**

1 critical finding (blocking), 3 warnings, 2 info-level observations. See inline comments for details.

**Critical:** `m.group(1) or m.group(2)` at `tools_execution.py:88` uses fragile positional group indices — use named groups instead.

**Note:** Multiple subagents incorrectly flagged `test_step_name_timing_recorded` as a failing test (claiming early return bypasses `finally`). This is a false positive — Python guarantees `finally` fires on `return` inside `try`. Prior review thread at L90 already confirmed this behavior.

src/autoskillit/server/tools_execution.py
Trecek: Agreed — fixed in 7b6806399952d51508fbe7dc0e07dc1788310e1b. Switched to named groups `(?P<py_secs>...)` and `(?P<sh_secs>...)` with `m.group('py_secs') or m.group('sh_secs')` access, making the mapping explicit and index-robust.

src/autoskillit/server/tools_execution.py
Trecek: Investigated — this is intentional. The fast-path for pure-sleep commands is designed to skip subprocess scaffolding entirely — there is no subprocess to inject env into, trace, or audit-log. The `finally` block at lines 115-117 correctly records step_name timing regardless of path. Skipping env injection (`SCENARIO_STEP_NAME_ENV`) and subprocess telemetry is correct behavior for an `asyncio.sleep()` replacement. The concern about future security checks is speculative; if such checks are added, they would be placed before the try block (lines 69-72 already demonstrate this pattern with `_require_orchestrator_or_higher` and `_require_enabled`).

tests/server/test_tools_run_cmd.py
Trecek: Valid observation — flagged for design decision. The substitution from `sleep 1` to `echo timeout_test` is functionally required (sleep commands are intercepted before reaching the subprocess runner). Whether to add an explanatory comment is a style judgment — the CLAUDE.md guidelines default to no comments unless the WHY is non-obvious.

tests/server/test_tools_run_cmd.py
Trecek: Agreed — fixed in 12dbab184f4ede4295add50502e3fef5895bc71e. Added `assert len(tool_ctx.runner.call_args_list) == 0` before the timing assertion to confirm interception occurred, consistent with the other tests in the same class.

tests/server/test_tools_run_cmd.py
Trecek, line 43: [warning] tests: `test_custom_timeout` uses `echo timeout_test` instead of a sleep command because sleep commands are now intercepted. The test correctly validates timeout parameter forwarding, but the name no longer reflects why sleep was replaced. A future reader may accidentally revert to a sleep command. Consider renaming to `test_custom_timeout_propagated` or similar to signal that echo is a deliberate substitute.

Trecek (commented):

Trecek (commented):

tests/server/test_tools_run_cmd.py
Trecek: Investigated — this is intentional. The test's mock setup (`tool_ctx.runner.push` + `assert call_args_list[-1][2] == 30.0`) makes the command choice self-documenting: any command that reaches the subprocess runner works, and pure-sleep commands are intercepted before the runner (commit fce5d574). The echo command is the correct choice for this invariant, not a workaround.

Trecek: Investigated — this is intentional. The rename to `test_custom_timeout_propagated` adds no clarity over the current name, which accurately describes the invariant: custom timeout forwarding. The risk of reverting to `sleep` is mitigated by the test structure itself — `tool_ctx.runner.push()` + `call_args_list[-1]` would fail immediately if the command were intercepted (fce5d574).

PR #1260, opened 2026-04-25
fix: fleet runtime gate consistency — three gate paths with unified authority

Trecek:
## Summary

The fleet feature had three runtime gate mechanisms consulting different data sources, creating a validated bypass: `SESSION_TYPE=fleet` + absent `AUTOSKILLIT_FEATURES__FLEET` env var + `features.fleet: false` in config = fleet tools callable. The fix establishes unified authority by (1) adding a config-based feature guard in `dispatch_food_truck`, (2) calling feature suppression in `_fleet_auto_gate_boot()` at lifespan time, and (3) documenting `_fleet_gate()`'s intentional env-var-only behavior. A new `FLEET_FEATURE_DISABLED` error code is added to `FleetErrorCode` to complete the path.

## Requirements

- [ ] REQ-RGC-001: Add `is_feature_enabled("fleet", config.features)` check inside `dispatch_food_truck` handler, after `_require_enabled()` — return `fleet_error(FleetErrorCode.FEATURE_DISABLED, ...)` if fleet is disabled
- [ ] REQ-RGC-002: Call `_redisable_subsets()` (or an equivalent feature-suppression pass) inside `_fleet_auto_gate_boot()` so fleet sessions enforce config-based feature suppression
- [ ] REQ-RGC-003: Document the intentional design of `_fleet_gate()` reading env vars directly (it runs at import time before config is available) — OR — add a deferred config check that fires after config is loaded
- [ ] REQ-RGC-004: Add test: `SESSION_TYPE=fleet` + `features.fleet: false` in config + no `AUTOSKILLIT_FEATURES__FLEET` env var → fleet tools must NOT be callable
- [ ] REQ-RGC-005: Add test: `_fleet_auto_gate_boot()` with `features.fleet: false` → fleet tools must be suppressed after boot

### Security (Trust Boundaries) Diagram

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 40, 'rankSpacing': 55, 'curve': 'basis'}}}%%
flowchart TB
    %% CLASS DEFINITIONS %%
    classDef cli fill:#1a237e,stroke:#7986cb,stroke-width:2px,color:#fff;
    classDef stateNode fill:#004d40,stroke:#4db6ac,stroke-width:2px,color:#fff;
    classDef handler fill:#e65100,stroke:#ffb74d,stroke-width:2px,color:#fff;
    classDef phase fill:#6a1b9a,stroke:#ba68c8,stroke-width:2px,color:#fff;
[...]

PR #1290, opened 2026-04-26
Implementation Plan: Feature Flag Lifecycle — EXPERIMENTAL on by Default, DISABLED State, Test Marker Parity

### LINKED ISSUE #1261: Feature Flag Lifecycle: EXPERIMENTAL on by default, main/stable opt-out, DISABLED state, test marker parity
## Summary

The feature flag lifecycle system has `FeatureLifecycle.EXPERIMENTAL` but doesn't enforce what "experimental" means end-to-end. The intent: experimental features are active by default on integration and any non-main/non-stable branch (including worktrees), and disabled only on main/stable. Currently each experimental feature must be individually listed in `.autoskillit/config.yaml`, `planner` was missed, and the lifecycle state is decorative rather than authoritative.

**The fix:** Make `EXPERIMENTAL` lifecycle the authority. A single `experimental_enabled: true` default enables all EXPERIMENTAL features everywhere. Main/stable opt out with `experimental_enabled: false` in their committed config. Register a feature as `EXPERIMENTAL` and it's active on integration. Period.

### Default is ON

`defaults.yaml` ships with `experimental_enabled: true` as the only entry in the `features:` section. The existing per-feature `fleet: false` and `planner: false` entries are removed — they would shadow the blanket toggle since per-feature entries take precedence.

This means:
- Integration: experimental features ON (inherits default)
- Worktrees off integration: ON (inherits default)
- Feature branches: ON (inherits default)

### Main/stable opt out

Main's committed `.autoskillit/config.yaml` adds:
```yaml
features:
  experimental_enabled: false
```

This commit must happen as part of this implementation to avoid a transition window where experimental features are ON on main.

### Per-feature overrides

Users can override individual features regardless of the blanket toggle:
- Per-feature entry in any config layer: `features: {planner: false}` disables planner even when `experimental_enabled: true`
- Env var: `AUTOSKILLIT_FEATURES__PLANNER=false` overrides everything (highest priority via dynaconf)
- User-level config (`~/.autoskillit/config.yaml`): machine-wide overrides

### Resolution order

`is_feature_enabled()` resolves in this order:
1. If `lifecycle == DISABLED` → `False` (hard override, config cannot change this)
2. If explicit per-feature entry exists in resolved config → use it (escape hatch)
3. If `experimental_enabled == True` and `lifecycle == EXPERIMENTAL` → `True`
4. Fall back to `FeatureDef.default_enabled`

`DEPRECATED` features are unaffected by `experimental_enabled` — they follow their `default_enabled` unless explicitly overridden per-feature.

### Storage model

`_build_features_dict()` extracts `experimental_enabled` from the raw features dict before validating per-feature keys against `FEATURE_REGISTRY`. It is stored as `AutomationConfig.experimental_enabled: bool` (a separate field, not in the `features` dict). `is_feature_enabled()` receives it either as a new parameter or reads it from a module-level/context source. This keeps the `features` dict clean — only real feature names live there.

### DISABLED lifecycle

`FeatureLifecycle.DISABLED` is for features registered in `FEATURE_REGISTRY` but not ready for any branch. `is_feature_enabled()` returns `False` regardless of config or `experimental_enabled`. `_build_features_dict()` rejects attempts to explicitly enable a DISABLED feature. The cook-session all-features-enabled dict in `session_skills.py` filters out DISABLED features.

### R1: Blanket experimental enablement

- Add `experimental_enabled: true` to `defaults.yaml` under `features:`
- Remove `fleet: false` and `planner: false` from `defaults.yaml`
- Add `AutomationConfig.experimental_enabled: bool` field
- `_build_features_dict()` extracts `experimental_enabled` before the per-feature validation loop
- `_build_config_schema()` adds `experimental_enabled` to the valid features sub-key set (both validators must accept it)
- `is_feature_enabled()` implements the resolution order above
- Remove `features: {fleet: true}` from integration's `.autoskillit/config.yaml`
- Add `features: {experimental_enabled: false}` to main's `.autoskillit/config.yaml`

### R2: Feature marker enforcement generalized

`tests/arch/test_feature_markers.py` currently hardcodes fleet-specific file lists. Generalize:
- Auto-discover `tests/{feature_name}/test_*.py` for every feature in `FEATURE_REGISTRY`
- Assert each carries `pytest.mark.feature("{name}")` in `pytestmark`
- Support cross-directory and class-level markers via a registry dict keyed by feature name
- Add missing fleet cross-dir entries: `tests/cli/test_reap.py` and `tests/cli/test_signal_guard.py`

### R3: Planner test files get feature markers

Add `pytest.mark.feature("planner")` to `pytestmark` in:
- `tests/planner/test_compiler.py`
- `tests/planner/test_manifests.py`
- `tests/planner/test_planner.py`
- `tests/planner/test_validation.py`

### R4: Add DISABLED lifecycle state

- Add `FeatureLifecycle.DISABLED = "disabled"` to the enum
- Update `FeatureLifecycle.EXPERIMENTAL` docstring to reflect "on by default" semantics
- `is_feature_enabled()` returns `False` for DISABLED regardless of config
- `_build_features_dict()` rejects explicit enable of DISABLED features
- `session_skills.py` cook-session dict filters out DISABLED features

### R5: Test infrastructure updates

- `_is_test_feature_enabled()` in `conftest.py` must call `is_feature_enabled()` with the new lifecycle logic rather than reading `cfg.features` directly and falling back to `defn.default_enabled` — otherwise experimental feature tests would be incorrectly skipped on integration
- Update `test_feature_registry.py`: lifecycle enum member count assertion (3 → 4), `is_feature_enabled` default behavior assertions
- Update `test_type_constants.py`: `default_enabled` assertions
- Update `test_conftest.py`: `_is_test_feature_enabled` coverage for `experimental_enabled` resolution

### R6: Server startup gate

`_fleet_gate()` in `_session_type.py` reads `AUTOSKILLIT_FEATURES__FLEET` directly from `os.environ` — it does not read config or know about `experimental_enabled`. Update it to respect the resolved config state so that on main (where `experimental_enabled: false`), fleet tools are correctly hidden at MCP startup. Coordinate with PR #1260 (fleet runtime gate consistency) which also modifies this function.

### R7: CLI and documentation

- `features list` shows lifecycle-aware status: EXPERIMENTAL features show `on (experimental)` when blanket is active, `off (override)` when per-feature disabled
- `features status` (single-feature detail) also shows lifecycle-aware source
- Document in `docs/configuration.md`: EXPERIMENTAL (on by default, main opts out), STABLE (on everywhere), DISABLED (off everywhere, cannot override), DEPRECATED (follows default_enabled)

## Open PR dependencies

This issue should be implemented after these open PRs merge (they modify overlapping files):
1. **#1257** — Fleet Import Isolation (modifies `_type_constants.py`, `test_type_constants.py`)
2. **#1259** — Feature gate for run_python callables (calls `is_feature_enabled()` in `rules_features.py`)
3. **#1260** — Fleet runtime gate consistency (modifies `_type_enums.py`, `_session_type.py`, `_lifespan.py`)
4. **#1258** — audit-feature-gates skill (adds adjacent arch test file)

## Non-goals

- **No runtime branch detection.** Config committed per-branch is the mechanism.
- **No config stripping during promotion.** Main has its own `experimental_enabled: false`.
- **No changes to env var override chain.** `AUTOSKILLIT_FEATURES__*` continues as highest-priority escape hatch.

## Files to change

| File | Change |
|---|---|
| `src/autoskillit/core/_type_enums.py` | Add `DISABLED`; update `EXPERIMENTAL` docstring |
| `src/autoskillit/core/feature_flags.py` | Lifecycle-aware `is_feature_enabled()` with `experimental_enabled` |
| `src/autoskillit/config/settings.py` | `_build_features_dict()` extracts `experimental_enabled`; `_build_config_schema()` accepts it |
| `src/autoskillit/config/defaults.yaml` | `experimental_enabled: true` only; remove per-feature entries |
| `.autoskillit/config.yaml` (integration) | Remove `features: {fleet: true}` |
| Main branch `.autoskillit/config.yaml` | Add `features: {experimental_enabled: false}` |
| `src/autoskillit/server/_session_type.py` | `_fleet_gate()` respects resolved config/lifecycle |
| `src/autoskillit/workspace/session_skills.py` | Cook-session dict filters DISABLED features |
| `tests/conftest.py` | `_is_test_feature_enabled()` calls `is_feature_enabled()` |
| `tests/planner/test_compiler.py` | Add `pytest.mark.feature("planner")` |
| `tests/planner/test_manifests.py` | Add `pytest.mark.feature("planner")` |
| `tests/planner/test_planner.py` | Add `pytest.mark.feature("planner")` |
| `tests/planner/test_validation.py` | Add `pytest.mark.feature("planner")` |
| `tests/arch/test_feature_markers.py` | Generalize for all features; add missing fleet cross-dir entries |
| `tests/arch/test_feature_registry.py` | Update enum count + default behavior assertions |
| `tests/core/test_type_constants.py` | Update `default_enabled` assertions |
| `tests/test_conftest.py` | Update `_is_test_feature_enabled` test coverage |
| `src/autoskillit/cli/_features.py` | Lifecycle-aware `features list` + `features status` |
| `docs/configuration.md` | Document convention |

---

**Motivation note:** A planner recipe test run (2026-04-25) confirmed this blocks recipe-dispatched headless sessions — `planner-analyze` couldn't load in sessions dispatched by `run_skill` due to the gate, causing the entire pipeline to fail silently.

Trecek:
## Summary

Make `FeatureLifecycle.EXPERIMENTAL` authoritative: a single `experimental_enabled: true`
in `defaults.yaml` enables all EXPERIMENTAL features on integration and non-main branches.
Add a `DISABLED` lifecycle hard-off state. Add `AutomationConfig.experimental_enabled: bool`
to carry the blanket toggle through the config pipeline into `is_feature_enabled()`. Update
`defaults.yaml`, integration's `.autoskillit/config.yaml`, and the test infrastructure to
reflect the new semantics. Generalize the feature-marker arch test to cover all registry
features rather than hardcoded fleet + planner.

### R1: Blanket experimental enablement

- Add `experimental_enabled: true` to `defaults.yaml` under `features:`
- Remove `fleet: false` and `planner: false` from `defaults.yaml`
- Add `AutomationConfig.experimental_enabled: bool` field
- `_build_features_dict()` extracts `experimental_enabled` before the per-feature validation loop
- `_build_config_schema()` adds `experimental_enabled` to the valid features sub-key set (both validators must accept it)
- `is_feature_enabled()` implements the resolution order above
- Remove `features: {fleet: true}` from integration's `.autoskillit/config.yaml`
- Add `features: {experimental_enabled: false}` to main's `.autoskillit/config.yaml`

### R2: Feature marker enforcement generalized

`tests/arch/test_feature_markers.py` currently hardcodes fleet-specific file lists. Generalize:
- Auto-discover `tests/{feature_name}/test_*.py` for every feature in `FEATURE_REGISTRY`
- Assert each carries `pytest.mark.feature("{name}")` in `pytestmark`
- Support cross-directory and class-level markers via a registry dict keyed by feature name
- Add missing fleet cross-dir entries: `tests/cli/test_reap.py` and `tests/cli/test_signal_guard.py`

### R3: Planner test files get feature markers

Add `pytest.mark.feature("planner")` to `pytestmark` in:
- `tests/planner/test_compiler.py`
- `tests/planner/test_manifests.py`
- `tests/planner/test_planner.py`
- `tests/planner/test_validation.py`

### R4: Add DISABLED lifecycle state

- Add `FeatureLifecycle.DISABLED = "disabled"` to the enum
- Update `FeatureLifecycle.EXPERIMENTAL` docstring to reflect "on by default" semantics
- `is_feature_enabled()` returns `False` for DISABLED regardless of config
- `_build_features_dict()` rejects explicit enable of DISABLED features
- `session_skills.py` cook-session dict filters out DISABLED features

### R5: Test infrastructure updates

- `_is_test_feature_enabled()` in `conftest.py` must call `is_feature_enabled()` with the new lifecycle logic rather than reading `cfg.features` directly and falling back to `defn.default_enabled` — otherwise experimental feature tests would be incorrectly skipped on integration
- Update `test_feature_registry.py`: lifecycle enum member count assertion (3 → 4), `is_feature_enabled` default behavior assertions
- Update `test_type_constants.py`: `default_enabled` assertions
- Update `test_conftest.py`: `_is_test_feature_enabled` coverage for `experimental_enabled` resolution

### R6: Server startup gate

`_fleet_gate()` in `_session_type.py` reads `AUTOSKILLIT_FEATURES__FLEET` directly from `os.environ` — it does not read config or know about `experimental_enabled`. Update it to respect the resolved config state so that on main (where `experimental_enabled: false`), fleet tools are correctly hidden at MCP startup. Coordinate with PR #1260 (fleet runtime gate consistency) which also modifies this function.

### R7: CLI and documentation

- `features list` shows lifecycle-aware status: EXPERIMENTAL features show `on (experimental)` when blanket is active, `off (override)` when per-feature disabled
- `features status` (single-feature detail) also shows lifecycle-aware source
- Document in `docs/configuration.md`: EXPERIMENTAL (on by default, main opts out), STABLE (on everywhere), DISABLED (off everywhere, cannot override), DEPRECATED (follows default_enabled)

### State Lifecycle Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    START([PROCESS START])

    %% ── LAYER 1: INIT_ONLY — Frozen Registry ── %%
    subgraph Registry ["INIT_ONLY — Frozen Feature Registry (module import)"]
        direction TB
        FD["● FeatureDef<br/>━━━━━━━━━━<br/>@dataclass(frozen=True)<br/>name · lifecycle · tool_tags<br/>skill_categories · default_enabled<br/>depends_on · since_version"]
        LC["● FeatureLifecycle<br/>━━━━━━━━━━<br/>EXPERIMENTAL · STABLE<br/>DEPRECATED · ● DISABLED<br/>(StrEnum — immutable values)"]
        FR["FEATURE_REGISTRY<br/>━━━━━━━━━━<br/>dict[str, FeatureDef]<br/>fleet · planner (EXPERIMENTAL)<br/>append-only at import time"]
        IG1["Import Guard 1<br/>━━━━━━━━━━<br/>tool_tags ⊆ TOOL_SUBSET_TAGS<br/>AssertionError at import"]
        IG2["Import Guard 2<br/>━━━━━━━━━━<br/>dict key == FeatureDef.name<br/>AssertionError at import"]
    end

    %% ── LAYER 2: Config Load + Validation Gates ── %%
    subgraph ConfigLoad ["● MUTABLE — Config Load + Validation (_build_features_dict)"]
        direction TB
        CG1["Gate 1: Key type<br/>━━━━━━━━━━<br/>Must be str<br/>→ ConfigSchemaError"]
        CG2["Gate 2: Registry membership<br/>━━━━━━━━━━<br/>Key ∈ FEATURE_REGISTRY<br/>→ ConfigSchemaError"]
        CG3["Gate 3: Value type<br/>━━━━━━━━━━<br/>Must be bool (strict)<br/>→ ConfigSchemaError"]
        CG4["● Gate 4: DISABLED lifecycle<br/>━━━━━━━━━━<br/>value=True + lifecycle==DISABLED<br/>→ ConfigSchemaError (cannot enable)"]
        CG5["Gate 5: Dependency<br/>━━━━━━━━━━<br/>depends_on features must<br/>also be enabled<br/>→ ConfigSchemaError"]
        CFG["● AutomationConfig<br/>━━━━━━━━━━<br/>features: dict[str, bool]<br/>● experimental_enabled: bool<br/>(from features.experimental_enabled YAML key)<br/>Read-only post-load"]
    end

    %% ── LAYER 3: Resolution Function (DERIVED) ── %%
    subgraph Resolution ["● DERIVED — is_feature_enabled() Resolution Chain"]
        direction TB
        R1["● Priority 1: DISABLED hard-off<br/>━━━━━━━━━━<br/>lifecycle==DISABLED → False<br/>No config can override"]
        R2["Priority 2: Explicit override<br/>━━━━━━━━━━<br/>name ∈ features → features[name]"]
        R3["● Priority 3: Blanket experimental<br/>━━━━━━━━━━<br/>experimental_enabled=True<br/>+ lifecycle==EXPERIMENTAL → True"]
        R4["Priority 4: Registry default<br/>━━━━━━━━━━<br/>defn.default_enabled (False)"]
    end

    %% ── LAYER 4: MCP Gate Application ── %%
    subgraph GateApply ["● MCP Tool Gate Application (Two Entry Points)"]
        direction LR
        OKG["● open_kitchen Gate<br/>━━━━━━━━━━<br/>_redisable_subsets()<br/>_collect_disabled_feature_tags()<br/>ctx.disable_components(tags)"]
        FLG["● Fleet Auto-Gate Boot<br/>━━━━━━━━━━<br/>_fleet_auto_gate_boot()<br/>in _lifespan<br/>_mcp.disable(tags)<br/>before lifespan yield"]
    end

    %% ── LAYER 5: INIT_PRESERVE — Kitchen Session State ── %%
    subgraph KitchenState ["INIT_PRESERVE — Kitchen Session State (open_kitchen → run_skill)"]
        direction TB
        KS1["● ctx.gate<br/>━━━━━━━━━━<br/>Set: open_kitchen<br/>Cleared: close_kitchen<br/>Read: _require_enabled()"]
        KS2["ctx.kitchen_id<br/>━━━━━━━━━━<br/>Set: open_kitchen<br/>Read: run_skill → executor<br/>Cleared: close_kitchen"]
        KS3["ctx.recipe_* hashes<br/>━━━━━━━━━━<br/>Set: open_kitchen after load<br/>Read: run_skill → executor<br/>Must match on resume or<br/>provenance inconsistency"]
        KS4["ctx.active_recipe_packs<br/>━━━━━━━━━━<br/>Reset to frozenset() first<br/>Set after recipe load<br/>Read: run_skill → init_session"]
    end

    %% ── LAYER 6: Session Skill Filtering ── %%
    subgraph SkillFilter ["● Session Skill Filtering (init_session)"]
        direction LR
        COOK["● Cook Session Path<br/>━━━━━━━━━━<br/>Force-enables all features<br/>lifecycle != DISABLED<br/>Ignores experimental_enabled<br/>(orchestrator sees all)"]
        PIPE["● Pipeline Session Path<br/>━━━━━━━━━━<br/>Uses config.features verbatim<br/>+ experimental_enabled<br/>Feature-gated skills suppressed<br/>from skill directory"]
    end

    %% ── LAYER 7: Runtime Tool Gates ── %%
    subgraph ToolGates ["Runtime Tool Gates (run_skill / dispatch_food_truck)"]
        direction TB
        TG1["Gate: Session type<br/>━━━━━━━━━━<br/>_require_orchestrator_or_higher()<br/>Headless → denied"]
        TG2["Gate: Kitchen open<br/>━━━━━━━━━━<br/>_require_enabled()<br/>ctx.gate closed → denied"]
        TG3["● Gate: Fleet feature<br/>━━━━━━━━━━<br/>is_feature_enabled('fleet')<br/>Dual-layer guard beyond MCP<br/>visibility (dispatch_food_truck)"]
    end

    END_OK(["TOOL EXECUTED"])
    END_FAIL(["GATE DENIED"])

    %% ── CONNECTIONS ── %%
    START --> Registry
    FD --> LC
    LC --> FR
    FR --> IG1
    FR --> IG2

    Registry --> ConfigLoad
    IG1 & IG2 -->|"Import assertion passes"| CG1
    CG1 --> CG2 --> CG3 --> CG4 --> CG5 --> CFG

    CFG --> Resolution
    FR --> R1
    CFG --> R2
    CFG --> R3
    FR --> R4
    R1 --> R2 --> R3 --> R4

    Resolution --> GateApply
    OKG -->|"Kitchen open path"| KitchenState
    FLG -->|"Fleet session path"| ToolGates

    KitchenState --> SkillFilter
    COOK & PIPE --> ToolGates

    TG1 --> TG2 --> TG3
    TG3 -->|"All gates pass"| END_OK
    TG1 & TG2 & TG3 -->|"Any gate fails"| END_FAIL

    CG1 & CG2 & CG3 & CG4 & CG5 -->|"Validation error"| END_FAIL

    %% ── CLASS ASSIGNMENTS ── %%
    class FD,LC detector;
    class FR stateNode;
    class IG1,IG2 detector;
    class CG1,CG2,CG3,CG4,CG5 stateNode;
    class CFG gap;
    class R1,R2,R3,R4 phase;
    class OKG,FLG handler;
    class KS1 detector;
    class KS2,KS3,KS4 gap;
    class COOK,PIPE handler;
    class TG1,TG2,TG3 stateNode;
    class START,END_OK,END_FAIL terminal;
```

Closes #1261

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260426-034917-731188/.autoskillit/temp/make-plan/feature_flag_lifecycle_plan_2026-04-26_000100.md`

## Token Usage Summary

| plan | 5.6k | 21.3k | 473.1k | 67.4k | 1 | 9m 8s |
| verify | 156 | 14.6k | 916.6k | 51.6k | 1 | 8m 16s |
| implement | 1.0k | 52.9k | 12.6M | 148.7k | 1 | 22m 15s |
| fix | 326 | 17.8k | 2.5M | 138.3k | 1 | 9m 58s |
| prepare_pr | 52 | 8.2k | 184.1k | 37.1k | 1 | 2m 16s |
| run_arch_lenses | 3.0k | 8.2k | 172.4k | 34.5k | 1 | 4m 5s |
| compose_pr | 59 | 5.4k | 201.3k | 26.4k | 1 | 1m 54s |
| **Total** | 10.3k | 128.4k | 17.1M | 504.0k | | 57m 55s |

### DIFF
--- a/src/autoskillit/config/defaults.yaml
+++ b/src/autoskillit/config/defaults.yaml
@@ -268,5 +268,4 @@ fleet:
   default_timeout_sec: 3600
 
 features:
-  fleet: false  # Disabled by default; enabled via project config on integration
-  planner: false  # EXPERIMENTAL — off by default
+  experimental_enabled: true
--- a/src/autoskillit/config/settings.py
+++ b/src/autoskillit/config/settings.py
@@ -373,7 +373,7 @@ class AutomationConfig:
                     f"got {type(value).__name__!r}: {value!r}"
                 )
             if value is True:
-                from autoskillit.core._type_enums import FeatureLifecycle
+                from autoskillit.core import FeatureLifecycle
 
                 if FEATURE_REGISTRY[name].lifecycle == FeatureLifecycle.DISABLED:
                     raise ConfigSchemaError(
--- a/src/autoskillit/config/settings.py
+++ b/src/autoskillit/config/settings.py
@@ -339,17 +339,23 @@ class AutomationConfig:
     workspace: WorkspaceConfig = field(default_factory=WorkspaceConfig)
     fleet: FleetConfig = field(default_factory=FleetConfig)
     features: dict[str, bool] = field(default_factory=dict)
+    experimental_enabled: bool = False
 
     @staticmethod
-    def _build_features_dict(raw: dict[str, Any]) -> dict[str, bool]:
+    def _build_features_dict(raw: dict[str, Any]) -> tuple[dict[str, bool], bool]:
         """Validate and coerce the features section from a raw config dict.
+
+        Returns (features_dict, experimental_enabled).
 
         Raises ConfigSchemaError for:
         - Unknown feature names (not in FEATURE_REGISTRY)
+        - Attempting to enable a DISABLED lifecycle feature
         - Dependency violations: enabling feature B without its required feature A
 
         Coerces all values to bool.
         """
+        raw = dict(raw)  # copy to avoid mutating caller's dict
+        experimental_enabled: bool = bool(raw.pop("experimental_enabled", False))
         result: dict[str, bool] = {}
         for name, value in raw.items():
             if not isinstance(name, str):
@@ -366,6 +372,14 @@ class AutomationConfig:
                     f"Feature {name!r} value must be a bool, "
                     f"got {type(value).__name__!r}: {value!r}"
                 )
+            if value is True:
+                from autoskillit.core._type_enums import FeatureLifecycle
+
+                if FEATURE_REGISTRY[name].lifecycle == FeatureLifecycle.DISABLED:
+                    raise ConfigSchemaError(
+                        f"Feature {name!r} has lifecycle DISABLED"
+                        " and cannot be explicitly enabled."
+                    )
             result[name] = value
 
         # Dependency validation
@@ -388,7 +402,7 @@ class AutomationConfig:
                         f"Enable {dep!r} first."
                     )
 
-        return result
+        return result, experimental_enabled
 
     @classmethod
     def from_dynaconf(cls, d: Dynaconf) -> AutomationConfig:
@@ -453,6 +467,10 @@ class AutomationConfig:
         _sk = _field_defaults(SkillsConfig)
         _wsc = _field_defaults(WorkspaceConfig)
         _fr = _field_defaults(FleetConfig)
+
+        _features_dict, _exp_enabled = AutomationConfig._build_features_dict(
+            dict(feat) if isinstance(feat, dict) else {}
+        )
 
         result = cls(
             test_check=TestCheckConfig(
@@ -602,12 +620,15 @@ class AutomationConfig:
                     val(fr, "default_timeout_sec", _fr["default_timeout_sec"])
                 ),
             ),
-            features=AutomationConfig._build_features_dict(
-                dict(feat) if isinstance(feat, dict) else {}
-            ),
+            features=_features_dict,
+            experimental_enabled=_exp_enabled,
         )
         try:
-            result.fleet.validate(is_feature_enabled("fleet", result.features))
+            result.fleet.validate(
+                is_feature_enabled(
+                    "fleet", result.features, experimental_enabled=result.experimental_enabled
+                )
+            )
         except ValueError as exc:
             raise ValueError(f"fleet config: {exc}") from exc
         return result
@@ -619,7 +640,12 @@ def _build_config_schema() -> dict[str, frozenset[str]]:
     for f in dataclasses.fields(AutomationConfig):
         # Special case: features is a dict[str, bool]; valid sub-keys come from FEATURE_REGISTRY
         if f.name == "features":
-            schema["features"] = frozenset(FEATURE_REGISTRY.keys())
+            schema["features"] = frozenset(FEATURE_REGISTRY.keys()) | frozenset(
+                {"experimental_enabled"}
+            )
+            continue
+        # Skip the scalar experimental_enabled field — it is handled under the features section
+        if f.name == "experimental_enabled":
             continue
         sub_type: type | None = None
         if f.default_factory is not dataclasses.MISSING:  # type: ignore[misc]
--- a/src/autoskillit/config/settings.py
+++ b/src/autoskillit/config/settings.py
@@ -22,6 +22,7 @@ from typing import TYPE_CHECKING, Any, ClassVar
 from autoskillit.core import (
     CATEGORY_TAGS,
     FEATURE_REGISTRY,
+    FeatureLifecycle,
     OutputFormat,
     atomic_write,
     dump_yaml_str,
@@ -373,8 +374,6 @@ class AutomationConfig:
                     f"got {type(value).__name__!r}: {value!r}"
                 )
             if value is True:
-                from autoskillit.core import FeatureLifecycle
-
                 if FEATURE_REGISTRY[name].lifecycle == FeatureLifecycle.DISABLED:
                     raise ConfigSchemaError(
                         f"Feature {name!r} has lifecycle DISABLED"
--- a/src/autoskillit/core/feature_flags.py
+++ b/src/autoskillit/core/feature_flags.py
@@ -44,16 +44,12 @@ def is_feature_enabled(
     defn = FEATURE_REGISTRY.get(name)
     if defn is None:
         raise KeyError(f"Unknown feature: {name!r}")
-    # Step 1: DISABLED is unconditional — cannot be overridden by any config
     if defn.lifecycle == FeatureLifecycle.DISABLED:
         return False
-    # Step 2: explicit per-feature entry always wins
     if name in features:
         return features[name]
-    # Step 3: blanket experimental enablement
     if experimental_enabled and defn.lifecycle == FeatureLifecycle.EXPERIMENTAL:
         return True
-    # Step 4: registry default
     return defn.default_enabled
 
 
--- a/src/autoskillit/core/feature_flags.py
+++ b/src/autoskillit/core/feature_flags.py
@@ -8,9 +8,15 @@ full AutomationConfig to keep core/ free of config/ imports.
 from __future__ import annotations
 
 from ._type_constants import FEATURE_REGISTRY
+from ._type_enums import FeatureLifecycle
 
 
-def is_feature_enabled(name: str, features: dict[str, bool]) -> bool:
+def is_feature_enabled(
+    name: str,
+    features: dict[str, bool],
+    *,
+    experimental_enabled: bool = False,
+) -> bool:
     """Check whether a named feature is enabled.
 
     Parameters
@@ -20,11 +26,15 @@ def is_feature_enabled(name: str, features: dict[str, bool]) -> bool:
     features:
         Resolved features dict from ``AutomationConfig.features``. Typically
         passed as ``config.features`` at call sites; never pass the full config.
+    experimental_enabled:
+        When True, all EXPERIMENTAL lifecycle features are enabled unless explicitly
+        overridden by a per-feature entry in ``features``.
 
     Returns
     -------
     bool
-        ``features[name]`` if the key is present; otherwise ``FeatureDef.default_enabled``.
+        Resolution order:
+        DISABLED hard-off → explicit override → experimental blanket → default_enabled.
 
     Raises
     ------
@@ -34,10 +44,24 @@ def is_feature_enabled(name: str, features: dict[str, bool]) -> bool:
     defn = FEATURE_REGISTRY.get(name)
     if defn is None:
         raise KeyError(f"Unknown feature: {name!r}")
-    return features.get(name, defn.default_enabled)
+    # Step 1: DISABLED is unconditional — cannot be overridden by any config
+    if defn.lifecycle == FeatureLifecycle.DISABLED:
+        return False
+    # Step 2: explicit per-feature entry always wins
+    if name in features:
+        return features[name]
+    # Step 3: blanket experimental enablement
+    if experimental_enabled and defn.lifecycle == FeatureLifecycle.EXPERIMENTAL:
+        return True
+    # Step 4: registry default
+    return defn.default_enabled
 
 
-def _collect_disabled_feature_tags(features: dict[str, bool]) -> frozenset[str]:
+def _collect_disabled_feature_tags(
+    features: dict[str, bool],
+    *,
+    experimental_enabled: bool = False,
+) -> frozenset[str]:
     """Return feature tags that should be suppressed.
 
     Single source of truth used by _fleet_auto_gate_boot and _redisable_subsets.
@@ -48,7 +72,7 @@ def _collect_disabled_feature_tags(features: dict[str, bool]) -> frozenset[str]:
     for name, defn in FEATURE_REGISTRY.items():
         if not defn.tool_tags:
             continue
-        if is_feature_enabled(name, features):
+        if is_feature_enabled(name, features, experimental_enabled=experimental_enabled):
             enabled_tags |= defn.tool_tags
         else:
             disabled_tags |= defn.tool_tags
--- a/src/autoskillit/server/tools_execution.py
+++ b/src/autoskillit/server/tools_execution.py
@@ -437,10 +437,14 @@ async def dispatch_food_truck(
         from autoskillit.server import _get_ctx as _get_ctx_for_feature_check
 
         _feature_ctx = _get_ctx_for_feature_check()
-        if not is_feature_enabled("fleet", _feature_ctx.config.features):
+        if not is_feature_enabled(
+            "fleet",
+            _feature_ctx.config.features,
+            experimental_enabled=_feature_ctx.config.experimental_enabled,
+        ):
             return fleet_error(
                 FleetErrorCode.FLEET_FEATURE_DISABLED,
-                "Fleet feature is disabled in configuration. Set features.fleet: true to enable.",
+                "Fleet feature is disabled. Set features.experimental_enabled: true to enable.",
             )
 
         if os.environ.get("AUTOSKILLIT_HEADLESS") == "1":
--- a/src/autoskillit/workspace/session_skills.py
+++ b/src/autoskillit/workspace/session_skills.py
@@ -402,7 +402,7 @@ class DefaultSessionSkillManager:
             effective_custom_tags = dict(config.subsets.custom_tags)
 
         packs_enabled: list[str] = [] if config is None else list(config.packs.enabled)
-        from autoskillit.core._type_enums import FeatureLifecycle
+        from autoskillit.core import FeatureLifecycle
 
         session_features: dict[str, bool] = (
             {
--- a/src/autoskillit/workspace/session_skills.py
+++ b/src/autoskillit/workspace/session_skills.py
@@ -152,6 +152,8 @@ def _is_skill_disabled(
     disabled: list[str],
     custom_tags: dict[str, list[str]],
     features: dict[str, bool],
+    *,
+    experimental_enabled: bool = False,
 ) -> bool:
     """Return True if skill should be excluded due to a disabled subset.
 
@@ -174,7 +176,7 @@ def _is_skill_disabled(
     enabled_cats: set[str] = set()
     disabled_cats: set[str] = set()
     for feat_name, feat_def in FEATURE_REGISTRY.items():
-        if is_feature_enabled(feat_name, features):
+        if is_feature_enabled(feat_name, features, experimental_enabled=experimental_enabled):
             enabled_cats |= feat_def.skill_categories
         else:
             disabled_cats |= feat_def.skill_categories
@@ -221,6 +223,7 @@ def _should_inject_skill(
     effective_disabled: frozenset[str],
     effective_custom_tags: dict[str, list[str]],
     features: dict[str, bool],
+    experimental_enabled: bool = False,
 ) -> bool:
     """Return True if this skill should be written to the ephemeral session dir.
 
@@ -235,7 +238,13 @@ def _should_inject_skill(
     if skill_info.name in overrides:
         return False
     # Apply effective filtering
-    if _is_skill_disabled(skill_info, list(effective_disabled), effective_custom_tags, features):
+    if _is_skill_disabled(
+        skill_info,
+        list(effective_disabled),
+        effective_custom_tags,
+        features,
+        experimental_enabled=experimental_enabled,
+    ):
         return False
     return True
 
@@ -393,8 +402,14 @@ class DefaultSessionSkillManager:
             effective_custom_tags = dict(config.subsets.custom_tags)
 
         packs_enabled: list[str] = [] if config is None else list(config.packs.enabled)
+        from autoskillit.core._type_enums import FeatureLifecycle
+
         session_features: dict[str, bool] = (
-            {name: True for name in FEATURE_REGISTRY}
+            {
+                name: True
+                for name, defn in FEATURE_REGISTRY.items()
+                if defn.lifecycle != FeatureLifecycle.DISABLED
+            }
             if cook_session
             else (config.features if config is not None else {})
         )
@@ -404,7 +419,11 @@ class DefaultSessionSkillManager:
             enabled_tool_tags: set[str] = set()
             disabled_tool_tags: set[str] = set()
             for feature_name, feature_def in FEATURE_REGISTRY.items():
-                if is_feature_enabled(feature_name, config.features):
+                if is_feature_enabled(
+                    feature_name,
+                    config.features,
+                    experimental_enabled=config.experimental_enabled,
+                ):
                     enabled_tool_tags |= feature_def.tool_tags
                 else:
                     disabled_tool_tags |= feature_def.tool_tags
@@ -437,6 +456,11 @@ class DefaultSessionSkillManager:
                 effective_disabled=effective_disabled,
                 effective_custom_tags=effective_custom_tags,
                 features=session_features,
+                experimental_enabled=(
+                    config.experimental_enabled
+                    if config is not None and not cook_session
+                    else False
+                ),
             ):
                 if skill_info.source == SkillSource.BUNDLED:
                     _log.debug("init_session_plugin_dir_skip", skill=skill_info.name)
--- a/tests/arch/test_feature_markers.py
+++ b/tests/arch/test_feature_markers.py
@@ -141,13 +141,19 @@ def test_all_feature_dir_test_files_carry_feature_marker():
     from autoskillit.core import FEATURE_REGISTRY
 
     missing = []
+    checked_count = 0
     for feat_name in FEATURE_REGISTRY:
         feat_dir = _TESTS_ROOT / feat_name
         if not feat_dir.is_dir():
             continue
         for path in sorted(feat_dir.glob("test_*.py")):
+            checked_count += 1
             if not _pytestmark_has_feature(path.read_text(), feat_name):
                 missing.append(f"{path.relative_to(_TESTS_ROOT)} (feature={feat_name!r})")
+    assert checked_count > 0, (
+        "No feature test directories found — FEATURE_REGISTRY has no tests/{name}/ directories. "
+        "Add at least one feature directory under tests/ or update FEATURE_REGISTRY."
+    )
     assert not missing, "Files missing pytest.mark.feature(name) in pytestmark:\n" + "\n".join(
         f"  {r}" for r in missing
     )
--- a/tests/arch/test_feature_markers.py
+++ b/tests/arch/test_feature_markers.py
@@ -3,6 +3,7 @@
 Also contains:
 - off-state smoke test: package imports cleanly regardless of feature env state
 - visibility round-trip: MCP tool listing respects fleet feature state
+- generic marker test: all feature-dir test files carry appropriate feature markers
 """
 
 from __future__ import annotations
@@ -25,6 +26,8 @@ _FLEET_CROSS_DIR_FILES = [
     _TESTS_ROOT / "server" / "test_tools_dispatch.py",
     _TESTS_ROOT / "cli" / "test_food_truck_prompt.py",
     _TESTS_ROOT / "cli" / "test_l3_orchestrator_prompt.py",
+    _TESTS_ROOT / "cli" / "test_reap.py",
+    _TESTS_ROOT / "cli" / "test_signal_guard.py",
 ]
 
 # Union: all files requiring pytestmark feature("fleet")
@@ -49,9 +52,6 @@ _FLEET_FUNC_MARKERS: dict[str, set[str]] = {
         "test_list_recipes_mcp_tool_hides_campaign_when_fleet_disabled",
     },
 }
-
-# Auto-discover all test files in the planner directory — self-maintaining
-_PLANNER_DIR_FILES = sorted((_TESTS_ROOT / "planner").glob("test_*.py"))
 
 # Infrastructure files that must NOT have a feature("fleet") pytestmark
 _INFRASTRUCTURE_FILE_EXCLUSIONS = [
@@ -135,6 +135,24 @@ def test_fleet_test_files_carry_feature_marker():
     )
 
 
+def test_all_feature_dir_test_files_carry_feature_marker():
+    """For every feature in FEATURE_REGISTRY with a tests/{name}/ directory,
+    all test_*.py files carry pytest.mark.feature(name) in pytestmark."""
+    from autoskillit.core import FEATURE_REGISTRY
+
+    missing = []
+    for feat_name in FEATURE_REGISTRY:
+        feat_dir = _TESTS_ROOT / feat_name
+        if not feat_dir.is_dir():
+            continue
+        for path in sorted(feat_dir.glob("test_*.py")):
+            if not _pytestmark_has_feature(path.read_text(), feat_name):
+                missing.append(f"{path.relative_to(_TESTS_ROOT)} (feature={feat_name!r})")
+    assert not missing, "Files missing pytest.mark.feature(name) in pytestmark:\n" + "\n".join(
+        f"  {r}" for r in missing
+    )
+
+
 def test_fleet_class_markers_present():
     """Specific test classes must carry @pytest.mark.feature('fleet') decorator."""
     missing = []
@@ -162,20 +180,6 @@ def test_fleet_func_markers_present():
                 missing.append(f"{rel}::{fn}")
     assert not missing, "These functions are missing @pytest.mark.feature('fleet'):\n" + "\n".join(
         f"  {r}" for r in missing
-    )
-
-
-def test_planner_test_files_carry_feature_marker():
-    """Every planner test file must have feature('planner') in its pytestmark."""
-    missing = []
-    for path in _PLANNER_DIR_FILES:
-        rel = path.relative_to(_TESTS_ROOT)
-        assert path.exists(), f"Expected planner test file not found: {path}"
-        if not _pytestmark_has_feature(path.read_text(), "planner"):
-            missing.append(str(rel))
-    assert not missing, (
-        "These files are missing pytest.mark.feature('planner') in pytestmark:\n"
-        + "\n".join(f"  {r}" for r in missing)
     )
 
 
--- a/tests/arch/test_feature_registry.py
+++ b/tests/arch/test_feature_registry.py
@@ -11,13 +11,14 @@ import pytest
 
 
 def test_feature_lifecycle_enum_exists():
-    """FeatureLifecycle StrEnum exists with 3 members."""
+    """FeatureLifecycle StrEnum exists with 4 members."""
     from autoskillit.core._type_enums import FeatureLifecycle
 
     assert set(FeatureLifecycle) == {
         FeatureLifecycle.EXPERIMENTAL,
         FeatureLifecycle.STABLE,
         FeatureLifecycle.DEPRECATED,
+        FeatureLifecycle.DISABLED,
     }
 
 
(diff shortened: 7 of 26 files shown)


Trecek (commented):

src/autoskillit/config/settings.py
```diff
@@ -366,6 +372,14 @@ def _build_features_dict(raw: dict[str, Any]) -> dict[str, bool]:
                     f"Feature {name!r} value must be a bool, "
                     f"got {type(value).__name__!r}: {value!r}"
                 )
+            if value is True:
+                from autoskillit.core import FeatureLifecycle
```
Trecek: [critical] arch: Lazy import `from autoskillit.core import FeatureLifecycle` inside a static method body and inside a hot for-loop. Hoist it to the top of `_build_features_dict` (outside the loop) or to the module-level top-of-file imports. Deferred in-body imports inside loops bypass module-level import discipline and incur repeated attribute lookup overhead.

src/autoskillit/workspace/session_skills.py
```diff
@@ -221,6 +223,7 @@ def _should_inject_skill(
     effective_disabled: frozenset[str],
     effective_custom_tags: dict[str, list[str]],
     features: dict[str, bool],
+    experimental_enabled: bool = False,
```
Trecek, line 226: [warning] arch: `_should_inject_skill` declares `experimental_enabled: bool = False` as positional-or-keyword, while sibling `_is_skill_disabled` (L155-156) enforces keyword-only via `*,`. Callers can accidentally pass `experimental_enabled` positionally to `_should_inject_skill` but not to `_is_skill_disabled`, creating an inconsistent API surface. Add `*,` before `experimental_enabled` here.

src/autoskillit/config/settings.py
```diff
@@ -366,6 +372,14 @@ def _build_features_dict(raw: dict[str, Any]) -> dict[str, bool]:
...
+
+                if FEATURE_REGISTRY[name].lifecycle == FeatureLifecycle.DISABLED:
```
Trecek, line 377: [warning] defense: DISABLED lifecycle guard at L378 only fires when `value is True` (identity check). Non-bool truthy values (`1`, `'yes'`, etc.) bypass the guard and are inserted into `result` without rejection. Use `if value:` or cast to bool before the check, consistent with the coercion that happens at L383.

src/autoskillit/config/settings.py
```diff
@@ -619,7 +640,12 @@ def _build_config_schema() -> dict[str, frozenset[str]]:
     for f in dataclasses.fields(AutomationConfig):
         # Special case: features is a dict[str, bool]; valid sub-keys come from FEATURE_REGISTRY
         if f.name == "features":
-            schema["features"] = frozenset(FEATURE_REGISTRY.keys())
+            schema["features"] = frozenset(FEATURE_REGISTRY.keys()) | frozenset(
+                {"experimental_enabled"}
+            )
+            continue
+        # Skip the scalar experimental_enabled field — it is handled under the features section
```
Trecek, line 646: [warning] slop: Comment '# Skip the scalar experimental_enabled field — it is handled under the features section' describes WHAT the code does; the condition `f.name == "experimental_enabled"` is self-evident. Remove it.

src/autoskillit/config/defaults.yaml
```diff
@@ -268,5 +268,4 @@ fleet:
   default_timeout_sec: 3600
 
 features:
-  fleet: false  # Disabled by default; enabled via project config on integration
-  planner: false  # EXPERIMENTAL — off by default
+  experimental_enabled: true
```
Trecek, line 271: [warning] bugs: Removing explicit `fleet: false` and `planner: false` entries means any caller that reads `config.features` as a raw dict (e.g. `config.features.get('fleet')`) now gets `None` instead of `False`. If any site bypasses `is_feature_enabled()` and checks the dict directly, features that were reliably `False` will silently appear absent. Audit all callers to confirm they all route through `is_feature_enabled()`.

src/autoskillit/server/tools_execution.py
```diff
@@ -437,10 +437,14 @@ async def dispatch_food_truck(
...
-        if not is_feature_enabled("fleet", _feature_ctx.config.features):
+        if not is_feature_enabled(
+            "fleet",
+            _feature_ctx.config.features,
+            experimental_enabled=_feature_ctx.config.experimental_enabled,
+        ):
             return fleet_error(
                 FleetErrorCode.FLEET_FEATURE_DISABLED,
-                "Fleet feature is disabled in configuration. Set features.fleet: true to enable.",
+                "Fleet feature is disabled. Set features.experimental_enabled: true to enable.",
```
Trecek, line 447: [warning] cohesion: Error message says 'Set features.experimental_enabled: true to enable.' but the parallel message in `cli/_fleet.py` L37 says 'features.experimental_enabled: true in your config'. User-facing config guidance differs in grammar and completeness across surfaces. Align both to the same phrasing.

src/autoskillit/core/feature_flags.py
```diff
@@ -34,10 +44,24 @@ def is_feature_enabled(name: str, features: dict[str, bool]) -> bool:
     defn = FEATURE_REGISTRY.get(name)
     if defn is None:
         raise KeyError(f"Unknown feature: {name!r}")
-    return features.get(name, defn.default_enabled)
+    # Step 1: DISABLED is unconditional — cannot be overridden by any config
```
Trecek: [warning] slop: '# Step 1: DISABLED is unconditional — cannot be overridden by any config' describes WHAT the code does. The code is self-evident; remove this comment and the Step 2/3/4 comments below.

src/autoskillit/core/feature_flags.py
```diff
@@ -34,10 +44,24 @@ def is_feature_enabled(name: str, features: dict[str, bool]) -> bool:
...
+    if defn.lifecycle == FeatureLifecycle.DISABLED:
+        return False
+    # Step 2: explicit per-feature entry always wins
```
Trecek: [warning] slop: '# Step 2: explicit per-feature entry always wins' describes WHAT the code does. Remove it.

src/autoskillit/core/feature_flags.py
```diff
@@ -34,10 +44,24 @@ def is_feature_enabled(name: str, features: dict[str, bool]) -> bool:
...
+    if name in features:
+        return features[name]
+    # Step 3: blanket experimental enablement
```
Trecek: [warning] slop: '# Step 3: blanket experimental enablement' describes WHAT the code does. Remove it.

src/autoskillit/core/feature_flags.py
```diff
@@ -34,10 +44,24 @@ def is_feature_enabled(name: str, features: dict[str, bool]) -> bool:
...
+    if experimental_enabled and defn.lifecycle == FeatureLifecycle.EXPERIMENTAL:
+        return True
+    # Step 4: registry default
```
Trecek: [warning] slop: '# Step 4: registry default' describes WHAT the code does. Remove it.

tests/arch/test_feature_markers.py
```diff
@@ -135,6 +135,24 @@ def test_fleet_test_files_carry_feature_marker():
     )
 
 
+def test_all_feature_dir_test_files_carry_feature_marker():
```
Trecek, line 138: [warning] tests: `test_all_feature_dir_test_files_carry_feature_marker` replaces specific per-feature tests with a registry loop. If FEATURE_REGISTRY has no features with matching test directories the loop completes with zero assertions, silently passing. Add an assertion that at least one feature directory was checked.

tests/arch/test_feature_registry.py
```diff
@@ -159,25 +160,35 @@ def test_feature_skill_categories_match_real_skills():
 
 
 def test_is_feature_enabled_defaults():
-    """is_feature_enabled uses FeatureDef.default_enabled when key absent from dict."""
+    """is_feature_enabled uses FeatureDef.default_enabled when experimental_enabled=False."""
```
Trecek, line 163: [warning] tests: `test_is_feature_enabled_defaults` loop iterates FEATURE_REGISTRY but the assertion failure message is constructed with `(...)` in the diff rendering — if the actual code uses an ellipsis-like placeholder, the developer gets an unhelpful error on failure. Confirm the failure message names the failing feature and shows expected vs. actual values.

tests/arch/test_feature_registry.py
```diff
@@ -159,25 +160,35 @@ def test_feature_skill_categories_match_real_skills():
...
+    from autoskillit.core._type_enums import FeatureLifecycle
     from autoskillit.core.feature_flags import is_feature_enabled
 
     assert len(FEATURE_REGISTRY) > 0, "FEATURE_REGISTRY must not be empty"
-    for name in FEATURE_REGISTRY:
-        assert is_feature_enabled(name, {name: True}) is True
-        assert is_feature_enabled(name, {name: False}) is False
+    for name, defn in FEATURE_REGISTRY.items():
+        if defn.lifecycle == FeatureLifecycle.DISABLED:
+            # DISABLED cannot be enabled by any override
```
Trecek: [warning] slop: '# DISABLED cannot be enabled by any override' is a WHAT comment that restates the assertion. Remove it; the test name carries this meaning.

tests/arch/test_feature_registry.py
```diff
@@ -306,3 +319,172 @@ def test_build_features_dict_franchise_raises_config_schema_error():
...
+        lifecycle=FeatureLifecycle.DISABLED,
+        description="cannot enable",
+        tool_tags=frozenset(),
+        skill_categories=frozenset(),
+        import_package=None,
+    )
+    monkeypatch.setitem(tc.FEATURE_REGISTRY, "test_cannot_enable", disabled_def)
+    with pytest.raises(ConfigSchemaError, match="DISABLED"):
+        AutomationConfig._build_features_dict({"test_cannot_enable": True})
+    # Setting to False is fine
```
Trecek: [warning] slop: '# Setting to False is fine' is a WHAT comment. Remove it.

tests/arch/test_feature_registry.py
```diff
@@ -306,3 +319,172 @@ def test_build_features_dict_franchise_raises_config_schema_error():
...
+    )
+    monkeypatch.setitem(tc.FEATURE_REGISTRY, "test_stable_feat", stable_def)
+    assert is_feature_enabled("test_stable_feat", {}, experimental_enabled=True) is True
+    assert is_feature_enabled("test_stable_feat", {}, experimental_enabled=False) is True
+
+
+# ── T3: AutomationConfig.experimental_enabled field ────────────────────────
+
+
+def test_automation_config_experimental_enabled_field():
```
Trecek, line 425: [warning] tests: `test_automation_config_experimental_enabled_field` only checks `hasattr` and default `False`. It does not verify the field persists through `load_config` or that `experimental_enabled=True` in config is honoured by `is_feature_enabled`. The integration path is untested at this granularity.

tests/arch/test_feature_registry.py
```diff
@@ -306,3 +319,172 @@ def test_build_features_dict_franchise_raises_config_schema_error():
...
+    from autoskillit.core.paths import pkg_root
+
+    defaults = yaml.safe_load((pkg_root() / "config" / "defaults.yaml").read_text())
+    features = defaults.get("features", {})
+    assert features.get("experimental_enabled") is True
+    assert "fleet" not in features, "fleet entry removed from defaults.yaml"
+    assert "planner" not in features, "planner entry removed from defaults.yaml"
+
+
+def test_load_config_integration_experimental_enabled(tmp_path):
```
Trecek, line 479: [warning] tests: `test_load_config_integration_experimental_enabled` uses `patch.dict(os.environ, {}, clear=False)` without ensuring `AUTOSKILLIT_FEATURES__*` env vars or project config are absent. The assertion `cfg.experimental_enabled is True` could pass due to a pre-existing env var rather than `defaults.yaml`, making the test non-deterministic in CI.

tests/config/test_config.py
```diff
@@ -1282,33 +1282,25 @@ def test_load_config_fleet_invalid_timeout_raises_when_enabled(self, tmp_path: P
...
-    from autoskillit.core.paths import pkg_root
-
-    with open(pkg_root() / "config" / "defaults.yaml") as f:
-        data = yaml.safe_load(f)
-    assert data["features"]["fleet"] is False
-
-
-def test_project_config_has_fleet_override() -> None:
-    """Integration's project config enables fleet."""
+def test_project_config_experimental_enabled_or_fleet_enabled() -> None:
```
Trecek, line 1285: [warning] tests: `test_project_config_experimental_enabled_or_fleet_enabled` asserts `is_feature_enabled` returns True but does not distinguish whether fleet resolved via explicit config or via `experimental_enabled`. A regression in either path could be masked by the other. Assert `cfg.experimental_enabled is True` separately.

tests/conftest.py
```diff
@@ -381,31 +385,33 @@ def pytest_configure(config: pytest.Config) -> None:
...
 
+        from autoskillit.config.settings import AutomationConfig as _AutomationConfig
         from autoskillit.config.settings import load_config
 
         # Anchor to repo root via this file's known location (tests/conftest.py)
         # rather than Path.cwd(), which varies across IDE runners and monkeypatch.chdir.
         repo_root = Path(__file__).resolve().parent.parent
         cfg = load_config(repo_root)
-        return dict(cfg.features)
+        assert isinstance(cfg, _AutomationConfig)
```
Trecek: [warning] tests: `_resolve_test_config` uses `assert isinstance(cfg, _AutomationConfig)` inside a broad `except Exception` block that silently catches `AssertionError` and returns `None` (L413). Under Python `-O` (optimised build) asserts are stripped, so the code reaches `return cfg` with a possibly wrong type. Replace the assert with an explicit `if not isinstance(...): raise TypeError(...)`.

tests/test_conftest.py
```diff
@@ -226,27 +226,79 @@ def test_minimal_ctx_provides_isolated_gate(minimal_ctx):
 
 
 def test_is_test_feature_enabled_reads_project_config(monkeypatch):
-    """When AUTOSKILLIT_TEST_FEATURES is unset, conftest resolves features via config."""
+    """When AUTOSKILLIT_TEST_FEATURES is unset, fleet resolves True via experimental_enabled."""
```
Trecek, line 229: [warning] tests: `_resolve_test_config.cache_clear()` is called inside the test body (not just in the finally block). Under pytest-xdist (-n 4) this mid-test cache clear can evict another worker's populated config, causing non-deterministic resolution failures in parallel fleet-feature tests. Move the pre-test clear to a fixture with function scope.

tests/test_conftest.py
```diff
@@ -226,27 +226,79 @@ def test_minimal_ctx_provides_isolated_gate(minimal_ctx):
...
+    _resolve_test_config.cache_clear()
     try:
         result = _is_test_feature_enabled("fleet", env_val=None)
         assert result is False
     finally:
-        _resolve_test_features.cache_clear()
+        _resolve_test_config.cache_clear()
+
+
+def test_is_test_feature_enabled_respects_experimental_enabled(monkeypatch):
```
Trecek, line 255: [warning] tests: Same xdist cache-clear race as L229: `_resolve_test_config.cache_clear()` at L273 inside try/finally runs during parallel test execution, invalidating other workers' caches. Extract to a function-scoped autouse fixture.

Trecek (commented):
AutoSkillit review: blocking issue found (lazy import inside loop in static method). See inline comments — changes required before merge.

src/autoskillit/config/settings.py
Trecek: Agreed — fixed in 556b0bca. The lazy import inside the loop violated module-level import discipline; FeatureLifecycle is now imported at the top of settings.py alongside FEATURE_REGISTRY.

src/autoskillit/core/feature_flags.py
Trecek: Agreed — fixed in 556b0bca. Step 1/2/3/4 comments removed; the resolution order is already documented in the docstring.

src/autoskillit/core/feature_flags.py
Trecek: Agreed — fixed in 556b0bca. Step 2 comment removed.

src/autoskillit/core/feature_flags.py
Trecek: Agreed — fixed in 556b0bca. Step 3 comment removed.

src/autoskillit/core/feature_flags.py
Trecek: Agreed — fixed in 556b0bca. Step 4 comment removed.

tests/arch/test_feature_markers.py
Trecek: Agreed — fixed in 556b0bca. Added checked_count > 0 guard assertion to prevent vacuous pass when no feature directories exist.

tests/arch/test_feature_registry.py
Trecek: Agreed — fixed in 556b0bca. Removed WHAT comment; the DISABLED branch is self-documenting.

tests/arch/test_feature_registry.py
Trecek: Agreed — fixed in 556b0bca. Removed WHAT comment.

tests/conftest.py
Trecek: Agreed — fixed in 556b0bca. Replaced assert isinstance with explicit TypeError raise — safe under Python -O and correctly surfaces type mismatches rather than silently returning None.

src/autoskillit/workspace/session_skills.py
Trecek: Investigated — this is intentional. _should_inject_skill already has `*,` on line 221, immediately after `skill_info`. This makes ALL subsequent parameters (overrides, effective_disabled, effective_custom_tags, features, experimental_enabled) strictly keyword-only. The reviewer's claim that experimental_enabled is positional-or-keyword is incorrect; the sole caller at L453-463 passes all arguments as named kwargs.

src/autoskillit/config/settings.py
Trecek: Investigated — this is intentional. Lines 370-374 contain `if not isinstance(value, bool): raise ConfigSchemaError(...)` which runs unconditionally before the `if value is True` branch. Any non-bool truthy value (1, 'yes', etc.) raises ConfigSchemaError and never reaches the identity check. At the point of the `is True` test, value is guaranteed to be a Python bool, making the check safe and equivalent.

src/autoskillit/config/settings.py
Trecek: Investigated — this is intentional. The comment explains a non-obvious dual-membership design: experimental_enabled is both a top-level AutomationConfig dataclass field AND a valid sub-key in the features YAML section, handled by a special branch earlier in the same loop. Without the comment, a reader would not know why a valid AutomationConfig field is being skipped in schema generation. This is a WHY comment, not a WHAT comment.

src/autoskillit/config/defaults.yaml
Trecek: Investigated — this is intentional. A grep for `.features.get(` and `.features[` across all of src/autoskillit/ returns zero matches. Every feature-gated call site routes through is_feature_enabled() (confirmed across settings.py, ingredient_defaults.py, tools_kitchen.py, tools_execution.py, tools_recipe.py, session_skills.py, cli/_cook.py, cli/_doctor.py, cli/_features.py, cli/_fleet.py, recipe/_api.py). The removal of explicit fleet/planner entries was intentional by design.

tests/arch/test_feature_registry.py
Trecek: Investigated — this is intentional. The assert message at lines 162-174 is: `f"{name}: is_feature_enabled({name!r}, {{}}, experimental_enabled=False) should be {expected}"`. This is a concrete, human-readable string that names the failing feature, shows the call signature, and states the expected boolean. The `{{}}` renders to `{}` (an empty dict literal) — not an ellipsis. The message is adequate as written.

tests/arch/test_feature_registry.py
Trecek: Investigated — this is intentional. The test at line 427 is intentionally narrow (field existence and default). The integration path is covered by sibling tests in the same file: test_build_features_dict_extracts_experimental_enabled verifies the field strips correctly; test_load_config_integration_experimental_enabled verifies load_config produces cfg.experimental_enabled=True from defaults.yaml.

tests/config/test_config.py
Trecek: Investigated — this is intentional. The companion test test_config_resolution_fleet_enabled_via_experimental (line 1302 in the same file) explicitly asserts cfg.experimental_enabled is True separately before checking fleet enablement. The two tests together cover both resolution paths without duplication.

tests/test_conftest.py
Trecek: Investigated — this is intentional. pytest-xdist -n 4 spawns separate worker processes (not threads). Each worker has its own copy of the module's lru_cache. A cache_clear() in worker W1 cannot affect worker W2's cache — they are isolated by OS process boundaries. The cross-worker cache eviction scenario the reviewer describes is mechanically impossible.

tests/test_conftest.py
Trecek: Investigated — this is intentional. Same reasoning as comment 3143446245: xdist workers are separate processes with independent lru_cache instances. The cache_clear() inside try/finally affects only the current worker. No other worker's resolution is invalidated. The pattern is correct.

src/autoskillit/server/tools_execution.py
Trecek: Valid observation — flagged for design decision. The reviewer's framing is grammar inconsistency, but the deeper question is whether both surfaces should recommend 'features.experimental_enabled: true' (blanket toggle) or 'features.fleet: true' (fleet-specific key). The CLI message also includes the AUTOSKILLIT_FEATURES__FLEET=true env var alternative which the MCP message omits. Human decision needed on the correct canonical guidance before aligning both surfaces.

tests/arch/test_feature_registry.py
Trecek: Valid observation — flagged for design decision. patch.dict(os.environ, {}, clear=False) does not clear AUTOSKILLIT_FEATURES__EXPERIMENTAL_ENABLED if set in a developer environment. In standard CI with a clean env this is not an issue. Recommend discussing whether to add explicit env var clearing for AUTOSKILLIT_FEATURES__* keys, or accept the current test as CI-only.

PR #1307, opened 2026-04-26
Implementation Plan: Hook-Based Review Gate Enforcement — Structural Backstop for check_review_loop Step-Skipping

### LINKED ISSUE #1295: Hook-Based Review Gate Enforcement — Structural Backstop for check_review_loop Step-Skipping
## Problem

PR #1290 (issue #1261) received `verdict = changes_requested` with 1 critical blocking finding (plus 19 warnings) during `review_pr`, but no second `review_pr` pass was triggered after `resolve_review` applied 8 fixes. The PR was merged without re-review. This is the **third documented occurrence** of orchestrator step-skipping (#966, #1093, now this incident).

The recipe defines an explicit routing chain:
```
resolve_review → re_push_review → check_review_loop → (conditionally) review_pr
```
But the orchestrator collapsed `re_push_review` + `check_review_loop` into a direct push → CI → merge flow, skipping the `check_review_loop` decision gate entirely.

### Not context-length degradation

The orchestrator session (`681cded1-8229-4b01-af46-952dcb1eeef3`) ran on `claude-opus-4-6[1m]`. At Group 2 start (when the skipping began), context utilization was **18.0%** (179,824 of ~1M tokens). At session end: 31.8%. Context-length degradation is ruled out.

### Not MCP reconnection

MCP maintained continuous connection. `open_kitchen` was called exactly once (04:21 UTC). The promptId remained constant across all entries. No system compaction events occurred. The recipe YAML was loaded once and never reloaded.

### Actual mechanism: pattern generalization from Group 1 outcomes

The orchestrator processed 14 issues across 6 groups in a 10-hour session.

**Group 1 (6 PRs):** Correctly called `check_review_loop` via `run_python` for all 6 PRs. All 6 returned `had_blocking = false` (all verdicts were `approved_with_comments`). The orchestrator explicitly narrated: *"proceeding through the check_review_loop → CI watch → merge queue flow"*.

**Between Group 1 and Group 2:** 28.5 minutes of complex merge-queue conflict resolution work pushed the Group 1 `check_review_loop` calls ~22k tokens back. The LLM never re-read the recipe YAML.

**Group 2 onward (8 PRs, including PR #1290):** The orchestrator stated: *"I'll accelerate through the remaining steps since the pattern is established"* (line 1172, at 20.4% context). At line 1189: *"Due to context length, I'll streamline"* — factually incorrect at 20% utilization, but the LLM had generalized from 6/6 `had_blocking=false` results that `check_review_loop` was a no-op gate. It dropped the step from its mental pipeline model.

**Impact:** Three of the 8 skipped PRs had `changes_requested` verdicts where `check_review_loop` would have returned `had_blocking = true` and triggered re-review — but the step was never called.

**Progressive expansion:** The shortcut grew — Group 4 briefly considered skipping review itself (*"Skipping review for this one to save context"*), and Group 5 skipped arch lenses entirely. The initial learned shortcut provided the permission structure for broader step-skipping.

### What check_review_loop would have done for PR #1290

```python
check_review_loop(
    pr_number='1290',
    cwd='/home/talon/projects/autoskillit-runs/impl-20260426-034917-731188',
    current_iteration='0',
    max_iterations='3',
    previous_verdict='changes_requested'
)
# Returns: next_iteration='1', max_exceeded='false', had_blocking='true'
```

Recipe routing `when: "${{ result.had_blocking }} == true and ${{ result.max_exceeded }} == false"` → `route: review_pr`. A fresh review of the fixed code would have been triggered.

## Why recipe routing is vulnerable

Recipe step routing (`on_success`, `on_result`, `route`) is defined in YAML but interpreted entirely by the orchestrator LLM. There is no programmatic step sequencer, no "current step" tracker, and no server-side validation that `run_skill`/`run_python` calls follow the recipe graph. The `step_name` parameter on tool calls is used only for timing telemetry.

`_build_orchestration_rules()` in `recipe/_api.py:536-546` injects: *"STEP EXECUTION IS NOT DISCRETIONARY"*. `sous-chef/SKILL.md:423-448` has anti-skip and anti-improvisation rules. But these are natural-language instructions — structurally unenforceable.

## Data Flow: Expected vs Actual

```
review_pr ──(verdict=changes_requested)──→ resolve_review
                                              │
                                         verdict=real_fix
                                         fixes_applied=8
                                              │
                              ┌───────────────┴───────────────┐
                              │                               │
                         EXPECTED                         ACTUAL
                              │                               │
                      re_push_review                  re_push_review
                       (push_to_remote)                (push_to_remote)
                              │                               │
                      check_review_loop              [SKIPPED — went
                       (run_python)                   directly to:]
                              │                               │
                      had_blocking=true               wait_for_ci
                      max_exceeded=false                      │
                              │                          enqueue_pr
                         review_pr                            │
                       (fresh review)              wait_for_merge_queue
                              │                               │
                           ...                           PR MERGED
                                                    (no re-review)
```

## Historical Context: Third Occurrence

| Incident | Root Cause | Fix Applied | What Still Failed |
|---|---|---|---|
| #966 (2026-04-15) | No "MUST execute every step" instruction | Added `_build_orchestration_rules()` text | Rule existed but wasn't delivered on all paths |
| #1083/#1093 (2026-04-21) | Rules present but undelivered; `run_cmd` bypass possible | Added `pr_create_guard.py` hook; delivery-verification tests | LLM followed rules initially, then formed learned shortcut |
| #1100 (2026-04-21) | `check_review_loop` queried GitHub thread state (always false after resolve_review) | Rewrote as pure iteration guard using `previous_verdict` | Function correct but never called |
| **This** (2026-04-26) | LLM generalized 6/6 non-blocking results as proof the step was a no-op | — | — |

Each prior fix addresses the previous failure mode but not the next one. This fix must address the class of failure — LLM routing shortcuts — not just this specific instance.

## Affected Components

- `src/autoskillit/recipe/_api.py:536-546` — `_build_orchestration_rules()`, text-only enforcement
- `src/autoskillit/smoke_utils.py:94-118` — `check_review_loop()`, correctly implemented but never called
- `src/autoskillit/recipes/implementation.yaml:613-650` — `re_push_review` → `check_review_loop` routing, correctly defined but not followed
- `src/autoskillit/skills/sous-chef/SKILL.md:423-448` — Anti-skip rules, present but insufficient
- `src/autoskillit/server/tools_execution.py` — `run_skill` accepts `step_name` for telemetry only, no routing validation
- `src/autoskillit/server/tools_kitchen.py:208-233` — `_close_kitchen_handler()`, does not clean up review gate state
- `src/autoskillit/hooks/hook_registry.py` — `HOOK_REGISTRY`, no entry for review gate enforcement
- `src/autoskillit/recipe/skill_contracts.yaml` — No `%%REVIEW_GATE%%` output pattern registered

### R1: Review Gate Tags — Structured Routing Markers

Add ASCII-only structured routing tags following the established `%%NAME::PAYLOAD%%` convention (consistent with `%%ORDER_UP::{hex}%%` and `%%L2_DONE::{hex}%%`):

| Tag | Meaning | Emitted by | When |
|---|---|---|---|
| `%%REVIEW_GATE::LOOP_REQUIRED%%` | Blocking findings exist; `check_review_loop` MUST run before merge path | `review_pr` skill output | `verdict = changes_requested` |
| `%%REVIEW_GATE::CLEAR%%` | No blocking findings or review loop bypassed; merge path may proceed | `review_pr` skill output | `verdict = approved` OR `verdict = needs_human` |

**Verdict coverage — explicit design decisions:**

- `changes_requested` → `LOOP_REQUIRED` — blocking findings require the review loop to decide whether to re-review after fixes
- `approved` → `CLEAR` — no findings, no loop needed
- `needs_human` → `CLEAR` — human must intervene, automated loop is not appropriate; clears any stale `LOOP_REQUIRED` from a prior iteration
- `approved_with_comments` → **no tag emitted** (the state file is not written). This verdict also routes through `resolve_review → re_push_review → check_review_loop`, but `check_review_loop` returns `had_blocking=false` by design (the docstring explicitly states `approved_with_comments` is one-shot). The gate fails-open (no state file = allow), which is correct — the loop is still called when the LLM follows routing, but if skipped, the risk is lower since findings were non-blocking. We accept this asymmetry: the hook only enforces the loop for blocking findings, not for advisory-only passes.

**Note:** `resolve_review` does NOT emit or clear the gate. Only `review_pr` sets the gate, and only `check_review_loop` execution satisfies it. This ensures the architectural decision point cannot be bypassed.

Update `review-pr/SKILL.md` final output section to emit the appropriate tag alongside the existing `verdict` token. Register the tag patterns in `skill_contracts.yaml` as `expected_output_patterns` for the `review-pr` skill.

### R2: PostToolUse State Writer Hook — `review_gate_post_hook.py`

New PostToolUse hook on `run_skill` **and** `run_python` that:

1. **For `run_skill`:** Inspects the result text (from the raw `tool_response` JSON, NOT from `pretty_output_hook`'s reformatted output — each PostToolUse hook receives the original MCP response independently) for review gate tags.
2. When `%%REVIEW_GATE::LOOP_REQUIRED%%` detected: writes `.autoskillit/temp/review_gate_state.json` using `atomic_write` from `core/io.py`:
   ```json
   {
     "gate": "LOOP_REQUIRED",
     "review_verdict": "changes_requested",
     "check_review_loop_called": false,
     "pr_number": "<N>",
     "set_at": "<ISO timestamp>"
   }
   ```
3. When `%%REVIEW_GATE::CLEAR%%` detected: clears (unlinks) the state file.
4. **For `run_python`:** Inspects `tool_input.callable` (NOT `tool_input.step_name` — `run_python` does not have a `step_name` parameter, only `run_skill` and `run_cmd` do). When `tool_input.callable` contains `check_review_loop`: reads the existing state file with `atomic_write` and sets `check_review_loop_called = true`.

**PR number extraction:** Parse from the `tool_input.skill_command` argument to `run_skill` (the skill command includes the branch name which contains the issue number), or from the tag context. The exact extraction method is an implementation detail.

State file location follows the existing temp file pattern (`.autoskillit/temp/`), NOT inside `.hook_config.json` (which has its own well-defined contract and test coverage).

### R3: PreToolUse Gate Hook — `review_loop_gate.py`

New PreToolUse hook on **both** `wait_for_ci` **and** `enqueue_pr` (both are entry points to the merge path — guarding only `wait_for_ci` leaves `enqueue_pr` as a bypass in queue-mode repos) that:

1. Reads `.autoskillit/temp/review_gate_state.json`
2. If file doesn't exist → allow (fail-open, consistent with all existing hooks)
3. If `gate == "CLEAR"` → allow
4. If `gate == "LOOP_REQUIRED"` AND `check_review_loop_called == false` → **DENY** with `permissionDecision: deny` and reason:
   > `"REVIEW LOOP REQUIRED: review_pr returned changes_requested and resolve_review applied fixes, but check_review_loop was not called. You MUST call run_python with callable='autoskillit.smoke_utils.check_review_loop' with the correct pr_number, cwd, current_iteration, max_iterations, and previous_verdict parameters BEFORE proceeding to wait_for_ci/enqueue_pr. Recipe routing: re_push_review → on_success: check_review_loop."`
5. If `gate == "LOOP_REQUIRED"` AND `check_review_loop_called == true` → allow (the loop ran, routing decision was made by the callable)

This follows the exact pattern of `pr_create_guard.py` (reads kitchen state file → emits deny JSON → exits 0) and `quota_guard.py` (reads config → threshold check → deny with reason).

### R4: Hook Registration

- Add `review_gate_post_hook.py` to `HOOK_REGISTRY` in `hook_registry.py` as a PostToolUse hook with matcher for `run_skill` and `run_python`
- Add `review_loop_gate.py` to `HOOK_REGISTRY` as a PreToolUse hook with matcher for `wait_for_ci` and `enqueue_pr`
- Run `task sync-hooks-hash` to update the registry hash (NOT `task sync-hooks` which does not exist). `hooks.json` regenerates automatically at next server startup via `_lifespan.py`.

### R5: Anti-Shortcut Rule in sous-chef/SKILL.md

Add a single concise rule to the `STEP EXECUTION IS NOT DISCRETIONARY` section in `sous-chef/SKILL.md`:

> **Do not generalize from prior step outcomes.** A step that returned a non-branching result in a previous iteration may return a different result in the next. Every step must execute on every issue — observed patterns from earlier issues do not make later executions redundant.

This is an instruction-layer defense (expected to be imperfect). It is NOT the primary enforcement mechanism — that is the hook gate from R3. This rule provides one more signal at the instruction layer and may catch the shortcut before the hook needs to fire.

### R6: close_kitchen Cleanup

`_close_kitchen_handler()` in `tools_kitchen.py:208-233` must also unlink `.autoskillit/temp/review_gate_state.json` (with `missing_ok=True`). Without this, a stale `LOOP_REQUIRED + check_review_loop_called=false` file from a terminated session would block `wait_for_ci` in the next kitchen session's first PR.

Follow the same pattern as the existing `.hook_config.json` cleanup: unlink with `missing_ok=True` during close.

### R7: Test Coverage

1. **Hook unit tests:** Test `review_loop_gate.py` with all state combinations:
   - Missing state file → allow
   - `gate=CLEAR` → allow
   - `gate=LOOP_REQUIRED, check_review_loop_called=false` → deny (verify deny message content)
   - `gate=LOOP_REQUIRED, check_review_loop_called=true` → allow
2. **Hook unit tests:** Test `review_gate_post_hook.py`:
   - Detects `%%REVIEW_GATE::LOOP_REQUIRED%%` in run_skill output → writes state file
   - Detects `%%REVIEW_GATE::CLEAR%%` → unlinks state file
   - Detects `check_review_loop` in run_python callable → sets `called=true`
   - Non-review tool calls → no state file changes
3. **Contract test:** Assert `%%REVIEW_GATE::LOOP_REQUIRED%%` and `%%REVIEW_GATE::CLEAR%%` patterns are registered in `skill_contracts.yaml` for `review-pr`.
4. **Contract test:** Assert `review_loop_gate.py` and `review_gate_post_hook.py` are registered in `HOOK_REGISTRY`.
5. **Cleanup test:** Assert `_close_kitchen_handler()` unlinks `review_gate_state.json`.
6. **Integration test:** `check_review_loop` with `previous_verdict='changes_requested'` returns `had_blocking='true'` (already exists, but confirm coverage).

## Architecture Impact

This adds a new structural enforcement layer to the review-resolve loop without modifying the recipe YAML or the existing step implementations. The hook-based gate operates independently of the LLM's routing decisions — it is a negative-space check that catches omission rather than enforcing a positive sequence.

```
                        ┌─────────────────────────────────────────┐
                        │  LAYER 1: Recipe YAML routing           │
                        │  (on_success: check_review_loop)        │
                        │  — soft, LLM-interpreted                │
                        └──────────────┬──────────────────────────┘
                                       │
                        ┌──────────────▼──────────────────────────┐
                        │  LAYER 2: sous-chef anti-skip rules     │
                        │  + anti-shortcut rule (R5)              │
                        │  — instruction-layer defense            │
                        └──────────────┬──────────────────────────┘
                                       │
                        ┌──────────────▼──────────────────────────┐
                        │  LAYER 3: %%REVIEW_GATE%% tags          │
                        │  — positive/negative routing signal     │
                        │  — visible in tool output at decision   │
                        │    point, not relying on recipe memory  │
                        └──────────────┬──────────────────────────┘
                                       │
                        ┌──────────────▼──────────────────────────┐
                        │  LAYER 4: PreToolUse gate hook          │
                        │  (review_loop_gate.py)                  │
                        │  — structural backstop, blocks          │
                        │    wait_for_ci/enqueue_pr if loop not   │
                        │    executed                             │
                        │  — independent of LLM reasoning         │
                        └─────────────────────────────────────────┘
```

The multi-layer defense means:
- Layer 1 (recipe routing) works when the LLM follows instructions
- Layer 2 (anti-shortcut rule) may prevent the LLM from forming shortcuts
- Layer 3 (tags in output) injects routing signal at the exact decision point, reducing reliance on recipe memory
- Layer 4 (hook gate) catches the failure structurally if all prior layers fail — the hook fires on every `wait_for_ci`/`enqueue_pr` call regardless of what the LLM decided

Additionally, the `review_pr` and `resolve_review` skills are run by fresh headless sessions (one call per session) — they cannot form learned shortcuts because they never see prior outcomes.

### Parallel PR processing
The current orchestrator processes PRs sequentially per recipe `kitchen_rules`. The single-file state design (`review_gate_state.json`) is correct for this model. If `run_mode=parallel` is ever exercised for the review loop, the state file should be keyed by PR number (e.g., `review_gate_state_{pr_number}.json`). This is a future concern, not a current blocker.

### Hook interaction with quota_post_hook
If `quota_post_hook.py` fires and replaces the full tool output with a quota warning (via `updatedMCPToolOutput`), the LLM will see the quota text instead of the `%%REVIEW_GATE%%` tags. Layer 3 (tag visibility) is defeated in this case, but Layer 4 (hook gate) still enforces — the `review_gate_post_hook.py` reads from the original `tool_response` (raw MCP JSON), not the reformatted output. This is acceptable: the hook gate is the primary enforcement mechanism, tags are supplementary.

### Session resume
The state file persists across session resumes (it lives in `.autoskillit/temp/` which is not cleaned on resume). If the orchestrator crashes after `review_pr` sets `LOOP_REQUIRED` but before `check_review_loop` runs, the gate correctly denies `wait_for_ci` on resume. This is the desired conservative behavior.

## Consistency with Existing Patterns

| Existing Pattern | This Approach |
|---|---|
| `pr_create_guard.py` blocks `gh pr create` when kitchen open | `review_loop_gate.py` blocks `wait_for_ci`/`enqueue_pr` when review loop not executed |
| `.hook_config.json` tracks kitchen state across calls | `review_gate_state.json` tracks review gate state across calls |
| `%%ORDER_UP::{hex}%%` structured completion marker | `%%REVIEW_GATE::LOOP_REQUIRED%%` structured routing marker |
| `quota_guard.py` reads state → threshold check → deny | `review_loop_gate.py` reads state → gate check → deny |
| `skill_contracts.yaml` defines expected output patterns | Gate tags registered as expected outputs for `review-pr` |
| `HOOK_REGISTRY` in `hook_registry.py` centralized registration | New hooks registered via same mechanism |
| `_close_kitchen_handler()` unlinks `.hook_config.json` | Also unlinks `review_gate_state.json` |

## Related Issues

- #1083 — Orchestrator step-skipping recurrence (OPEN/staged) — parent issue for structural enforcement
- #1096 — check_review_loop routing fix (OPEN/staged, code merged via #1100)
- #968 — Review retry loop (OPEN/staged, code merged via #982)
- #1112 — Non-blocking routing through resolve_review (OPEN/staged, code merged via #1115)

## Investigation Artifacts

- Orchestrator session: `681cded1-8229-4b01-af46-952dcb1eeef3` (2007 lines, 10+ hours)
- review_pr session: `583d3975-557f-4ad7-9ee2-423eee3aab8d` (verdict=changes_requested, 1 blocking/19 warning)
- resolve_review session: `5cd3b34e-53f4-46eb-94ac-7a66a53c2d67` (verdict=real_fix, fixes_applied=8, ACCEPT:8/REJECT:9/DISCUSS:2)
- Kitchen ID: `efdf583b-1078-4093-90cf-844f4e4fc8ce`
- CWD: `/home/talon/projects/autoskillit-runs/impl-20260426-034917-731188`

Trecek:
## Summary

Add a multi-layer structural enforcement mechanism that prevents the orchestrator LLM from skipping the `check_review_loop` step after `review_pr` returns `changes_requested`. The system uses three new components working together:

1. **Review gate tags** (`%%REVIEW_GATE::LOOP_REQUIRED%%` / `%%REVIEW_GATE::CLEAR%%`) emitted by the `review-pr` skill alongside the existing verdict token — these inject routing signal at the decision point
2. **PostToolUse state writer hook** (`review_gate_post_hook.py`) that captures gate tags from `run_skill` output and `check_review_loop` calls from `run_python`, persisting state to `.autoskillit/temp/review_gate_state.json`
3. **PreToolUse gate hook** (`review_loop_gate.py`) that blocks `wait_for_ci` and `enqueue_pr` when a `LOOP_REQUIRED` gate is active and `check_review_loop` has not been called

This addresses the class of failure — LLM routing shortcuts — not just the specific incident, by placing a structural backstop that operates independently of the LLM's reasoning.

### R1: Review Gate Tags — Structured Routing Markers

Add ASCII-only structured routing tags following the established `%%NAME::PAYLOAD%%` convention (consistent with `%%ORDER_UP::{hex}%%` and `%%L2_DONE::{hex}%%`):

| Tag | Meaning | Emitted by | When |
|---|---|---|---|
| `%%REVIEW_GATE::LOOP_REQUIRED%%` | Blocking findings exist; `check_review_loop` MUST run before merge path | `review_pr` skill output | `verdict = changes_requested` |
| `%%REVIEW_GATE::CLEAR%%` | No blocking findings or review loop bypassed; merge path may proceed | `review_pr` skill output | `verdict = approved` OR `verdict = needs_human` |

**Verdict coverage — explicit design decisions:**

- `changes_requested` → `LOOP_REQUIRED` — blocking findings require the review loop to decide whether to re-review after fixes
- `approved` → `CLEAR` — no findings, no loop needed
- `needs_human` → `CLEAR` — human must intervene, automated loop is not appropriate; clears any stale `LOOP_REQUIRED` from a prior iteration
- `approved_with_comments` → **no tag emitted** (the state file is not written). This verdict also routes through `resolve_review → re_push_review → check_review_loop`, but `check_review_loop` returns `had_blocking=false` by design. The gate fails-open (no state file = allow), which is correct.

Update `review-pr/SKILL.md` final output section to emit the appropriate tag alongside the existing `verdict` token. Register the tag patterns in `skill_contracts.yaml` as `expected_output_patterns` for the `review-pr` skill.

### R2: PostToolUse State Writer Hook — `review_gate_post_hook.py`

New PostToolUse hook on `run_skill` **and** `run_python` that:

1. **For `run_skill`:** Inspects the result text for review gate tags.
2. When `%%REVIEW_GATE::LOOP_REQUIRED%%` detected: writes `.autoskillit/temp/review_gate_state.json` using `atomic_write` from `core/io.py`.
3. When `%%REVIEW_GATE::CLEAR%%` detected: clears (unlinks) the state file.
4. **For `run_python`:** When `tool_input.callable` contains `check_review_loop`: reads the existing state file and sets `check_review_loop_called = true`.

### R3: PreToolUse Gate Hook — `review_loop_gate.py`

New PreToolUse hook on **both** `wait_for_ci` **and** `enqueue_pr` that:

1. Reads `.autoskillit/temp/review_gate_state.json`
2. If file doesn't exist → allow (fail-open)
3. If `gate == "CLEAR"` → allow
4. If `gate == "LOOP_REQUIRED"` AND `check_review_loop_called == false` → **DENY**
5. If `gate == "LOOP_REQUIRED"` AND `check_review_loop_called == true` → allow

### R4: Hook Registration

- Add `review_gate_post_hook.py` to `HOOK_REGISTRY` as a PostToolUse hook with matcher for `run_skill` and `run_python`
- Add `review_loop_gate.py` to `HOOK_REGISTRY` as a PreToolUse hook with matcher for `wait_for_ci` and `enqueue_pr`
- Run `task sync-hooks-hash` to update the registry hash.

### R5: Anti-Shortcut Rule in sous-chef/SKILL.md

Add a concise rule to the `STEP EXECUTION IS NOT DISCRETIONARY` section:

> **Do not generalize from prior step outcomes.** A step that returned a non-branching result in a previous iteration may return a different result in the next. Every step must execute on every issue — observed patterns from earlier issues do not make later executions redundant.

### R6: close_kitchen Cleanup

`_close_kitchen_handler()` in `tools_kitchen.py` must also unlink `.autoskillit/temp/review_gate_state.json` (with `missing_ok=True`).

### R7: Test Coverage

1. Hook unit tests for `review_loop_gate.py` (T1-1 through T1-9)
2. Hook unit tests for `review_gate_post_hook.py` (T2-1 through T2-9)
3. Contract tests for `%%REVIEW_GATE%%` tags in `skill_contracts.yaml` (T3-1, T3-2)
4. Hook registration tests for both hooks in `HOOK_REGISTRY` (T4-1 through T4-3)
5. Cleanup tests for `_close_kitchen_handler()` unlinks state file (T5-1, T5-2)
6. Existing coverage verification for `check_review_loop` (T6-1)

### State Lifecycle Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% ENTRY EVENTS %%
    EV1(["PostToolUse: run_skill"])
    EV2(["PostToolUse: run_python"])
    EV3(["close_kitchen MCP tool"])
    EV4(["PreToolUse: wait_for_ci / enqueue_pr"])

    subgraph PostHook ["★ review_gate_post_hook.py — Write Path"]
        direction TB
        B1["Branch 1: LOOP_REQUIRED tag<br/>━━━━━━━━━━<br/>Detects %%REVIEW_GATE::LOOP_REQUIRED%%<br/>in run_skill result"]
        B2["Branch 2: CLEAR tag<br/>━━━━━━━━━━<br/>Detects %%REVIEW_GATE::CLEAR%%<br/>in run_skill result"]
        B3["Branch 3: check_review_loop call<br/>━━━━━━━━━━<br/>Detects run_python callable<br/>containing 'check_review_loop'"]
        ATOMIC["_atomic_write()<br/>━━━━━━━━━━<br/>mkstemp → fsync → os.replace<br/>Prevents partial-write corruption"]
    end

    subgraph StateFile ["review_gate_state.json — Field Contracts"]
        direction TB
        GATE["gate: 'LOOP_REQUIRED'<br/>━━━━━━━━━━<br/>INIT_ONLY — set once<br/>Never overwritten"]
        VERDICT["review_verdict: 'changes_requested'<br/>━━━━━━━━━━<br/>INIT_ONLY — set once<br/>Audit trail only"]
        SET_AT["set_at: ISO-8601 UTC<br/>━━━━━━━━━━<br/>INIT_ONLY — set once<br/>Diagnostic timestamp"]
        CHECK_CALLED["check_review_loop_called: bool<br/>━━━━━━━━━━<br/>MUTABLE: False → True<br/>Gate-satisfied signal"]
        PR_NUM["pr_number: str<br/>━━━━━━━━━━<br/>MUTABLE — extracted from<br/>skill_command; refined by args"]
    end

    subgraph GateRead ["★ review_loop_gate.py — Read / Enforce Path"]
        direction TB
        EXIST_CHECK{"state_file.exists()?"}
        GATE_CHECK{"gate == 'LOOP_REQUIRED'?"}
        CALLED_CHECK{"check_review_loop_called?"}
        DENY["Deny tool call<br/>━━━━━━━━━━<br/>permissionDecision: deny<br/>Recipe routing hint injected"]
        ALLOW["Allow tool call<br/>━━━━━━━━━━<br/>sys.exit(0)"]
    end

    subgraph ClearPaths ["Clear / Teardown Paths"]
        direction LR
        CLR_TAG["CLEAR tag path<br/>━━━━━━━━━━<br/>Branch 2 unlinks file<br/>on explicit CLEAR signal"]
        CLR_KITCHEN["● close_kitchen handler<br/>━━━━━━━━━━<br/>_close_kitchen_handler() unlinks<br/>review_gate_state.json<br/>at kitchen teardown"]
    end

    %% WRITE PATH %%
    EV1 --> B1
    EV1 --> B2
    EV2 --> B3

    B1 --> ATOMIC
    ATOMIC -->|"Writes full state object"| GATE
    ATOMIC --> VERDICT
    ATOMIC --> SET_AT
    ATOMIC -->|"check_review_loop_called=False"| CHECK_CALLED
    ATOMIC -->|"Extracted from skill_command"| PR_NUM

    B3 -->|"File exists AND gate=LOOP_REQUIRED"| CHECK_CALLED
    B3 -->|"Refine from args.pr_number"| PR_NUM

    %% READ / ENFORCE PATH %%
    EV4 --> EXIST_CHECK
    EXIST_CHECK -->|"absent"| ALLOW
    EXIST_CHECK -->|"present"| GATE_CHECK
    GATE_CHECK -->|"≠ LOOP_REQUIRED"| ALLOW
    GATE_CHECK -->|"= LOOP_REQUIRED"| CALLED_CHECK
    CALLED_CHECK -->|"True — satisfied"| ALLOW
    CALLED_CHECK -->|"False — not yet called"| DENY

    %% CLEAR PATHS %%
    B2 --> CLR_TAG
    EV3 --> CLR_KITCHEN

    %% CLASS ASSIGNMENTS %%
    class EV1,EV2,EV3,EV4 terminal;
    class B1,B2,B3,ATOMIC newComponent;
    class GATE,SET_AT,VERDICT detector;
    class CHECK_CALLED,PR_NUM phase;
    class EXIST_CHECK,GATE_CHECK,CALLED_CHECK stateNode;
    class DENY gap;
    class ALLOW output;
    class CLR_TAG newComponent;
    class CLR_KITCHEN handler;
```

### Security Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    subgraph Registry ["● HOOK REGISTRY — Authorization Surface"]
        HOOK_REG["● hook_registry.py<br/>━━━━━━━━━━<br/>PostToolUse: run_skill|run_python<br/>→ review_gate_post_hook.py<br/>PreToolUse: wait_for_ci|enqueue_pr<br/>→ review_loop_gate.py"]
    end

    subgraph PostBoundary ["TRUST BOUNDARY 1: PostToolUse — Gate State Writer"]
        RUN_SKILL["run_skill / run_python<br/>━━━━━━━━━━<br/>Tool execution output<br/>(untrusted skill output)"]
        POST_HOOK["★ review_gate_post_hook.py<br/>━━━━━━━━━━<br/>PostToolUse listener<br/>Scans output for sentinel tags"]
        TAG_LOOP["%%REVIEW_GATE::LOOP_REQUIRED%%<br/>━━━━━━━━━━<br/>Arms gate<br/>review_verdict: changes_requested"]
        TAG_CLEAR["%%REVIEW_GATE::CLEAR%%<br/>━━━━━━━━━━<br/>Clears gate<br/>unlinks state file"]
        CHECK_LOOP["check_review_loop callable<br/>━━━━━━━━━━<br/>run_python branch<br/>Sets satisfied=true"]
    end

    subgraph GateState ["GATE STATE — .autoskillit/temp/review_gate_state.json"]
        STATE["★ review_gate_state.json<br/>━━━━━━━━━━<br/>gate: LOOP_REQUIRED<br/>check_review_loop_called: bool<br/>pr_number: int<br/>set_at: ISO timestamp<br/>atomic write via os.replace + fsync"]
    end

    subgraph PreBoundary ["TRUST BOUNDARY 2: PreToolUse — Gate Enforcer"]
        PRE_HOOK["★ review_loop_gate.py<br/>━━━━━━━━━━<br/>PreToolUse enforcer<br/>Read-only state reader"]
        GATE_CHECK{"Gate == LOOP_REQUIRED<br/>AND<br/>check_review_loop_called == False?"}
        DENY["DENY<br/>━━━━━━━━━━<br/>permissionDecision: deny<br/>REVIEW LOOP REQUIRED<br/>Must call check_review_loop<br/>before wait_for_ci/enqueue_pr"]
        ALLOW["ALLOW — fail-open<br/>━━━━━━━━━━<br/>No state file → exit(0)<br/>JSON error → exit(0)<br/>gate != LOOP_REQUIRED → exit(0)<br/>satisfied → exit(0)"]
    end

    subgraph Targets ["PROTECTED TARGETS"]
        WAIT_CI["wait_for_ci<br/>━━━━━━━━━━<br/>CI gate advancement"]
        ENQUEUE["enqueue_pr<br/>━━━━━━━━━━<br/>Merge queue entry"]
    end

    CLEANUP["● close_kitchen<br/>━━━━━━━━━━<br/>Unconditional gate clear<br/>on kitchen close"]

    %% REGISTRATION FLOW %%
    HOOK_REG -->|"registers PostToolUse"| POST_HOOK
    HOOK_REG -->|"registers PreToolUse"| PRE_HOOK

    %% STATE ARMING FLOW %%
    RUN_SKILL --> POST_HOOK
    POST_HOOK --> TAG_LOOP
    POST_HOOK --> TAG_CLEAR
    POST_HOOK --> CHECK_LOOP

    TAG_LOOP -->|"atomic write"| STATE
    TAG_CLEAR -->|"unlink"| STATE
    CHECK_LOOP -->|"update satisfied=true"| STATE

    %% ENFORCEMENT FLOW %%
    STATE -->|"read-only"| PRE_HOOK
    PRE_HOOK --> GATE_CHECK
    GATE_CHECK -->|"armed + unsatisfied"| DENY
    GATE_CHECK -->|"absent / satisfied / error"| ALLOW

    DENY -.->|"blocks"| WAIT_CI
    DENY -.->|"blocks"| ENQUEUE
    ALLOW --> WAIT_CI
    ALLOW --> ENQUEUE

    %% CLEANUP %%
    CLEANUP -->|"unlink on close"| STATE

    %% CLASS ASSIGNMENTS %%
    class HOOK_REG handler;
    class RUN_SKILL cli;
    class POST_HOOK newComponent;
    class TAG_LOOP gap;
    class TAG_CLEAR output;
    class CHECK_LOOP phase;
    class STATE stateNode;
    class PRE_HOOK newComponent;
    class GATE_CHECK detector;
    class DENY detector;
    class ALLOW output;
    class WAIT_CI,ENQUEUE phase;
    class CLEANUP handler;
```

### Process Flow Diagram

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 42, 'rankSpacing': 54, 'curve': 'basis'}}}%%
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINALS %%
    START([START])
    DENIED([DENIED — wait_for_ci blocked])
    ALLOWED([ALLOWED — wait_for_ci / enqueue_pr proceeds])

    subgraph P1 ["Phase 1 — ● review_pr Execution"]
        direction TB
        RunSkill["● run_skill<br/>━━━━━━━━━━<br/>tool: run_skill review-pr<br/>forwards to headless Claude"]
        ReviewPRSkill["● review-pr skill<br/>━━━━━━━━━━<br/>Analyzes PR diff, posts inline comments<br/>Writes verdict line to output"]
        VerdictDecision{"Verdict tag<br/>in output?<br/>━━━━━━━━━━<br/>LOOP_REQUIRED / CLEAR / none"}
    end

    subgraph P2 ["Phase 2 — ★ PostToolUse Hook: review_gate_post_hook"]
        direction TB
        PostHookRunSkill["★ review_gate_post_hook<br/>━━━━━━━━━━<br/>PostToolUse: matcher run_skill|run_python<br/>reads tool_response JSON"]
        TagBranch{"Tag type?<br/>━━━━━━━━━━<br/>_extract_run_skill_result()"}
        WriteStateFile["★ Write review_gate_state.json<br/>━━━━━━━━━━<br/>gate=LOOP_REQUIRED<br/>check_review_loop_called=false<br/>pr_number extracted from skill_command"]
        DeleteStateFile["Delete review_gate_state.json<br/>━━━━━━━━━━<br/>CLEAR tag — gate disarmed"]
        NoOpTag["No-op exit 0<br/>━━━━━━━━━━<br/>No gate tag in output"]
    end

    subgraph P3 ["Phase 3 — LLM Routing"]
        direction TB
        LLMRoute{"LLM routes<br/>━━━━━━━━━━<br/>verdict = changes_requested?"}
        ResolveReview["resolve_review skill<br/>━━━━━━━━━━<br/>Applies fix instructions<br/>from review summary"]
        RePush["re_push_review skill<br/>━━━━━━━━━━<br/>Commits fixes, pushes branch<br/>on_success → check_review_loop"]
    end

    subgraph P4 ["Phase 4 — ★ check_review_loop Detection"]
        direction TB
        CallCRL["run_python check_review_loop<br/>━━━━━━━━━━<br/>callable=autoskillit.smoke_utils.check_review_loop<br/>args: pr_number, cwd, current_iteration, max_iterations, previous_verdict"]
        PostHookCRL["★ review_gate_post_hook<br/>━━━━━━━━━━<br/>PostToolUse: run_python branch<br/>detects callable contains check_review_loop"]
        UpdateStateCalled["★ Update review_gate_state.json<br/>━━━━━━━━━━<br/>check_review_loop_called=true<br/>pr_number updated from args if present"]
        IterDecision{"Loop continue?<br/>━━━━━━━━━━<br/>had_blocking=true AND<br/>max_exceeded=false?"}
    end

    subgraph P5 ["Phase 5 — ★ PreToolUse Gate: review_loop_gate"]
        direction TB
        WaitForCIAttempt["wait_for_ci / enqueue_pr attempt<br/>━━━━━━━━━━<br/>LLM calls CI progression tool"]
        PreHookGate["★ review_loop_gate<br/>━━━━━━━━━━<br/>PreToolUse: matcher wait_for_ci|enqueue_pr<br/>reads review_gate_state.json"]
        GateActive{"Gate armed?<br/>━━━━━━━━━━<br/>gate=LOOP_REQUIRED AND<br/>check_review_loop_called=false?"}
    end

    subgraph P6 ["Phase 6 — ● close_kitchen Cleanup"]
        direction TB
        CloseKitchen["● close_kitchen<br/>━━━━━━━━━━<br/>Unregisters kitchen, clears hook config<br/>resets recipe context"]
        CleanupState["● Delete review_gate_state.json<br/>━━━━━━━━━━<br/>.autoskillit/temp/review_gate_state.json<br/>Unconditional cleanup on close"]
    end

    %% FLOW %%
    START --> RunSkill
    RunSkill --> ReviewPRSkill
    ReviewPRSkill --> VerdictDecision

    VerdictDecision -->|"LOOP_REQUIRED emitted"| PostHookRunSkill
    VerdictDecision -->|"CLEAR emitted"| PostHookRunSkill
    VerdictDecision -->|"approved_with_comments — no tag"| PostHookRunSkill

    PostHookRunSkill --> TagBranch
    TagBranch -->|"LOOP_REQUIRED"| WriteStateFile
    TagBranch -->|"CLEAR"| DeleteStateFile
    TagBranch -->|"no tag"| NoOpTag

    WriteStateFile --> LLMRoute
    DeleteStateFile --> WaitForCIAttempt
    NoOpTag --> WaitForCIAttempt

    LLMRoute -->|"yes — changes_requested"| ResolveReview
    LLMRoute -->|"no — approved / needs_human"| WaitForCIAttempt
    ResolveReview --> RePush
    RePush --> CallCRL

    CallCRL --> PostHookCRL
    PostHookCRL --> UpdateStateCalled
    UpdateStateCalled --> IterDecision

    IterDecision -->|"true — re-review needed"| RunSkill
    IterDecision -->|"false — proceed to CI"| WaitForCIAttempt

    WaitForCIAttempt --> PreHookGate
    PreHookGate --> GateActive
    GateActive -->|"yes — gate armed, deny"| DENIED
    GateActive -->|"no — gate clear or already called"| ALLOWED

    ALLOWED --> CloseKitchen
    CloseKitchen --> CleanupState

    %% CLASS ASSIGNMENTS %%
    class START,DENIED,ALLOWED terminal;
    class RunSkill,ReviewPRSkill,ResolveReview,RePush,CallCRL,WaitForCIAttempt handler;
    class VerdictDecision,TagBranch,LLMRoute,GateActive,IterDecision stateNode;
    class PostHookRunSkill,PostHookCRL,PreHookGate newComponent;
    class WriteStateFile,UpdateStateCalled output;
    class DeleteStateFile,NoOpTag phase;
    class CloseKitchen integration;
    class CleanupState phase;
```

Closes #1295

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260426-092115-537305/.autoskillit/temp/make-plan/hook_based_review_gate_enforcement_plan_2026-04-26_092500.md`

## Token Usage Summary

| analyze | 22 | 2.8k | 104.0k | 40.7k | 1 | 3m 44s |
| extract_domain | 33 | 7.3k | 170.0k | 28.2k | 1 | 3m 39s |
| generate_phases | 28 | 3.8k | 319.8k | 23.9k | 1 | 1m 47s |
| reconcile_deps | 23 | 745 | 135.1k | 14.1k | 1 | 28s |
| **Total** | 106 | 14.6k | 728.9k | 106.9k | | 9m 40s |


src/autoskillit/hooks/review_gate_post_hook.py
```diff
@@ -0,0 +1,113 @@
...
+import os
+import re
+import sys
+import tempfile
+from datetime import UTC, datetime
+from pathlib import Path
+
+_TAG_LOOP_REQUIRED = "%%REVIEW_GATE::LOOP_REQUIRED%%"
+_TAG_CLEAR = "%%REVIEW_GATE::CLEAR%%"
+_STATE_FILE_RELPATH = (".autoskillit", "temp", "review_gate_state.json")
```
Trecek, line 15: [warning] cohesion: _STATE_FILE_RELPATH is defined identically here and in review_loop_gate.py (L20). Extract to a shared constant (e.g. a new _review_gate_state.py helper or existing hooks utility) to avoid path drift when changed.

src/autoskillit/hooks/review_gate_post_hook.py
```diff
@@ -0,0 +1,113 @@
...
+        os.replace(tmp, path)
+    except Exception:
+        try:
+            os.unlink(tmp)
+        except OSError:
+            pass
+        raise
+
+
+def _extract_run_skill_result(tool_response: str | dict) -> str:  # type: ignore[return]
```
Trecek, line 36: [warning] defense: `# type: ignore[return]` on `_extract_run_skill_result` suppresses a mypy error rather than fixing it. The `except Exception: return ""` already covers all exit paths; remove the suppressor and rely on the explicit fallback.

src/autoskillit/hooks/review_gate_post_hook.py
```diff
@@ -0,0 +1,113 @@
...
+    """Unwrap double-wrapped run_skill JSON to get the inner result string."""
+    try:
+        outer = json.loads(tool_response) if isinstance(tool_response, str) else tool_response
+        inner_str = outer["result"]
+        inner = json.loads(inner_str)
```
Trecek: [warning] bugs: `json.loads(inner_str)` is called without first checking `isinstance(inner_str, str)`. If `outer["result"]` is already a dict (single-wrapped response), this raises `TypeError` which is silently swallowed, causing gate tags to be missed. Guard: `if isinstance(inner_str, str): inner = json.loads(inner_str)` (matches the pattern in `token_summary_hook._unwrap_mcp_response`).

src/autoskillit/hooks/review_gate_post_hook.py
```diff
@@ -0,0 +1,113 @@
...
+def _extract_pr_number(skill_command: str) -> str:
+    """Best-effort PR number extraction from a skill_command string."""
+    m = _PR_NUMBER_RE.search(skill_command)
+    return m.group(1) if m else ""
+
+
+def main() -> None:
+    try:
+        data = json.loads(sys.stdin.read())
+    except Exception:
```
Trecek, line 56: [warning] defense: Malformed stdin is silently swallowed via `sys.exit(0)`, making the hook indistinguishable from a successful no-op. Write a diagnostic to stderr before exiting so truncated/corrupt payloads are observable.

src/autoskillit/hooks/review_gate_post_hook.py
```diff
@@ -0,0 +1,113 @@
...
+            state = {
+                "gate": "LOOP_REQUIRED",
+                "review_verdict": "changes_requested",
+                "check_review_loop_called": False,
+                "pr_number": pr_number,
+                "set_at": datetime.now(UTC).isoformat(),
+            }
+            try:
+                _atomic_write(state_file, json.dumps(state))
+            except Exception:
```
Trecek: [warning] defense: `_atomic_write` failure for the LOOP_REQUIRED gate state is silently swallowed (`except Exception: pass`). If the write fails, the enforcement backstop is never set and the hook exits 0 silently. At minimum write the error to stderr.

src/autoskillit/hooks/review_gate_post_hook.py
```diff
@@ -0,0 +1,113 @@
...
+                try:
+                    state = json.loads(state_file.read_text(encoding="utf-8"))
+                    if state.get("gate") == "LOOP_REQUIRED":
+                        state["check_review_loop_called"] = True
+                        # Update pr_number from args if available
+                        args = tool_input.get("args") or {}
+                        if isinstance(args, dict) and args.get("pr_number"):
+                            state["pr_number"] = str(args["pr_number"])
+                        _atomic_write(state_file, json.dumps(state))
+                except Exception:
```
Trecek: [warning] defense: `_atomic_write` failure when marking `check_review_loop_called=True` is silently swallowed. If the write fails, the gate state retains `check_review_loop_called: False`, causing `review_loop_gate.py` to incorrectly block `wait_for_ci`/`enqueue_pr` after `check_review_loop` was legitimately invoked. Surface to stderr.

src/autoskillit/hooks/review_loop_gate.py
```diff
@@ -0,0 +1,59 @@
...
+    "resolve_review applied fixes, but check_review_loop was not called. "
+    "You MUST call run_python with callable="
+    "'autoskillit.smoke_utils.check_review_loop' with the correct "
+    "pr_number, cwd, current_iteration, max_iterations, and "
+    "previous_verdict parameters BEFORE proceeding to "
+    "wait_for_ci/enqueue_pr. "
+    "Recipe routing: re_push_review → on_success: check_review_loop."
+)
+
+_STATE_FILE_RELPATH = (".autoskillit", "temp", "review_gate_state.json")
```
Trecek, line 20: [warning] cohesion: `_STATE_FILE_RELPATH` is defined identically here and in `review_gate_post_hook.py` (L15). Extract to a shared constant to avoid drift.

src/autoskillit/hooks/review_loop_gate.py
```diff
@@ -0,0 +1,59 @@
...
+    except Exception:
+        sys.exit(0)
+
+    state_file = Path.cwd().joinpath(*_STATE_FILE_RELPATH)
+    if not state_file.exists():
+        sys.exit(0)
+
+    try:
+        state = json.loads(state_file.read_text(encoding="utf-8"))
+    except Exception:
```
Trecek, line 35: [warning] defense: A corrupt or unreadable `review_gate_state.json` causes `sys.exit(0)`, allowing `wait_for_ci`/`enqueue_pr` to proceed even when a LOOP_REQUIRED gate may be active. Consider whether to deny (safe) or allow (current) on state corruption; the silent pass-through is a defense gap.

src/autoskillit/hook_registry.py
```diff
@@ -98,6 +98,15 @@ def __post_init__(self) -> None:
         matcher=r"mcp__.*autoskillit.*__run_skill.*",
         scripts=["token_summary_hook.py", "quota_post_hook.py"],
     ),
+    HookDef(
+        event_type="PostToolUse",
+        matcher=r"mcp__.*autoskillit.*__(run_skill|run_python).*",
+        scripts=["review_gate_post_hook.py"],
+    ),
+    HookDef(
```
Trecek, line 106: [warning] cohesion: `HookDef` for `review_loop_gate.py` omits `event_type=` keyword argument. All other PreToolUse entries in `HOOK_REGISTRY` set `event_type` explicitly. Add `event_type="PreToolUse"` for consistency.

tests/infra/test_review_loop_gate.py
```diff
@@ -0,0 +1,152 @@
...
+    from autoskillit.hooks.review_loop_gate import main  # noqa: PLC0415
+
+    if raw_stdin is not None:
+        stdin_content = raw_stdin
+    else:
+        stdin_content = json.dumps({"tool_name": tool_name, "tool_input": {}})
+
+    buf = io.StringIO()
+    with contextlib.redirect_stdout(buf):
+        with unittest.mock.patch("sys.stdin", io.StringIO(stdin_content)):
```
Trecek, line 52: [warning] tests: `unittest.mock.patch("pathlib.Path.cwd", ...)` patches the classmethod globally. Prefer patching at the module-level name (`autoskillit.hooks.review_loop_gate.Path.cwd`) to avoid inadvertently affecting stdlib internals.

tests/infra/test_review_loop_gate.py
```diff
@@ -0,0 +1,152 @@
...
+            with unittest.mock.patch("pathlib.Path.cwd", return_value=tmp_dir):
+                try:
+                    main()
+                except SystemExit as exc:
```
Trecek, line 58: [warning] tests: `assert exc.code == 0` is embedded inside `_run_gate()`, firing unconditionally for all callers. If the hook exits non-zero, the failure manifests as an `AssertionError` from inside the helper with no test context. Move assertions to individual test bodies.

tests/infra/test_review_loop_gate.py
```diff
@@ -0,0 +1,152 @@
...
+        assert out.strip() == ""
+
+
+# ---------------------------------------------------------------------------
+# T1-7 through T1-8: Matcher coverage
+# ---------------------------------------------------------------------------
+
+
+class TestReviewLoopGateMatcherCoverage:
+    def test_fires_on_wait_for_ci(self, tmp_path):
```
Trecek, line 144: [warning] tests: `test_fires_on_wait_for_ci` (T1-7) is identical to `test_denies_when_loop_required_and_not_called` (T1-3): same state setup, same tool, same assertion. T1-7 is redundant. Remove it or merge concerns into T1-3.

tests/infra/test_review_gate_post_hook.py
```diff
@@ -0,0 +1,274 @@
...
+    """Run review_gate_post_hook.main() and return (stdout, exit_code)."""
+    from autoskillit.hooks.review_gate_post_hook import main  # noqa: PLC0415
+
+    stdin_text = raw_stdin if raw_stdin is not None else json.dumps(event or {})
+
+    buf = io.StringIO()
+    exit_code = 0
+    with contextlib.redirect_stdout(buf):
+        with unittest.mock.patch("sys.stdin", io.StringIO(stdin_text)):
+            with unittest.mock.patch("pathlib.Path.cwd", return_value=tmp_dir):
```
Trecek: [warning] tests: `unittest.mock.patch("pathlib.Path.cwd", ...)` patches globally. Prefer patching at `autoskillit.hooks.review_gate_post_hook.Path.cwd` to avoid inadvertently affecting stdlib or third-party code within the test block.

tests/contracts/test_skill_contracts.py
```diff
@@ -175,3 +175,22 @@ def test_skill_contracts_yaml_includes_compose_research_pr(skills):
 def test_skill_contracts_yaml_open_research_pr_removed(skills):
     """open-research-pr must no longer be registered — it has been retired."""
     assert "open-research-pr" not in skills
+
+
+# T3-1
+def test_review_gate_loop_required_pattern_in_review_pr_contracts(skills):
```
Trecek, line 181: [warning] tests: T3-1 and T3-2 assert overlapping facts. T3-1 already asserts the regex covering both LOOP_REQUIRED and CLEAR; T3-2 re-asserts the same. Consolidate into one test.

tests/server/test_tools_kitchen.py
```diff
@@ -1398,3 +1398,56 @@ def test_display_categories_includes_fleet_when_enabled() -> None:
...
+                "review_verdict": "changes_requested",
+                "check_review_loop_called": False,
+                "pr_number": "1290",
+                "set_at": "2026-04-26T04:30:00+00:00",
+            }
+        )
+    )
+    assert state_path.exists(), "State file must exist before close_kitchen"
+
+    with patch("autoskillit.server._get_ctx", return_value=mock_ctx):
```
Trecek, line 1432: [warning] tests: `_close_kitchen_handler` is imported inside the `with patch(...)` context. If the module was already cached in `sys.modules`, the import is a no-op and the patch may not affect the already-bound function. Move the import outside the patch context.

Trecek (commented):
AutoSkillit PR Review — Verdict: approved_with_comments. Warning-only findings detected (15 warnings, 0 criticals). See inline comments — no blocking changes required.

src/autoskillit/hooks/review_gate_post_hook.py
Trecek: Agreed — fixed in c307ccf5. If outer["result"] is already a dict (single-wrapped MCP response), json.loads() raises TypeError silently swallowed by except-clause, causing gate tags to be missed. Added isinstance(inner_str, str) guard matching the token_summary_hook._unwrap_mcp_response pattern.

src/autoskillit/hooks/review_gate_post_hook.py
Trecek: Investigated — this is intentional. The # type: ignore[return] suppressor is narrowly scoped to the [return] error. mypy requires it because it cannot statically verify that the try/except exhaustively covers all return paths for the str annotation. The suppressor was present in the original commit and is the minimal targeted fix rather than a broad ignore.

src/autoskillit/hooks/review_gate_post_hook.py
Trecek: Investigated — this is intentional. Silently exiting 0 on malformed stdin is the project's explicit fail-open hook design documented in tests/arch/_rules.py (_BROAD_EXCEPT_EXEMPT). All other hook scripts follow the same pattern (quota_guard.py, skill_command_guard.py, etc.). Writing to stderr on bad stdin would violate the intentional design and produce noise in Claude Code's hook stderr capture.

src/autoskillit/hooks/review_gate_post_hook.py
Trecek: Agreed — fixed in a3010340. Replaced except Exception: pass with except Exception as e: print(..., file=sys.stderr) so disk/permission failures are diagnosable. This is distinct from the fail-open input-parsing pattern — here it is a critical state mutation silently dropped.

src/autoskillit/hooks/review_gate_post_hook.py
Trecek: Agreed — fixed in a3010340. Applied the same stderr surfacing fix for the check_review_loop_called=True write path. If this write fails, the gate retains False causing false blocks on legitimate check_review_loop invocations.

src/autoskillit/hooks/review_loop_gate.py
Trecek: Valid observation — flagged for design decision. _STATE_FILE_RELPATH is byte-for-byte identical in both hook files. A shared _review_gate_state.py helper following the _hook_settings.py / _fmt_primitives.py precedent would eliminate drift risk, but whether to introduce a new shared helper for this two-file scope is a maintainer judgment call.

src/autoskillit/hooks/review_loop_gate.py
Trecek: Investigated — this is intentional. Failing open on corrupt state is the project's deliberate PreToolUse hook design — failing closed (deny) would block legitimate operations on transient I/O errors. Both review hook files are in _BROAD_EXCEPT_EXEMPT (added in a35c11d6). The reviewer's 'safe = deny' framing contradicts the project's fail-open mandate for hook scripts.

src/autoskillit/hook_registry.py
Trecek: Investigated — this is intentional. HookDef.event_type defaults to 'PreToolUse' (hook_registry.py:24). Inspecting the full HOOK_REGISTRY, none of the other PreToolUse entries set event_type explicitly — all rely on the default. Only PostToolUse and SessionStart entries set it explicitly. The review_loop_gate.py HookDef is consistent with every other PreToolUse entry in the registry. The reviewer's claim is factually incorrect.

tests/infra/test_review_loop_gate.py
Trecek: Agreed — fixed in cfbb250f. Changed patch target from pathlib.Path.cwd to autoskillit.hooks.review_loop_gate.Path.cwd to scope the classmethod patch to only the module under test.

tests/infra/test_review_loop_gate.py
Trecek: Investigated — this is intentional. The hook does not call sys.exit() for allow/deny decisions — it outputs JSON to stdout for deny and nothing for allow, then returns normally. SystemExit only occurs on unexpected error paths. The assert exc.code == 0 is a defensive guard against unexpected non-zero exits, not conflated with allow/deny semantics. The deny path is handled by _is_denied(out) in individual test bodies.

tests/infra/test_review_loop_gate.py
Trecek: Investigated — this is intentional. T1-7 and T1-3 have different taxonomic purposes: T1-3 (TestReviewLoopGateWithStateFile) tests the core enforcement invariant; T1-7 (TestReviewLoopGateMatcherCoverage) explicitly verifies that the wait_for_ci matcher fires — paired with T1-8 for enqueue_pr. The class name documents the intent. Removing T1-7 would leave matcher coverage for wait_for_ci implicit and not independently verified alongside its enqueue_pr counterpart.

tests/infra/test_review_gate_post_hook.py
Trecek: Agreed — fixed in cfbb250f. Changed patch target from pathlib.Path.cwd to autoskillit.hooks.review_gate_post_hook.Path.cwd to scope the classmethod patch to only the module under test.

tests/contracts/test_skill_contracts.py
Trecek: Investigated — this is intentional. T3-1 and T3-2 test different invariants: T3-1 asserts a single pattern containing the LOOP_REQUIRED|CLEAR alternation regex; T3-2 asserts that across all patterns with REVIEW_GATE, both strings appear (accepting two separate patterns). T3-2 would pass if skills emit separate %%REVIEW_GATE::LOOP_REQUIRED%% and %%REVIEW_GATE::CLEAR%% patterns while T3-1 would fail. They are complementary guards, not redundant.

tests/server/test_tools_kitchen.py
Trecek: Investigated — this is intentional. The concern is unfounded for two reasons: (1) autoskillit.server.tools_kitchen is not imported at module level in the test file — the FROM import inside the patch context always executes and binds the local name; (2) _close_kitchen_handler imports _get_ctx deferred inside its own function body (tools_kitchen.py), not at module import time, so the patch on autoskillit.server._get_ctx intercepts it correctly regardless of import order. This pattern is used consistently across 8+ other tests in the file.

src/autoskillit/hooks/review_gate_post_hook.py
Trecek: Valid observation — flagged for design decision. _STATE_FILE_RELPATH = ('.autoskillit', 'temp', 'review_gate_state.json') is byte-for-byte identical in both hook files. A shared _review_gate_state.py helper following the _hook_settings.py precedent (stdlib-only, sys.path insertion pattern used by token_summary_hook.py for _fmt_primitives) would eliminate drift risk. Whether to introduce a new shared file for this two-file scope is a maintainer judgment call.

PR #1480, opened 2026-04-28
Rectify: Kitchen-Open Guard Over-Scoped — Blocks Updates and Auto-Checks Across Unrelated Sessions

### LINKED ISSUE #1415: Kitchen-open guard over-scoped: blocks updates and auto-checks across unrelated sessions
## Summary

The `any_kitchen_open()` guard introduced in commit `aa302873` (PR #1066, April 19) blocks **all** update paths whenever **any** session has an open kitchen — even unrelated sessions running in separate terminals. This produces two user-visible regressions:

1. **`autoskillit update` hard-blocks with `SystemExit(1)`** when any kitchen is open anywhere on the machine
2. **Automatic update checks on `order`/`cook` are silently suppressed** — no prompt, no signals gathered, no notification

The guard uses a global, home-scoped registry (`~/.autoskillit/active_kitchens.json`) with PID liveness verification. Because any interactive or fleet session with an open kitchen registers an entry, and users routinely run multiple sessions, at least one kitchen is almost always registered — making updates permanently blocked and auto-checks permanently suppressed.

## Root Cause

The `any_kitchen_open()` function (`core/_plugin_cache.py:240`) reads `~/.autoskillit/active_kitchens.json` and returns `True` if any live-PID kitchen entry exists. This gates two paths:

- **`run_update_command()`** (`cli/_update.py:44-46`) — hard block:
  ```python
  if any_kitchen_open():
      print("A kitchen is currently open. Close it or wait for the pipeline to finish.")
      raise SystemExit(1)
  ```

- **`run_update_checks()`** (`cli/_update_checks.py:691-694`) — silent skip:
  ```python
  if any_kitchen_open():
      return
  ```

The guard was introduced to prevent plugin cache corruption: `_clear_plugin_cache()` uses `shutil.rmtree` to destroy the plugin cache directory, and running sessions hold in-memory references to the old version's `installPath` which they check on every hook dispatch. Without the guard, all hooks fail-open when the directory disappears.

**However, the guard is over-scoped.** It blocks the *calling process's* update when a *different process's* kitchen is open. The same commit also introduced cache rotation with a 2-hour grace period (REQ-CACHE-001), which retires old directories instead of deleting them — making the kitchen guard unnecessary defense-in-depth for cross-session safety.

## Contributing Factor: Fleet Auto-Gate Broadened Registration

Commit `2460e24d` (PR #1246, April 25) added `_fleet_auto_gate_boot()` to `_lifespan.py`, which calls `register_active_kitchen()` at MCP server startup for fleet sessions. This means fleet sessions register a kitchen entry immediately on `autoskillit serve`, before `open_kitchen` is even called — broadening the set of sessions that trigger the guard.

## Affected Components

| File | Line(s) | Role |
|------|---------|------|
| `src/autoskillit/cli/_update.py` | 42-46 | Hard kitchen guard on `autoskillit update` |
| `src/autoskillit/cli/_update_checks.py` | 691-694 | Silent kitchen guard on startup auto-check |
| `src/autoskillit/core/_plugin_cache.py` | 240-269 | `any_kitchen_open()` implementation |
| `src/autoskillit/server/_lifespan.py` | 187-188 | Fleet auto-gate registration broadens entries |
| `src/autoskillit/server/tools_kitchen.py` | 162-164 | Interactive session kitchen registration |

## Data Flow

```
Session A (interactive, kitchen open):
  open_kitchen → register_active_kitchen(kitchen_id, pid_A, project_path)
    → writes to ~/.autoskillit/active_kitchens.json

Session B (user terminal, runs "autoskillit update"):
  run_update_command() → any_kitchen_open()
    → reads ~/.autoskillit/active_kitchens.json
    → finds pid_A alive via os.kill(pid, 0) + psutil create_time
    → returns True → SystemExit(1)

Session B (user terminal, runs "autoskillit order"):
  main() → run_update_checks() → any_kitchen_open()
    → returns True → silent return (no prompt, no signals)
```

## Architectural Tension

The cache rotation with grace period (REQ-CACHE-001, same commit `aa302873`) was designed to solve the plugin cache corruption problem *permissively* — old directories survive for 2 hours after retirement, so running sessions continue referencing retired directories safely. The kitchen-open guard (REQ-GUARD-001) was added as defense-in-depth but has *restrictive* properties that conflict:

- Cache rotation: **permissive** — updates proceed, old directories survive
- Kitchen guard: **restrictive** — updates blocked entirely

If cache rotation works correctly, the guard is redundant for cross-session safety. Its value is only as a last-resort defense against edge cases where `shutil.rmtree` might still be called on the active directory.

## Test Gaps

1. **No test verifies `run_update_command()` blocks when kitchen is open.** All tests patch `any_kitchen_open` to return `False`.
2. **No test verifies `run_update_checks()` returns silently when kitchen is open.** The `_setup_run_checks()` helper never patches `any_kitchen_open`; tests pass incidentally because no `active_kitchens.json` exists in `tmp_path`.
3. **No integration test exercises `main()` → update check flow for `order`/`cook`.** Tests call `cli.order()` directly, bypassing `main()`.
4. **No cross-session isolation test** simulates "session A has kitchen open, session B tries to update."

## Recommended Fix

**Scope the guard to same-project only.** Modify `any_kitchen_open()` to accept an optional `project_path` parameter. In both `run_update_command()` and `run_update_checks()`, pass `str(Path.cwd())` so the guard only fires when a kitchen for the *same project* is open. Kitchens from unrelated projects/sessions would not block updates.

Additionally, consider downgrading the `run_update_command()` guard from a hard `SystemExit(1)` to a warning — the user is making an explicit choice to update, and the cache rotation mechanism handles safety. The `run_update_checks()` silent skip is fine (don't interrupt a pipeline with prompts) but should also be project-scoped.

### Test Plan

- Add test: `run_update_command()` raises `SystemExit(1)` when same-project kitchen is open
- Add test: `run_update_command()` proceeds when only different-project kitchens are open
- Add test: `run_update_checks()` returns silently when same-project kitchen is open
- Add test: `run_update_checks()` proceeds normally when only different-project kitchens are open
- Fix `_setup_run_checks()` to explicitly patch `any_kitchen_open` to `False` (remove implicit dependency)

## Introducing Commits

| Commit | Date | Description |
|--------|------|-------------|
| `aa302873` | 2026-04-19 | Introduced the kitchen-open guard on both update paths (PR #1066) |
| `2460e24d` | 2026-04-25 | Fleet auto-gate boot broadened kitchen registration (PR #1246) |
| `a7bb27b6` | 2026-04-26 | `perform_restart()` with `os.execv` — sets `AUTOSKILLIT_SKIP_UPDATE_CHECK` (PR #1341) |

### Factual Corrections

Two claims in the original report were inaccurate:

1. **Test helper name**: The helper in `test_update_checks.py` is named `_setup_run_checks` (correct), but the helper in `test_update_command.py` is named `_setup_run_update` (incorrectly called `_setup_run_checks` in one reference). The substantive finding (neither patches `any_kitchen_open`) is correct for both.

2. **`_clear_plugin_cache()` no longer calls `shutil.rmtree` directly**: Since commit `aa302873`, `_clear_plugin_cache()` delegates to `_retire_old_versions()` which uses the grace-period rotation for cross-version transitions. However, `_retire_old_versions()` still calls `shutil.rmtree` **immediately** for same-version directory slots (line 129 of `_plugin_cache.py`) — no grace period for reinstalls.

### Critical Finding: Project-Scoping is Insufficient

The recommended fix ("scope `any_kitchen_open()` by `project_path`") is **logically incorrect** because the plugin cache is a **single global directory**:

```
~/.claude/plugins/cache/autoskillit-local/autoskillit/
```

This directory is shared across all projects on the machine. Scoping the guard by project would allow project-A updates to destroy the cache that project-B's running kitchens depend on. The guard needs to remain global because the resource it protects is global.

### Real Failure Mode: Food Truck Sessions, Not Hooks

The original bug framing (hooks failing when cache disappears) is partially wrong. Hook scripts use **hardcoded absolute paths to the source tree** (e.g., `python3 /path/to/src/autoskillit/hooks/quota_guard.py`) and are stdlib-only — they do not reference the plugin cache at all. They are immune to cache directory changes.

The actual failure mode is **food truck (headless) child sessions** launched with `--plugin-dir <old_cache_path>` embedded in the subprocess command at construction time (`execution/commands.py:431-432`). If the cache directory is deleted before the subprocess starts (or during its execution), Claude Code fails to load the plugin.

### Same-Version `rmtree` Bypasses Grace Period

`_retire_old_versions()` at `_plugin_cache.py:129`:
```python
if subdir.name == new_version:
    shutil.rmtree(subdir)  # immediate, no grace period
```

Same-version reinstalls (e.g., `autoskillit install` after a failed upgrade, or a reinstall of the current version) destroy the active cache directory immediately. The 2-hour grace period only applies to cross-version transitions.

### Race Window

The zero-cache window between destruction and recreation is bounded by two `subprocess.run` calls with 30-second timeouts each — up to 60 seconds where the plugin cache directory may not exist.

### Revised Recommendation

The project-scoping approach should be abandoned. Instead:

1. **Make `autoskillit install` write-before-delete**: Create the new cache directory alongside the old one, update `installed_plugins.json` to point to the new path, then retire the old directory. This eliminates the zero-cache window entirely and makes the kitchen guard unnecessary for safety.

2. **Fix same-version `rmtree`**: Change `_retire_old_versions()` to use `append_retiring_entry()` for same-version directories too, instead of calling `shutil.rmtree` immediately. The grace period should apply universally.

3. **Once write-before-delete is in place**: The kitchen guard in `run_update_command()` can be safely downgraded from `SystemExit(1)` to a warning, and the guard in `run_update_checks()` can be removed entirely — allowing auto-update checks and explicit updates to proceed regardless of kitchen state.

4. **Food truck resilience**: Consider resolving `--plugin-dir` lazily (at subprocess launch time from `installed_plugins.json`) rather than embedding the path at command construction time. This would make child sessions resilient to cache transitions.

### Cross-Validator Consensus

Four independent auditors examined the issue's claims, the revised recommendation, and the codebase. Their findings converge on several critical points that invalidate the revised recommendation and reshape the solution space.

### Finding 1: All Four Revised Recommendations Are Blocked or Infeasible

| Recommendation | Verdict | Blocker |
|---|---|---|
| 1. Write-before-delete install | **BLOCKED** | `claude plugin install` is a Claude Code CLI tool that manages its own cache paths. Autoskillit cannot pre-populate the new cache directory — it has no API to replicate Claude Code's internal plugin snapshotting logic. Additionally, REQ-CACHE-003 mandates `InstalledPluginsFile().remove()` BEFORE `claude plugin install` — the old entry must be absent for Claude Code to treat it as a fresh install. |
| 2. Same-version rmtree → grace period | **BLOCKED** | REQ-CACHE-002 is deliberate and technically mandatory. `claude plugin install` writes to `<cache_dir>/<version>/`. If the same-version directory still exists on disk, Claude Code may reuse stale contents or abort. The immediate rmtree clears the slot so a fresh, coherent snapshot can be written. |
| 3. Downgrade guard to warning | **BLOCKED** | Depends on Recs 1 & 2 being implemented first. Without write-before-delete, downgrading the guard re-exposes the vulnerability documented in `aa302873`. |
| 4. Lazy --plugin-dir resolution | **BLOCKED** | `MarketplaceInstall` is a frozen dataclass. `ToolContext` is shared across concurrent async tasks. Lazy resolution would require unfreezing the dataclass or creating per-dispatch inconsistency (some food trucks use old path, some use new). The existing grace-period mechanism is the more robust approach. |

### Finding 2: hooks.json Discovery Mechanism Was Misidentified

**Refuted claim (from Round 1):** "Hook scripts are cache-immune because they use hardcoded source tree paths."

**Actual behavior:** In a marketplace install (the production case), hooks.json lives inside the versioned plugin cache directory at `~/.claude/plugins/cache/autoskillit-local/autoskillit/<version>/hooks/hooks.json`. Claude Code reads this file from `<installPath>/hooks/hooks.json` to discover which hooks to register. The hook commands inside that file point to **uv site-packages** paths (e.g., `python3 /home/talon/.local/share/uv/tools/autoskillit/lib/python3.13/site-packages/autoskillit/hooks/quota_guard.py`) — not the source tree.

This means:
- Hook **scripts** survive cache deletion (they're at site-packages paths)
- Hook **definitions** (which scripts to run, for which matchers) are lost when the cache directory is deleted — Claude Code can't find hooks.json, so all hooks fail open
- The grace period in `_retire_old_versions()` is what keeps hooks.json accessible for running sessions during cross-version transitions

The Round 1 claim that "hooks are cache-immune" was wrong. The hooks.json discovery mechanism IS the vulnerability.

### Finding 3: The Guard IS Load-Bearing — But Only for Same-Version Reinstalls

Cross-validator analysis of `_retire_old_versions()` reveals a critical nuance:

**For cross-version upgrades (e.g., 0.9.209 → 0.9.210):**
- `autoskillit update` runs `uv tool upgrade` first, changing the installed version
- `autoskillit install` then calls `_retire_old_versions(cache_dir, "0.9.210")`
- Old `0.9.209/` directory: `subdir.name != new_version` → `append_retiring_entry()` → **2-hour grace period**
- No existing `0.9.210/` directory (fresh upgrade) → nothing to rmtree
- Running sessions continue referencing `0.9.209/` which **survives for 2 hours** ✓
- `claude plugin install` creates fresh `0.9.210/` cache ✓

**For same-version reinstalls (e.g., `autoskillit install` without prior upgrade):**
- `_retire_old_versions(cache_dir, "0.9.209")` is called with the CURRENT version
- Existing `0.9.209/` directory: `subdir.name == new_version` → **immediate `shutil.rmtree`**
- Running sessions' hooks.json is destroyed → hooks fail open ✗

**Conclusion:** The cache rotation mechanism **already protects running sessions during genuine upgrades**. The kitchen guard is only truly needed for same-version reinstalls — which is NOT what `autoskillit update` does (it always upgrades first).

### Finding 4: Corrected Failure Mode Chain

The original commit message for `aa302873` describes a failure chain that was partially misidentified by the investigation. The validated chain is:

1. `autoskillit update` / `autoskillit install` runs
2. `_clear_plugin_cache()` → `_retire_old_versions()` processes cache directories
3. For cross-version: old directories get 2-hour grace period (safe)
4. For same-version: directory is immediately `rmtree`'d (unsafe)
5. Claude Code's in-memory `installPath` now points to a deleted directory
6. **hooks.json** at `<installPath>/hooks/hooks.json` is gone → hook dispatcher can't find definitions → hooks fail open silently
7. **Food truck sessions** launched with `--plugin-dir <old_path>` reference a deleted directory → Claude Code fails to load plugin → child session has no autoskillit tools

### Finding 5: Minor Factual Corrections

- `psutil` is a **hard dependency** in `_plugin_cache.py` (unconditional import, no `ImportError` guard) — claim C1 implied graceful degradation
- The `create_time` check in `_pid_alive()` fires for all successful `os.kill()` calls, not only on `PermissionError` — it's a PID-reuse guard in all non-dead cases
- `sweep_retiring_cache()` called at the end of `_retire_old_versions()` does NOT delete newly-added entries (they're younger than the 2-hour cutoff) — it only sweeps entries from prior retire cycles that have aged out

### Revised Recommendation (Post-Round-2)

Given that:
1. Write-before-delete is blocked by Claude Code's `plugin install` semantics
2. Same-version grace period is blocked by the same constraint
3. The cache rotation mechanism **already protects** cross-version upgrades
4. `autoskillit update` always does a version upgrade before install

The correct fix is:

**A. Remove the kitchen guard from `run_update_command()` only.** The explicit `autoskillit update` command always runs `uv tool upgrade` first, which changes the installed version. The subsequent `autoskillit install` call passes the NEW version to `_retire_old_versions()`, and old-version directories receive the 2-hour grace period. Running sessions are protected by cache rotation without needing the guard.

**B. Keep the kitchen guard on `run_update_checks()` but change from silent return to passive notification.** Instead of silently skipping, show: "Update available but cannot auto-apply while a kitchen is open. Run `autoskillit update` manually." This preserves the safety property (no auto-update mid-pipeline) while informing the user.

**C. Keep (or add) a kitchen guard on standalone `autoskillit install`.** Same-version reinstalls use immediate rmtree and WOULD corrupt running sessions. The guard belongs on `install`, not on `update` (which always changes the version first).

**D. Improve error messaging in the current guard.** If the guard does fire, show which sessions have kitchens open (PIDs, project paths, opened_at timestamps) so the user can close them. Currently it just says "A kitchen is currently open" with no actionable detail.

### Confidence Assessment

| Finding | Confidence | Evidence |
|---|---|---|
| Write-before-delete is blocked | HIGH | 4/4 validators confirmed; REQ-CACHE-003 is a hard constraint |
| Same-version rmtree grace period is blocked | HIGH | 3/4 validators confirmed; `claude plugin install` requires empty slot |
| Cross-version upgrades are protected by cache rotation | HIGH | Direct code trace through `_retire_old_versions()` logic confirmed by 2 validators |
| hooks.json is the vulnerability, not hook scripts | HIGH | 2/4 validators independently discovered the hooks.json-in-cache mechanism |
| Removing guard from `run_update_command()` is safe | MEDIUM | Depends on `uv tool upgrade` always changing the version — edge case: upgrade to same version (no-op) would still trigger same-version rmtree |

Trecek:
## Summary

The `any_kitchen_open()` function in `core/_plugin_cache.py` was a global boolean guard that reads `~/.autoskillit/active_kitchens.json` and returns `True` if *any* live-PID kitchen entry exists across the entire machine — regardless of which project that kitchen belongs to. This blocked `autoskillit update` (hard `SystemExit(1)`) and silently suppressed startup update checks for all sessions whenever any session on any project had a kitchen open.

The fix propagates the project-scoping pattern (already present in `mcp_health_guard.py`) into `any_kitchen_open()` and all its callers, adds the missing kitchen guard to `install()`, and introduces a structural arch test (`test_kitchen_guard_scoping.py`) that AST-walks all production call sites to enforce the `project_path=` kwarg contract permanently.

### Process Flow Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    %% TERMINALS %%
    START([PreToolUse: open_kitchen called])
    COMPLETE([Kitchen Open — marker written])
    ERROR([Denied / Blocked])

    %% ── PHASE 1: Kitchen Open Hook ── %%
    subgraph HookPhase ["● open_kitchen_guard.py — PreToolUse Hook"]
        direction TB
        H1{"Headless<br/>session?"}
        H2{"Session type<br/>== fleet?"}
        H3{"Session type<br/>== orchestrator?"}
        H4["Write kitchen marker<br/>━━━━━━━━━━<br/>kitchen_state/{session_id}.json<br/>includes project_path"]
    end

    %% ── PHASE 2: Registry (any_kitchen_open) ── %%
    subgraph RegistryPhase ["● _plugin_cache.py — any_kitchen_open() Guard"]
        direction TB
        R1["Read ~/.autoskillit/<br/>active_kitchens.json<br/>━━━━━━━━━━<br/>fcntl LOCK_EX"]
        R2["Iterate entries<br/>━━━━━━━━━━<br/>prune dead PIDs<br/>via _pid_alive()"]
        R3{"project_path<br/>arg provided?"}
        R4{"Any entry matches<br/>project_path?"}
        R5{"Any live<br/>entry exists?"}
        R6["Write survivors back<br/>━━━━━━━━━━<br/>active_kitchens.json"]
    end
    OPEN_TRUE(["→ True: kitchen open"])
    OPEN_FALSE(["→ False: no kitchen open"])

    %% ── PHASE 3: Update Command ── %%
    subgraph UpdateCmd ["● _update.py — run_update_command()"]
        direction TB
        U1["● any_kitchen_open<br/>━━━━━━━━━━<br/>project_path=cwd"]
        U2{"Kitchen open<br/>for project?"}
        U3["Run upgrade subprocess<br/>+ autoskillit install<br/>━━━━━━━━━━<br/>SKIP_UPDATE_CHECK env set"]
        U4["Invalidate fetch cache<br/>+ perform_restart()"]
    end

    %% ── PHASE 4: Update Checks ── %%
    subgraph UpdateChecks ["● _update_checks.py — run_update_checks()"]
        direction TB
        C1{"CLAUDECODE / CI /<br/>skip env / non-TTY?"}
        C2["● any_kitchen_open<br/>━━━━━━━━━━<br/>project_path=cwd"]
        C3{"Kitchen open<br/>for project?"}
        C4["Gather 4 signals<br/>━━━━━━━━━━<br/>binary / hooks /<br/>source drift / dual-mcp"]
        C5{"Undismissed<br/>signals?"}
        C6["30s [Y/n] prompt<br/>━━━━━━━━━━<br/>timeout = silent return"]
    end

    %% ── PHASE 5: Install / Cache Clear ── %%
    subgraph InstallPhase ["● _marketplace.py — install()"]
        direction TB
        I1["● any_kitchen_open<br/>━━━━━━━━━━<br/>project_path=cwd<br/>(under _InstallLock)"]
        I2{"Kitchen open<br/>for project?"}
        I3["SKIP cache clear<br/>━━━━━━━━━━<br/>avoid disrupting<br/>running sessions"]
        I4["Clear plugin cache<br/>━━━━━━━━━━<br/>remove snapshot +<br/>installed_plugins.json entry"]
        I5["Write symlink,<br/>marketplace.json,<br/>hooks.json, settings"]
    end

    %% ── PHASE 6: Arch Test Enforcement ── %%
    subgraph ArchTest ["★ test_kitchen_guard_scoping.py — Arch Test"]
        direction TB
        T1["AST walk src/**/*.py<br/>━━━━━━━━━━<br/>find all any_kitchen_open()<br/>call sites"]
        T2{"project_path=<br/>kwarg present<br/>at every call?"}
        T3["Verify test helpers<br/>monkeypatch any_kitchen_open<br/>━━━━━━━━━━<br/>prevent real FS reads in tests"]
    end

    %% ── FLOWS ── %%

    START --> H1
    H1 -->|"yes"| H2
    H1 -->|"no"| H4
    H2 -->|"yes"| ERROR
    H2 -->|"no"| H3
    H3 -->|"no (leaf/unset)"| ERROR
    H3 -->|"yes"| H4
    H4 --> COMPLETE

    %% Registry consumed by callers %%
    COMPLETE -.->|"marker written; callers read registry"| R1
    R1 --> R2
    R2 --> R3
    R3 -->|"yes"| R4
    R3 -->|"no (global check)"| R5
    R4 -->|"match found"| R6
    R4 -->|"no match"| R6
    R5 -->|"live entry found"| R6
    R5 -->|"none"| R6
    R6 -->|"match found"| OPEN_TRUE
    R6 -->|"no match / none"| OPEN_FALSE

    %% Update command flow %%
    U1 --> U2
    U2 -->|"open"| ERROR
    U2 -->|"not open"| U3
    U3 --> U4

    %% Update checks flow %%
    C1 -->|"yes"| OPEN_FALSE
    C1 -->|"no"| C2
    C2 --> C3
    C3 -->|"open"| OPEN_TRUE
    C3 -->|"not open"| C4
    C4 --> C5
    C5 -->|"yes"| C6
    C5 -->|"all dismissed"| OPEN_FALSE

    %% Install flow %%
    I1 --> I2
    I2 -->|"open"| I3
    I2 -->|"not open"| I4
    I3 --> I5
    I4 --> I5

    %% Arch test flow %%
    T1 --> T2
    T2 -->|"all present — PASS"| T3
    T2 -->|"missing kwarg — FAIL"| ERROR

    %% CLASS ASSIGNMENTS %%
    class START,COMPLETE terminal;
    class ERROR detector;
    class H1,H2,H3 stateNode;
    class H4 handler;
    class R1,R2,R6 handler;
    class R3,R4,R5 stateNode;
    class OPEN_TRUE,OPEN_FALSE output;
    class U1,C2,I1 phase;
    class U2,C1,C3,C5,I2 stateNode;
    class U3,U4,C4,C6,I3,I4,I5 handler;
    class T1,T3 newComponent;
    class T2 detector;
```

### State Lifecycle Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    START([OPEN / UPDATE / INSTALL OPERATION])

    %% ── KITCHEN MARKER STATE ──────────────────────────────────────────── %%
    subgraph KitchenMarkerState ["KITCHEN MARKER STATE  (kitchen_state.py)"]
        direction TB
        KM["KitchenMarker ●<br/>━━━━━━━━━━<br/>INIT_ONLY (frozen=True)<br/>session_id · opened_at<br/>recipe_name · marker_version<br/>content_hash · composite_hash"]
        KMWrite["● write_marker()<br/>━━━━━━━━━━<br/>mkstemp + os.replace<br/>Atomic overwrite only"]
        KMRead["read_marker(session_id)<br/>━━━━━━━━━━<br/>Returns None on:<br/>missing · bad JSON · schema mismatch"]
        KMTTL["is_marker_fresh()<br/>━━━━━━━━━━<br/>TTL gate: age < 24 h<br/>Stale → treat as absent"]
        KMSweep["sweep_stale_markers()<br/>━━━━━━━━━━<br/>Deletes age ≥ 24 h<br/>or malformed files"]
    end

    %% ── ACTIVE KITCHENS REGISTRY ─────────────────────────────────────── %%
    subgraph ActiveKitchenReg ["● ACTIVE KITCHENS REGISTRY  (_plugin_cache.py)"]
        direction TB
        AKEntry["● active_kitchens.json entry<br/>━━━━━━━━━━<br/>MUTABLE (atomic rewrite)<br/>kitchen_id · pid · create_time<br/>project_path · opened_at"]
        PIDGate["_pid_alive(pid, create_time)<br/>━━━━━━━━━━<br/>os.kill probe → PermissionError<br/>→ psutil create_time ±1 s check<br/>Guards PID-reuse false positives"]
        AKOpen["● any_kitchen_open(project_path=)<br/>━━━━━━━━━━<br/>LOCK_EX flock<br/>Lazy GC: prune dead PIDs<br/>Filter: entry[project_path] == arg<br/>Returns bool (scoped to project)"]
        AKArch["★ test_kitchen_guard_scoping<br/>━━━━━━━━━━<br/>AST-scans all callers<br/>VIOLATION if project_path= absent<br/>Enforces no bare any_kitchen_open()"]
    end

    %% ── PLUGIN CACHE LIFECYCLE ───────────────────────────────────────── %%
    subgraph PluginCache ["PLUGIN CACHE LIFECYCLE  (_plugin_cache.py / _marketplace.py)"]
        direction TB
        RetCache["retiring_cache.json<br/>━━━━━━━━━━<br/>APPEND_ONLY entries<br/>version · path · retired_at"]
        RetAppend["append_retiring_entry()<br/>━━━━━━━━━━<br/>Read-append-atomic-write<br/>Never overwrites existing"]
        RetSweep["sweep_retiring_cache(grace_hours=2)<br/>━━━━━━━━━━<br/>Deletes dirs older than grace period<br/>Survivors rewritten atomically"]
        CacheClearGate["● _clear_plugin_cache() gate<br/>━━━━━━━━━━<br/>any_kitchen_open(project_path=cwd)<br/>→ True: SKIP cache clear<br/>→ False: retire old dirs + sweep"]
    end

    %% ── UPDATE CHECK STATE ───────────────────────────────────────────── %%
    subgraph UpdateCheckState ["● UPDATE CHECK STATE  (_update_checks.py / _update.py)"]
        direction TB
        DismissState["update_check.json<br/>━━━━━━━━━━<br/>MUTABLE (key-level replace)<br/>update_prompt: {dismissed_at,<br/>dismissed_version, conditions}<br/>binary_snoozed (INIT_PRESERVE)"]
        FetchCache["github_fetch_cache.json<br/>━━━━━━━━━━<br/>MUTABLE per-URL entry<br/>body · etag · cached_at<br/>installed_version (version guard)"]
        DismissExpiry["● _is_dismissed() gate<br/>━━━━━━━━━━<br/>Expired if: time window elapsed<br/>OR version upgraded<br/>OR new Signal.kind not in conditions"]
        KitchenGate["● Kitchen-scope update gate<br/>━━━━━━━━━━<br/>any_kitchen_open(project_path=cwd)<br/>_update_checks: soft skip + msg<br/>_marketplace: skip cache-clear only<br/>_update: hard SystemExit(1)"]
        UpdateSuccess["● On successful update<br/>━━━━━━━━━━<br/>pop(update_prompt)<br/>pop(binary_snoozed)<br/>invalidate_fetch_cache()<br/>perform_restart()"]
    end

    %% ── VALIDATION GATE FLOW ─────────────────────────────────────────── %%
    subgraph ValidationGates ["VALIDATION GATES (ordered)"]
        direction TB
        G1["Env / TTY guards<br/>━━━━━━━━━━<br/>CLAUDECODE · CI · SKIP_ vars<br/>stdin/stdout isatty()"]
        G2["● Project-scope kitchen guard<br/>━━━━━━━━━━<br/>any_kitchen_open(project_path=cwd)<br/>Blocks mutation when kitchen live"]
        G3["Install type guard<br/>━━━━━━━━━━<br/>LOCAL_EDITABLE · LOCAL_PATH<br/>UNKNOWN → skip unless FORCE"]
        G4["Worktree guard<br/>━━━━━━━━━━<br/>is_git_worktree(pkg_dir)<br/>→ SystemExit (no symlink from worktree)"]
        G5["PID-reuse guard<br/>━━━━━━━━━━<br/>create_time ±1 s tolerance<br/>Prevents ghost kitchen entries"]
    end

    %% ── CONNECTIONS ─────────────────────────────────────────────────── %%
    START --> G1
    G1 -->|"pass"| G2
    G2 -->|"no kitchen open<br/>(scoped to project)"| G3
    G2 -->|"kitchen OPEN → block / skip"| KitchenGate
    G3 -->|"installable type"| G4
    G4 -->|"not worktree"| KMWrite
    KMWrite --> KM
    KM --> KMRead
    KMRead --> KMTTL
    KMTTL -->|"fresh"| AKEntry
    KMTTL -->|"stale"| KMSweep

    AKEntry --> PIDGate
    PIDGate -->|"alive + create_time match"| AKOpen
    PIDGate -->|"dead / PID-reused"| G5
    G5 --> AKOpen
    AKOpen -->|"project_path= required"| AKArch
    AKOpen -->|"returns bool"| G2

    G3 -->|"cache ops allowed"| CacheClearGate
    CacheClearGate -->|"no kitchen open"| RetAppend
    RetAppend --> RetCache
    RetCache --> RetSweep

    G1 -->|"pass"| DismissState
    DismissState --> DismissExpiry
    DismissExpiry -->|"not dismissed"| FetchCache
    FetchCache -->|"update available"| KitchenGate
    KitchenGate -->|"kitchen closed"| UpdateSuccess
    UpdateSuccess --> DismissState

    %% CLASS ASSIGNMENTS %%
    class KM stateNode;
    class KMWrite,KMRead handler;
    class KMTTL,KMSweep detector;
    class AKEntry stateNode;
    class PIDGate,G5 detector;
    class AKOpen phase;
    class AKArch newComponent;
    class RetCache stateNode;
    class RetAppend,RetSweep handler;
    class CacheClearGate phase;
    class DismissState,FetchCache stateNode;
    class DismissExpiry,KitchenGate detector;
    class UpdateSuccess output;
    class G1,G2,G3,G4 detector;
    class START terminal;
```

### Operational Diagram

```mermaid
flowchart TB
    %% CLASS DEFINITIONS %%

    subgraph CLILayer ["CLI ENTRY POINTS"]
        direction TB
        UPDATE["● autoskillit update<br/>━━━━━━━━━━<br/>_update.py<br/>Blocks if kitchen open<br/>(project-scoped)"]
        INSTALL["● autoskillit install<br/>━━━━━━━━━━<br/>_marketplace.py<br/>Skips cache clear<br/>if kitchen open"]
        SESSION["autoskillit cook / order<br/>━━━━━━━━━━<br/>Interactive or headless<br/>Triggers startup checks"]
    end

    subgraph StartupCheck ["STARTUP UPDATE CHECK (_update_checks.py ●)"]
        direction TB
        RUN_CHECKS["● run_update_checks()<br/>━━━━━━━━━━<br/>Signals: binary · hooks · drift · dual-mcp<br/>Skips (non-fatal) if kitchen open<br/>Skips entirely: CI=1, CLAUDECODE=1,<br/>non-TTY, SKIP_UPDATE_CHECK=1"]
    end

    subgraph GuardLayer ["KITCHEN GUARD (_plugin_cache.py ●)"]
        direction TB
        ANY_KITCHEN["● any_kitchen_open(<br/>  project_path=str(Path.cwd()))<br/>━━━━━━━━━━<br/>Prunes dead PIDs on each read<br/>Scoped to current project<br/>(PR change: was global)"]
        REGISTRY["active_kitchens.json<br/>━━━━━━━━━━<br/>~/.autoskillit/<br/>PID · kitchen_id · project_path<br/>+ psutil create time<br/>(fcntl-locked)"]
    end

    subgraph StateFiles ["STATE & CACHE (read/write)"]
        direction TB
        UPDATE_STATE["update_check.json<br/>━━━━━━━━━━<br/>~/.autoskillit/<br/>update_prompt · binary_snoozed<br/>Dismissal windows: 7d / 12h"]
        GH_CACHE["github_fetch_cache.json<br/>━━━━━━━━━━<br/>~/.autoskillit/<br/>GitHub API response cache<br/>TTL 30 min · ETag-conditional"]
        KITCHEN_MARKERS["kitchen_state/{session}.json<br/>━━━━━━━━━━<br/>.autoskillit/temp/<br/>Session open marker<br/>(written by open_kitchen_guard hook)"]
    end

    subgraph ArchGuard ["ARCHITECTURE TEST (★ new)"]
        direction TB
        SCOPE_TEST["★ test_kitchen_guard_scoping.py<br/>━━━━━━━━━━<br/>AST scan: ALL any_kitchen_open callers<br/>in src/ must pass project_path=<br/>Test helpers must mock any_kitchen_open<br/>(prevents real ~/.autoskillit reads in tests)"]
    end

    %% FLOWS %%
    UPDATE -->|"SystemExit(1)<br/>if open"| ANY_KITCHEN
    INSTALL -->|"skips _clear_plugin_cache<br/>if open"| ANY_KITCHEN
    SESSION --> RUN_CHECKS
    RUN_CHECKS -->|"silent skip<br/>if open"| ANY_KITCHEN
    ANY_KITCHEN -->|"reads + prunes"| REGISTRY
    KITCHEN_MARKERS -.->|"register_active_kitchen<br/>on open_kitchen call"| REGISTRY
    RUN_CHECKS -->|"reads dismissal<br/>writes snooze"| UPDATE_STATE
    RUN_CHECKS -->|"reads cached<br/>GitHub data"| GH_CACHE
    UPDATE -->|"clears on<br/>successful upgrade"| UPDATE_STATE
    UPDATE -->|"invalidates<br/>on upgrade"| GH_CACHE

    SCOPE_TEST -.->|"enforces scoping<br/>contract at test-time"| ANY_KITCHEN

    %% CLASS ASSIGNMENTS %%
    class UPDATE,INSTALL,SESSION cli;
    class RUN_CHECKS handler;
    class ANY_KITCHEN detector;
    class REGISTRY stateNode;
    class UPDATE_STATE,GH_CACHE,KITCHEN_MARKERS output;
    class SCOPE_TEST newComponent;
```

Closes #1415

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260428-070957-123956/.autoskillit/temp/rectify/rectify_kitchen-open-guard-over-scoped_2026-04-28_143500.md`

## Token Usage Summary

| plan | 77 | 25.7k | 1.5M | 85.0k | 1 | 14m 51s |
| verify | 4.1k | 11.0k | 1.2M | 55.0k | 1 | 5m 13s |
| implement | 4.1k | 33.5k | 7.2M | 127.7k | 1 | 21m 26s |
| **Total** | 8.3k | 70.2k | 9.9M | 267.8k | | 41m 31s |

### DIFF
--- a/src/autoskillit/core/_plugin_cache.py
+++ b/src/autoskillit/core/_plugin_cache.py
@@ -237,7 +237,7 @@ def clear_kitchens_for_pid(pid: int) -> None:
         fh.close()
 
 
-def any_kitchen_open() -> bool:
+def any_kitchen_open(project_path: str | None = None) -> bool:
     akp = _active_kitchens_path()
     lock = _active_kitchens_lock()
     if not akp.exists():
@@ -264,6 +264,8 @@ def any_kitchen_open() -> bool:
                 write_versioned_json(akp, {"kitchens": survivors}, schema_version=_SCHEMA_VERSION)
             except OSError as exc:
                 logger.warning("any_kitchen_open: failed to persist pruned kitchens: %s", exc)
+        if project_path is not None:
+            return any(entry.get("project_path") == project_path for entry in survivors)
         return len(survivors) > 0
     finally:
         fh.close()
--- a/src/autoskillit/core/_plugin_cache.py
+++ b/src/autoskillit/core/_plugin_cache.py
@@ -265,7 +265,10 @@ def any_kitchen_open(project_path: str | None = None) -> bool:
             except OSError as exc:
                 logger.warning("any_kitchen_open: failed to persist pruned kitchens: %s", exc)
         if project_path is not None:
-            return any(entry.get("project_path") == project_path for entry in survivors)
+            return any(
+                entry.get("project_path") is None or entry.get("project_path") == project_path
+                for entry in survivors
+            )
         return len(survivors) > 0
     finally:
         fh.close()
--- a/tests/arch/test_kitchen_guard_scoping.py
+++ b/tests/arch/test_kitchen_guard_scoping.py
@@ -1,0 +1,86 @@
+"""Architectural enforcement: any_kitchen_open call-site scoping and test helper isolation."""
+
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+import pytest
+
+pytestmark = [pytest.mark.layer("arch")]
+
+_SRC_ROOT = Path(__file__).parents[2] / "src" / "autoskillit"
+_TEST_ROOT = Path(__file__).parents[2] / "tests"
+
+# File that defines any_kitchen_open — not a call site.
+_DEFINITION_FILE = "core/_plugin_cache.py"
+
+
+def _has_project_path_kwarg(call_node: ast.Call) -> bool:
+    return any(kw.arg == "project_path" for kw in call_node.keywords)
+
+
+def _is_any_kitchen_open_call(call_node: ast.Call) -> bool:
+    func = call_node.func
+    if isinstance(func, ast.Name):
+        return func.id == "any_kitchen_open"
+    if isinstance(func, ast.Attribute):
+        return func.attr == "any_kitchen_open"
+    return False
+
+
+def test_any_kitchen_open_callers_pass_project_path() -> None:
+    violations: list[str] = []
+    for py_file in sorted(_SRC_ROOT.rglob("*.py")):
+        rel = py_file.relative_to(_SRC_ROOT)
+        if str(rel) == _DEFINITION_FILE:
+            continue
+        source = py_file.read_text(encoding="utf-8")
+        if "any_kitchen_open" not in source:
+            continue
+        tree = ast.parse(source, filename=str(py_file))
+        for node in ast.walk(tree):
+            if isinstance(node, ast.Call) and _is_any_kitchen_open_call(node):
+                if not _has_project_path_kwarg(node):
+                    violations.append(f"{rel}:{node.lineno}")
+    assert not violations, (
+        f"any_kitchen_open called without project_path= at: {violations}. "
+        "All production callers must pass project_path to scope the guard to the current project."
+    )
+
+
+def _function_body_has_any_kitchen_open_patch(func_node: ast.FunctionDef) -> bool:
+    """Return True if the function contains monkeypatch.setattr with 'any_kitchen_open'."""
+    for node in ast.walk(func_node):
+        if not isinstance(node, ast.Call):
+            continue
+        func = node.func
+        if not (isinstance(func, ast.Attribute) and func.attr == "setattr"):
+            continue
+        for arg in node.args:
+            if isinstance(arg, ast.Constant) and "any_kitchen_open" in str(arg.value):
+                return True
+    return False
+
+
+def test_setup_helpers_must_patch_any_kitchen_open() -> None:
+    helpers = [
+        (_TEST_ROOT / "cli" / "test_update_command.py", "_setup_run_update"),
+        (_TEST_ROOT / "cli" / "test_update_checks_prompt.py", "_setup_run_checks"),
+    ]
+    missing: list[str] = []
+    for file_path, func_name in helpers:
+        source = file_path.read_text(encoding="utf-8")
+        tree = ast.parse(source, filename=str(file_path))
+        for node in ast.walk(tree):
+            if isinstance(node, ast.FunctionDef) and node.name == func_name:
+                if not _function_body_has_any_kitchen_open_patch(node):
+                    missing.append(f"{file_path.name}::{func_name}")
+                break
+        else:
+            missing.append(f"{file_path.name}::{func_name} (function not found)")
+    assert not missing, (
+        f"Test helpers do not patch any_kitchen_open: {missing}. "
+        "Add monkeypatch.setattr('autoskillit.core.any_kitchen_open', lambda **kw: False) "
+        "to prevent silent dependency on real $HOME kitchen state."
+    )
--- a/tests/arch/test_kitchen_guard_scoping.py
+++ b/tests/arch/test_kitchen_guard_scoping.py
@@ -57,9 +57,12 @@ def _function_body_has_any_kitchen_open_patch(func_node: ast.FunctionDef) -> boo
         func = node.func
         if not (isinstance(func, ast.Attribute) and func.attr == "setattr"):
             continue
-        for arg in node.args:
-            if isinstance(arg, ast.Constant) and "any_kitchen_open" in str(arg.value):
-                return True
+        if (
+            len(node.args) >= 2
+            and isinstance(node.args[1], ast.Constant)
+            and "any_kitchen_open" in str(node.args[1].value)
+        ):
+            return True
     return False
 
 
--- a/tests/arch/test_kitchen_guard_scoping.py
+++ b/tests/arch/test_kitchen_guard_scoping.py
@@ -57,10 +57,13 @@ def _function_body_has_any_kitchen_open_patch(func_node: ast.FunctionDef) -> boo
         func = node.func
         if not (isinstance(func, ast.Attribute) and func.attr == "setattr"):
             continue
+        # 2-arg string form: setattr("dotted.path.name", value) → name at args[0]
+        # 3-arg form: setattr(target, "name", value) → name at args[1]
+        name_arg_idx = 0 if len(node.args) == 2 else 1
         if (
             len(node.args) >= 2
-            and isinstance(node.args[1], ast.Constant)
-            and "any_kitchen_open" in str(node.args[1].value)
+            and isinstance(node.args[name_arg_idx], ast.Constant)
+            and "any_kitchen_open" in str(node.args[name_arg_idx].value)
         ):
             return True
     return False
--- a/src/autoskillit/cli/_marketplace.py
+++ b/src/autoskillit/cli/_marketplace.py
@@ -165,7 +165,12 @@ def install(*, scope: str = "user") -> bool:
     from autoskillit.core import _InstallLock
 
     with _InstallLock():
-        _clear_plugin_cache()
+        from autoskillit.core import any_kitchen_open
+
+        if any_kitchen_open(project_path=str(Path.cwd())):
+            print("Kitchen open for this project — skipping plugin cache clear.")
+        else:
+            _clear_plugin_cache()
 
         # Regenerate hooks.json from the canonical registry with absolute paths
         hooks_json_path = pkg_root() / "hooks" / "hooks.json"
--- a/src/autoskillit/cli/_update.py
+++ b/src/autoskillit/cli/_update.py
@@ -41,8 +41,11 @@ def run_update_command(home: Path | None = None) -> None:
 
     from autoskillit.core import any_kitchen_open
 
-    if any_kitchen_open():
-        print("A kitchen is currently open. Close it or wait for the pipeline to finish.")
+    if any_kitchen_open(project_path=str(Path.cwd())):
+        print(
+            "A kitchen is currently open for this project. "
+            "Close it or wait for the pipeline to finish.",
+        )
         raise SystemExit(1)
 
     info = detect_install()
--- a/src/autoskillit/cli/_update_checks.py
+++ b/src/autoskillit/cli/_update_checks.py
@@ -690,7 +690,11 @@ def run_update_checks(home: Path | None = None) -> None:
 
     from autoskillit.core import any_kitchen_open  # noqa: PLC0415
 
-    if any_kitchen_open():
+    if any_kitchen_open(project_path=str(Path.cwd())):
+        print(
+            "Skipping update check: a kitchen is open for this project. "
+            "Run 'autoskillit update' manually after the pipeline finishes.",
+        )
         return
 
     _skip_env: dict[str, str] = {
--- a/src/autoskillit/recipes/diagrams/implementation.md
+++ b/src/autoskillit/recipes/diagrams/implementation.md
@@ -1,158 +1,31 @@
-<!-- autoskillit-recipe-hash: sha256:dcb5955bae118041160b3100f4837bb735caadfcd1aad6cb40b1ba346a11902f -->
+<!-- autoskillit-recipe-hash: sha256:d47cab45ad7b2d7b64bb3dd723cdbbf2cbc98dceb71ceea8b533646da81bbdda -->
 <!-- autoskillit-diagram-format: v7 -->
 # implementation
 
-```mermaid
-flowchart TD
-    S0[clone]
-    S1[capture_base_sha]
-    S0 --> S1
-    S2[get_issue_title]
-    S1 --> S2
-    S3[claim_issue]
-    S2 --> S3
-    S4[compute_branch]
-    S3 --> S4
-    S5[create_branch]
-    S4 --> S5
-    S6[push_merge_target]
-    S5 --> S6
-    S7[plan]
-    S6 --> S7
-    S8[review]
-    S7 --> S8
-    S9[verify]
-    S8 --> S9
-    S10[implement]
-    S9 --> S10
-    S11[retry_worktree]
-    S10 --> S11
-    S12[test]
-    S11 --> S12
-    S13[commit_guard]
-    S12 --> S13
-    S14[merge]
-    S13 --> S14
-    S15[push]
-    S14 --> S15
-    S16[fix]
-    S15 --> S16
-    S17[next_or_done]
-    S16 --> S17
-    S18[audit_impl]
-    S17 --> S18
-    S19[remediate]
-    S18 --> S19
-    S20[prepare_pr]
-    S19 --> S20
-    S21[run_arch_lenses]
-    S20 --> S21
-    S22[compose_pr]
-    S21 --> S22
-    S23[extract_pr_number]
-    S22 --> S23
-    S24[annotate_pr_diff]
-    S23 --> S24
-    S25[review_pr]
-    S24 --> S25
-    S26[resolve_review]
-    S25 --> S26
-    S27[re_push_review]
-    S26 --> S27
-    S28[check_review_loop]
-    S27 --> S28
-    S29[check_repo_ci_event]
-    S28 --> S29
-    S30[check_pr_state]
-    S29 --> S30
-    S31[ci_watch]
-    S30 --> S31
-    S32[handle_no_ci_runs]
-    S31 --> S32
-    S33[check_ci_loop]
-    S32 --> S33
-    S34[check_active_trigger_loop]
-    S33 --> S34
-    S35[trigger_ci_actively]
-    S34 --> S35
-    S36[escalate_stop_no_ci]
-    S35 --> S36
-    S37[check_repo_merge_state]
-    S36 --> S37
-    S38[route_queue_mode]
-    S37 --> S38
-    S39[enqueue_to_queue]
-    S38 --> S39
-    S40[verify_queue_enrollment]
-    S39 --> S40
-    S41[wait_for_queue]
-    S40 --> S41
-    S42[reenroll_stalled_pr]
-    S41 --> S42
-    S43[check_stall_loop]
-    S42 --> S43
-    S44[check_eject_limit]
-    S43 --> S44
-    S45[queue_ejected_fix]
-    S44 --> S45
-    S46[resolve_queue_merge_conflicts]
-    S45 --> S46
-    S47[re_push_queue_fix]
-    S46 --> S47
-    S48[ci_watch_post_queue_fix]
-    S47 --> S48
-    S49[reenter_merge_queue]
-    S48 --> S49
-    S50[reenter_merge_queue_cheap]
-    S49 --> S50
-    S51[direct_merge]
-    S50 --> S51
-    S52[wait_for_direct_merge]
-    S51 --> S52
-    S53[direct_merge_conflict_fix]
-    S52 --> S53
-    S54[resolve_direct_merge_conflicts]
-    S53 --> S54
-    S55[re_push_direct_fix]
-    S54 --> S55
-    S56[redirect_merge]
-    S55 --> S56
-    S57[immediate_merge]
-    S56 --> S57
-    S58[wait_for_immediate_merge]
-    S57 --> S58
-    S59[immediate_merge_conflict_fix]
-    S58 --> S59
-    S60[resolve_immediate_merge_conflicts]
-    S59 --> S60
-    S61[re_push_immediate_fix]
-    S60 --> S61
-    S62[remerge_immediate]
-    S61 --> S62
-    S63[diagnose_ci]
-    S62 --> S63
-    S64[resolve_ci]
-    S63 --> S64
-    S65[pre_resolve_rebase]
-    S64 --> S65
-    S66[re_push]
-    S65 --> S66
-    S67[detect_ci_conflict]
-    S66 --> S67
-    S68[ci_conflict_fix]
-    S67 --> S68
-    S69[release_issue_success]
-    S68 --> S69
-    S70[register_clone_unconfirmed]
-    S69 --> S70
-    S71[release_issue_failure]
-    S70 --> S71
-    S72[register_clone_success]
-    S71 --> S72
-    S73[register_clone_failure]
-    S72 --> S73
-    S74[done]
-    S73 --> S74
-    S75[escalate_stop]
-    S74 --> S75
+```
+      clone → get_issue_title → claim_issue → compute_branch
+      |
+      +-- create_branch (optional)
+      |
+      plan
+      |
+      +-- review (optional)
+      |
+ +----+ FOR EACH PLAN PART:
+ |    |
+ |    verify → implement → test ↔ [fix on failure]
+ |    |
+ |    merge → push → next_or_done
+ |
+ +----+
+      |
+      +-- audit_impl → remediate (optional)
+      |
+      +-- [open-pr] (optional):
+      |     prepare_pr → compose_pr → review_pr → resolve_review
+      |     ci_watch → check_repo_merge_state
+      |     → [queue | direct | immediate] merge path
+      |     → diagnose_ci → resolve_ci (on CI failure)
+      |
+      release_issue_success / release_issue_failure
 ```
--- a/src/autoskillit/recipes/diagrams/implementation.md
+++ b/src/autoskillit/recipes/diagrams/implementation.md
@@ -1,31 +1,158 @@
-<!-- autoskillit-recipe-hash: sha256:d47cab45ad7b2d7b64bb3dd723cdbbf2cbc98dceb71ceea8b533646da81bbdda -->
+<!-- autoskillit-recipe-hash: sha256:dcb5955bae118041160b3100f4837bb735caadfcd1aad6cb40b1ba346a11902f -->
 <!-- autoskillit-diagram-format: v7 -->
 # implementation
 
-```
-      clone → get_issue_title → claim_issue → compute_branch
-      |
-      +-- create_branch (optional)
-      |
-      plan
-      |
-      +-- review (optional)
-      |
- +----+ FOR EACH PLAN PART:
- |    |
- |    verify → implement → test ↔ [fix on failure]
- |    |
- |    merge → push → next_or_done
- |
- +----+
-      |
-      +-- audit_impl → remediate (optional)
-      |
-      +-- [open-pr] (optional):
-      |     prepare_pr → compose_pr → review_pr → resolve_review
-      |     ci_watch → check_repo_merge_state
-      |     → [queue | direct | immediate] merge path
-      |     → diagnose_ci → resolve_ci (on CI failure)
-      |
-      release_issue_success / release_issue_failure
+```mermaid
+flowchart TD
+    S0[clone]
+    S1[capture_base_sha]
+    S0 --> S1
+    S2[get_issue_title]
+    S1 --> S2
+    S3[claim_issue]
+    S2 --> S3
+    S4[compute_branch]
+    S3 --> S4
+    S5[create_branch]
+    S4 --> S5
+    S6[push_merge_target]
+    S5 --> S6
+    S7[plan]
+    S6 --> S7
+    S8[review]
+    S7 --> S8
+    S9[verify]
+    S8 --> S9
+    S10[implement]
+    S9 --> S10
+    S11[retry_worktree]
+    S10 --> S11
+    S12[test]
+    S11 --> S12
+    S13[commit_guard]
+    S12 --> S13
+    S14[merge]
+    S13 --> S14
+    S15[push]
+    S14 --> S15
+    S16[fix]
+    S15 --> S16
+    S17[next_or_done]
+    S16 --> S17
+    S18[audit_impl]
+    S17 --> S18
+    S19[remediate]
+    S18 --> S19
+    S20[prepare_pr]
+    S19 --> S20
+    S21[run_arch_lenses]
+    S20 --> S21
+    S22[compose_pr]
+    S21 --> S22
+    S23[extract_pr_number]
+    S22 --> S23
+    S24[annotate_pr_diff]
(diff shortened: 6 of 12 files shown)


src/autoskillit/core/_plugin_cache.py
```diff
@@ -264,6 +264,8 @@ def any_kitchen_open() -> bool:
                 write_versioned_json(akp, {"kitchens": survivors}, schema_version=_SCHEMA_VERSION)
             except OSError as exc:
                 logger.warning("any_kitchen_open: failed to persist pruned kitchens: %s", exc)
+        if project_path is not None:
```
Trecek, line 267: [warning] bugs: Old-format kitchen entries (written before this PR) lack the project_path field; entry.get("project_path") returns None, so any_kitchen_open(project_path=x) returns False for them even if they represent a kitchen for that project. This creates a backward compatibility gap: any in-flight kitchen started before upgrading will not block the update check.

tests/arch/test_kitchen_guard_scoping.py
```diff
@@ -0,0 +1,86 @@
...
+            if isinstance(node, ast.Call) and _is_any_kitchen_open_call(node):
+                if not _has_project_path_kwarg(node):
+                    violations.append(f"{rel}:{node.lineno}")
+    assert not violations, (
+        f"any_kitchen_open called without project_path= at: {violations}. "
+        "All production callers must pass project_path to scope the guard to the current project."
+    )
+
+
+def _function_body_has_any_kitchen_open_patch(func_node: ast.FunctionDef) -> bool:
```
Trecek, line 52: [warning] arch: _function_body_has_any_kitchen_open_patch scans all positional args for the substring "any_kitchen_open", matching even comment-like constants unrelated to the patch target. Consider checking only keyword arg or the attribute-name position (args[1] for 3-arg setattr form) to avoid false positives.

Trecek (commented):

Trecek (commented):

**Info findings (not blocking):**

**src/autoskillit/cli/_update_checks.py L693**
- [info/arch]: `Path.cwd()` is used as project_path for the kitchen guard, but if the CLI is invoked from a subdirectory the cwd will not match the project root stored in the kitchen registry, silently skipping the guard. Same concern applies to `_update.py` L44 and `_marketplace.py` L170.

**src/autoskillit/cli/_marketplace.py L171**
- [info/arch]: When `any_kitchen_open` returns True, `_clear_plugin_cache()` is skipped but the rest of `install()` (hooks.json regeneration, subprocess.run, etc.) continues to execute. If cache clearing is a prerequisite for idempotent hook regeneration, this partial execution could leave the system in an inconsistent state. Consider returning early or determining whether a full skip is safer.

src/autoskillit/core/_plugin_cache.py
Trecek: Agreed — fixed in a79aeac1. Old-format entries (lacking project_path) now match conservatively — treated as matching any project so in-flight kitchens from before the upgrade still block the guard. See _plugin_cache.py:267-270.

tests/arch/test_kitchen_guard_scoping.py
Trecek: Agreed — fixed in 877950a4. Now checks only the name-position arg: args[0] for 2-arg string form (monkeypatch.setattr("dotted.path", val)) and args[1] for 3-arg form (monkeypatch.setattr(target, "name", val)), avoiding false positives from constants in other positions.

Issue #1581, opened 2026-05-01
Fix: Add hooks/leaf_orchestration_guard.py to import-layer label test skip list


Trecek:
## Problem

PR #1580 (import-layer IL-N rename, issue #1574) is repeatedly ejected from the merge queue. The PR's own branch CI passes, but the merge-group branch CI fails every time — 5 consecutive failures recorded.

## Root Cause

A **cross-PR test conflict** between #1579 and #1580:

- **PR #1579** (commit `52973582`) updated the docstring in `src/autoskillit/hooks/leaf_orchestration_guard.py:8` from `"Tier invariant: orchestrator and fleet tiers may call orchestration tools."` to `"L2+ invariant: orchestrator (L2) and fleet (L3) sessions may call orchestration tools."` — correctly adopting the canonical L0–L3 orchestration-level vocabulary from `docs/orchestration-levels.md`.

- **PR #1580** introduced a new architectural test `test_no_bare_import_layer_labels_in_src` in `tests/arch/test_import_layer_labels.py` that scans all `.py` source files for bare `L[0-3]` patterns (intended to enforce the IL-N naming convention for import-layer labels). The test has a `_SKIP_PATHS` frozenset (lines 61–77) that exempts files where bare L-numbers legitimately refer to orchestration levels — currently covering `fleet/*.py` and `cli/_prompts.py`.

- Each PR passes CI independently on its own branch. But when the merge queue creates the `gh-readonly-queue/integration/pr-1580-*` branch (combining #1580 onto the current integration HEAD which already includes #1579), the new test catches `leaf_orchestration_guard.py:8` as a violation:

```
FAILED tests/arch/test_import_layer_labels.py::test_no_bare_import_layer_labels_in_src
AssertionError: Found 1 bare import-layer L-number label(s) — use IL-N instead:
  hooks/leaf_orchestration_guard.py:8: 'L2+ invariant: orchestrator (L2) and fleet (L3) sessions may call orchestration tools.'
```

## Merge Queue Ejection Timeline

| Time (UTC) | Event | CI Run ID |
|---|---|---|
| 07:44:14 | added_to_merge_queue | 25206945521 (failure) |
| 07:50:59 | removed_from_merge_queue | — |
| 07:51:32 | re-enqueued | 25207133291 (failure) |
| 07:57:54 | removed_from_merge_queue | — |
| 07:58:27 | re-enqueued | 25207312309 (failure) |
[...]

PR #1583, opened 2026-05-01
Rectify: Skip-List Self-Validation

Trecek:
## Summary

`tests/arch/test_import_layer_labels.py` maintains a `_SKIP_PATHS` frozenset as a manual
path-based allowlist to exempt files from the bare-import-layer-label scanner. The list
has grown silently over time: 9 of its 12 entries are either unnecessary (the file contains
no matching patterns) or protective against regex phrasing that currently escapes by
accident. There is no inverse verification — the test never checks that skipped files
still contain matching content, that every path in the list resolves to a real file, or
that a skip entry is still warranted. The architectural weakness is that the allowlist
conflates two semantically different categories (load-bearing exemptions vs. precautionary
guards) into one untyped frozenset, with no mechanism to detect when the list becomes stale.
The fix is to split the frozenset into two typed sets, add inverse verification for
load-bearing entries, and add file-existence assertions for all entries.

Part B (separate task) covers the merge queue `DROPPED_HEALTHY` infinite loop.

Closes #1581

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260501-075358-469666/.autoskillit/temp/rectify/rectify_skip_paths_self_validation_2026-05-01_080000_part_a.md`

## Token Usage Summary

| investigate | 26 | 3.9k | 274.3k | 50.6k | 1 | 2m 55s |
| rectify | 105 | 23.1k | 567.9k | 78.8k | 1 | 11m 9s |
| dry_walkthrough | 264 | 34.4k | 1.7M | 129.4k | 2 | 10m 48s |
| implement | 858 | 63.0k | 9.6M | 175.2k | 2 | 24m 57s |
| prepare_pr | 60 | 6.5k | 225.1k | 30.4k | 1 | 1m 46s |
| compose_pr | 59 | 2.0k | 198.7k | 19.8k | 1 | 44s |
| review_pr | 828 | 75.4k | 1.7M | 184.3k | 2 | 20m 36s |
| resolve_review | 688 | 63.3k | 5.5M | 172.9k | 2 | 29m 5s |
| **Total** | 2.9k | 271.5k | 19.7M | 841.3k | | 1h 42m |

### DIFF
--- a/src/autoskillit/execution/_merge_queue_classifier.py
+++ b/src/autoskillit/execution/_merge_queue_classifier.py
@@ -38,6 +38,7 @@ class PRFetchState(TypedDict):
     in_queue: bool
     queue_state: str | None
     checks_state: str | None  # statusCheckRollup.state; None = no checks configured
+    merge_group_checks_state: str | None  # None = not queried; populated on in_queue True→False
 
 
 # Maps PRFetchState keys to GraphQL source paths. Validated at import time.
@@ -52,6 +53,7 @@ _QUERY_FIELD_MAP: dict[str, str] = {
     "in_queue": "<computed>",
     "queue_state": "<computed>",
     "checks_state": "statusCheckRollup.state",
+    "merge_group_checks_state": "<computed>",
 }
 
 _ALL_FETCH_STATE_KEYS = PRFetchState.__required_keys__ | PRFetchState.__optional_keys__
@@ -110,6 +112,13 @@ def _is_positive_dropped_healthy(state: PRFetchState, *, ever_enrolled: bool) ->
     )
 
 
+def _is_positive_dropped_merge_group_ci(state: PRFetchState, *, ever_enrolled: bool) -> bool:
+    """True when the PR was dropped from the queue and merge-group CI is confirmed failed."""
+    return _is_positive_dropped_healthy(state, ever_enrolled=ever_enrolled) and state[
+        "merge_group_checks_state"
+    ] in ("FAILURE", "ERROR")
+
+
 def _is_not_enrolled(state: PRFetchState, *, ever_enrolled: bool) -> bool:
     """True when the PR is healthy but was never enrolled in the merge queue."""
     if ever_enrolled:
@@ -145,6 +154,12 @@ def _classify_pr_state(state: PRFetchState, *, ever_enrolled: bool) -> Classific
 
     if state["mergeable"] == "CONFLICTING":
         return ClassificationResult(PRState.EJECTED, "conflicting changes prevent merge")
+
+    if _is_positive_dropped_merge_group_ci(state, ever_enrolled=ever_enrolled):
+        return ClassificationResult(
+            PRState.DROPPED_MERGE_GROUP_CI,
+            "merge-group CI failed; PR ejected from queue (merge-group check run FAILURE)",
+        )
 
     if _is_positive_dropped_healthy(state, ever_enrolled=ever_enrolled):
         return ClassificationResult(PRState.DROPPED_HEALTHY, "PR healthy but auto_merge cleared")
--- a/src/autoskillit/execution/_merge_queue_classifier.py
+++ b/src/autoskillit/execution/_merge_queue_classifier.py
@@ -113,10 +113,16 @@ def _is_positive_dropped_healthy(state: PRFetchState, *, ever_enrolled: bool) ->
 
 
 def _is_positive_dropped_merge_group_ci(state: PRFetchState, *, ever_enrolled: bool) -> bool:
-    """True when the PR was dropped from the queue and merge-group CI is confirmed failed."""
-    return _is_positive_dropped_healthy(state, ever_enrolled=ever_enrolled) and state[
-        "merge_group_checks_state"
-    ] in ("FAILURE", "ERROR")
+    """True when the PR was dropped from the queue and merge-group CI is confirmed failed.
+
+    Extends `_is_positive_dropped_healthy` as its first conjunct — all preconditions of
+    that predicate (including `checks_state in (None, 'SUCCESS')`) are inherited here.
+    `_query_merge_group_ci` never returns 'ERROR', so this checks only 'FAILURE'.
+    """
+    return (
+        _is_positive_dropped_healthy(state, ever_enrolled=ever_enrolled)
+        and state["merge_group_checks_state"] == "FAILURE"
+    )
 
 
 def _is_not_enrolled(state: PRFetchState, *, ever_enrolled: bool) -> bool:
--- a/src/autoskillit/execution/_merge_queue_classifier.py
+++ b/src/autoskillit/execution/_merge_queue_classifier.py
@@ -113,16 +113,10 @@ def _is_positive_dropped_healthy(state: PRFetchState, *, ever_enrolled: bool) ->
 
 
 def _is_positive_dropped_merge_group_ci(state: PRFetchState, *, ever_enrolled: bool) -> bool:
-    """True when the PR was dropped from the queue and merge-group CI is confirmed failed.
-
-    Extends `_is_positive_dropped_healthy` as its first conjunct — all preconditions of
-    that predicate (including `checks_state in (None, 'SUCCESS')`) are inherited here.
-    `_query_merge_group_ci` never returns 'ERROR', so this checks only 'FAILURE'.
-    """
-    return (
-        _is_positive_dropped_healthy(state, ever_enrolled=ever_enrolled)
-        and state["merge_group_checks_state"] == "FAILURE"
-    )
+    """True when PR dropped from queue with CI failure; extends _is_positive_dropped_healthy."""
+    return _is_positive_dropped_healthy(state, ever_enrolled=ever_enrolled) and state[
+        "merge_group_checks_state"
+    ] in ("FAILURE", "ERROR")
 
 
 def _is_not_enrolled(state: PRFetchState, *, ever_enrolled: bool) -> bool:
--- a/src/autoskillit/execution/_merge_queue_group_ci.py
+++ b/src/autoskillit/execution/_merge_queue_group_ci.py
@@ -91,7 +91,7 @@ async def _query_merge_group_ci(
     """Query the most recent workflow run on the merge-group ref for this PR.
 
     Uses `gh run list` with branch prefix matching for the gh-readonly-queue ref.
-    Returns 'SUCCESS', 'FAILURE', 'ERROR', or None (not found / still running).
+    Returns 'SUCCESS', 'FAILURE', or None (not found / still running / query failed).
     Never raises.
     """
     branch_prefix = f"gh-readonly-queue/{base_branch}/pr-{pr_number}-"
@@ -123,6 +123,6 @@ async def _query_merge_group_ci(
                 if conclusion == "success":
                     return "SUCCESS"
         return None
-    except Exception:
-        logger.warning("_query_merge_group_ci failed", exc_info=True)
+    except Exception as e:
+        logger.warning("_query_merge_group_ci failed: %s", type(e).__name__, exc_info=True)
         return None
--- a/src/autoskillit/execution/_merge_queue_group_ci.py
+++ b/src/autoskillit/execution/_merge_queue_group_ci.py
@@ -1,0 +1,128 @@
+"""Merge-group CI helpers and GraphQL mutation/query strings — private sub-module.
+
+Extracted from merge_queue.py to satisfy the 500-line size budget.
+Import public symbols from autoskillit.execution.merge_queue, not here.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import os
+import re
+
+from autoskillit.core import get_logger
+from autoskillit.execution._merge_queue_classifier import _QUERY_FIELD_MAP
+
+logger = get_logger(__name__)
+
+_QUERY = """
+query GetPRAndQueueState($owner: String!, $repo: String!, $prNumber: Int!, $branch: String!) {
+  repository(owner: $owner, name: $repo) {
+    pullRequest(number: $prNumber) {
+      id
+      merged
+      state
+      mergeable
+      mergeStateStatus
+      autoMergeRequest {
+        enabledAt
+      }
+      statusCheckRollup {
+        state
+      }
+    }
+    mergeQueue(branch: $branch) {
+      entries(first: 100) {
+        nodes {
+          pullRequest { number }
+          state
+        }
+      }
+    }
+  }
+}
+"""
+
+# Part 2 of the _QUERY_FIELD_MAP validation: checks that every non-computed field
+# path head appears as a GraphQL field name in _QUERY.
+# Part 1 (keys match PRFetchState) lives in _merge_queue_classifier.py.
+for _key, _path in _QUERY_FIELD_MAP.items():
+    if _path.startswith("<"):
+        continue
+    _head = _path.split(".", 1)[0]
+    # Word-boundary search prevents "state" from matching inside "mergeStateStatus".
+    if not re.search(r"\b" + re.escape(_head) + r"\b", _QUERY):
+        raise RuntimeError(
+            f"_QUERY is missing GraphQL field {_head!r} required by PRFetchState[{_key!r}]"
+        )
+
+_MUTATION_DISABLE_AUTO_MERGE = """
+mutation DisableAutoMerge($prId: ID!) {
+  disablePullRequestAutoMerge(input: {pullRequestId: $prId}) {
+    pullRequest { number }
+  }
+}
+"""
+
+_MUTATION_ENABLE_AUTO_MERGE = """
+mutation EnableAutoMerge($prId: ID!, $mergeMethod: PullRequestMergeMethod!) {
+  enablePullRequestAutoMerge(input: {pullRequestId: $prId, mergeMethod: $mergeMethod}) {
+    pullRequest { number }
+  }
+}
+"""
+
+_MUTATION_ENQUEUE_PR = """
+mutation EnqueuePullRequest($prId: ID!) {
+  enqueuePullRequest(input: {pullRequestId: $prId}) {
+    mergeQueueEntry { id }
+  }
+}
+"""
+
+
+async def _query_merge_group_ci(
+    repo: str,
+    pr_number: int,
+    base_branch: str,
+    github_token: str | None,
+) -> str | None:
+    """Query the most recent workflow run on the merge-group ref for this PR.
+
+    Uses `gh run list` with branch prefix matching for the gh-readonly-queue ref.
+    Returns 'SUCCESS', 'FAILURE', 'ERROR', or None (not found / still running).
+    Never raises.
+    """
+    branch_prefix = f"gh-readonly-queue/{base_branch}/pr-{pr_number}-"
+    env = {**os.environ}
+    if github_token:
+        env["GH_TOKEN"] = github_token
+    try:
+        proc = await asyncio.create_subprocess_exec(
+            "gh",
+            "run",
+            "list",
+            "--repo",
+            repo,
+            "--json",
+            "conclusion,headBranch",
+            "--limit",
+            "10",
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+            env=env,
+        )
+        stdout, _ = await asyncio.wait_for(proc.communicate(), timeout=15)
+        runs: list[dict[str, str]] = json.loads(stdout.decode())
+        for run in runs:
+            if run.get("headBranch", "").startswith(branch_prefix):
+                conclusion = run.get("conclusion", "")
+                if conclusion in ("failure", "cancelled", "timed_out", "action_required"):
+                    return "FAILURE"
+                if conclusion == "success":
+                    return "SUCCESS"
+        return None
+    except Exception:
+        logger.warning("_query_merge_group_ci failed", exc_info=True)
+        return None
--- a/src/autoskillit/execution/merge_queue.py
+++ b/src/autoskillit/execution/merge_queue.py
@@ -18,6 +18,7 @@ import httpx
 
 from autoskillit.core import PRState, get_logger
 from autoskillit.execution._merge_queue_classifier import (
+    _QUERY_FIELD_MAP,  # noqa: F401 — re-export: tests access merge_queue._QUERY_FIELD_MAP
     KNOWN_MQ_MERGE_STATE_STATUSES,  # noqa: F401 — re-export for callers
     CIStillRunning,
     ClassificationResult,  # noqa: F401 — re-export for callers
--- a/src/autoskillit/execution/merge_queue.py
+++ b/src/autoskillit/execution/merge_queue.py
@@ -289,7 +289,9 @@ class DefaultMergeQueueWatcher:
                 )
             was_in_queue = False
             if merge_group_ci_cache is not None:
-                state = cast(PRFetchState, {**state, "merge_group_checks_state": merge_group_ci_cache})
+                state = cast(
+                    PRFetchState, {**state, "merge_group_checks_state": merge_group_ci_cache}
+                )
 
             try:
                 classification = _classify_pr_state(state, ever_enrolled=ever_enrolled)
--- a/src/autoskillit/execution/merge_queue.py
+++ b/src/autoskillit/execution/merge_queue.py
@@ -8,10 +8,7 @@ Facade: re-exports from _merge_queue_classifier and _merge_queue_repo_state.
 from __future__ import annotations
 
 import asyncio
-import json
-import os
 import random  # noqa: F401 — re-exported for test monkeypatching (_mq.random.uniform)
-import re
 import time
 from collections.abc import Callable
 from datetime import UTC, datetime
@@ -21,7 +18,6 @@ import httpx
 
 from autoskillit.core import PRState, get_logger
 from autoskillit.execution._merge_queue_classifier import (
-    _QUERY_FIELD_MAP,
     KNOWN_MQ_MERGE_STATE_STATUSES,  # noqa: F401 — re-export for callers
     CIStillRunning,
     ClassificationResult,  # noqa: F401 — re-export for callers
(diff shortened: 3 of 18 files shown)


src/autoskillit/execution/merge_queue.py
```diff
@@ -226,6 +172,19 @@ def _make_result(
 
             not_in_queue_cycles += 1
 
+            if was_in_queue and state["checks_state"] not in ("FAILURE", "ERROR"):
```
Trecek, line 175: [critical] bugs: The `was_in_queue` transition gate: verify that `merge_group_ci_cache` is not stale across re-enrollment cycles. `merge_group_ci_cache` is reset to None at line 165 when `in_queue` becomes True (correct), but on the cycle the PR exits the queue the cache is populated and then applied at line 184. On the next cycle `was_in_queue` is False (reset at 182) and the cache is still set — but a subsequent re-enrollment would reset it at line 165. This is structurally sound; however the conditional at line 175 (`was_in_queue and checks_state not in (FAILURE, ERROR)`) skips the query when PR-level CI already failed. This means if `checks_state` transitions to FAILURE between when the PR exits the queue and when the poller observes it, the merge-group CI query is skipped and `DROPPED_MERGE_GROUP_CI` is silently mis-classified as something else. Consider querying regardless of `checks_state` and letting the classifier sort it out — or document this as an intentional tradeoff.

tests/arch/test_import_layer_labels.py
```diff
@@ -93,32 +135,67 @@ def test_no_bare_import_layer_labels_in_src() -> None:
...
+    """Files in _PRECAUTIONARY_SKIP_PATHS must NOT currently match the import-layer
+    pattern. If a precautionary-listed file gains matching content, it must be moved
+    to _LOAD_BEARING_SKIP_PATHS (which triggers test_no_bare_import_layer_labels_in_src
+    if absent from that set, providing the right signal)."""
+    from autoskillit.core.paths import pkg_root
+
+    src_root = pkg_root()
+    for skip_path in sorted(_PRECAUTIONARY_SKIP_PATHS):
+        full_path = src_root / skip_path
+        if not full_path.exists():
```
Trecek, line 195: [critical] tests: Dead code — `_scan_file_for_violations(full_path)` at L197 and `assert not violations` at L198 are unreachable because `continue` at L196 unconditionally skips the rest of the loop body when `not full_path.exists()`. Fix: restructure as `if full_path.exists(): violations = _scan_file_for_violations(full_path); assert not violations, (...)`

src/autoskillit/recipes/merge-prs.yaml
```diff
@@ -308,6 +311,34 @@ steps:
       Circuit breaker for the queue ejection loop. After 3 ejections, routes
       to register_clone_failure to prevent infinite re-enqueue cycles.
 
+  check_dropped_healthy_loop:
+    tool: run_cmd
+    with:
+      cmd: >
+        COUNTER_FILE="${{ context.work_dir }}/{{AUTOSKILLIT_TEMP}}/dropped_healthy_count" &&
```
Trecek, line 318: [critical] defense: `COUNTER_FILE` path `${{ context.work_dir }}/{{AUTOSKILLIT_TEMP}}/dropped_healthy_count` is malformed when `AUTOSKILLIT_TEMP` expands to an absolute path. Bash string concatenation does not override the base path for absolute segments (unlike Python `os.path.join`), producing a broken path like `/clone/dir/home/user/.autoskillit/temp/dropped_healthy_count`. Use only `{{AUTOSKILLIT_TEMP}}/dropped_healthy_count` if `AUTOSKILLIT_TEMP` is already absolute, or hardcode a relative sub-path. The same bug exists in `implementation.yaml`, `implementation-groups.yaml`, and `remediation.yaml`.

src/autoskillit/execution/_merge_queue_group_ci.py
```diff
@@ -0,0 +1,128 @@
...
+      }
+    }
+  }
+}
+"""
+
+# Part 2 of the _QUERY_FIELD_MAP validation: checks that every non-computed field
+# path head appears as a GraphQL field name in _QUERY.
+# Part 1 (keys match PRFetchState) lives in _merge_queue_classifier.py.
+for _key, _path in _QUERY_FIELD_MAP.items():
```
Trecek, line 50: [warning] arch: Module-level loop with `raise RuntimeError` executes at import time. This was extracted from `merge_queue.py` as-is. While the startup assertion is intentional, running it unconditionally at import time means every consumer (test collection, `importlib.reload`) triggers it. Consider wrapping in a function called once at module bottom with a sentinel guard to make the semantics explicit.

src/autoskillit/execution/_merge_queue_group_ci.py
```diff
@@ -0,0 +1,128 @@
...
+            "--limit",
+            "10",
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+            env=env,
+        )
+        stdout, _ = await asyncio.wait_for(proc.communicate(), timeout=15)
+        runs: list[dict[str, str]] = json.loads(stdout.decode())
+        for run in runs:
+            if run.get("headBranch", "").startswith(branch_prefix):
```
Trecek, line 119: [warning] arch: Conclusions `"skipped"`, `"neutral"`, `"startup_failure"`, and `"stale"` are unmapped and silently return `None` (treated as 'not found / still running'). A PR dropped because its merge-group CI run was skipped or neutral will be misclassified as `DROPPED_HEALTHY` and re-enrolled rather than routed to `diagnose_ci`. Document which conclusions intentionally map to None vs. which are gaps.

src/autoskillit/execution/_merge_queue_classifier.py
```diff
@@ -110,6 +112,13 @@ def _is_positive_dropped_healthy(state: PRFetchState, *, ever_enrolled: bool) ->
     )
 
 
+def _is_positive_dropped_merge_group_ci(state: PRFetchState, *, ever_enrolled: bool) -> bool:
+    """True when the PR was dropped from the queue and merge-group CI is confirmed failed."""
+    return _is_positive_dropped_healthy(state, ever_enrolled=ever_enrolled) and state[
```
Trecek, line 117: [warning] bugs: `_is_positive_dropped_merge_group_ci` delegates to `_is_positive_dropped_healthy` as its first conjunct. Verify that `_is_positive_dropped_healthy` does not reject when `checks_state` is SUCCESS (it should accept SUCCESS since that's the normal case for a healthy drop). If `_is_positive_dropped_healthy` requires `auto_merge_present=False` and `in_queue=False`, confirm those are always true when `merge_group_checks_state` is populated. The implicit dependency between the caller populating `merge_group_checks_state` and the classifier relying on it should be documented.

tests/execution/test_merge_queue_polling.py
```diff
@@ -595,3 +595,66 @@ async def test_wait_accepts_stall_grace_period_and_max_stall_retries_params(self
...
+        import autoskillit.execution.merge_queue as _mq
+
+        watcher = _make_watcher()
+        call_count = 0
+
+        async def _fetch(*_a, **_kw):
+            nonlocal call_count
+            call_count += 1
+            if call_count == 1:
+                return _queue_state(in_queue=True, checks_state="SUCCESS")
```
Trecek, line 648: [warning] tests: `patch.object(_mq, '_query_merge_group_ci', ..., return_value='FAILURE')` passes `...` (Ellipsis) as the `new` positional argument. This is not a valid mock factory and will raise `TypeError` at test runtime. Use `new_callable=AsyncMock` instead, matching the sibling test at L623.

src/autoskillit/execution/_merge_queue_group_ci.py
```diff
@@ -0,0 +1,128 @@
...
+                conclusion = run.get("conclusion", "")
+                if conclusion in ("failure", "cancelled", "timed_out", "action_required"):
+                    return "FAILURE"
+                if conclusion == "success":
+                    return "SUCCESS"
+        return None
+    except Exception:
```
Trecek: [warning] defense: Bare `except Exception` swallows `asyncio.TimeoutError`, `json.JSONDecodeError`, and subprocess spawn failures, returning `None` for all. The 'Never raises' contract is intentional, but callers receive `None` with no distinction between 'no matching run', 'timed out after 15s', or 'gh CLI not found'. The `exc_info=True` on the warning is good — consider adding the exception class to the log message for faster diagnosis in production.

src/autoskillit/recipes/merge-prs.yaml
```diff
@@ -308,6 +311,34 @@ steps:
...
+        mkdir -p "$(dirname "$COUNTER_FILE")" &&
+        COUNT=$(cat "$COUNTER_FILE" 2>/dev/null || echo 0) &&
+        COUNT=$((COUNT + 1)) &&
+        echo "$COUNT" > "$COUNTER_FILE" &&
+        if [ "$COUNT" -gt 2 ]; then
```
Trecek, line 323: [warning] defense: Circuit breaker threshold `COUNT -gt 2` escalates on the 3rd drop, but the note says 'After 2 consecutive unexplained drops'. These disagree: with `> 2`, drops 1 and 2 re-enroll; only drop 3 escalates (3 total). If intent is 2 drops before escalation, use `COUNT -gt 1`. Fix the threshold or the note. Same discrepancy in `implementation.yaml`, `implementation-groups.yaml`, `remediation.yaml`.

Trecek (commented):

3 blocking findings require resolution before this PR can merge.

src/autoskillit/execution/merge_queue.py
Trecek: Investigated — this is intentional. When checks_state=FAILURE, _classify_pr_state at L149 returns EJECTED_CI_FAILURE immediately before _is_positive_dropped_merge_group_ci can fire. Additionally, that check requires checks_state in (None, 'SUCCESS') via _is_positive_dropped_healthy, so DROPPED_MERGE_GROUP_CI cannot fire when checks_state=FAILURE. The gate skip is correct; test test_transition_gate_not_triggered_when_checks_already_failed validates this design (commit e3558c26).

tests/arch/test_import_layer_labels.py
Trecek: Investigated — this is intentional. The continue at L196 is guarded by if not full_path.exists(): — it only fires when the file is missing. When the file exists, execution falls through to L197-L198 (violations = _scan_file_for_violations(full_path) and assert not violations). Both lines are reachable. The pattern is a standard guard-clause: skip missing files (delegated to a sibling test), assert on existing files (commit 65b4378c).

src/autoskillit/recipes/merge-prs.yaml
Trecek: Investigated — this is intentional. {{AUTOSKILLIT_TEMP}} is substituted with a relative path via temp_dir_display_str() which returns override or '.autoskillit/temp' — always relative by default. The lint rule in rules_temp_path.py explicitly requires the context.work_dir prefix before {{AUTOSKILLIT_TEMP}} (bare paths are rejected as a validation error). After substitution the command produces '$work_dir/.autoskillit/temp/dropped_healthy_count' — correct Bash concatenation. The reviewer's premise that AUTOSKILLIT_TEMP would be absolute does not hold for this codebase's substitution machinery (commit 4d1f39a4).

tests/execution/test_merge_queue_polling.py
Trecek: Investigated — this is intentional. The actual code at L626-628 and L654-656 both use new_callable=AsyncMock, return_value='FAILURE' — exactly the correct pattern for patching async functions. There is no Ellipsis (...) positional argument anywhere in this file. Both tests already follow the new_callable=AsyncMock form that the reviewer references as correct. The described bug does not exist in the code (commit 983b8773).

src/autoskillit/execution/_merge_queue_group_ci.py
Trecek: Valid observation — flagged for design decision. The import-time validation loop is intentional (startup assertion to catch _QUERY drift), acknowledged as such in the reviewer's comment. Whether to refactor to a sentinel-guarded function is a design/style decision for the maintainers. The current pattern is consistent with similar validation blocks elsewhere in the codebase.

src/autoskillit/execution/_merge_queue_group_ci.py
Trecek: Valid observation — flagged for design decision. Conclusions 'skipped', 'neutral', 'startup_failure', 'stale' currently return None (treated as not-found/still-running). Whether these are intentional (e.g., a skipped CI run is not a merge-group failure) or gaps (should route to diagnose_ci) is a product decision that requires explicit resolution from the maintainers.

src/autoskillit/execution/_merge_queue_classifier.py
Trecek: Valid observation — flagged for design decision. The factual questions are answerable: (1) _is_positive_dropped_healthy accepts checks_state=SUCCESS at L109; (2) in_queue=False is guaranteed by the polling flow; (3) auto_merge_present=False is expected post-ejection but not structurally enforced. The request to document the implicit dependency is valid; a human decision is needed on where to add this documentation (docstring, inline comment, or type annotation).

src/autoskillit/execution/_merge_queue_group_ci.py
Trecek: Valid observation — flagged for design decision. The 'Never raises' contract is intentional and exc_info=True already logs the full traceback (including exception class). Whether to additionally embed the exception class name in the log message string is a minor logging ergonomics preference. A human should decide if the marginal diagnostic value justifies the change.

src/autoskillit/recipes/merge-prs.yaml
Trecek: Valid observation — flagged for design decision. The threshold COUNT -gt 2 is consistent with the sibling check_eject_limit convention: the note number reflects drops *tolerated* (drops 1 and 2 re-enroll; drop 3 escalates), matching 'COUNT -gt N' where N drops are tolerated. However the wording 'After 2 consecutive unexplained drops' is genuinely ambiguous and could be read as escalating on the 2nd drop. A human decision is needed on whether to clarify the note wording or change the threshold to -gt 1.

src/autoskillit/execution/_merge_queue_group_ci.py
Trecek, line 50: [warning] arch: Module-level for-loop with `raise RuntimeError` executes at import time. This causes failures during test collection and `importlib.reload` if `_QUERY_FIELD_MAP` or `_QUERY` is malformed at that moment. Consider wrapping in a `_validate_query_field_map()` function called once at module bottom with a sentinel guard to make import-time failures easier to isolate and diagnose.

src/autoskillit/execution/_merge_queue_group_ci.py
```diff
@@ -0,0 +1,128 @@
...
+async def _query_merge_group_ci(
+    repo: str,
+    pr_number: int,
+    base_branch: str,
+    github_token: str | None,
+) -> str | None:
+    """Query the most recent workflow run on the merge-group ref for this PR.
+
+    Uses `gh run list` with branch prefix matching for the gh-readonly-queue ref.
+    Returns 'SUCCESS', 'FAILURE', 'ERROR', or None (not found / still running).
```
Trecek: [warning] cohesion: Docstring declares return values `'SUCCESS'`, `'FAILURE'`, `'ERROR'`, or `None`, but the implementation never returns `'ERROR'` — the `except` block returns `None` for all exceptions. The downstream classifier (`_is_positive_dropped_merge_group_ci`) checks `merge_group_checks_state in ("FAILURE", "ERROR")`, making the `"ERROR"` branch dead code. Either return `'ERROR'` on exception instead of `None`, or remove `'ERROR'` from the docstring and drop the `"ERROR"` check from the classifier.

src/autoskillit/execution/_merge_queue_group_ci.py
Trecek, line 119: [warning] arch: GitHub Actions conclusions `'skipped'`, `'neutral'`, `'startup_failure'`, and `'stale'` are unmapped and silently return `None` (treated as not-found / still running). A PR dropped because its merge-group CI run was skipped or neutral will be misclassified as `DROPPED_HEALTHY` and re-enrolled rather than routed to `diagnose_ci`. Document which conclusions intentionally map to `None` vs. which are gaps that should map to `'FAILURE'`.

src/autoskillit/execution/_merge_queue_group_ci.py
Trecek: [warning] defense: Bare `except Exception` returns `None` for all failures (`asyncio.TimeoutError`, `json.JSONDecodeError`, subprocess spawn errors), giving callers no way to distinguish 'no matching run', 'timed out after 15s', or 'gh CLI not found'. The 'Never raises' contract is intentional and `exc_info=True` already logs the traceback — adding the exception class to the log message (e.g. `logger.warning("_query_merge_group_ci failed: %s", type(e).__name__, exc_info=True)`) would speed up production diagnosis.

src/autoskillit/execution/_merge_queue_classifier.py
```diff
@@ -110,6 +112,13 @@ def _is_positive_dropped_healthy(state: PRFetchState, *, ever_enrolled: bool) ->
     )
 
 
+def _is_positive_dropped_merge_group_ci(state: PRFetchState, *, ever_enrolled: bool) -> bool:
```
Trecek, line 115: [warning] cohesion: `_is_positive_dropped_merge_group_ci` is a strict refinement of `_is_positive_dropped_healthy` (calls it as its first conjunct and inherits all its preconditions including `checks_state in (None, 'SUCCESS')`), but the naming implies peer-level predicates. Future callers may treat them as independent gates. Add a docstring note that this predicate extends `_is_positive_dropped_healthy` and that its `checks_state` restriction is inherited.

src/autoskillit/execution/_merge_queue_classifier.py
Trecek, line 117: [warning] bugs: `_is_positive_dropped_merge_group_ci` delegates to `_is_positive_dropped_healthy`, which requires `checks_state in (None, 'SUCCESS')`. If a PR exits the queue with `checks_state='PENDING'` or another transient value, `merge_group_ci_cache` may be set to `'FAILURE'` but this predicate returns `False` because the parent predicate rejects the non-SUCCESS state — causing silent misclassification as `DROPPED_HEALTHY`. Confirm the poller guarantees `checks_state` is always `SUCCESS` or `None` when `in_queue` transitions False, or document this as an accepted gap.

src/autoskillit/execution/merge_queue.py
```diff
@@ -226,6 +172,19 @@ def _make_result(
...
+                merge_group_ci_cache = await _query_merge_group_ci(
+                    repo=repo,
+                    pr_number=pr_number,
+                    base_branch=target_branch,
+                    github_token=None,
+                )
+            was_in_queue = False
```
Trecek: [warning] bugs: `was_in_queue` is reset to `False` unconditionally at L182, immediately after a single `_query_merge_group_ci` call. If the merge-group CI run hasn't finished yet (`_query_merge_group_ci` returns `None`), `merge_group_ci_cache` stays `None` and the query is never retried on subsequent poll cycles. A CI run that completes after the first poll following queue exit will permanently miss `DROPPED_MERGE_GROUP_CI` classification, causing the PR to fall back to `DROPPED_HEALTHY` re-enrollment. Consider retrying while `merge_group_ci_cache is None` for N cycles after exit, or documenting this as a known single-shot limitation.

tests/arch/test_import_layer_labels.py
```diff
@@ -57,34 +58,75 @@
...
+    in_triple_double = False
+    in_triple_single = False
+    for lineno, line in enumerate(
+        path.read_text(encoding="utf-8", errors="replace").splitlines(), 1
+    ):
+        stripped = line.strip()
+        if not in_triple_single and '"""' in stripped:
+            count = stripped.count('"""')
+            if count % 2 == 1:
+                in_triple_double = not in_triple_double
```
Trecek, line 107: [warning] tests: `_scan_file_for_violations` uses `elif` for the triple-single-quote toggle, so a line containing both `"""` and `'''` only fires the `in_triple_double` toggle and skips the `'''` check. This mirrors the original test logic and is likely intentional (pathological lines are rare), but confirm this is a deliberate tradeoff rather than an oversight — a sequential `if`/`if` structure would handle both on the same line.

tests/arch/test_import_layer_labels.py
```diff
@@ -57,34 +58,75 @@
...
+        elif not in_triple_double and "'''" in stripped:
+            count = stripped.count("'''")
+            if count % 2 == 1:
+                in_triple_single = not in_triple_single
+        inside_docstring = in_triple_double or in_triple_single
+        if (
+            stripped.startswith("#")
+            or '"""' in stripped
+            or stripped.startswith("'")
```
Trecek, line 116: [warning] tests: The scan filter includes `stripped.startswith("'")` to catch single-quoted string literals, but this also matches ordinary single-quoted strings and f-strings that are not docstrings, which could yield false positives for non-docstring content containing IL-N patterns. Since `inside_docstring` already covers triple-single-quoted blocks, verify that the `startswith("'")` condition is intentional and that the test never fires on non-docstring content in practice.

Trecek (commented):

Trecek (commented):

### Info-level findings (not posted as inline comments)

**tests/arch/test_import_layer_labels.py**
- **L61** [info/arch]: `_LOAD_BEARING_SKIP_PATHS` lists `hooks/leaf_orchestration_guard.py` with an inline comment describing the skip reason. Adequate as-is; no action required — the companion test `test_load_bearing_skip_paths_still_match` enforces liveness.
- **L132** [info/tests]: `rel_str` is used inside the loop body at L138-L140 but its assignment is not visible in the diff hunk. Confirm it is assigned in the elided lines; if not, this would be a `NameError` at runtime.

**src/autoskillit/recipes/merge-prs.yaml**
- **L339** [info/cohesion]: Step note references `DROPPED_MERGE_GROUP_CI` in uppercase (Python enum name style) while all routing conditions in the same file use the canonical lowercase form `dropped_merge_group_ci`. Normalise to lowercase for consistency with sibling steps (e.g. `check_eject_limit`).

**tests/execution/test_merge_queue_classifier.py**
- **L583** [info/tests]: New parametrize entry adds `PRState.DROPPED_MERGE_GROUP_CI` (a terminal state) inside a hunk whose `@@` context marker shows `test_classifier_raises_inconclusive_when_no_positive_signal_matches`. Verify this entry is attached to a *terminal-state* parametrize list and not to the inconclusive-raise test, as the latter would be an incorrect assertion.

### Suppressed findings

- `src/autoskillit/execution/merge_queue.py:175` — transition-gate design; resolved in prior review cycle.
- `tests/execution/test_merge_queue_polling.py:648` — within ±5 lines of a prior-resolved thread; suppressed to avoid drift noise.
- `src/autoskillit/recipes/merge-prs.yaml:323` — COUNT -gt 2 threshold vs. note wording; within ±5 of a prior-resolved thread; suppressed.

src/autoskillit/execution/_merge_queue_group_ci.py
Trecek: Agreed — fixed in 3f732ea6. Removed 'ERROR' from _query_merge_group_ci docstring (the except block never returns 'ERROR'; it returns None). The 'ERROR' check in the classifier is retained for forward compatibility. Evidence: implementation L126 returns None on all exceptions; test_dropped_merge_group_ci_also_fires_on_error_conclusion verifies the classifier still accepts 'ERROR' as a valid merge_group_checks_state value.

src/autoskillit/execution/_merge_queue_group_ci.py
Trecek: Agreed — fixed in 3f732ea6. Changed `logger.warning("_query_merge_group_ci failed", exc_info=True)` to `logger.warning("_query_merge_group_ci failed: %s", type(e).__name__, exc_info=True)`. Exception class name now appears in the warning message for faster production diagnosis.

src/autoskillit/execution/_merge_queue_classifier.py
Trecek: Agreed — fixed in 7bd652c1. Updated _is_positive_dropped_merge_group_ci docstring to explicitly note it extends _is_positive_dropped_healthy and inherits its preconditions. Evidence: new docstring at L116.

src/autoskillit/execution/merge_queue.py
Trecek: Agreed — fixed in 7bd652c1. was_in_queue is now only cleared when merge_group_ci_cache is not None (a result was obtained) or when not_in_queue_cycles reaches not_in_queue_confirmation_cycles (the same window used for classification readiness). This ensures the CI query is retried on subsequent polls while the run is still pending. Evidence: merge_queue.py L175-182.

src/autoskillit/execution/_merge_queue_group_ci.py
Trecek: Valid observation — flagged for design decision. The import-time validation loop is intentional: it catches _QUERY_FIELD_MAP/PRFetchState drift at startup rather than silently returning wrong results at runtime. importlib.reload is not a supported code path; test isolation is handled by mocking before the module is loaded.

src/autoskillit/execution/_merge_queue_group_ci.py
Trecek: Valid observation — flagged for design decision. Conclusions 'skipped', 'neutral', 'startup_failure', 'stale' intentionally map to None (not-found / still-running). These are non-terminal states for our use case. A PR dropped after a 'skipped' merge-group run is handled by re-enrollment on the next cycle. Documenting the intentional exclusion in the source is tracked for a follow-up.

src/autoskillit/execution/_merge_queue_classifier.py
Trecek: Valid observation — flagged for design decision. The poller does not guarantee checks_state is SUCCESS or None on queue exit; PENDING is possible if main-branch checks are still running. In that edge case the PR would not match DROPPED_MERGE_GROUP_CI (due to the parent predicate's checks_state constraint) and would re-enroll on the next cycle. This is an accepted gap — the PR will be retried rather than silently misclassified as DROPPED_HEALTHY.

tests/arch/test_import_layer_labels.py
Trecek: Valid observation — flagged for design decision. The elif is a deliberate tradeoff: Python lines containing both triple-double and triple-single quotes in the same line are pathological and impossible in practice. Switching to sequential if/if would add complexity with zero real-world benefit. This mirrors the logic in the main test for the same reason.

tests/arch/test_import_layer_labels.py
Trecek: Valid observation — flagged for design decision. The startswith("'") condition is intentional — it targets lines that open with a single-quoted string literal (which could contain an import-layer label). A false positive requires both the filter match AND the _IMPORT_LAYER_PATTERNS regex to fire, which is highly unlikely in production code. The condition is retained to avoid silent false negatives on single-quoted docstring-like content.

PR #1596, opened 2026-05-02
Rectify: Kitchen-Open Guard in run_update_checks() Blocks Informational Prompts

### LINKED ISSUE #1594: Kitchen-open guard in run_update_checks() blocks informational update prompt for non-mutating commands
## Problem

When running `autoskillit order` from a terminal where an AutoSkillit MCP server is already running (kitchen open) for the same project, the startup update check prints:

```
Skipping update check: a kitchen is open for this project. Run 'autoskillit update' manually after the pipeline finishes.
```

This message appears before the recipe selection menu. The update check is fully suppressed — no update signals are gathered, no prompt is shown. The `order` command itself proceeds normally after the message, but the user sees a confusing notification that implies they should wait for a pipeline to finish before they can check for updates.

This is **not a regression** from any recent commit. The guard code at `_update_checks.py:320-327` is byte-for-byte identical to what was committed in the prior fix (`77edfff3`, PR #1480). All 108 commits between `77edfff3` and HEAD were verified — none altered this code path. This is a **residual design flaw** from the original kitchen guard introduced in `aa302873` (#1066).

Related: #1415 (Kitchen-open guard over-scoped — remains OPEN)

## Root Cause

**The `main()` function at `app.py:436-447` applies `run_update_checks()` unconditionally to every non-`serve` CLI command.** The only filter is `_first_arg != "serve"` at line 439 — no per-command exemptions exist. Every command from `order` to `doctor` to `recipes list` goes through the same update check path.

The kitchen-open guard inside `run_update_checks()` at `_update_checks.py:320-327` was designed to protect against `autoskillit update` corrupting the plugin cache while a pipeline is running, but it gates the update *check* (a read-only informational prompt) for all commands including `order`.

There are three independent kitchen guards in the codebase, each with different severity:

| Guard location | Behavior when kitchen open | Purpose |
|---|---|---|
| `_update.py:42-49` | Hard `SystemExit(1)` | Prevent `autoskillit update` from mutating packages mid-pipeline |
| `_marketplace.py:167-173` | Skip `_clear_plugin_cache()` only | Prevent cache corruption during `install` |
| `_update_checks.py:320-327` | Print message, return early | Suppress informational update prompt — **no mutation prevented** |

The third guard is the one causing the user-facing issue. It provides no safety benefit for non-mutating commands like `order`, `cook`, `doctor`, or `recipes list`.

## Data Flow

```
User terminal: autoskillit order
  -> main()                                    # app.py:436
  -> _first_arg = "order" != "serve"           # app.py:438-439
  -> evict_direct_mcp_entry()                  # app.py:442
  -> run_update_checks()                       # app.py:446
     -> 7 env/TTY guards pass                  # _update_checks.py:309-317
        (CLAUDECODE, CI, SKIP_STALE_CHECK,
         SKIP_UPDATE_CHECK, SKIP_SOURCE_DRIFT_CHECK,
         stdin.isatty(), stdout.isatty())
     -> any_kitchen_open(project_path=cwd)     # _update_checks.py:322
        -> reads ~/.autoskillit/active_kitchens.json (fcntl LOCK_EX)
        -> iterates entries, calls _pid_alive() for each
        -> finds MCP server PID alive + create_time match
        -> project_path matches
        -> returns True
     -> prints "Skipping update check: ..."    # _update_checks.py:323-326
     -> return (update check suppressed)       # _update_checks.py:327
  -> app() -> order()                          # app.py:447 -> _order.py:91
     -> (proceeds normally, user sees recipe menu)
```

## Affected Components

- `src/autoskillit/cli/app.py:436-447`: `main()` — unconditional `run_update_checks()` for all non-`serve` commands
- `src/autoskillit/cli/_update_checks.py:320-327`: Kitchen-open guard — scoped to project but not to command type
- `src/autoskillit/core/_plugin_cache.py:240-274`: `any_kitchen_open()` — correctly scoped, correctly prunes dead PIDs
- `src/autoskillit/core/_plugin_cache.py:177-203`: `register_active_kitchen()` — called by MCP server on `open_kitchen`, stays registered until `close_kitchen` or server teardown
- `src/autoskillit/core/_plugin_cache.py:155-174`: `_pid_alive()` — uses `os.kill(pid, 0)` + psutil create-time (1s tolerance) for PID-reuse safety
- `src/autoskillit/server/tools_kitchen.py:161-167`: `_open_kitchen_handler()` — sole interactive registration call site
- `src/autoskillit/server/tools_kitchen.py:202-216`: `_close_kitchen_handler()` — sole unregistration call site (unregister call at line 214)
- `src/autoskillit/server/_lifespan.py:247-252`: `clear_kitchens_for_pid()` — teardown fallback cleanup

## Design Intent

- **`any_kitchen_open()` guard in `run_update_checks()`**: Introduced in commit `aa302873` (Apr 19, 2026) as part of "Resilient Plugin Cache Lifecycle" (#1066). Requirement REQ-GUARD-001 stated: "autoskillit update and the auto-update check MUST be blocked while any kitchen is open, using a global (home-scoped) registry with PID liveness verification." The requirement names only `update` and `auto-update check` — not `order` or other CLI commands. No ADR exists for this decision. REQ-GUARD-001 appears only in commit messages, not in any living artifact.
- **Project-path scoping**: Introduced in commit `77edfff3` (Apr 28, 2026) as part of "Rectify: Kitchen-Open Guard Over-Scoped" (#1480). Fixed global-to-project scoping. The architecture diagram in this commit groups `cook/order` as "SESSION" commands that "Trigger startup checks" and are "non-fatally skipped" — this grouping was intentional but did not question whether `order` should be subject to the guard at all.
- **Kitchen registry has no session-type field**: `register_active_kitchen()` stores `kitchen_id`, `pid`, `create_time`, `project_path`, `opened_at`. No field distinguishes MCP server kitchen from fleet kitchen from interactive kitchen. `any_kitchen_open()` cannot filter by session type.

## Historical Context

This is a recurring pattern. A prior investigation (`investigation_update_kitchen_guard_regression_2026-04-27`) identified the same root cause (kitchen guard too coarse) and recommended 4 fixes:

1. Scope guard to same-project — **IMPLEMENTED in 77edfff3**
2. Remove kitchen guard from `run_update_command()` entirely — **NOT IMPLEMENTED**
3. Downgrade `_update.py` guard from hard block to warning — **NOT IMPLEMENTED**
4. Exempt `autoskillit update` from startup auto-check kitchen guard via `sys.argv[1]` — **NOT IMPLEMENTED**

The current issue is a direct consequence of recommendations 2-4 not being implemented. The fix addressed only the cross-project dimension (Session A on project X blocking updates for project Y). It did not address the cross-command dimension (kitchen open blocking `order` in the same project).

## Command Classification

All CLI commands were classified by whether they perform package/cache mutations:

**Mutation commands (should RETAIN the kitchen guard):**

| Command | Reason |
|---|---|
| `update` | Runs pip/uv upgrade + `autoskillit install` — already has its own hard `SystemExit(1)` guard in `_update.py:42-49` |
| `install` | Calls `_clear_plugin_cache()`, registers plugin — already has its own skip-cache-clear guard in `_marketplace.py:167-173` |
| `init` | Writes `.autoskillit/config.yaml`, registers hooks in `settings.json` |

**Non-mutation commands (should BYPASS the kitchen guard):**

| Command | Reason |
|---|---|
| `order` | Session launcher — starts Claude Code subprocess, no package mutation |
| `cook` | Interactive session launcher — same as `order` |
| `doctor` | Read-only diagnostics |
| `upgrade` | YAML migration (renames scripts/ to recipes/, rewrites keys) — does NOT touch pip, plugin cache, or hooks |
| `migrate` | Reads and reports pending migrations — does not apply them |
| `config show` | Pure read |
| `skills list` | Pure read |
| `recipes list/show/render` | Pure read |
| `quota_status` | Pure read |
| `features list/status` | Pure read |
| `fleet dispatch/campaign/list/status` | Orchestration launchers or read-only |
| `workspace init/clean` | Local directory management, not package state |
| `sessions analyze` | Pure read |

## Recommended Fix

**Add `skip_kitchen_guard: bool = False` keyword-only parameter to `run_update_checks()`.**

In `main()`, pass `skip_kitchen_guard=True` when `_first_arg` is NOT in a mutation-requiring set (`{"update", "install", "init"}`). Inside `run_update_checks()`, wrap the kitchen guard (lines 320-327) in `if not skip_kitchen_guard:`.

Signature: `def run_update_checks(home: Path | None = None, *, skip_kitchen_guard: bool = False)`

### Why this is the right fix

- ~4 lines of production code changed
- Default `False` preserves existing behavior for all callers and existing tests
- Existing kitchen test at `test_update_checks_prompt.py:822-843` survives unchanged (calls `run_update_checks(home=tmp_path)` without the new kwarg)
- Arch test `test_kitchen_guard_scoping.py` is unaffected (AST scan still finds `any_kitchen_open(project_path=...)` inside the conditional)
- The `any_kitchen_open()` call site remains in `run_update_checks()` with `project_path=` kwarg (no arch test violation)
- `_update.py` retains its own independent hard-block (defense-in-depth preserved)

### Mid-pipeline update prompt risk

When `skip_kitchen_guard=True` is passed for `order`, `run_update_checks()` proceeds past the kitchen guard. If update signals fire, the user sees a 30-second `[Y/n]` prompt. If the user answers "Y", `_run_update_sequence()` at `_update_checks.py:255-291` runs pip upgrade and `autoskillit install`. The subprocess `install` has its own independent kitchen guard at `_marketplace.py:167-173` that skips `_clear_plugin_cache()`, so the cache is not corrupted. However, the pip binary replacement itself is unguarded. This risk is bounded: it requires deliberate user action (answering "Y"), the 30-second timeout defaults to no-action (skip), and the scenario only arises when the user is launching a new session (not running an update). This is an acceptable tradeoff — the current behavior (suppressing all update information) is worse for UX.

### Killed alternatives

- **Move guard to main():** Breaks existing kitchen test at `test_update_checks_prompt.py:822-843`, creates arch test drift in `test_kitchen_guard_scoping.py` (helpers test would need updating), splits guard logic across two files — more disruptive for no benefit.
- **Pass command= param:** Wrong default (`command=""`) silently disables the guard for direct callers who forget to pass the arg — a correctness trap.
- **Remove guard entirely:** Exposes "upgrade mid-pipeline" risk without the bounded protection of the interactive prompt timeout. Less safe than the recommended fix because ALL commands including programmatic callers would bypass the guard.

## Test Gap Analysis

1. **No test verifies per-command exemption behavior.** `test_update_checks_prompt.py:822-843` tests that when `any_kitchen_open` returns `True`, the message is printed and no prompt is shown — but it does not test which commands should or shouldn't be subject to the guard.
2. **No integration test for `main()` → update check → command dispatch.** `test_app_main.py` has exactly one test that patches `run_update_checks` entirely. No test exercises the real `run_update_checks` from `main()`.
3. **The arch test `test_kitchen_guard_scoping.py` enforces only `project_path=` kwarg presence** — it has no per-command exemption mechanism.
4. **`_setup_run_checks` helper patches `any_kitchen_open` to return `False` by default** (line 116), meaning all ~30 update check tests never exercise the kitchen-open code path.

### New tests needed

1. Test in `test_update_checks_prompt.py`: when `skip_kitchen_guard=True` and kitchen is open, `run_update_checks()` should proceed to signal evaluation (not return early).
2. Test in `test_app_main.py`: `main()` with `sys.argv = ["autoskillit", "order"]` should pass `skip_kitchen_guard=True` to `run_update_checks()`.
3. Test in `test_app_main.py`: `main()` with `sys.argv = ["autoskillit", "update"]` should NOT pass `skip_kitchen_guard=True`.

## Breakage Analysis

**New risk surface:** With `skip_kitchen_guard=True` for session-launcher commands, the `_run_update_sequence()` code path becomes reachable while a separate kitchen is open. The subprocess `autoskillit install` independently skips `_clear_plugin_cache()` via its own kitchen guard at `_marketplace.py:167-173`, so the plugin cache is protected. The binary replacement (`uv tool upgrade`) is not protected, but requires the user to actively answer "Y" to a 30-second timeout prompt. Risk level: LOW.

**Multiple concurrent kitchens:** `register_active_kitchen()` appends entries with no duplicate-project check. Two concurrent `order` sessions for the same project would each register a kitchen. `any_kitchen_open()` uses `any()` and short-circuits on the first match, so behavioral correctness is maintained. The proposed fix does not alter this behavior.

**Latent vectors:** The `project_path is None` wildcard match at `_plugin_cache.py:269` is a false-positive vector inherited from `77edfff3`. Current code cannot register a `project_path=None` entry (the parameter is non-optional), but if a future code path did, it would match ALL project-scoped queries. Not triggered by this bug.

Trecek:
## Summary

The `run_update_checks()` function at `_update_checks.py:322` uses `any_kitchen_open()` as an entry-point guard that suppresses all update signals — including purely informational ones — for every CLI command. Only mutation commands (`update`, `install`, `init`) actually need this protection. The guard lacks caller-context discrimination, a class of bug the codebase already prevents elsewhere via typed frozenset registries (e.g., `_ORCHESTRATION_TOOLS`, `GATED_TOOLS`). The fix applies that same registry pattern: define a `KITCHEN_GUARDED_COMMANDS` frozenset, pass the CLI command name into `run_update_checks()`, and gate only when the command is in the set. This makes the guard structurally immune to over-scoping — new commands default to unguarded, and mutation commands must be explicitly registered.

Closes #1594

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260501-214425-611916/.autoskillit/temp/rectify/rectify_kitchen_guard_update_checks_2026-05-01_214425.md`

## Token Usage Summary

| rectify | 59 | 12.8k | 664.8k | 72.7k | 1 | 5m 27s |
| dry_walkthrough | 38 | 8.0k | 1.0M | 51.4k | 1 | 4m 13s |
| implement | 180 | 8.7k | 1.1M | 63.2k | 1 | 3m 27s |
| prepare_pr | 68 | 5.5k | 273.4k | 31.9k | 1 | 1m 33s |
| compose_pr | 59 | 2.2k | 194.6k | 19.7k | 1 | 58s |
| review_pr | 100 | 30.2k | 522.7k | 93.8k | 1 | 5m 57s |
| resolve_review | 283 | 21.1k | 2.0M | 73.8k | 1 | 10m 23s |
| **Total** | 787 | 88.6k | 5.7M | 406.5k | | 31m 59s |

### DIFF
--- a/src/autoskillit/cli/_update_checks.py
+++ b/src/autoskillit/cli/_update_checks.py
@@ -49,6 +49,8 @@ _DISMISS_FILE = "update_check.json"
 _STABLE_DISMISS_WINDOW = timedelta(days=7)
 _DEV_DISMISS_WINDOW = timedelta(hours=12)
 
+KITCHEN_GUARDED_COMMANDS: frozenset[str] = frozenset({"update", "install", "init"})
+
 
 @dataclass(frozen=True)
 class Signal:
@@ -291,7 +293,7 @@ def _run_update_sequence(
         perform_restart()
 
 
-def run_update_checks(home: Path | None = None) -> None:
+def run_update_checks(home: Path | None = None, *, command: str = "") -> None:
     """Run the unified update check on interactive CLI invocations.
 
     At most one ``[Y/n]`` prompt is shown per invocation.  Guards:
@@ -317,14 +319,15 @@ def run_update_checks(home: Path | None = None) -> None:
     ):
         return
 
-    from autoskillit.core import any_kitchen_open  # noqa: PLC0415
+    if command in KITCHEN_GUARDED_COMMANDS:
+        from autoskillit.core import any_kitchen_open  # noqa: PLC0415
 
-    if any_kitchen_open(project_path=str(Path.cwd())):
-        print(
-            "Skipping update check: a kitchen is open for this project. "
-            "Run 'autoskillit update' manually after the pipeline finishes.",
-        )
-        return
+        if any_kitchen_open(project_path=str(Path.cwd())):
+            print(
+                "Skipping update check: a kitchen is open for this project. "
+                "Run 'autoskillit update' manually after the pipeline finishes.",
+            )
+            return
 
     _skip_env: dict[str, str] = {
         **os.environ,
--- a/tests/arch/test_kitchen_guard_scoping.py
+++ b/tests/arch/test_kitchen_guard_scoping.py
@@ -102,13 +102,9 @@ def _if_body_has_any_kitchen_open_call(if_node: ast.If) -> bool:
 
 def _if_test_references_kitchen_guarded_commands(if_node: ast.If) -> bool:
     """Return True if the if-node's test references KITCHEN_GUARDED_COMMANDS."""
-    test_src = ast.dump(if_node.test)
-    return "KITCHEN_GUARDED_COMMANDS" in test_src or (
-        isinstance(if_node.test, ast.Compare)
-        and any(
-            isinstance(c, ast.Name) and c.id == "KITCHEN_GUARDED_COMMANDS"
-            for c in ast.walk(if_node.test)
-        )
+    return any(
+        isinstance(node, ast.Name) and node.id == "KITCHEN_GUARDED_COMMANDS"
+        for node in ast.walk(if_node.test)
     )
 
 
--- a/tests/arch/test_kitchen_guard_scoping.py
+++ b/tests/arch/test_kitchen_guard_scoping.py
@@ -69,6 +69,72 @@ def _function_body_has_any_kitchen_open_patch(func_node: ast.FunctionDef) -> boo
     return False
 
 
+_UPDATE_CHECKS_FILE = _SRC_ROOT / "cli" / "_update_checks.py"
+
+
+def test_run_update_checks_has_command_parameter() -> None:
+    """run_update_checks must declare a 'command' parameter."""
+    source = _UPDATE_CHECKS_FILE.read_text(encoding="utf-8")
+    tree = ast.parse(source, filename=str(_UPDATE_CHECKS_FILE))
+    for node in ast.walk(tree):
+        if isinstance(node, ast.FunctionDef) and node.name == "run_update_checks":
+            param_names = [arg.arg for arg in node.args.args + node.args.kwonlyargs]
+            assert "command" in param_names, (
+                f"run_update_checks must have a 'command' parameter, got: {param_names}"
+            )
+            return
+    raise AssertionError("run_update_checks not found in _update_checks.py")
+
+
+def _find_function(tree: ast.AST, name: str) -> ast.FunctionDef | None:
+    for node in ast.walk(tree):
+        if isinstance(node, ast.FunctionDef) and node.name == name:
+            return node
+    return None
+
+
+def _if_body_has_any_kitchen_open_call(if_node: ast.If) -> bool:
+    for node in ast.walk(if_node):
+        if isinstance(node, ast.Call) and _is_any_kitchen_open_call(node):
+            return True
+    return False
+
+
+def _if_test_references_kitchen_guarded_commands(if_node: ast.If) -> bool:
+    """Return True if the if-node's test references KITCHEN_GUARDED_COMMANDS."""
+    test_src = ast.dump(if_node.test)
+    return "KITCHEN_GUARDED_COMMANDS" in test_src or (
+        isinstance(if_node.test, ast.Compare)
+        and any(
+            isinstance(c, ast.Name) and c.id == "KITCHEN_GUARDED_COMMANDS"
+            for c in ast.walk(if_node.test)
+        )
+    )
+
+
+def test_kitchen_guard_gated_by_kitchen_guarded_commands() -> None:
+    """The any_kitchen_open call in run_update_checks must be inside an if block
+    whose test references KITCHEN_GUARDED_COMMANDS."""
+    source = _UPDATE_CHECKS_FILE.read_text(encoding="utf-8")
+    tree = ast.parse(source, filename=str(_UPDATE_CHECKS_FILE))
+    func = _find_function(tree, "run_update_checks")
+    assert func is not None, "run_update_checks not found in _update_checks.py"
+
+    for node in ast.walk(func):
+        if not isinstance(node, ast.If):
+            continue
+        if not _if_test_references_kitchen_guarded_commands(node):
+            continue
+        if _if_body_has_any_kitchen_open_call(node):
+            return
+
+    raise AssertionError(
+        "any_kitchen_open call in run_update_checks is not inside an if block "
+        "whose test references KITCHEN_GUARDED_COMMANDS. "
+        "The guard must be gated: `if command in KITCHEN_GUARDED_COMMANDS: any_kitchen_open(...)`"
+    )
+
+
 def test_setup_helpers_must_patch_any_kitchen_open() -> None:
     helpers = [
         (_TEST_ROOT / "cli" / "test_update_command.py", "_setup_run_update"),
--- a/tests/cli/test_app_main.py
+++ b/tests/cli/test_app_main.py
@@ -22,7 +22,7 @@ def test_main_does_not_call_app_after_update_restart(monkeypatch: pytest.MonkeyP
         "autoskillit.cli._init_helpers.evict_direct_mcp_entry", lambda *a, **kw: None
     )
 
-    def fake_run_update_checks() -> None:
+    def fake_run_update_checks(**kwargs: object) -> None:
         raise SystemExit(0)
 
     monkeypatch.setattr("autoskillit.cli._update_checks.run_update_checks", fake_run_update_checks)
@@ -34,3 +34,37 @@ def test_main_does_not_call_app_after_update_restart(monkeypatch: pytest.MonkeyP
         pass
 
     assert not app_called, "app() must not be called when update path exits the process"
+
+
+@pytest.mark.parametrize(
+    "argv_command",
+    [
+        "order",
+        "update",
+    ],
+)
+def test_main_passes_command_to_run_update_checks(
+    monkeypatch: pytest.MonkeyPatch, argv_command: str
+) -> None:
+    """main() must pass _first_arg as command= to run_update_checks()."""
+    app_module = importlib.import_module("autoskillit.cli.app")
+
+    monkeypatch.setattr(app_module, "app", lambda: None)
+    monkeypatch.setattr(
+        "autoskillit.cli._init_helpers.evict_direct_mcp_entry", lambda *a, **kw: None
+    )
+
+    captured_kwargs: list[dict[str, object]] = []
+
+    def spy_run_update_checks(**kwargs: object) -> None:
+        captured_kwargs.append(dict(kwargs))
+
+    monkeypatch.setattr("autoskillit.cli._update_checks.run_update_checks", spy_run_update_checks)
+    monkeypatch.setattr(sys, "argv", ["autoskillit", argv_command])
+
+    app_module.main()
+
+    assert len(captured_kwargs) == 1, "run_update_checks must be called exactly once"
+    assert captured_kwargs[0].get("command") == argv_command, (
+        f"run_update_checks must receive command={argv_command!r}, got {captured_kwargs[0]!r}"
+    )
--- a/tests/cli/test_update_checks_prompt.py
+++ b/tests/cli/test_update_checks_prompt.py
@@ -831,7 +831,7 @@ def test_run_update_checks_shows_notification_when_kitchen_open(
         lambda *a, **kw: binary_signal_calls.append(True) or None,
     )
 
-    run_update_checks(home=tmp_path)
+    run_update_checks(home=tmp_path, command="update")
 
     combined = " ".join(printed)
     assert "kitchen" in combined.lower(), (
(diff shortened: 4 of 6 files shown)


src/autoskillit/cli/_update_checks.py
```diff
@@ -317,14 +319,15 @@ def run_update_checks(home: Path | None = None) -> None:
     ):
         return
 
-    from autoskillit.core import any_kitchen_open  # noqa: PLC0415
+    if command in KITCHEN_GUARDED_COMMANDS:
```
Trecek, line 322: [warning] bugs: The kitchen-guard branch is only entered when command is in KITCHEN_GUARDED_COMMANDS. For all other commands any_kitchen_open check is silently skipped. If a command not in KITCHEN_GUARDED_COMMANDS is invoked while a kitchen is open the update check will proceed and may interfere with a running pipeline. The prior code guarded unconditionally — narrowing the guard to three commands is an intentional behavioral change; confirm this is the desired semantics. (requires_decision: true)

tests/arch/test_kitchen_guard_scoping.py
```diff
@@ -69,6 +69,72 @@ def _function_body_has_any_kitchen_open_patch(func_node: ast.FunctionDef) -> boo
...
+
+
+def _if_body_has_any_kitchen_open_call(if_node: ast.If) -> bool:
+    for node in ast.walk(if_node):
+        if isinstance(node, ast.Call) and _is_any_kitchen_open_call(node):
+            return True
+    return False
+
+
+def _if_test_references_kitchen_guarded_commands(if_node: ast.If) -> bool:
```
Trecek, line 103: [warning] tests: _if_test_references_kitchen_guarded_commands uses ast.dump string-contains as primary check. The ast.dump path will match any AST node whose dump string contains "KITCHEN_GUARDED_COMMANDS", including string literals with that text. Prefer a pure AST walk (isinstance(node, ast.Name) and node.id == "KITCHEN_GUARDED_COMMANDS") across all nodes in the test expression to avoid false positives.

tests/cli/test_app_main.py
```diff
@@ -34,3 +34,37 @@ def fake_run_update_checks() -> None:
...
+    app_module = importlib.import_module("autoskillit.cli.app")
+
+    monkeypatch.setattr(app_module, "app", lambda: None)
+    monkeypatch.setattr(
+        "autoskillit.cli._init_helpers.evict_direct_mcp_entry", lambda *a, **kw: None
+    )
+
+    captured_kwargs: list[dict[str, object]] = []
+
+    def spy_run_update_checks(**kwargs: object) -> None:
```
Trecek, line 59: [warning] bugs: spy_run_update_checks accepts only **kwargs but run_update_checks has a positional-or-keyword home parameter. If home= is ever passed positionally the spy will silently drop it from captured_kwargs. The assertion at L68 only checks command=, so a missing home argument would go undetected.

tests/cli/test_update_checks_prompt.py
```diff
@@ -841,3 +841,64 @@ def test_run_update_checks_shows_notification_when_kitchen_open(
...
         "run_update_checks must not proceed to signal checks when kitchen is open"
     )
+
+
+def test_run_update_checks_kitchen_open_bypassed_for_non_mutation_command(
+    monkeypatch: pytest.MonkeyPatch, tmp_path: Path
+) -> None:
+    """run_update_checks with command='order' proceeds past kitchen guard even when open."""
+    printed, input_calls = _setup_run_checks(monkeypatch, tmp_path, binary_signal=True)
+    monkeypatch.setattr("autoskillit.core.any_kitchen_open", lambda **kw: True)
```
Trecek: [warning] bugs: monkeypatch patches autoskillit.core.any_kitchen_open but for the test at L859 that passes command="order", the import inside the if-block is never executed so the patch is never reached. The test may pass for the wrong reason: the kitchen check is skipped not because the guard logic is correct but because the import is never reached. This could mask a regression if the import location changes.

tests/cli/test_update_checks_prompt.py
```diff
@@ -841,3 +841,64 @@ def test_run_update_checks_shows_notification_when_kitchen_open(
...
+        "run_update_checks must call _binary_signal (signal gathering proceeds) for 'order'"
+    )
+
+
+def test_run_update_checks_kitchen_guard_active_for_update_command(
+    monkeypatch: pytest.MonkeyPatch, tmp_path: Path
+) -> None:
+    """run_update_checks with command='update' retains kitchen guard suppression."""
+    printed, input_calls = _setup_run_checks(monkeypatch, tmp_path, binary_signal=True)
+    monkeypatch.setattr("autoskillit.core.any_kitchen_open", lambda **kw: True)
```
Trecek, line 874: [warning] tests: monkeypatch patches autoskillit.core.any_kitchen_open but the production code imports it locally inside the conditional branch. This is fragile: if the import is ever hoisted out of the if-block the patch will silently stop working. Consider patching autoskillit.cli._update_checks.any_kitchen_open or documenting the dependency on the late-import pattern.

tests/cli/test_update_checks_prompt.py
```diff
@@ -841,3 +841,64 @@ def test_run_update_checks_shows_notification_when_kitchen_open(
...
+    )
+    assert not input_calls, (
+        "run_update_checks must not prompt interactively for 'update' + kitchen"
+    )
+    assert not binary_signal_calls, (
+        "run_update_checks must not reach signal gathering for 'update' + kitchen"
+    )
+
+
+def test_kitchen_guarded_commands_registry() -> None:
```
Trecek, line 896: [warning] tests: test_kitchen_guarded_commands_registry hard-codes the exact frozenset {"update", "install", "init"}. This makes the test a change-detector: adding a legitimately guarded command requires updating both production code and this assertion in lock-step with no safety net. Consider asserting invariants (required members present, all lowercase strings) rather than exact equality, or add a comment explaining why exact membership is intentional. (requires_decision: true)

Trecek (commented):

Trecek (commented):

src/autoskillit/cli/_update_checks.py
Trecek: Investigated — this is intentional. The narrowing to KITCHEN_GUARDED_COMMANDS is the explicit purpose of this PR (commit a214c860: 'any_kitchen_open guard fires only for mutation commands (update, install, init). Non-mutation commands now see informational update signals even when a kitchen is open.'). The branch name encodes the intent and test_kitchen_guard_gated_by_kitchen_guarded_commands enforces the gated structure as an architectural invariant.

tests/arch/test_kitchen_guard_scoping.py
Trecek: Agreed — fixed in 85239b16. Replaced the ast.dump string-contains check (which could match Constant nodes with the literal text 'KITCHEN_GUARDED_COMMANDS') with a pure ast.walk checking isinstance(node, ast.Name) and node.id == 'KITCHEN_GUARDED_COMMANDS' across all nodes in if_node.test.

tests/cli/test_app_main.py
Trecek: Investigated — this is intentional. app.main() always calls run_update_checks(command=_first_arg) with home omitted (defaults to None). The home= parameter is never passed positionally from the actual call site (app.py:446), so the **kwargs spy correctly captures all arguments. The test's sole purpose is to verify command= propagation, which the spy handles correctly.

tests/cli/test_update_checks_prompt.py
Trecek: Agreed — fixed in 7bb953c4. The monkeypatch.setattr('autoskillit.core.any_kitchen_open', ...) at line 851 was dead code for command='order': execution never enters the if command in KITCHEN_GUARDED_COMMANDS block, so any_kitchen_open is never imported or called. The patch was removed to clarify that the test verifies command-gating behavior, not kitchen-open state.

tests/cli/test_update_checks_prompt.py
Trecek: Investigated — this is intentional. For a late import (from autoskillit.core import any_kitchen_open inside the conditional branch), patching autoskillit.core.any_kitchen_open is the correct and only viable approach. The name any_kitchen_open is not bound in autoskillit.cli._update_checks module namespace until the late import fires, so patching autoskillit.cli._update_checks.any_kitchen_open would raise AttributeError at test time. Patching the source module before the late import binds the name is the standard pattern for this pattern.

tests/cli/test_update_checks_prompt.py
Trecek: Investigated — this is intentional. The exact frozenset equality is a deliberate change-detector: the commit message explicitly enumerates {'update', 'install', 'init'} as the complete and intentional membership of mutation commands. Exact equality correctly enforces 'exactly these three commands and no others' — adding a new guarded command should require a conscious, visible update to both production code and test. This is the intended safety invariant of KITCHEN_GUARDED_COMMANDS.

Issue #1586, opened 2026-05-02
CI watch pipeline failure: orchestrator event hallucination + loop guard null crash + missing failure route


Trecek:
## Summary

The implementation pipeline for [TalonT-Org/the-token-reserve PR #119](https://github.com/TalonT-Org/the-token-reserve/pull/119) (issue #117) stalled at the CI watch stage and terminated without attempting test-failure remediation. CI actually ran and completed on GitHub (`cargo nextest` FAILED, `pytest e2e` passed), but the pipeline never had the opportunity to diagnose or fix the test failures.

Three interacting problems caused a cascade failure:

1. **No null-handling instruction for LLM template substitution** — the model passed `event: "push"` when `context.ci_event` was `None`, with no guard rail to catch or prevent this
2. **`check_loop_iteration` crashes on `None` input** — a code bug that bypassed the entire retry loop
3. **`check_ci_already_passed` has no route to CI failure remediation** — treats "CI failed" the same as "CI never ran"

### How recipe template substitution works

AutoSkillit recipes use `${{ context.var }}` template syntax. **There is no server-side template engine.** The raw YAML is delivered to the Claude orchestrator model, which is instructed via the system prompt to perform substitution itself:

> "Substitute `${{ context.var_name }}` and `${{ inputs.var_name }}` placeholders with their resolved values."
>
> — `cli/_prompts.py:610`

There is **no instruction** on what to do when a context variable is `None`/null. No guidance like "pass null as-is" or "omit the parameter when null." The model must infer behavior.

### What happened

**Evidence from orchestrator JSONL log** (`e1f75eaa`):

| Log Line | Event | Detail |
|---|---|---|
| 146 | `check_repo_merge_state` result | `ci_event: None` (correct — `push.branches: ["main"]` doesn't match feature branch) |
| 149 | Orchestrator text | "Captured `ci_event` = `None`" |
| 155 | Orchestrator `wait_for_ci` call | `"event": "push"` — should have been `null` or omitted |
| 156 | `wait_for_ci` result | `conclusion: no_runs`, `scope_used: {event: push, workflow: test.yml}` |
| 191 | User correction | User told the model about the wrong event value |
| 211 | Second attempt | Model correctly **omitted** the `event` parameter |

The model acknowledged `ci_event = None`, then passed `event: "push"` instead of null/omitting it. After user correction, the second attempt was correct.

### Why the model likely chose "push"

[...]

PR #1602, opened 2026-05-02
Rectify: CI Watch Pipeline Null Immunity & Event Validation

Trecek:
## Summary

Two of the three interacting bugs that stalled the CI watch pipeline share a common architectural weakness: the `run_python` dispatch layer passes LLM-provided arguments to Python callables without any null-safety enforcement, and the `wait_for_ci` tool accepts an `event` parameter without validating it against known GitHub Actions trigger events. This plan introduces structural immunity at two chokepoints — the `_import_and_call` dispatch boundary and the `CIRunScope` construction — plus extends the `callable_contracts` schema to declare nullable/non-nullable input constraints enforced by a new semantic rule.

Part B will cover recipe `on_result` routing completeness enforcement and orchestrator prompt null-handling instructions — implement as a separate task.

Closes #1586

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260501-215359-106845/.autoskillit/temp/rectify/rectify_ci_watch_pipeline_null_immunity_2026-05-01_215359_part_a.md`

## Token Usage Summary

| rectify | 55 | 18.4k | 1.1M | 65.2k | 1 | 13m 2s |
| dry_walkthrough | 381 | 34.1k | 5.2M | 176.3k | 2 | 15m 31s |
| implement | 1.3k | 105.0k | 15.3M | 246.6k | 2 | 31m 47s |
| prepare_pr | 60 | 8.1k | 242.8k | 36.8k | 1 | 2m 2s |
| compose_pr | 67 | 1.8k | 225.3k | 19.5k | 1 | 38s |
| review_pr | 4.1k | 73.2k | 2.0M | 185.2k | 2 | 18m 48s |
| resolve_review | 1.1k | 54.5k | 9.0M | 157.6k | 2 | 28m 24s |
| **Total** | 7.1k | 295.1k | 33.0M | 887.2k | | 1h 50m |

### DIFF
--- a/src/autoskillit/core/_type_results.py
+++ b/src/autoskillit/core/_type_results.py
@@ -12,6 +12,7 @@ from enum import Enum
 from pathlib import Path
 from typing import Any, Generic, Literal, TypedDict, TypeVar
 
+from ._type_constants import KNOWN_CI_EVENTS
 from ._type_enums import KillReason, RetryReason, SessionOutcome
 
 T = TypeVar("T")
@@ -272,19 +273,6 @@ class CIRunScope:
             raise ValueError(
                 f"Invalid CI event {self.event!r}. Valid events: {sorted(KNOWN_CI_EVENTS)}"
             )
-
-
-KNOWN_CI_EVENTS: frozenset[str] = frozenset(
-    {
-        "push",
-        "pull_request",
-        "pull_request_target",
-        "merge_group",
-        "workflow_dispatch",
-        "schedule",
-        "workflow_call",
-    }
-)
 
 
 class CloneSuccessResult(TypedDict):
--- a/src/autoskillit/core/_type_results.py
+++ b/src/autoskillit/core/_type_results.py
@@ -27,7 +27,6 @@ __all__ = [
     "SkillResult",
     "CleanupResult",
     "CIRunScope",
-    "KNOWN_CI_EVENTS",
     "CloneSuccessResult",
     "CloneGateUncommitted",
     "CloneGateUnpublished",
--- a/src/autoskillit/core/_type_results.py
+++ b/src/autoskillit/core/_type_results.py
@@ -26,6 +26,7 @@ __all__ = [
     "SkillResult",
     "CleanupResult",
     "CIRunScope",
+    "KNOWN_CI_EVENTS",
     "CloneSuccessResult",
     "CloneGateUncommitted",
     "CloneGateUnpublished",
@@ -265,6 +266,25 @@ class CIRunScope:
     workflow: str | None = None  # workflow filename, e.g. "tests.yml"
     head_sha: str | None = None  # commit SHA to pin results to
     event: str | None = None  # trigger event, e.g. "push", "pull_request"
+
+    def __post_init__(self) -> None:
+        if self.event is not None and self.event not in KNOWN_CI_EVENTS:
+            raise ValueError(
+                f"Invalid CI event {self.event!r}. Valid events: {sorted(KNOWN_CI_EVENTS)}"
+            )
+
+
+KNOWN_CI_EVENTS: frozenset[str] = frozenset(
+    {
+        "push",
+        "pull_request",
+        "pull_request_target",
+        "merge_group",
+        "workflow_dispatch",
+        "schedule",
+        "workflow_call",
+    }
+)
 
 
 class CloneSuccessResult(TypedDict):
--- a/src/autoskillit/recipe/_cmd_rpc.py
+++ b/src/autoskillit/recipe/_cmd_rpc.py
@@ -16,7 +16,7 @@ def compute_branch(
     run_name: str = "",
     issue_number: str = "",
 ) -> dict[str, str]:
-    """Compute branch name from slug + issue or date. Callable via run_python."""
+    """Compute branch name from slug + issue or date."""
     prefix = issue_slug or run_name
     if issue_number:
         return {"branch_name": f"{prefix}/{issue_number}"}
@@ -27,7 +27,7 @@ def check_eject_limit(
     counter_file: str,
     max_ejects: str = "3",
 ) -> dict[str, str]:
-    """Increment counter file; return EJECT_OK or EJECT_LIMIT_EXCEEDED. Callable via run_python."""
+    """Increment counter file; return EJECT_OK or EJECT_LIMIT_EXCEEDED."""
     max_ejects = max_ejects or "3"
     path = Path(counter_file)
     path.parent.mkdir(parents=True, exist_ok=True)
@@ -60,7 +60,7 @@ def check_dropped_healthy_loop(
 
 
 def commit_guard(worktree_path: str) -> dict[str, str]:
-    """Auto-commit pending changes if worktree is dirty. Callable via run_python."""
+    """Auto-commit pending changes if worktree is dirty."""
     result = subprocess.run(
         ["git", "status", "--porcelain"],
         cwd=worktree_path,
@@ -95,7 +95,7 @@ def queue_ejected_fix(
     work_dir: str,
     base_branch: str,
 ) -> dict[str, str]:
-    """Fetch and rebase onto base branch; return clean or conflicts. Callable via run_python."""
+    """Fetch and rebase onto base branch; return clean or conflicts."""
     remote = _detect_remote(work_dir)
     fetch = subprocess.run(
         ["git", "fetch", remote, base_branch],
@@ -125,7 +125,7 @@ def direct_merge_conflict_fix(
     work_dir: str,
     base_branch: str,
 ) -> dict[str, str]:
-    """Attempt rebase for direct-merge path; return clean or conflicts. Callable via run_python."""
+    """Attempt rebase for direct-merge path; return clean or conflicts."""
     return queue_ejected_fix(work_dir=work_dir, base_branch=base_branch)
 
 
@@ -142,7 +142,7 @@ def wait_for_direct_merge(
     max_polls: str = "90",
     poll_interval: str = "10",
 ) -> dict[str, str]:
-    """Poll PR state until merged/closed/timeout. Callable via run_python."""
+    """Poll PR state until merged/closed/timeout."""
     import time  # noqa: PLC0415
 
     max_polls = max_polls or "90"
@@ -170,7 +170,7 @@ def wait_for_immediate_merge(
     max_polls: str = "30",
     poll_interval: str = "10",
 ) -> dict[str, str]:
-    """Poll PR state until merged/closed/timeout (shorter). Callable via run_python."""
+    """Poll PR state until merged/closed/timeout (shorter)."""
     return wait_for_direct_merge(
         pr_number=pr_number, max_polls=max_polls, poll_interval=poll_interval
     )
@@ -181,7 +181,7 @@ def attempt_cheap_rebase(
     ejected_pr_branch: str,
     base_branch: str,
 ) -> dict[str, str]:
-    """Checkout ejected branch and attempt rebase. Callable via run_python."""
+    """Checkout ejected branch and attempt rebase."""
     remote = _detect_remote(work_dir)
     subprocess.run(
         ["git", "fetch", remote, ejected_pr_branch],
@@ -216,7 +216,7 @@ def wait_for_review_pr_mergeability(
     max_polls: str = "12",
     poll_interval: str = "15",
 ) -> dict[str, str]:
-    """Extract PR number and poll until mergeability resolves. Callable via run_python."""
+    """Extract PR number and poll until mergeability resolves."""
     import time  # noqa: PLC0415
 
     max_polls = max_polls or "12"
@@ -289,7 +289,7 @@ def force_push_and_wait_mergeability(
     max_polls: str = "12",
     poll_interval: str = "15",
 ) -> dict[str, str]:
-    """Force-push integration branch and wait for mergeability. Callable via run_python."""
+    """Force-push integration branch and wait for mergeability."""
     import time  # noqa: PLC0415
 
     max_polls = max_polls or "12"
@@ -325,7 +325,7 @@ def advance_queue_pr(
     current_pr_number: str,
     pr_order_file: str,
 ) -> dict[str, str]:
-    """Find next PR in queue order file. Callable via run_python."""
+    """Find next PR in queue order file."""
     if not current_pr_number:
         return {"error": f"current_pr_number is required, got {current_pr_number!r}"}
     try:
@@ -351,7 +351,7 @@ def proactive_rebase_next_pr(
     next_pr_branch: str,
     base_branch: str,
 ) -> dict[str, str]:
-    """Fetch, checkout, and rebase next PR branch. Callable via run_python."""
+    """Fetch, checkout, and rebase next PR branch."""
     remote = _detect_remote(work_dir)
     subprocess.run(
         ["git", "fetch", remote, next_pr_branch],
@@ -382,7 +382,7 @@ def proactive_rebase_next_pr(
 
 
 def refetch_issues(issue_urls: str) -> dict[str, str]:
-    """Build GraphQL query from issue URLs, fetch open issues. Callable via run_python."""
+    """Build GraphQL query from issue URLs, fetch open issues."""
     urls = issue_urls.split(",")
     parts = []
     for i, url in enumerate(urls):
@@ -425,7 +425,7 @@ def emit_fallback_map(
     issue_urls: str,
     temp_dir: str,
 ) -> dict[str, str]:
-    """Build fallback execution map JSON from issue URLs. Callable via run_python."""
+    """Build fallback execution map JSON from issue URLs."""
     nums: list[int] = []
     for url in issue_urls.split(","):
         m = re.search(r"issues/(\d+)", url.strip())
@@ -468,7 +468,7 @@ def export_local_bundle(
     source_dir: str,
     research_dir: str,
 ) -> dict[str, str]:
-    """Copy research dir to source_dir/research-bundles/{slug}/. Callable via run_python."""
+    """Copy research dir to source_dir/research-bundles/{slug}/."""
     import shutil  # noqa: PLC0415
 
     local_root = Path(source_dir) / "research-bundles"
--- a/src/autoskillit/recipe/_cmd_rpc.py
+++ b/src/autoskillit/recipe/_cmd_rpc.py
@@ -327,10 +327,7 @@ def advance_queue_pr(
 ) -> dict[str, str]:
     """Find next PR in queue order file. Callable via run_python."""
     if not current_pr_number:
-        return {
-            "error": f"current_pr_number is required, got {current_pr_number!r}",
-            "current_pr_number": "done",
-        }
+        return {"error": f"current_pr_number is required, got {current_pr_number!r}"}
     try:
         with open(pr_order_file) as f:
             order = json.load(f)
--- a/src/autoskillit/recipe/_cmd_rpc.py
+++ b/src/autoskillit/recipe/_cmd_rpc.py
@@ -28,6 +28,7 @@ def check_eject_limit(
     max_ejects: str = "3",
 ) -> dict[str, str]:
     """Increment counter file; return EJECT_OK or EJECT_LIMIT_EXCEEDED. Callable via run_python."""
+    max_ejects = max_ejects or "3"
     path = Path(counter_file)
     path.parent.mkdir(parents=True, exist_ok=True)
     try:
@@ -44,7 +45,8 @@ def check_dropped_healthy_loop(
     counter_file: str,
     max_drops: str = "2",
 ) -> dict[str, str]:
-    """Increment dropped-healthy counter; return DROPPED_OK or DROPPED_LIMIT_EXCEEDED. Callable via run_python."""
+    """Increment dropped-healthy counter; return DROPPED_OK or DROPPED_LIMIT_EXCEEDED."""
+    max_drops = max_drops or "2"
     path = Path(counter_file)
     path.parent.mkdir(parents=True, exist_ok=True)
     try:
@@ -131,7 +133,7 @@ def immediate_merge_conflict_fix(
     work_dir: str,
     base_branch: str,
 ) -> dict[str, str]:
-    """Attempt rebase for immediate-merge path; return clean or conflicts. Callable via run_python."""
+    """Attempt rebase for immediate-merge path; return clean or conflicts."""
     return queue_ejected_fix(work_dir=work_dir, base_branch=base_branch)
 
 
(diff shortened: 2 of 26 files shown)


src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -318,6 +326,11 @@ def advance_queue_pr(
     pr_order_file: str,
 ) -> dict[str, str]:
     """Find next PR in queue order file. Callable via run_python."""
+    if not current_pr_number:
+        return {
+            "error": f"current_pr_number is required, got {current_pr_number!r}",
+            "current_pr_number": "done",
```
Trecek: [critical] bugs: The error-path return includes "current_pr_number": "done" alongside "error". Any on_result routing that checks current_pr_number == "done" as the terminal sentinel will silently treat this validation failure as successful queue exhaustion, masking the error entirely. Use a value distinct from the normal termination sentinel (e.g., "error") or omit current_pr_number from the error return.

src/autoskillit/core/_type_results.py
```diff
@@ -266,6 +267,25 @@ class CIRunScope:
...
     event: str | None = None  # trigger event, e.g. "push", "pull_request"
 
+    def __post_init__(self) -> None:
+        if self.event is not None and self.event not in KNOWN_CI_EVENTS:
+            raise ValueError(
+                f"Invalid CI event {self.event!r}. Valid events: {sorted(KNOWN_CI_EVENTS)}"
+            )
+
+
+KNOWN_CI_EVENTS: frozenset[str] = frozenset(
```
Trecek: [warning] arch: KNOWN_CI_EVENTS is a frozenset of domain constants, but all other frozenset constants (GATED_TOOLS, FREE_RANGE_TOOLS, SKILL_TOOLS, etc.) live in _type_constants.py. Placing it in _type_results.py breaks the established pattern. Move it to _type_constants.py and import from there.

src/autoskillit/core/_type_results.py
```diff
@@ -266,6 +267,25 @@ class CIRunScope:
     head_sha: str | None = None  # commit SHA to pin results to
     event: str | None = None  # trigger event, e.g. "push", "pull_request"
 
+    def __post_init__(self) -> None:
+        if self.event is not None and self.event not in KNOWN_CI_EVENTS:
```
Trecek, line 271: [warning] defense: KNOWN_CI_EVENTS is defined after the class that references it in __post_init__ (L277 vs L270). This is a forward reference -- valid at runtime since __post_init__ is called at instantiation time, but fragile if definition order ever changes. Define KNOWN_CI_EVENTS before CIRunScope.

src/autoskillit/recipe/contracts.py
```diff
@@ -230,13 +244,34 @@ def get_callable_contract(
...
+        return None
+    fields = {
+        field_name: ToolOutputFieldDef(
+            allowed_values=tuple(field_data.get("allowed_values", [])),
+            terminal_values=frozenset(field_data.get("terminal_values", [])),
+            recoverable_values=frozenset(field_data.get("recoverable_values", [])),
+        )
+        for field_name, field_data in entry.get("fields", {}).items()
+    }
+    return ToolOutputContractDef(result_field=entry.get("result_field", ""), fields=fields)
```
Trecek: [warning] defense: entry.get('result_field', '') falls back to an empty string when result_field is absent from the manifest entry. An empty result_field in ToolOutputContractDef is semantically invalid and silently produces a broken contract object. Return None or raise when result_field is missing rather than defaulting to "".

src/autoskillit/recipe/contracts.py
```diff
@@ -72,6 +73,19 @@ class SkillContract:
     result_fields: list[ResultFieldSpec] = dataclasses.field(default_factory=list)
 
 
+@dataclasses.dataclass(frozen=True)
```
Trecek, line 76: [warning] cohesion: ToolOutputFieldDef and ToolOutputContractDef use the *Def suffix, which per project conventions designates 'static definition of a registered entity... Lives in core/'. These classes reside in recipe/contracts.py, creating a naming convention violation. Either rename to *Spec (since they configure a validation pipeline) or move to core/.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -318,6 +326,11 @@ def advance_queue_pr(
     pr_order_file: str,
 ) -> dict[str, str]:
     """Find next PR in queue order file. Callable via run_python."""
+    if not current_pr_number:
```
Trecek, line 329: [warning] defense: current_pr_number is validated for emptiness here at point-of-use rather than at function entry like the other null-defaulted params (max_polls, poll_interval use early or-defaulting). The inconsistency means the guard could be missed if the function body is restructured. Move the guard to function entry alongside the other null checks.

src/autoskillit/server/tools_execution.py
```diff
@@ -74,6 +74,23 @@ async def _import_and_call(
     if not callable(func):
         return {"success": False, "error": f"{dotted_path!r} is not callable"}
 
+    sig = inspect.signature(func)
+    coerced: dict[str, object] = {}
+    for key, val in args.items():
+        if val is None and key in sig.parameters:
+            param = sig.parameters[key]
+            if param.default is not inspect.Parameter.empty and param.default is not None:
```
Trecek, line 82: [warning] arch: inspect.signature introspection and default-substitution logic (L77-L92) is inlined in the server/MCP handler. Argument coercion policy is business/contract logic that belongs in a shared utility (e.g., execution/ or recipe/), not in the MCP tool handler. The server layer should be a thin dispatch layer.

Trecek (commented):

tests/cli/test_cli_prompts.py
```diff
@@ -573,6 +573,17 @@ def test_orchestrator_prompt_contains_skill_command_format_guidance():
...
 
 
+def test_orchestrator_prompt_includes_null_context_handling():
+    """System prompt must instruct the model on null/None context variable behavior."""
+    from autoskillit.cli._mcp_names import DIRECT_PREFIX
+    from autoskillit.cli._prompts import _build_orchestrator_prompt
+
+    prompt = _build_orchestrator_prompt("my-recipe", mcp_prefix=DIRECT_PREFIX)
+    prompt_lower = prompt.lower()
+    assert "null" in prompt_lower or "none" in prompt_lower
```
Trecek: [warning] tests: Assertions check for 'null' or 'none' anywhere in the lowercased prompt -- an extremely broad match that would pass if those words appeared in any unrelated part of the prompt. Tighten to a specific substring unique to the null-handling instruction (e.g., the exact phrase added in _prompts.py).

tests/recipe/test_check_ci_already_passed_routing.py
```diff
@@ -14,21 +17,82 @@ def recipe(request):
     return load_recipe(builtin_recipes_dir() / f"{request.param}.yaml")
 
 
+def _run_jq(rollup: list[dict]) -> str:
```
Trecek, line 25: [warning] tests: _run_jq() calls load_recipe() and parses the cmd string on every invocation. All four test functions that call this helper re-parse the recipe independently. Consider a session-scoped fixture to load the recipe once and extract the jq expression once, reducing fragility against cmd string formatting changes.

tests/recipe/test_check_ci_already_passed_routing.py
```diff
@@ -14,21 +17,82 @@ def recipe(request):
...
+    """Extract the jq expression from the recipe and run it against test data."""
+    recipe = load_recipe(builtin_recipes_dir() / "implementation.yaml")
+    step = recipe.steps["check_ci_already_passed"]
+    cmd_str = step.with_args["cmd"]
+    jq_start = cmd_str.index("--jq '") + len("--jq '")
+    jq_end = cmd_str.rindex("'")
+    jq_expr = cmd_str[jq_start:jq_end]
+    input_json = json.dumps({"statusCheckRollup": rollup})
+    result = subprocess.run(
```
Trecek, line 34: [warning] tests: subprocess.run(["jq", ...]) invokes a system binary. If jq is not installed in the test environment, tests fail with a confusing FileNotFoundError. Add a jq availability skip guard (e.g., shutil.which("jq") or pytest.skip("jq not available")).

tests/server/test_tools_ci_watch.py
```diff
@@ -0,0 +1,29 @@
...
+    assert result["conclusion"] == "error"
+    assert "event" in result.get("error", "").lower()
+
+
+@pytest.mark.anyio
+async def test_wait_for_ci_coerces_string_none_to_null(tool_ctx) -> None:
+    tool_ctx.ci_watcher = InMemoryCIWatcher(
+        wait_result={"run_id": 42, "conclusion": "success", "failed_jobs": []}
+    )
+    result = json.loads(await wait_for_ci("main", event="None"))
```
Trecek, line 28: [warning] tests: test_wait_for_ci_coerces_string_none_to_null asserts a disjunction: result["conclusion"] != "error" OR "event" not in error. This passes trivially without any coercion -- any non-error conclusion satisfies the first disjunct. Rewrite to positively assert the coercion occurred: e.g., result["conclusion"] == "success".

src/autoskillit/recipe/rules_graph.py
```diff
@@ -10,7 +10,7 @@
     get_logger,
 )
 from autoskillit.recipe._analysis import ValidationContext
-from autoskillit.recipe.contracts import _CONTEXT_REF_RE
+from autoskillit.recipe.contracts import _CONTEXT_REF_RE, get_tool_output_contract
```
Trecek, line 13: [info] arch: _CONTEXT_REF_RE is a private symbol (underscore-prefixed) in contracts.py imported directly into rules_graph.py (and also rules_dataflow.py). Private symbols should not cross module boundaries. Promote it to a public name or move it to a shared internal helpers module.

src/autoskillit/recipe/rules_dataflow.py
```diff
@@ -13,6 +13,7 @@
 )
 from autoskillit.recipe._analysis import ValidationContext
 from autoskillit.recipe.contracts import (
+    _CONTEXT_REF_RE,
```
Trecek, line 16: [info] arch: Same as rules_graph.py -- _CONTEXT_REF_RE (private symbol from contracts.py) imported across module boundaries. Two rules modules depending on a private implementation detail creates tight coupling.

tests/recipe/test_check_ci_already_passed_routing.py
```diff
@@ -14,21 +17,82 @@ def recipe(request):
...
+        ["jq", "-r", jq_expr],
+        input=input_json,
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+    return result.stdout.strip()
+
+
+@pytest.mark.medium
```
Trecek, line 44: [info] tests: @pytest.mark.medium is applied both at module level via pytestmark (L10) and individually on each test function (L39, L45, L51, L56). The per-function markers are redundant with the module-level pytestmark.

tests/recipe/test_cmd_rpc_null_safety.py
```diff
@@ -0,0 +1,21 @@
...
+
+from autoskillit.recipe._cmd_rpc import check_dropped_healthy_loop, check_eject_limit
+
+pytestmark = [pytest.mark.layer("recipe"), pytest.mark.small]
+
+
+def test_check_eject_limit_none_max(tmp_path) -> None:
+    counter = tmp_path / "counter"
+    result = check_eject_limit(counter_file=str(counter), max_ejects=None)  # type: ignore[arg-type]
+    assert "status" in result
```
Trecek, line 15: [info] tests: Asserts only that 'status' is a key in the result -- the minimum possible assertion. Does not verify the status value or that the function behaves correctly when max_ejects=None. Consider asserting the expected status value (e.g., == "EJECT_OK").

Trecek (commented):
AutoSkillit review: blocking issues found. 1 critical finding requires changes before merge. See inline comments for details — all findings are anchored to specific diff lines.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 60069783. The error path returned both "error" key and "current_pr_number": "done", which collided with the terminal sentinel used by on_result routing — silently treating the error as queue exhaustion. Removed the sentinel from the error return.

src/autoskillit/core/_type_results.py
Trecek: Agreed — fixed in 4e9398a7. Moved KNOWN_CI_EVENTS to _type_constants.py alongside all other frozenset domain constants (GATED_TOOLS, FREE_RANGE_TOOLS, SKILL_TOOLS, etc.). _type_results.py re-imports it to preserve public re-exports without breaking callers.

src/autoskillit/core/_type_results.py
Trecek: Investigated — this is intentional. Python dataclass __post_init__ is called at instantiation time (when CIRunScope() is constructed), not at class definition time. KNOWN_CI_EVENTS is fully defined before any CIRunScope() call is ever made. The ordering L270→L277 is valid Python semantics and works correctly at runtime. This concern is addressed as a side-effect of moving KNOWN_CI_EVENTS to _type_constants.py (4e9398a7), which makes it available before the class definition.

src/autoskillit/recipe/contracts.py
Trecek: Agreed — fixed in 0a2c9ae2. entry.get('result_field', '') silently produced an unusable contract (empty string is not a valid field name for routing). Now raises ValueError with a descriptive message at manifest-load time when result_field is absent or empty.

src/autoskillit/recipe/contracts.py
Trecek: Agreed — fixed in 0a2c9ae2. Renamed ToolOutputFieldDef/ToolOutputContractDef to ToolOutputFieldSpec/ToolOutputContractSpec per CLAUDE.md naming convention: *Spec designates behavioral specification objects configuring validation pipelines and lives in recipe/; *Def is reserved for registry-entry definitions in core/.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Investigated — this is intentional. Line 329 IS function entry (the first executable statement after the docstring). The max_polls/poll_interval pattern uses early or-defaulting as null-coercion for parameters with declared defaults; current_pr_number has no default and is validated as a required input. These are distinct patterns serving distinct purposes — null-coercion vs. required-input validation.

src/autoskillit/server/tools_execution.py
Trecek: Investigated — this is intentional. _import_and_call is the sole callsite for this null-arg coercion pattern; it's tightly coupled to the run_python dispatch mechanism. Extracting to a shared utility for a single callsite adds indirection with no concrete benefit. The 15-line implementation is self-contained and readable.

tests/cli/test_cli_prompts.py
Trecek: Agreed — fixed in 8345e361. Replaced broad 'null'/'none' substring match with assertion on the specific section header 'NULL/NONE CONTEXT VARIABLES' and the 'do not guess' instruction, anchoring the test to the actual instruction content rather than incidental word occurrences.

tests/recipe/test_check_ci_already_passed_routing.py
Trecek: Investigated — this is intentional. Per tests/CLAUDE.md: session-scoped fixtures run once per worker process under xdist (-n 4), not once globally. _run_jq also always loads implementation.yaml (not the parametrized recipe fixture), so session scope would add xdist-safety complexity with no correctness benefit for a helper invoked 4 times total.

tests/recipe/test_check_ci_already_passed_routing.py
Trecek: Agreed — fixed in 8345e361. Added _requires_jq = pytest.mark.skipif(shutil.which('jq') is None, reason='jq binary not available in PATH') and applied it to all four jq-invoking test functions.

tests/server/test_tools_ci_watch.py
Trecek: Agreed — fixed in 8345e361. Replaced the trivially-passing disjunction with assert result['conclusion'] == 'success'. This requires the 'None' → null coercion to have occurred: if coercion failed, KNOWN_CI_EVENTS would reject 'None' as an invalid event and return error conclusion.

src/autoskillit/recipe/rules_graph.py
Trecek: Acknowledged — minor suggestion noted.

src/autoskillit/recipe/rules_dataflow.py
Trecek: Acknowledged — minor suggestion noted.

tests/recipe/test_check_ci_already_passed_routing.py
Trecek: Acknowledged — minor suggestion noted.

tests/recipe/test_cmd_rpc_null_safety.py
Trecek: Acknowledged — minor suggestion noted.

Trecek (commented):

src/autoskillit/core/_type_results.py
```diff
@@ -12,6 +12,7 @@
 from pathlib import Path
 from typing import Any, Generic, Literal, TypedDict, TypeVar
 
+from ._type_constants import KNOWN_CI_EVENTS
```
Trecek, line 15: [warning] cohesion: KNOWN_CI_EVENTS is defined in _type_constants.py but re-exported via _type_results.py __all__ (L30). All other frozenset domain constants (GATED_TOOLS, FREE_RANGE_TOOLS, SKILL_TOOLS) are exported only from _type_constants — re-exporting KNOWN_CI_EVENTS from a results module breaks the single-source-of-truth pattern. Remove it from _type_results.__all__ and update callers to import from autoskillit.core._type_constants or autoskillit.core directly.

src/autoskillit/server/tools_ci_watch.py
```diff
@@ -78,6 +78,18 @@ async def wait_for_ci(
     try:
         _start = time.monotonic()
         _timing_ctx = None
+        if event == "None":
```
Trecek: [warning] cohesion: The string-sentinel coercion "event == 'None'" is a wire-format sanitization concern that belongs in _wire_compat.py alongside other wire sanitization middleware, not inlined in the tool implementation. This also silently masks mis-typed callers; the warning log should use a consistent coercion vocabulary shared with other tools. Consider moving to _wire_compat or at least extracting to a shared coerce_none_string() helper.

src/autoskillit/recipe/contracts.py
```diff
@@ -230,13 +244,39 @@ def get_callable_contract(
...
+def get_tool_output_contract(
+    tool_name: str, manifest: dict[str, Any] | None = None
+) -> ToolOutputContractSpec | None:
+    """Return the ToolOutputContractSpec for a named MCP tool, or None if not declared."""
+    if manifest is None:
+        manifest = load_bundled_manifest()
+    entry = manifest.get("tool_output_contracts", {}).get(tool_name)
+    if entry is None:
+        return None
+    fields = {
```
Trecek: [warning] defense: The dict comprehension at L264 iterates entry.get("fields", {}).items() with no type guard on field_data. If YAML parsing yields a non-dict value for a field entry (e.g. a null or scalar), ToolOutputFieldSpec construction will raise a TypeError with no diagnostic about which field or manifest entry is malformed. Add an isinstance(field_data, dict) guard and raise ValueError with the field name when the guard fails.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -44,7 +45,8 @@ def check_dropped_healthy_loop(
     counter_file: str,
     max_drops: str = "2",
 ) -> dict[str, str]:
-    """Increment dropped-healthy counter; return DROPPED_OK or DROPPED_LIMIT_EXCEEDED. Callable via run_python."""
+    """Increment dropped-healthy counter; return DROPPED_OK or DROPPED_LIMIT_EXCEEDED."""
```
Trecek, line 48: [warning] slop: This PR removes the "Callable via run_python." suffix from 4 docstrings (L48, L136/L254 after shifting, L454) but leaves it on several peer functions in the same file (e.g. L145 wait_for_direct_merge still has it). Either finish removing the suffix uniformly across the file or revert these partial removals — partial cleanup leaves the module internally inconsistent.

tests/execution/test_ci.py
```diff
@@ -632,3 +632,16 @@ async def mock_fetch_completed(client, headers, repo, branch, scope, cutoff_dt):
     assert result["run_id"] == 77
     assert result["conclusion"] == "failure"
     assert "build" in result["failed_jobs"]
+
+
+def test_known_ci_events_frozenset_exists() -> None:
+    from autoskillit.core._type_results import KNOWN_CI_EVENTS
```
Trecek: [warning] tests: test_known_ci_events_frozenset_exists imports KNOWN_CI_EVENTS from autoskillit.core._type_results (the re-export shim), not from the canonical autoskillit.core._type_constants where it is defined. The test should import from the authoritative source: from autoskillit.core._type_constants import KNOWN_CI_EVENTS. Using the re-export shim couples the test to the intermediate re-export rather than the definition.

Trecek (commented):

src/autoskillit/core/_type_results.py
Trecek: Agreed — fixed in b3d771bc. Removed KNOWN_CI_EVENTS from _type_results.__all__; the import at L15 stays for L272 internal use. Updated test_known_ci_events_frozenset_exists to import from the authoritative _type_constants.

src/autoskillit/server/tools_ci_watch.py
Trecek: Agreed — fixed in 3a8d3580. Extracted _coerce_none_string(value, *, tool) module-level helper in tools_ci_watch.py. Moving to _wire_compat.py would be a mismatch (it is a FastMCP middleware class, not a utility module); the named helper provides the consistent coercion vocabulary requested.

src/autoskillit/recipe/contracts.py
Trecek: Agreed — fixed in 7316d160. The dict comprehension was replaced with an explicit loop that checks isinstance(field_data, dict) before constructing ToolOutputFieldSpec, raising ValueError with the tool name and field name when the guard fails.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in d087eb2a. Removed the 'Callable via run_python.' suffix from all 15 remaining docstrings in the file uniformly. The module docstring ('run_python entry points') already documents this, making per-function suffixes redundant.

tests/execution/test_ci.py
Trecek: Agreed — fixed in b3d771bc. Changed the import from autoskillit.core._type_results (re-export shim) to autoskillit.core._type_constants (authoritative source).

Issue #1595, opened 2026-05-02
review-pr / resolve-review: duplicate comments on multi-round review cycles


Trecek:
## Summary

Two bugs in the review-pr / resolve-review loop cause duplicate GitHub PR comments when multiple review rounds occur on the same PR.

### Bug 1: resolve-review re-replies to already-handled comments

When resolve-review runs a second time (after a second review-pr pass), it fetches ALL inline comments — including ones it already replied to in a previous round. There is no filter checking whether a reply from the AutoSkillit bot already exists on a comment thread. This causes duplicate "Acknowledged — minor suggestion noted" replies on info findings, duplicate "Investigated — this is intentional" replies on REJECT findings, and duplicate "Valid observation — flagged for design decision" replies on DISCUSS findings.

**Evidence from PR #1588:**
- Comment 3176019318 (`_VAR_ASSIGN_RE` info) received "Acknowledged" at 03:44 AND again at 04:09
- Comment 3176019319 (allowlist info) received "Acknowledged" at 03:44 AND again at 04:09  
- Comment 3176019320 (`commit_guard` info) received "Acknowledged" at 03:44 AND again at 04:09
- Comment 3176019322 (remote detection info) received "Acknowledged" at 03:44 AND again at 04:09
- Comment 3176019311 (forwarding wrappers) received "Valid observation" at 03:44 AND again at 04:09

### Bug 2: review-pr re-raises findings already addressed by resolve-review

The `prior_resolved_findings` suppression in review-pr Step 1.5 relies on `isResolved` thread state from the GraphQL API. But resolve-review only resolves threads for ACCEPT and REJECT verdicts — DISCUSS and INFO findings are intentionally NOT resolved (per the thread_node_id tracking table in resolve-review Step 4). So when review-pr re-runs, these unresolved threads aren't in `prior_resolved_findings`, and semantically identical findings get re-raised as new inline comments.

**Evidence from PR #1588:**
- Round-2 review at 04:25 re-raised "forwarding wrappers" (already discussed), "check_eject_limit duplication" (already flagged), "wait_for_immediate_merge wrapper" (already flagged)

## Root Cause

Neither skill has a mechanism to detect "this comment already has a resolve-review reply." The suppression in review-pr (Step 1.5) and the thread-skip in resolve-review (Step 2) both key off `isResolved` thread state, but DISCUSS and INFO threads are never resolved by design. There is no hidden marker or reply-presence check to fill this gap.

[...]

PR #1604, opened 2026-05-02
resolve-review / review-pr — Duplicate Comments on Multi-Round Review Cycles

Trecek:
## Summary

Two skills cooperate in a review loop: `review-pr` posts findings as inline PR comments, then `resolve-review` applies fixes and posts reply acknowledgements. When the loop runs more than once (review-pr → resolve-review → review-pr → resolve-review …), two bugs cause duplicate replies:

- **Bug 1 (resolve-review):** Every invocation fetches ALL inline comments and posts a reply to each, regardless of whether a reply was already posted in a prior round. The GraphQL query only fetches the original comment (`comments(first:1)`) — it never sees existing replies, so there is no guard against re-replying.

- **Bug 2 (review-pr):** `prior_resolved_findings` is built by checking `isResolved=true` thread state. However, DISCUSS and INFO findings are intentionally never resolved by `resolve-review`, so they always remain `isResolved=false`. On a re-run, review-pr re-raises these findings as new inline comments.

The fix introduces a hidden HTML marker embedded in every reply posted by resolve-review. Both skills read this marker to make idempotent decisions:
- resolve-review skips any comment whose thread already has a marker-bearing reply
- review-pr treats any thread with a marker-bearing reply as suppressed (equivalent to `isResolved=true`) for the purpose of filtering `prior_resolved_findings`

This follows the existing project convention (`<!-- autoskillit:pipeline-signature ... -->` in PR bodies, `<!-- autoskillit-recipe-hash: ... -->` in diagram files).

Closes #1595

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260502-000721-219151/.autoskillit/temp/make-plan/resolve_review_duplicate_comments_plan_2026-05-02_001628.md`

## Token Usage Summary

| plan | 153 | 12.2k | 810.0k | 58.1k | 1 | 10m 54s |
| verify | 132 | 13.7k | 726.9k | 46.6k | 1 | 5m 52s |
| implement | 284 | 19.1k | 2.4M | 79.0k | 1 | 7m 52s |
| prepare_pr | 60 | 4.3k | 221.9k | 30.2k | 1 | 1m 23s |
| compose_pr | 59 | 2.5k | 194.5k | 19.9k | 1 | 51s |
| review_pr | 124 | 44.3k | 758.8k | 73.1k | 1 | 8m 43s |
| resolve_review | 243 | 15.2k | 1.5M | 62.2k | 1 | 9m 31s |
| **Total** | 1.1k | 111.3k | 6.6M | 369.1k | | 45m 8s |

### DIFF
--- a/tests/skills/test_resolve_review_duplicate_comments.py
+++ b/tests/skills/test_resolve_review_duplicate_comments.py
@@ -138,5 +138,5 @@ def test_step15_marker_bearing_threads_are_resolved() -> None:
 def test_marker_format_consistent_across_skills() -> None:
     resolve_mentions = RESOLVE_SKILL_MD.count("autoskillit:resolved")
     review_mentions = REVIEW_PR_SKILL_MD.count("autoskillit:resolved")
-    assert resolve_mentions >= 4, "resolve-review must reference the marker at least 4 times"
-    assert review_mentions >= 2, "review-pr must reference the marker at least 2 times"
+    assert resolve_mentions >= 5, "resolve-review must reference the marker at least 5 times"
+    assert review_mentions >= 3, "review-pr must reference the marker at least 3 times"
--- a/tests/skills/test_resolve_review_duplicate_comments.py
+++ b/tests/skills/test_resolve_review_duplicate_comments.py
@@ -1,7 +1,6 @@
 import re
 from pathlib import Path
 
-
 RESOLVE_SKILL_MD = (
     Path(__file__).parent.parent.parent
     / "src"
@@ -57,17 +56,23 @@ def _extract_reply_block(text: str, verdict: str) -> str:
 
 def test_accept_reply_has_marker() -> None:
     accept_block = _extract_reply_block(RESOLVE_SKILL_MD, "ACCEPT")
-    assert MARKER_RE.search(accept_block), "ACCEPT reply template missing autoskillit:resolved marker"
+    assert MARKER_RE.search(accept_block), (
+        "ACCEPT reply template missing autoskillit:resolved marker"
+    )
 
 
 def test_reject_reply_has_marker() -> None:
     reject_block = _extract_reply_block(RESOLVE_SKILL_MD, "REJECT")
-    assert MARKER_RE.search(reject_block), "REJECT reply template missing autoskillit:resolved marker"
+    assert MARKER_RE.search(reject_block), (
+        "REJECT reply template missing autoskillit:resolved marker"
+    )
 
 
 def test_discuss_reply_has_marker() -> None:
     discuss_block = _extract_reply_block(RESOLVE_SKILL_MD, "DISCUSS")
-    assert MARKER_RE.search(discuss_block), "DISCUSS reply template missing autoskillit:resolved marker"
+    assert MARKER_RE.search(discuss_block), (
+        "DISCUSS reply template missing autoskillit:resolved marker"
+    )
 
 
 def test_info_reply_has_marker() -> None:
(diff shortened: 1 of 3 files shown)


Trecek (commented):

tests/skills/test_resolve_review_duplicate_comments.py
```diff
@@ -0,0 +1,142 @@
...
+        return m.group(1)
+    # Fall back to any triple-backtick block containing 'query' or 'mutation'
+    for block_match in re.finditer(r"```[^\n]*\n(.*?)```", section_text, re.DOTALL):
+        content = block_match.group(1)
+        if "query" in content or "mutation" in content:
+            return content
+    return ""
+
+
+def _extract_reply_block(text: str, verdict: str) -> str:
```
Trecek, line 47: [warning] tests: _extract_reply_block uses regex r"#\\s+" + re.escape(verdict) with `$` as end terminator in DOTALL mode — will match end-of-string rather than end-of-line, potentially capturing too much content across verdict blocks. If the pattern fails to match (e.g., due to whitespace variation), the helper raises AssertionError rather than a descriptive failure message.

tests/skills/test_resolve_review_duplicate_comments.py
```diff
@@ -0,0 +1,142 @@
...
+        "DISCUSS reply template missing autoskillit:resolved marker"
+    )
+
+
+def test_info_reply_has_marker() -> None:
+    info_block = _extract_reply_block(RESOLVE_SKILL_MD, "INFO")
+    assert MARKER_RE.search(info_block), "INFO reply template missing autoskillit:resolved marker"
+
+
+def test_step2_graphql_fetches_five_comments_with_body() -> None:
```
Trecek, line 83: [warning] tests: test_step2_graphql_fetches_five_comments_with_body asserts `body in graphql_block` but the resolve-review Step 2 uses a bash fenced block (not graphql), so _extract_graphql_block falls back to the generic scanner. The test passes only because the inline bash query string contains `body`. This is fragile — the test name implies checking a GraphQL schema block but actually scans a bash block.

tests/skills/test_resolve_review_duplicate_comments.py
```diff
@@ -0,0 +1,142 @@
...
+def test_step15_marker_bearing_threads_are_resolved() -> None:
+    step15_section = _extract_section(
+        REVIEW_PR_SKILL_MD, "Step 1.5: Fetch Prior Review Thread Context"
+    )
+    assert "autoskillit:resolved" in step15_section
+    assert "prior_resolved_findings" in step15_section
+    assert "has_marker_reply" in step15_section or "marker" in step15_section.lower()
+
+
+def test_marker_format_consistent_across_skills() -> None:
```
Trecek, line 138: [warning] tests: test_marker_format_consistent_across_skills asserts `resolve_mentions >= 4` and `review_mentions >= 2` via raw substring count of entire SKILL.md files. These magic-number thresholds will silently pass even if the marker count drops below the expected structural usage, and do not verify the marker appears in each required location.

Trecek (commented):

tests/skills/test_resolve_review_duplicate_comments.py
Trecek: Investigated — this is intentional. `$` in the lookahead `(?=\n#\s+\w|```|$)` is the correct end-of-string fallback terminator for the last verdict block in the file. With Python `re` (no MULTILINE), `$` matches end-of-string, which is the desired behavior. The assertion at line 53 already has a descriptive message: `f"Reply block for verdict {verdict!r} not found"`.

tests/skills/test_resolve_review_duplicate_comments.py
Trecek: Investigated — this is intentional. The `_extract_graphql_block` fallback to bash blocks containing GraphQL queries is an explicit design choice introduced in commit 507e509 ("correct _extract_graphql_block fallback to filter for query/mutation blocks"). The resolve-review Step 2 section embeds the GraphQL query inside a bash `gh api graphql` call (standard SKILL.md convention), and the fallback correctly identifies it via the `query` keyword. The test verifies that `body` appears in the actual GraphQL query string — not incidentally.

tests/skills/test_resolve_review_duplicate_comments.py
Trecek: Agreed — fixed in ef1096bc7384546b658ce5c03313c25c9f45be21. Updated thresholds from `>= 4` to `>= 5` (resolve-review, actual count: 5) and from `>= 2` to `>= 3` (review-pr, actual count: 3), closing the regression gap where a marker could be removed without triggering a failure.

PR #1616, opened 2026-05-02
Implementation Plan: Install-Context Gating for Experimental Features

### LINKED ISSUE #1611: Experimental features lack install-context gating — will leak to stable releases
## Problem

`FeatureLifecycle.EXPERIMENTAL` documents a contract it doesn't enforce. The enum docstring at `src/autoskillit/core/_type_enums.py:395` states:

> EXPERIMENTAL — On by default when experimental_enabled=True; **disabled on main/stable.**

But the bundled `src/autoskillit/config/defaults.yaml:280` ships `experimental_enabled: true` unconditionally, and the config loader has no awareness of install type, git branch, or release context. If the current `integration` branch is promoted to `main` and packaged as a release, every user gets all EXPERIMENTAL features (currently `fleet` and `planner`) enabled with no warning.

## Current behavior

The config resolution chain has no install-context gate:

1. `defaults.yaml` sets `features.experimental_enabled: true`
2. `_config_loader.py` merges 4 layers (defaults → user → project → secrets) with no branch/install-type awareness
3. `_build_features_dict()` in `settings.py` extracts `experimental_enabled` as a plain bool
4. `is_feature_enabled()` in `feature_flags.py` blanket-enables all EXPERIMENTAL features when `experimental_enabled=True`
5. `detect_install()` / `InstallType` from `_install_info.py` exists but is completely disconnected from config loading — used only for update checks

There is no code path that checks whether the running install is editable/source vs. pip/marketplace before resolving `experimental_enabled`.

## Expected behavior

EXPERIMENTAL features should be enabled by default only in development/source installs. Production/stable installs should have them disabled unless the user explicitly opts in via config or env var (`AUTOSKILLIT_FEATURES__EXPERIMENTAL_ENABLED=true`).

## Affected features

Both currently registered EXPERIMENTAL features would leak:

| Feature | Registry location | `default_enabled` |
|---|---|---|
| `fleet` | `_type_constants.py:386` | `False` |
| `planner` | `_type_constants.py:396` | `False` |

Any future features added with `lifecycle=EXPERIMENTAL` would also be affected.

## Key files

| File | Role |
|---|---|
| `src/autoskillit/config/defaults.yaml:280` | Ships `experimental_enabled: true` unconditionally |
| `src/autoskillit/config/_config_loader.py` | Config layer merge — no install-context awareness |
| `src/autoskillit/config/settings.py:140-201` | `_build_features_dict()` — extracts the bool, no secondary checks |
| `src/autoskillit/core/feature_flags.py:14-53` | `is_feature_enabled()` — the gating primitive |
| `src/autoskillit/core/_type_enums.py:392-404` | `FeatureLifecycle` — documents the unfulfilled contract |
| `src/autoskillit/core/_type_constants.py:385-409` | `FEATURE_REGISTRY` — the two affected features |
| `src/autoskillit/cli/_install_info.py` | `detect_install()` / `InstallType` — exists but not wired to config |

### What the solution needs to accomplish

- EXPERIMENTAL features must be off by default for non-development installs
- Development installs (editable/source) should continue to get them enabled automatically — the current dev workflow should not degrade
- Users must always be able to explicitly opt in or out via config or env var, regardless of install type
- The `FeatureLifecycle` docstring contract must be honored by actual runtime behavior

### Constraints

- `defaults.yaml` is bundled inside the package — it ships with every install. Any approach that relies solely on changing `defaults.yaml` must account for the fact that dev installs read the same file
- `detect_install()` already distinguishes `GIT_VCS`, `LOCAL_EDITABLE`, `LOCAL_PATH`, and `UNKNOWN` — this machinery exists but is not wired into config resolution
- The config loader's 4-layer merge (defaults → user → project → secrets) plus env var overlay is the established pattern. The solution should work within this layering, not around it
- `is_feature_enabled()` is the single gating primitive (IL-0, zero autoskillit imports) — it should remain the decision point, not be bypassed
- Tests exist that hardcode version strings and feature expectations — check for `experimental_enabled` in test fixtures

### What to avoid

- Don't add a git-branch check at runtime — packaged installs have no `.git` directory, and branch detection is fragile
- Don't break the env var override path (`AUTOSKILLIT_FEATURES__EXPERIMENTAL_ENABLED`) — it's documented and used
- Don't scatter install-type checks across multiple call sites — the resolution should happen once during config loading

## Related issues

- #1261 — covers the same problem space and was implemented via PR #1290 (merged 2026-04-26). That design relies on a per-branch committed config approach: main's `.autoskillit/config.yaml` must contain `features: {experimental_enabled: false}` to disable experimental features on stable. **However, that critical config commit on main was never made.** As of today, `git show main:.autoskillit/config.yaml` has no `features` section at all — meaning if integration were promoted right now, experimental features would leak to every user. This is precisely why an automatic install-context gate is needed: the manual config approach already failed in practice. The solution here should not repeat this pattern — the safety must be enforced by code, not by remembering to commit a config line.
- #1101 — Feature Registry Foundation (foundational, largely implemented)

### Group GATE — Install-context resolution

**REQ-GATE-001:** The config loading system must determine the install type (editable, site-packages, source-tree, unknown) during config resolution and use it to derive the effective `experimental_enabled` default.

**REQ-GATE-002:** When no explicit `experimental_enabled` value is set in any config layer or env var, editable and source-tree installs must default to `experimental_enabled=true` and all other install types must default to `experimental_enabled=false`.

**REQ-GATE-003:** An explicit `experimental_enabled` value in any config layer (user, project, secrets) or env var must override the install-type-derived default.

### Group COMPAT — Backward compatibility

**REQ-COMPAT-001:** The env var override path (`AUTOSKILLIT_FEATURES__EXPERIMENTAL_ENABLED`) must continue to work identically to its current behavior.

**REQ-COMPAT-002:** Per-feature explicit overrides (e.g., `features: {fleet: false}`) must continue to take precedence over the `experimental_enabled` blanket, regardless of install type.

**REQ-COMPAT-003:** The existing dev workflow on editable/source installs must not require any new configuration to maintain current behavior.

### Group TEST — Verification

**REQ-TEST-001:** Tests must verify that `is_feature_enabled()` returns `False` for EXPERIMENTAL features when `experimental_enabled` resolves to `false` (non-dev install scenario).

**REQ-TEST-002:** Tests must verify that `is_feature_enabled()` returns `True` for EXPERIMENTAL features when `experimental_enabled` resolves to `true` (dev install scenario).

**REQ-TEST-003:** Tests must verify that explicit config and env var overrides take precedence over the install-type-derived default in both directions.

Trecek:
## Summary

`defaults.yaml` ships `experimental_enabled: true` unconditionally, meaning every install — including stable releases — gets all EXPERIMENTAL features enabled. The `FeatureLifecycle.EXPERIMENTAL` docstring promises these features are "disabled on main/stable" but no code enforces it. The fix introduces install-type awareness into config resolution: when no config layer or env var explicitly sets `experimental_enabled`, the system auto-detects whether the install is a development install (editable) and defaults accordingly. Explicit overrides in any config layer or env var continue to work unchanged.

### Group GATE — Install-context resolution

**REQ-GATE-001:** The config loading system must determine the install type (editable, site-packages, source-tree, unknown) during config resolution and use it to derive the effective `experimental_enabled` default.

**REQ-GATE-002:** When no explicit `experimental_enabled` value is set in any config layer or env var, editable and source-tree installs must default to `experimental_enabled=true` and all other install types must default to `experimental_enabled=false`.

**REQ-GATE-003:** An explicit `experimental_enabled` value in any config layer (user, project, secrets) or env var must override the install-type-derived default.

### Group COMPAT — Backward compatibility

**REQ-COMPAT-001:** The env var override path (`AUTOSKILLIT_FEATURES__EXPERIMENTAL_ENABLED`) must continue to work identically to its current behavior.

**REQ-COMPAT-002:** Per-feature explicit overrides (e.g., `features: {fleet: false}`) must continue to take precedence over the `experimental_enabled` blanket, regardless of install type.

**REQ-COMPAT-003:** The existing dev workflow on editable/source installs must not require any new configuration to maintain current behavior.

### Group TEST — Verification

**REQ-TEST-001:** Tests must verify that `is_feature_enabled()` returns `False` for EXPERIMENTAL features when `experimental_enabled` resolves to `false` (non-dev install scenario).

**REQ-TEST-002:** Tests must verify that `is_feature_enabled()` returns `True` for EXPERIMENTAL features when `experimental_enabled` resolves to `true` (dev install scenario).

**REQ-TEST-003:** Tests must verify that explicit config and env var overrides take precedence over the install-type-derived default in both directions.

Closes #1611

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260502-113623-724734/.autoskillit/temp/make-plan/experimental_features_install_context_gating_plan_2026-05-02_114500.md`

## Token Usage Summary

| plan | 68 | 12.7k | 1.3M | 82.7k | 1 | 10m 52s |
| verify | 52 | 11.7k | 2.0M | 140.3k | 1 | 8m 45s |
| implement | 276 | 17.8k | 1.8M | 58.7k | 1 | 7m 44s |
| prepare_pr | 68 | 6.0k | 258.9k | 31.3k | 1 | 1m 57s |
| compose_pr | 59 | 2.4k | 197.0k | 20.3k | 1 | 42s |
| review_pr | 108 | 22.9k | 595.9k | 56.9k | 1 | 5m 18s |
| resolve_review | 261 | 17.3k | 1.6M | 57.0k | 1 | 8m 18s |
| **Total** | 892 | 90.8k | 7.8M | 447.3k | | 43m 40s |
## Token Efficiency

| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| implement | 0 | — | — | — |
| fix | 13 | 166288.9 | 8503.9 | 1296.9 |
| **Total** | **13** | 590540.2 | 34150.8 | 5190.2 |

### DIFF
--- a/src/autoskillit/config/settings.py
+++ b/src/autoskillit/config/settings.py
@@ -156,7 +156,9 @@ class AutomationConfig:
         Coerces all values to bool.
         """
         raw = dict(raw)  # copy to avoid mutating caller's dict
-        _raw_exp = raw.pop("experimental_enabled", raw.pop("EXPERIMENTAL_ENABLED", _UNSET))
+        _raw_exp = raw.pop("experimental_enabled", _UNSET)
+        if _raw_exp is _UNSET:
+            _raw_exp = raw.pop("EXPERIMENTAL_ENABLED", _UNSET)
         experimental_enabled: bool = is_dev_install() if _raw_exp is _UNSET else bool(_raw_exp)
         result: dict[str, bool] = {}
         for name, value in raw.items():
--- a/src/autoskillit/config/settings.py
+++ b/src/autoskillit/config/settings.py
@@ -156,7 +156,7 @@ class AutomationConfig:
         Coerces all values to bool.
         """
         raw = dict(raw)  # copy to avoid mutating caller's dict
-        _raw_exp = raw.pop("experimental_enabled", _UNSET)
+        _raw_exp = raw.pop("experimental_enabled", raw.pop("EXPERIMENTAL_ENABLED", _UNSET))
         experimental_enabled: bool = is_dev_install() if _raw_exp is _UNSET else bool(_raw_exp)
         result: dict[str, bool] = {}
         for name, value in raw.items():
--- a/src/autoskillit/config/settings.py
+++ b/src/autoskillit/config/settings.py
@@ -58,6 +58,7 @@ from autoskillit.core import (
     atomic_write,
     dump_yaml_str,
     get_logger,
+    is_dev_install,
     is_feature_enabled,
 )
 
@@ -65,6 +66,8 @@ if TYPE_CHECKING:
     from dynaconf import Dynaconf
 
 logger = get_logger(__name__)
+
+_UNSET = object()
 
 __all__ = [
     "AutomationConfig",
@@ -153,7 +156,8 @@ class AutomationConfig:
         Coerces all values to bool.
         """
         raw = dict(raw)  # copy to avoid mutating caller's dict
-        experimental_enabled: bool = bool(raw.pop("experimental_enabled", False))
+        _raw_exp = raw.pop("experimental_enabled", _UNSET)
+        experimental_enabled: bool = is_dev_install() if _raw_exp is _UNSET else bool(_raw_exp)
         result: dict[str, bool] = {}
         for name, value in raw.items():
             if not isinstance(name, str):
--- a/src/autoskillit/core/_install_detect.py
+++ b/src/autoskillit/core/_install_detect.py
@@ -1,0 +1,38 @@
+"""Install-type detection for feature gating — IL-0 (zero autoskillit imports).
+
+is_dev_install() is the canonical predicate for determining whether the
+current autoskillit install is a development (editable) install. Used by
+config resolution to auto-detect the experimental_enabled default.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+
+logger = logging.getLogger(__name__)  # noqa: TID251 — IL-0 module, no autoskillit imports allowed
+
+
+def is_dev_install() -> bool:
+    """Return True if autoskillit is installed in editable (development) mode.
+
+    Reads direct_url.json from package metadata — the same mechanism used
+    by _version_snapshot._install_info() and cli._install_info.detect_install().
+
+    Returns False on any error (missing metadata, malformed JSON, etc.).
+    """
+    try:
+        import importlib.metadata
+
+        dist = importlib.metadata.Distribution.from_name("autoskillit")
+        raw = dist.read_text("direct_url.json")
+        if not raw:
+            return False
+        data = json.loads(raw)
+        dir_info = data.get("dir_info", {})
+        if isinstance(dir_info, dict) and dir_info.get("editable") is True:
+            return True
+        return False
+    except Exception:
+        logger.debug("install type detection failed", exc_info=True)
+        return False
--- a/src/autoskillit/core/_install_detect.py
+++ b/src/autoskillit/core/_install_detect.py
@@ -1,9 +1,4 @@
-"""Install-type detection for feature gating — IL-0 (zero autoskillit imports).
-
-is_dev_install() is the canonical predicate for determining whether the
-current autoskillit install is a development (editable) install. Used by
-config resolution to auto-detect the experimental_enabled default.
-"""
+"""Install-type detection for feature gating — IL-0."""
 
 from __future__ import annotations
 
@@ -14,13 +9,7 @@ logger = logging.getLogger(__name__)  # noqa: TID251 — IL-0 module, no autoski
 
 
 def is_dev_install() -> bool:
-    """Return True if autoskillit is installed in editable (development) mode.
-
-    Reads direct_url.json from package metadata — the same mechanism used
-    by _version_snapshot._install_info() and cli._install_info.detect_install().
-
-    Returns False on any error (missing metadata, malformed JSON, etc.).
-    """
+    """Return True if installed in editable mode; False on any error."""
     try:
         import importlib.metadata
 
--- a/tests/arch/test_feature_registry.py
+++ b/tests/arch/test_feature_registry.py
@@ -417,13 +417,43 @@ def test_build_features_dict_extracts_experimental_enabled():
     assert "experimental_enabled" not in result
 
 
-def test_build_features_dict_experimental_enabled_defaults_false():
-    """_build_features_dict returns experimental_enabled=False when key absent."""
+def test_build_features_dict_absent_experimental_enabled_auto_detects(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """_build_features_dict calls is_dev_install() when experimental_enabled absent."""
     from autoskillit.config.settings import AutomationConfig
 
+    monkeypatch.setattr("autoskillit.config.settings.is_dev_install", lambda: True)
+    result, exp_enabled = AutomationConfig._build_features_dict({})
+    assert exp_enabled is True
+    assert result == {}
+
+    monkeypatch.setattr("autoskillit.config.settings.is_dev_install", lambda: False)
     result, exp_enabled = AutomationConfig._build_features_dict({})
     assert exp_enabled is False
     assert result == {}
+
+
+def test_build_features_dict_explicit_true_overrides_auto_detect(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """_build_features_dict returns True when explicit True, ignoring is_dev_install."""
+    from autoskillit.config.settings import AutomationConfig
+
+    monkeypatch.setattr("autoskillit.config.settings.is_dev_install", lambda: False)
+    _, exp_enabled = AutomationConfig._build_features_dict({"experimental_enabled": True})
+    assert exp_enabled is True
+
+
+def test_build_features_dict_explicit_false_overrides_auto_detect(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """_build_features_dict returns False when explicit False, ignoring is_dev_install."""
+    from autoskillit.config.settings import AutomationConfig
+
+    monkeypatch.setattr("autoskillit.config.settings.is_dev_install", lambda: True)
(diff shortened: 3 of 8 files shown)


tests/arch/test_feature_registry.py
```diff
@@ -437,29 +468,103 @@ def test_build_config_schema_accepts_experimental_enabled():
...
-    with patch.dict(os.environ, {}, clear=False):
-        cfg = load_config(tmp_path)
+    monkeypatch.setattr("autoskillit.config.settings.is_dev_install", lambda: False)
+    fake_home = tmp_path / "home"
+    user_config_dir = fake_home / ".autoskillit"
+    user_config_dir.mkdir(parents=True)
+    (user_config_dir / "config.yaml").write_text(
+        yaml.dump({"features": {"experimental_enabled": True}})
+    )
+    monkeypatch.setattr("pathlib.Path.home", lambda: fake_home)
```
Trecek, line 568: [warning] tests: monkeypatch.setattr("pathlib.Path.home", lambda: fake_home) patches a global classmethod on pathlib.Path. Under xdist -n 4, other tests in the same worker process that call Path.home() concurrently can observe the fake value. Patch the config loader's specific call site (e.g., "autoskillit.config._config_loader.Path.home") instead of the global stdlib classmethod to avoid cross-test interference.

tests/arch/test_feature_registry.py
```diff
@@ -417,15 +418,45 @@ def test_build_features_dict_extracts_experimental_enabled():
     assert "experimental_enabled" not in result
 
 
-def test_build_features_dict_experimental_enabled_defaults_false():
-    """_build_features_dict returns experimental_enabled=False when key absent."""
+def test_build_features_dict_absent_experimental_enabled_auto_detects(
```
Trecek, line 421: [warning] tests: test_build_features_dict_absent_experimental_enabled_auto_detects applies two consecutive monkeypatch.setattr calls on the same target without any reset between them. The second setattr silently overwrites the first patch. Split into two separate tests (one for is_dev_install=True, one for False) to make intent clear and match the pattern at L487.

src/autoskillit/config/settings.py
```diff
@@ -153,7 +156,8 @@ def _build_features_dict(raw: dict[str, Any]) -> tuple[dict[str, bool], bool]:
         Coerces all values to bool.
         """
         raw = dict(raw)  # copy to avoid mutating caller's dict
-        experimental_enabled: bool = bool(raw.pop("experimental_enabled", False))
+        _raw_exp = raw.pop("experimental_enabled", raw.pop("EXPERIMENTAL_ENABLED", _UNSET))
```
Trecek: [warning] bugs: Double-pop side effect: raw.pop("experimental_enabled", raw.pop("EXPERIMENTAL_ENABLED", _UNSET)) always evaluates the inner pop, removing "EXPERIMENTAL_ENABLED" from raw even when "experimental_enabled" is present. If both keys coexist, the uppercase value is silently consumed without being used. Whether this is intentional aliasing or a bug depends on intent — if "EXPERIMENTAL_ENABLED" should only be consumed when "experimental_enabled" is absent, use an explicit two-step lookup.

src/autoskillit/core/_install_detect.py
```diff
@@ -0,0 +1,38 @@
+"""Install-type detection for feature gating — IL-0 (zero autoskillit imports).
```
Trecek: [warning] slop: Module docstring is over-verbose. The IL-0 label and "Used by config resolution" sentence duplicate what the function docstring already covers. Reduce to a single line: """Install-type detection for feature gating — IL-0."""

src/autoskillit/core/_install_detect.py
```diff
@@ -0,0 +1,38 @@
...
+from __future__ import annotations
+
+import json
+import logging
+
+logger = logging.getLogger(__name__)  # noqa: TID251 — IL-0 module, no autoskillit imports allowed
+
+
+def is_dev_install() -> bool:
+    """Return True if autoskillit is installed in editable (development) mode.
```
Trecek: [warning] slop: Function docstring references internal implementation details of other modules (_version_snapshot._install_info(), cli._install_info.detect_install()). Cross-module references in docstrings are maintenance burden. Keep only the contract line: "Returns True if installed in editable mode; False on any error."

Trecek (commented):

Trecek (commented):

tests/arch/test_feature_registry.py
Trecek: Investigated — this is intentional. xdist uses separate worker *processes* (not threads), so there is no shared memory across workers. Within a single worker, tests run sequentially and monkeypatch is test-scoped, restoring the original classmethod at teardown. The patch on pathlib.Path.home directly is correct: the production code calls Path.home() as a classmethod, so a module-local alias patch would not intercept calls from other imports. The current pattern is safe and consistent with tests/CLAUDE.md which mandates monkeypatch for module-level state mutations.

tests/arch/test_feature_registry.py
Trecek: Investigated — this is intentional. The evidence cited ("match the pattern at L487") is factually incorrect: lines 487-499 (test_load_config_integration_experimental_auto_detects) uses the identical two-consecutive-setattr pattern. The test at L421 is consistent with the established codebase convention. pytest monkeypatch records the original value before the first setattr and restores it at teardown regardless of intermediate overwrites.

src/autoskillit/config/settings.py
Trecek: Agreed — fixed in b75b80f1. Replaced the nested pop expression with an explicit two-step lookup: pop "experimental_enabled" first; only if _UNSET, pop "EXPERIMENTAL_ENABLED". This ensures the uppercase alias is never consumed when the lowercase key is present.

src/autoskillit/core/_install_detect.py
Trecek: Agreed — fixed in 32069896. Reduced module docstring to a single descriptive line: """Install-type detection for feature gating — IL-0."""

src/autoskillit/core/_install_detect.py
Trecek: Agreed — fixed in 32069896. Removed cross-module implementation references; function docstring now states only the public contract: "Return True if installed in editable mode; False on any error."

PR #1638, opened 2026-05-03
Implementation Plan: Collapse task + task_file Into a Single task Ingredient

### LINKED ISSUE #1634: Planner recipe: collapse task + task_file into a single task ingredient
## Problem

The planner recipe has two ingredients for the same thing: `task` (required) and `task_file` (optional). When `task_file` is provided, the user is still forced to provide a redundant `task` one-liner. This is broken UX — there's one task. It can be inline text, a file path, or whatever. The system should figure out what it is. The user provides one thing.

The implementation recipe gets this right: one `task` ingredient.

## What Happens Today

1. User provides `task_file` pointing to a detailed scoping document
2. Recipe forces user to also type a `task` one-liner
3. Every Python callable does `if task_file: task = Path(task_file).read_text()` — discarding the one-liner
4. The two entry-point skills receive ONLY the file path — the `task` string never reaches them
5. The bare `task` string survives only as a `# Plan: {task_label}` heading in `plan.md`

## Design Constraints (validated)

1. **Context bloat is real**: interpolating a 5000-word document 7× into the orchestrator's context (via `${{ inputs.task }}` in step `with:` blocks) wastes tens of thousands of tokens. The file-path indirection exists to avoid this — a short path string is interpolated instead.
2. **Skills need a file path**: `planner-generate-phases` and `planner-extract-domain` read from disk via `$3`/`$2`. They need a path, not inline content.
3. **`plan.md` needs a short heading**: `compiler.py` uses `task_label` for the `# Plan:` title.

## Fix

**One user-facing ingredient: `task`.** Accepts anything — file path, inline text, URL, whatever.

The `init` step resolves it internally:
- If `task` is a path to an existing file → use that path as the internal task file; derive label from filename or first heading
- If `task` is inline text → write it to a temp file in `planner_dir`; use that path as the internal task file; derive label from first line or truncation

After `init`, the pipeline has two context variables:
- `context.task_file_path` — always a file path (short string, safe to interpolate N times)
- `context.task_label` — short heading string derived from content

Skills continue receiving a file path (no change to `planner-generate-phases` or `planner-extract-domain`). Python callables receive the file path and read it themselves (no change to their `if task_file:` pattern, just rename the param). The `compile_plan` heading uses `context.task_label`.

The user provides ONE thing. The system handles the rest.

### Affected code

1. `src/autoskillit/recipes/planner.yaml` — remove `task_file` ingredient; add resolution logic to `init` step (run_python callable that detects input type, writes temp file if needed, returns `task_file_path` + `task_label`); update step `with:` blocks to use `context.task_file_path`
2. `src/autoskillit/planner/merge.py` — rename `task_file` param to match new context var name (or keep it, just receives the path from a different source)
3. `src/autoskillit/planner/compiler.py` — use `task_label` from context instead of the pre-resolution `task` capture
4. Skills unchanged — still receive a file path as positional arg
5. Tests — update to reflect single-ingredient contract

### Pattern to follow

The implementation recipe: one `task` ingredient, no type detection at the recipe layer, opaque passthrough. The planner adds one extra step: because its pipeline reuses the task 7× and skills read from disk, the init step materializes the task to a file path once.

Trecek:
## Summary

Replace the planner recipe's two task-related ingredients (`task` required + `task_file` optional) with a single `task` ingredient that accepts either inline text or a file path. Add a new `resolve_task` pipeline step immediately after `init` that detects the input type, materializes inline text to a file when needed, and captures two context variables (`task_file_path` and `task_label`) used by all downstream steps. Remove `task_file` from the recipe, Python callables, and tests. Skills remain unchanged since they already receive a file path positionally.

Closes #1634

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260502-174057-807186/.autoskillit/temp/make-plan/collapse_task_task_file_plan_2026-05-02_180500.md`

## Token Usage Summary

| plan | 2.1k | 18.6k | 1.7M | 93.4k | 1 | 8m 30s |
| verify | 56 | 23.9k | 2.1M | 78.8k | 1 | 11m 2s |
| implement | 800 | 41.0k | 9.0M | 128.3k | 1 | 13m 29s |
| prepare_pr | 68 | 5.9k | 274.3k | 34.8k | 1 | 1m 53s |
| compose_pr | 59 | 2.2k | 195.2k | 20.7k | 1 | 47s |
| review_pr | 268 | 70.1k | 2.0M | 180.2k | 2 | 17m 15s |
| resolve_review | 552 | 51.1k | 3.8M | 202.6k | 2 | 22m 58s |
| ci_conflict_fix | 172 | 7.7k | 935.6k | 43.4k | 1 | 2m 55s |
| diagnose_ci | 34 | 1.0k | 134.0k | 2.7k | 1 | 6m 40s |
| resolve_ci | 421 | 3.7k | 334.5k | 38.0k | 1 | 3m 41s |
| **Total** | 4.5k | 225.2k | 20.4M | 822.9k | | 1h 29m |

### DIFF
--- a/src/autoskillit/planner/compiler.py
+++ b/src/autoskillit/planner/compiler.py
@@ -123,7 +123,7 @@ def _render_issue_body(wp: dict, phase: dict, assignment: dict) -> str:
 def compile_plan(
     output_dir: str, task_file_path: str, source_dir: str, task_label: str = "", **kwargs: Any
 ) -> dict[str, str]:
-    task = Path(task_file_path).read_text() if task_file_path else ""
+    task = Path(task_file_path).read_text(encoding="utf-8") if task_file_path else ""
     task_label = task_label or _derive_label(task, "")
     root = Path(output_dir)
 
--- a/src/autoskillit/planner/compiler.py
+++ b/src/autoskillit/planner/compiler.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from typing import Any
 
 from autoskillit.core import atomic_write, get_logger, write_versioned_json
+from autoskillit.planner.manifests import _derive_label
 from autoskillit.planner.validation import (
     _load_assignment_results,
     _load_phase_results,
@@ -120,11 +121,10 @@ def _render_issue_body(wp: dict, phase: dict, assignment: dict) -> str:
 
 
 def compile_plan(
-    output_dir: str, task: str, source_dir: str, task_file: str = "", **kwargs: Any
+    output_dir: str, task_file_path: str, source_dir: str, task_label: str = "", **kwargs: Any
 ) -> dict[str, str]:
-    task_label = task
-    if task_file:
-        task = Path(task_file).read_text()
+    task = Path(task_file_path).read_text() if task_file_path else ""
+    task_label = task_label or _derive_label(task, "")
     root = Path(output_dir)
 
     phase_results = _load_phase_results(root)
--- a/src/autoskillit/planner/manifests.py
+++ b/src/autoskillit/planner/manifests.py
@@ -435,6 +435,6 @@ def resolve_task_input(task: str, planner_dir: str) -> TaskResolutionResult:
         label = _derive_label(content, task_path.stem)
         return TaskResolutionResult(task_file_path=str(task_path), task_label=label)
     out = Path(planner_dir) / "task_input.md"
-    out.write_text(task)
+    atomic_write(out, task)
     label = _derive_label(task, "")
     return TaskResolutionResult(task_file_path=str(out), task_label=label)
--- a/src/autoskillit/planner/manifests.py
+++ b/src/autoskillit/planner/manifests.py
@@ -415,15 +415,14 @@ def expand_wps(refined_assignments_path: str, output_dir: str, **kwargs: object)
 
 
 def _derive_label(content: str, filename_stem: str) -> str:
+    first_non_empty: str | None = None
     for line in content.splitlines():
         stripped = line.strip()
         if stripped.startswith("# "):
             return stripped[2:].strip()
-    for line in content.splitlines():
-        stripped = line.strip()
-        if stripped:
-            return stripped[:80]
-    return filename_stem or "Untitled"
+        if first_non_empty is None and stripped:
+            first_non_empty = stripped[:80]
+    return first_non_empty or filename_stem or "Untitled"
 
 
 def resolve_task_input(task: str, planner_dir: str) -> TaskResolutionResult:
@@ -431,7 +430,10 @@ def resolve_task_input(task: str, planner_dir: str) -> TaskResolutionResult:
         raise ValueError("task must be a non-empty string")
     task_path = Path(task)
     if task_path.is_file():
-        content = task_path.read_text()
+        try:
+            content = task_path.read_text(encoding="utf-8")
+        except OSError as exc:
+            raise OSError(f"Cannot read task file {task_path}: {exc}") from exc
         label = _derive_label(content, task_path.stem)
         return TaskResolutionResult(task_file_path=str(task_path), task_label=label)
     out = Path(planner_dir) / "task_input.md"
--- a/src/autoskillit/planner/manifests.py
+++ b/src/autoskillit/planner/manifests.py
@@ -10,6 +10,7 @@ from typing import TypedDict
 from autoskillit.core import atomic_write, write_versioned_json
 from autoskillit.planner.schema import (
     RunDirResult,
+    TaskResolutionResult,
     validate_assignment_result,
     validate_phase_result,
     validate_refined_assignments,
@@ -411,3 +412,29 @@ def expand_wps(refined_assignments_path: str, output_dir: str, **kwargs: object)
         "context_paths": ",".join(context_paths),
         "item_ids": ",".join(item_ids),
     }
+
+
+def _derive_label(content: str, filename_stem: str) -> str:
+    for line in content.splitlines():
+        stripped = line.strip()
+        if stripped.startswith("# "):
+            return stripped[2:].strip()
+    for line in content.splitlines():
+        stripped = line.strip()
+        if stripped:
+            return stripped[:80]
+    return filename_stem or "Untitled"
+
+
+def resolve_task_input(task: str, planner_dir: str) -> TaskResolutionResult:
+    if not task:
+        raise ValueError("task must be a non-empty string")
+    task_path = Path(task)
+    if task_path.is_file():
+        content = task_path.read_text()
+        label = _derive_label(content, task_path.stem)
+        return TaskResolutionResult(task_file_path=str(task_path), task_label=label)
+    out = Path(planner_dir) / "task_input.md"
+    out.write_text(task)
+    label = _derive_label(task, "")
+    return TaskResolutionResult(task_file_path=str(out), task_label=label)
--- a/src/autoskillit/planner/merge.py
+++ b/src/autoskillit/planner/merge.py
@@ -22,7 +22,7 @@ def merge_files(
     strict: bool = True,
     **kwargs: Any,
 ) -> dict[str, Any]:
-    task = Path(task_file_path).read_text() if task_file_path else ""
+    task = Path(task_file_path).read_text(encoding="utf-8") if task_file_path else ""
     if key not in _TIER_KEYS:
         raise ValueError(f"Invalid key {key!r}; must be one of {_TIER_KEYS}")
 
@@ -187,7 +187,7 @@ def build_plan_snapshot(
     source_dir: str = "",
     **kwargs: Any,
 ) -> dict[str, Any]:
-    task = Path(task_file_path).read_text() if task_file_path else ""
+    task = Path(task_file_path).read_text(encoding="utf-8") if task_file_path else ""
     phase_pairs: list[tuple[int, dict[str, Any]]] = []
     for p in sorted(Path(phases_dir).glob("*_result.json")):
         try:
--- a/src/autoskillit/planner/merge.py
+++ b/src/autoskillit/planner/merge.py
@@ -17,14 +17,12 @@ def merge_files(
     file_paths: list[str],
     output_path: str,
     key: str,
-    task: str = "",
+    task_file_path: str = "",
     source_dir: str = "",
     strict: bool = True,
-    task_file: str = "",
     **kwargs: Any,
 ) -> dict[str, Any]:
-    if task_file:
-        task = Path(task_file).read_text()
+    task = Path(task_file_path).read_text() if task_file_path else ""
     if key not in _TIER_KEYS:
         raise ValueError(f"Invalid key {key!r}; must be one of {_TIER_KEYS}")
 
@@ -166,9 +164,8 @@ def merge_tier_dir(
     results_dir: str,
     output_path: str,
     key: str,
-    task: str = "",
+    task_file_path: str = "",
     source_dir: str = "",
-    task_file: str = "",
     **kwargs: Any,
 ) -> dict[str, Any]:
     paths = sorted(Path(results_dir).glob("*_result.json"))
@@ -178,22 +175,19 @@ def merge_tier_dir(
         file_paths=[str(p) for p in paths],
         output_path=output_path,
         key=key,
-        task=task,
+        task_file_path=task_file_path,
         source_dir=source_dir,
-        task_file=task_file,
     )
 
 
 def build_plan_snapshot(
     phases_dir: str,
     output_path: str,
-    task: str = "",
+    task_file_path: str = "",
     source_dir: str = "",
-    task_file: str = "",
     **kwargs: Any,
 ) -> dict[str, Any]:
-    if task_file:
-        task = Path(task_file).read_text()
+    task = Path(task_file_path).read_text() if task_file_path else ""
     phase_pairs: list[tuple[int, dict[str, Any]]] = []
     for p in sorted(Path(phases_dir).glob("*_result.json")):
         try:
--- a/tests/planner/conftest.py
+++ b/tests/planner/conftest.py
@@ -61,3 +61,9 @@ def write_json(path: Path, data: object) -> None:
     """Write ``data`` as JSON to ``path``, creating parent dirs as needed."""
     path.parent.mkdir(parents=True, exist_ok=True)
     path.write_text(json.dumps(data))
+
+
+def write_task_file(tmp_path: Path, content: str = "test task") -> str:
+    f = tmp_path / "task_input.md"
+    f.write_text(content)
+    return str(f)
--- a/tests/planner/test_manifests.py
+++ b/tests/planner/test_manifests.py
@@ -513,3 +513,62 @@ def test_finalize_wp_manifest_nonexistent_dir_raises(tmp_path):
 
     with pytest.raises(FileNotFoundError, match="work_packages_dir does not exist"):
         finalize_wp_manifest(str(tmp_path / "nonexistent"), str(output_dir))
+
+
+# --- resolve_task_input tests (T1a-T1e) ---
+
+
+def test_resolve_task_input_file_with_heading(tmp_path):
+    from autoskillit.planner import resolve_task_input
+
+    task_file = tmp_path / "task.md"
+    task_file.write_text("# Deploy Auth Service\n\nDetailed description...")
+    planner_dir = tmp_path / "planner"
+    planner_dir.mkdir()
+    result = resolve_task_input(str(task_file), str(planner_dir))
+    assert result["task_file_path"] == str(task_file)
+    assert result["task_label"] == "Deploy Auth Service"
+
+
+def test_resolve_task_input_file_no_heading(tmp_path):
+    from autoskillit.planner import resolve_task_input
+
+    task_file = tmp_path / "task.txt"
+    task_file.write_text("Implement the feature flag system for gradual rollout")
+    planner_dir = tmp_path / "planner"
+    planner_dir.mkdir()
+    result = resolve_task_input(str(task_file), str(planner_dir))
+    assert result["task_file_path"] == str(task_file)
+    assert result["task_label"] == "Implement the feature flag system for gradual rollout"
+
+
+def test_resolve_task_input_inline_text(tmp_path):
+    from autoskillit.planner import resolve_task_input
+
+    planner_dir = tmp_path / "planner"
+    planner_dir.mkdir()
+    result = resolve_task_input("Add dark mode toggle", str(planner_dir))
+    assert result["task_file_path"] == str(planner_dir / "task_input.md")
+    assert Path(result["task_file_path"]).read_text() == "Add dark mode toggle"
+    assert result["task_label"] == "Add dark mode toggle"
+
+
+def test_resolve_task_input_inline_with_heading(tmp_path):
+    from autoskillit.planner import resolve_task_input
+
+    planner_dir = tmp_path / "planner"
+    planner_dir.mkdir()
+    text = "# Auth Overhaul\n\nRebuild the entire authentication layer..."
+    result = resolve_task_input(text, str(planner_dir))
+    assert result["task_label"] == "Auth Overhaul"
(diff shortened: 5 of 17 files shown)


src/autoskillit/planner/compiler.py
```diff
@@ -12,6 +12,7 @@
 from typing import Any
 
 from autoskillit.core import atomic_write, get_logger, write_versioned_json
+from autoskillit.planner.manifests import _derive_label
```
Trecek, line 15: [warning] arch: compiler.py imports private symbol _derive_label directly from manifests.py, creating tight intra-layer coupling. _derive_label should be published via __init__.py or compiler.py should use only the public resolve_task_input API.

src/autoskillit/planner/manifests.py
```diff
@@ -411,3 +412,29 @@ def expand_wps(refined_assignments_path: str, output_dir: str, **kwargs: object)
...
+        if stripped.startswith("# "):
+            return stripped[2:].strip()
+    for line in content.splitlines():
+        stripped = line.strip()
+        if stripped:
+            return stripped[:80]
+    return filename_stem or "Untitled"
+
+
+def resolve_task_input(task: str, planner_dir: str) -> TaskResolutionResult:
```
Trecek, line 428: [warning] arch: resolve_task_input performs file I/O (path detection, atomic write of task_input.md) inside the manifests module, which is a data-transformation layer. Side-effectful resolution logic belongs in an orchestration or I/O boundary layer.

src/autoskillit/planner/manifests.py
```diff
@@ -411,3 +412,29 @@ def expand_wps(refined_assignments_path: str, output_dir: str, **kwargs: object)
...
+    if not task:
+        raise ValueError("task must be a non-empty string")
+    task_path = Path(task)
+    if task_path.is_file():
+        content = task_path.read_text()
```
Trecek: [critical] bugs: task_path.read_text() called without encoding='utf-8' and no exception handling. A permissions error, broken symlink, or locale mismatch will propagate as a bare OSError with no identifying context. Add encoding='utf-8' and catch OSError to re-raise with context.

src/autoskillit/planner/compiler.py
```diff
@@ -120,11 +121,10 @@ def _render_issue_body(wp: dict, phase: dict, assignment: dict) -> str:
 
 
 def compile_plan(
-    output_dir: str, task: str, source_dir: str, task_file: str = "", **kwargs: Any
+    output_dir: str, task_file_path: str, source_dir: str, task_label: str = "", **kwargs: Any
 ) -> dict[str, str]:
-    task_label = task
-    if task_file:
-        task = Path(task_file).read_text()
+    task = Path(task_file_path).read_text() if task_file_path else ""
```
Trecek: [warning] bugs: Path(task_file_path).read_text() called without encoding='utf-8'. On systems where locale encoding differs from UTF-8, task files may be silently misread. Use encoding='utf-8' explicitly.

src/autoskillit/planner/merge.py
```diff
@@ -17,14 +17,12 @@ def merge_files(
...
-    task: str = "",
+    task_file_path: str = "",
     source_dir: str = "",
     strict: bool = True,
-    task_file: str = "",
     **kwargs: Any,
 ) -> dict[str, Any]:
-    if task_file:
-        task = Path(task_file).read_text()
+    task = Path(task_file_path).read_text() if task_file_path else ""
```
Trecek: [warning] bugs: Path(task_file_path).read_text() called without encoding='utf-8'. Same locale-encoding risk as compiler.py L126.

src/autoskillit/planner/merge.py
```diff
@@ -178,22 +175,19 @@ def merge_tier_dir(
...
     output_path: str,
-    task: str = "",
+    task_file_path: str = "",
     source_dir: str = "",
-    task_file: str = "",
     **kwargs: Any,
 ) -> dict[str, Any]:
-    if task_file:
-        task = Path(task_file).read_text()
+    task = Path(task_file_path).read_text() if task_file_path else ""
```
Trecek: [warning] bugs: Path(task_file_path).read_text() called without encoding='utf-8' in build_plan_snapshot. Same locale-encoding risk as L25.

src/autoskillit/planner/manifests.py
```diff
@@ -411,3 +412,29 @@ def expand_wps(refined_assignments_path: str, output_dir: str, **kwargs: object)
...
         "item_ids": ",".join(item_ids),
     }
+
+
+def _derive_label(content: str, filename_stem: str) -> str:
+    for line in content.splitlines():
+        stripped = line.strip()
+        if stripped.startswith("# "):
+            return stripped[2:].strip()
+    for line in content.splitlines():
```
Trecek: [warning] bugs: _derive_label calls content.splitlines() twice in sequence. A single pass can locate both '# '-headed and first non-empty line, avoiding the redundant scan.

tests/planner/conftest.py
```diff
@@ -61,3 +61,9 @@ def write_json(path: Path, data: object) -> None:
     """Write ``data`` as JSON to ``path``, creating parent dirs as needed."""
     path.parent.mkdir(parents=True, exist_ok=True)
     path.write_text(json.dumps(data))
+
+
+def write_task_file(tmp_path: Path, content: str = "test task") -> str:
```
Trecek, line 66: [warning] tests: write_task_file always writes to 'task_input.md' in the given directory. If a test calls this helper twice with different content in the same base directory, the second call silently overwrites the first with no diagnostic.

tests/planner/test_manifests.py
```diff
@@ -513,3 +513,62 @@ def test_finalize_wp_manifest_nonexistent_dir_raises(tmp_path):
...
+
+    planner_dir = tmp_path / "planner"
+    planner_dir.mkdir()
+    text = "# Auth Overhaul\n\nRebuild the entire authentication layer..."
+    result = resolve_task_input(text, str(planner_dir))
+    assert result["task_label"] == "Auth Overhaul"
+    assert Path(result["task_file_path"]).read_text() == text
+
+
+def test_resolve_task_input_long_inline_truncates_label(tmp_path):
```
Trecek, line 565: [warning] tests: test_resolve_task_input_long_inline_truncates_label only asserts len(result['task_label']) <= 80 with no check that the label is a prefix of the input text. The test would pass for any string of ≤80 chars, including an empty string.

tests/recipe/test_planner_recipe.py
```diff
@@ -272,43 +273,66 @@ def test_planner_recipe_assess_review_approach_routes_to_validate_on_failure(pla
...
     step = planner_recipe.steps["compile"]
-    assert "task_file" in step.with_args, "compile must receive task_file"
-    assert "inputs.task_file" in step.with_args["task_file"]
+    assert "task_file_path" in step.with_args
+    assert "task_label" in step.with_args
+    assert "context.task_file_path" in step.with_args["task_file_path"]
+    assert "context.task_label" in step.with_args["task_label"]
+
+
+def test_no_step_references_inputs_task_file(planner_recipe):
```
Trecek, line 330: [warning] tests: test_no_step_references_inputs_task_file uses str(step.with_args) for substring matching. Objects with custom __repr__ may produce false negatives. Prefer json.dumps or a recursive dict walk.

tests/planner/test_manifests.py
```diff
@@ -513,3 +513,62 @@ def test_finalize_wp_manifest_nonexistent_dir_raises(tmp_path):
 
     with pytest.raises(FileNotFoundError, match="work_packages_dir does not exist"):
         finalize_wp_manifest(str(tmp_path / "nonexistent"), str(output_dir))
+
+
+# --- resolve_task_input tests (T1a-T1e) ---
```
Trecek: [warning] slop: Section comment '# --- resolve_task_input tests (T1a-T1e) ---' uses opaque ticket-style labels (T1a-T1e) with no issue reference. Remove the label anchors; the test names are self-descriptive.

tests/recipe/test_planner_recipe.py
```diff
@@ -272,43 +273,66 @@ def test_planner_recipe_assess_review_approach_routes_to_validate_on_failure(pla
     assert step.on_failure == "validate"
 
 
-# --- task_file ingredient tests ---
+# --- resolve_task ingredient and step tests ---
```
Trecek: [warning] slop: Section comment '# --- resolve_task ingredient and step tests ---' adds no information beyond the test names themselves. Remove it.

Trecek (commented):

Trecek (commented):
AutoSkillit review: **changes_requested** (self-review mode — blocked changes detected, posted as comments).

**1 critical finding** requires a fix before merge; **10 warning findings** also inline.

Key blocking issue:
- `src/autoskillit/planner/manifests.py` L434 — `task_path.read_text()` called without `encoding='utf-8'` and no exception handling. A permissions error or locale mismatch will propagate as a bare OSError with no identifying context.

src/autoskillit/planner/compiler.py
Trecek: Investigated — this is intentional. compiler.py already imports three private symbols (_load_assignment_results, _load_phase_results, _load_wp_results) from validation.py — established intra-package convention documented in the compiler.py module docstring. _derive_label is correctly kept private; no public API wrapping is warranted.

src/autoskillit/planner/manifests.py
Trecek: Investigated — this is intentional. manifests.py has always been an I/O boundary layer — every public function (expand_assignments, build_phase_assignment_manifest, finalize_wp_manifest, etc.) performs file I/O via atomic_write. resolve_task_input is intentionally the materialization entry point for the resolve_task pipeline step, not an accidental side effect in a pure transform layer.

src/autoskillit/planner/manifests.py
Trecek: Agreed — fixed in c2457c69. Fixed in c2457c69: resolve_task_input now reads with encoding='utf-8' and wraps OSError with file path context (e.g., 'Cannot read task file {path}: {exc}').

src/autoskillit/planner/compiler.py
Trecek: Agreed — fixed in c2457c69. Fixed in c2457c69: compile_plan now reads task file with read_text(encoding='utf-8').

src/autoskillit/planner/merge.py
Trecek: Agreed — fixed in c2457c69. Fixed in c2457c69: merge_files now reads task file with read_text(encoding='utf-8').

src/autoskillit/planner/merge.py
Trecek: Agreed — fixed in c2457c69. Fixed in c2457c69: build_plan_snapshot now reads task file with read_text(encoding='utf-8').

src/autoskillit/planner/manifests.py
Trecek: Agreed — fixed in c2457c69. Fixed in c2457c69: _derive_label now uses a single pass tracking both the first heading line and first non-empty line simultaneously, eliminating the redundant second scan.

tests/planner/conftest.py
Trecek: Valid observation — flagged for design decision. write_task_file always writes to 'task_input.md' in the given directory. In practice every caller uses pytest's tmp_path (unique per test), but within a single test sequential calls would silently overwrite. Whether this warrants a guard or doc-comment requires a human decision.

tests/planner/test_manifests.py
Trecek: Agreed — fixed in 47d33ce4. Fixed in 47d33ce4: test now also asserts text.startswith(result['task_label']), ensuring the label is a proper prefix of the input, not just any string <=80 chars.

tests/recipe/test_planner_recipe.py
Trecek: Investigated — this is intentional. with_args is typed as dict[str, str] per recipe/schema.py (RecipeStep.with_args). Python's str() on a plain dict is deterministic; no custom __repr__ applies. The substring check for 'inputs.task_file' absence is reliable for this data type.

tests/planner/test_manifests.py
Trecek: Agreed — fixed in 47d33ce4. Fixed in 47d33ce4: removed the opaque ticket-style section comment '# --- resolve_task_input tests (T1a-T1e) ---'. Test names are self-descriptive.

tests/recipe/test_planner_recipe.py
Trecek: Agreed — fixed in 47d33ce4. Fixed in 47d33ce4: removed the vacuous section comment '# --- resolve_task ingredient and step tests ---'.

tests/planner/test_manifests.py
```diff
@@ -513,3 +513,60 @@ def test_finalize_wp_manifest_nonexistent_dir_raises(tmp_path):
...
+def test_resolve_task_input_file_no_heading(tmp_path):
+    from autoskillit.planner import resolve_task_input
+
+    task_file = tmp_path / "task.txt"
+    task_file.write_text("Implement the feature flag system for gradual rollout")
+    planner_dir = tmp_path / "planner"
+    planner_dir.mkdir()
+    result = resolve_task_input(str(task_file), str(planner_dir))
+    assert result["task_file_path"] == str(task_file)
+    assert result["task_label"] == "Implement the feature flag system for gradual rollout"
```
Trecek, line 540: [warning] tests: test_resolve_task_input_file_no_heading asserts label equals full content but the file content is a single line, so it cannot distinguish 'label is derived first line' from 'label is full content'. Add a multi-line file fixture to verify only the first line is used as label.

tests/planner/test_manifests.py
```diff
@@ -513,3 +513,60 @@ def test_finalize_wp_manifest_nonexistent_dir_raises(tmp_path):
...
+
+
+def test_resolve_task_input_long_inline_truncates_label(tmp_path):
+    from autoskillit.planner import resolve_task_input
+
+    planner_dir = tmp_path / "planner"
+    planner_dir.mkdir()
+    text = "A" * 120
+    result = resolve_task_input(text, str(planner_dir))
+    assert len(result["task_label"]) <= 80
```
Trecek, line 572: [warning] tests: test_resolve_task_input_long_inline_truncates_label asserts text.startswith(result['task_label']) but 'A'*120 trivially starts with any of its own prefixes — this adds no value over the len check. Assert result['task_label'] == text[:80] to pin the exact truncation boundary and catch off-by-one errors.

tests/planner/test_manifests.py
```diff
@@ -513,3 +513,60 @@ def test_finalize_wp_manifest_nonexistent_dir_raises(tmp_path):
...
+def test_resolve_task_input_file_with_heading(tmp_path):
+    from autoskillit.planner import resolve_task_input
+
+    task_file = tmp_path / "task.md"
+    task_file.write_text("# Deploy Auth Service\n\nDetailed description...")
+    planner_dir = tmp_path / "planner"
+    planner_dir.mkdir()
+    result = resolve_task_input(str(task_file), str(planner_dir))
+    assert result["task_file_path"] == str(task_file)
+    assert result["task_label"] == "Deploy Auth Service"
```
Trecek, line 528: [warning] tests: test_resolve_task_input_file_with_heading creates planner_dir but does not assert whether a task_input.md file was or was not created there (when input is a real file, resolve_task_input should NOT write to planner_dir). Adding an assertion that planner_dir remains empty after the call would verify this path correctly.

tests/recipe/test_planner_recipe.py
```diff
@@ -272,43 +273,63 @@ def test_planner_recipe_assess_review_approach_routes_to_validate_on_failure(pla
...
-def test_planner_recipe_has_task_file_ingredient(planner_recipe):
-    ingredients = planner_recipe.ingredients
-    assert "task_file" in ingredients, "planner recipe must have task_file ingredient"
-    assert ingredients["task_file"].required is False
+def test_planner_recipe_has_resolve_task_step(planner_recipe):
+    assert "resolve_task" in planner_recipe.steps
+    step = planner_recipe.steps["resolve_task"]
+    assert step.tool == "run_python"
+    assert step.with_args.get("callable") == "autoskillit.planner.resolve_task_input"
+    assert "task_file_path" in (step.capture or {})
```
Trecek, line 285: [warning] tests: test_planner_recipe_has_resolve_task_step asserts 'task_file_path' in (step.capture or {}). If step.capture is None the fallback {} means the assertion silently passes without reporting the missing capture block. Assert step.capture is not None first to surface this as a clear failure.

tests/planner/test_merge.py
```diff
@@ -471,21 +479,21 @@ def test_replace_item_corrupt_replacement_json_raises(tmp_path) -> None:
...
+def test_build_plan_snapshot_reads_task_from_task_file_path(tmp_path) -> None:
     phases_dir = tmp_path / "phases"
     phases_dir.mkdir()
     (phases_dir / "P1_result.json").write_text(json.dumps(make_phase_result(1)))
     out = tmp_path / "snapshot.json"
     task_file = tmp_path / "task_desc.txt"
     task_file.write_text("Full task from file")
 
-    build_plan_snapshot(str(phases_dir), str(out), task="label", task_file=str(task_file))
+    build_plan_snapshot(str(phases_dir), str(out), task_file_path=str(task_file))
```
Trecek, line 490: [warning] tests: test_build_plan_snapshot_reads_task_from_task_file_path: the 'out' variable is not visible in the diff hunk starting at L482 — verify 'out' is defined inside this test function (not shared from a preceding scope in the original file). If it is defined before the diff hunk, this is a false positive.

Trecek (commented):

Trecek (commented):

tests/planner/test_manifests.py
Trecek: Agreed — fixed in 42c1d8f9e1b7c6ff18c197a2927325bc6fe88629. Added `assert list(planner_dir.iterdir()) == []` after the `resolve_task_input` call to verify the function does not write `task_input.md` to `planner_dir` when given an existing file path. The implementation correctly returns early at L432–438 of manifests.py (before `atomic_write`), so this assertion now pins that invariant.

tests/recipe/test_planner_recipe.py
Trecek: Investigated — this is intentional. The reviewer's claim that the assertion "silently passes" when `step.capture is None` is factually incorrect. When `capture` is `None`, the expression `step.capture or {}` evaluates to `{}`, and `"task_file_path" in {}` evaluates to `False` — causing `assert False` which raises `AssertionError` with a clear failure. There is no silent pass. Python's `in` operator on an empty dict returns `False` for all keys. The current guard is sufficient.

tests/planner/test_manifests.py
Trecek: Acknowledged — minor suggestion noted. The single-line fixture cannot distinguish 'first-line-only' from 'full-content' derivation. Flagged for a follow-up test that uses a multi-line file to verify `_derive_label` returns only the first non-empty line.

tests/planner/test_manifests.py
Trecek: Acknowledged — minor suggestion noted. `'A'*120.startswith('A'*N)` is trivially `True` for any `N <= 120`, so the assertion adds no discriminating value. Replacing with `result['task_label'] == text[:80]` would pin the exact 80-char truncation boundary and catch off-by-one errors. Flagged for follow-up.

tests/planner/test_merge.py
Trecek: Acknowledged — minor suggestion noted. Confirmed not a false positive: `out = tmp_path / "snapshot.json"` is defined at line 486 inside the test function body. The diff hunk in the review started at L482 which is before the `out` assignment, creating the appearance that `out` was undefined in the hunk. The variable is fully in scope.

Issue #1636, opened 2026-05-03
Planner: replace fixed phase-count range with intent-driven decomposition


Trecek:
## Problem

The `planner-generate-phases` skill hardcodes a "3–6 phases" constraint repeated four times in the SKILL.md (lines 4, 17, 34, 66). Investigation of all completed planner runs confirms every one produced exactly 5 phases regardless of task complexity — a textbook LLM anchoring effect where the model gravitates to the safe center-high of a narrow stated range.

Three factors compound the bias:
1. **Narrow range + hard NEVER bounds** — with only [3,4,5,6] as options and absolute prohibitions at both endpoints, the LLM avoids edges and anchors at 5
2. **No task-size calibration** — a single-file rename and a 50-file campaign build receive identical "3–6" guidance
3. **Architectural example priming** — the JSON example shows "Database Layer" with `["models", "migrations", "repositories"]`, priming layer-based decomposition that naturally maps to ~5 layers (pink elephant problem)

**Evidence:** 3/3 completed runs (run-20260429, run-20260502, run-20260503) produced P1–P5. The 4th run aborted before phase generation.

**Investigation report:** `.autoskillit/temp/investigate/investigation_planner-five-phase-bias_2026-05-02_234500.md`

## Desired Behavior

Phase count should be determined entirely by the nature of the work, not by a stated numeric range. A phase is an **independently verifiable checkpoint** — a set of features whose functionality can be fully tested once implemented. The line is drawn where you can stop and validate that everything up to that point works.

For example, in a multi-recipe campaign, each recipe is naturally its own phase because it can be independently tested (including via API simulator). A simple single-component task might legitimately be 1 phase.

### 1. Rewrite phase definition in `planner-generate-phases/SKILL.md`

**Remove:**
- All four occurrences of "3–6" (lines 4, 17, 34, 66)
- The NEVER constraint "Produce fewer than 3 or more than 6 phases" (line 34)
- The "Database Layer" JSON example (lines 86–94) and the auth decomposition example (line 70)

**Replace with intent-driven decomposition guidance:**
- Define a phase as: an independently verifiable checkpoint where a coherent set of features can be fully tested
- The number of phases is determined by the task — wherever you can draw a line and say "everything up to here can be validated independently"
- A single-component task may warrant 1 phase; a multi-component campaign may warrant one phase per independently testable component
[...]

PR #1639, opened 2026-05-03
Planner — Replace Fixed Phase-Count Range with Intent-Driven Decomposition

Trecek:
## Summary

Replace the hardcoded "3–6 phases" constraint in `planner-generate-phases/SKILL.md` with intent-driven decomposition guidance. The LLM-facing SKILL.md currently repeats a narrow numeric range (3–6) four times plus uses domain-specific examples ("Database Layer", "Auth API Endpoints") that cause anchoring bias — every completed run produced exactly 5 phases regardless of task complexity. This plan removes all fixed-range and domain-anchoring content, replaces it with checkpoint-based decomposition guidance, neutralizes domain-specific examples in `planner-elaborate-assignments/SKILL.md`, and adds a batch ceiling to `planner-refine-phases/SKILL.md` (which currently spawns all phases in parallel with no size guard, relying on the now-removed 3–6 assumption).

No Python backend changes are required — `validation.py`, `schema.py`, and `manifests.py` impose no phase-count bounds. The contract card SHA hash for `planner-generate-phases` must be regenerated after the SKILL.md edit.

Closes #1636

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260502-181823-129753/.autoskillit/temp/make-plan/planner-replace-fixed-phase-count-range-with-intent-driven-d_plan_2026-05-02_190000.md`

## Token Usage Summary

| plan | 72 | 11.1k | 1.3M | 85.5k | 1 | 8m 31s |
| verify | 55 | 13.4k | 1.2M | 65.5k | 1 | 5m 31s |
| implement | 292 | 13.2k | 2.4M | 74.4k | 1 | 8m 21s |
| prepare_pr | 60 | 4.5k | 220.8k | 29.8k | 1 | 1m 23s |
| compose_pr | 51 | 2.4k | 163.8k | 20.0k | 1 | 43s |
| review_pr | 116 | 24.9k | 616.3k | 57.7k | 1 | 6m 20s |
| resolve_review | 311 | 19.4k | 2.4M | 79.2k | 1 | 9m 13s |
| **Total** | 957 | 88.9k | 8.3M | 412.1k | | 40m 5s |

### DIFF
--- a/tests/skills/test_planner_skill_contracts.py
+++ b/tests/skills/test_planner_skill_contracts.py
@@ -390,3 +390,84 @@ def test_generate_phases_skill_no_env_var_delivery():
     content = (SKILLS_ROOT / "planner-generate-phases" / "SKILL.md").read_text()
     assert "PLANNER_TASK_FILE" not in content, "Must not reference PLANNER_TASK_FILE env var"
     assert "PLANNER_TASK" not in content, "Must not reference PLANNER_TASK env var"
+
+
+# --- intent-driven decomposition tests (T1–T7) ---
+
+
+def test_generate_phases_no_fixed_range_in_description():
+    """Description frontmatter must not contain a fixed numeric phase range."""
+    content = (SKILLS_ROOT / "planner-generate-phases" / "SKILL.md").read_text()
+    parts = content.split("---", 2)
+    data = yaml.safe_load(parts[1]) or {}
+    description = data.get("description", "")
+    assert "3-6" not in description, "description must not hardcode '3-6' phase range"
+    assert "3–6" not in description, "description must not hardcode '3–6' phase range"
+
+
+def test_generate_phases_no_fixed_range_in_body():
+    """SKILL.md body must not contain a fixed '3-6' or '3–6' phase range."""
+    content = (SKILLS_ROOT / "planner-generate-phases" / "SKILL.md").read_text()
+    assert "3-6" not in content, "SKILL.md must not contain '3-6'"
+    assert "3–6" not in content, "SKILL.md must not contain '3–6'"
+
+
+def test_generate_phases_no_phase_count_never_constraint():
+    """The NEVER block must not impose a min/max phase count."""
+    content = (SKILLS_ROOT / "planner-generate-phases" / "SKILL.md").read_text()
+    assert "fewer than 3" not in content.lower()
+    assert "more than 6" not in content.lower()
+
+
+def test_generate_phases_has_intent_driven_guidance():
+    """SKILL.md must contain intent-driven decomposition guidance."""
+    content = (SKILLS_ROOT / "planner-generate-phases" / "SKILL.md").read_text().lower()
+    assert "independently verifiable" in content or "verifiable checkpoint" in content, (
+        "Must define phases as independently verifiable checkpoints"
+    )
+
+
+ANCHORING_TERMS = [
+    "Database Layer",
+    "Auth Data Model",
+    "Auth API Endpoints",
+    "Auth UI Integration",
+    "user authentication",
+    "Schema design",
+    "Migration framework",
+    "Repository pattern",
+]
+
+
+@pytest.mark.parametrize("term", ANCHORING_TERMS)
+def test_generate_phases_no_domain_anchoring(term: str):
+    """SKILL.md must not contain domain-specific example names that cause LLM anchoring."""
+    content = (SKILLS_ROOT / "planner-generate-phases" / "SKILL.md").read_text()
+    assert term not in content, (
+        f"Domain-specific example '{term}' causes LLM anchoring bias — "
+        f"use generic placeholders instead"
+    )
+
+
+ELABORATE_ANCHORING_TERMS = [
+    "Session Management",
+    "user session persistence",
+    "SQLite-backed session table",
+]
+
+
+@pytest.mark.parametrize("term", ELABORATE_ANCHORING_TERMS)
+def test_elaborate_assignments_no_domain_anchoring(term: str):
+    """elaborate-assignments SKILL.md must not use domain-specific examples."""
+    content = (SKILLS_ROOT / "planner-elaborate-assignments" / "SKILL.md").read_text()
+    assert term not in content, f"Domain-specific example '{term}' causes LLM anchoring bias"
+
+
+def test_refine_phases_has_batch_ceiling():
+    """refine-phases must handle >6 phases via sequential batching (no unbounded spawn)."""
+    content = (SKILLS_ROOT / "planner-refine-phases" / "SKILL.md").read_text()
+    has_batch_guidance = "batch" in content.lower() and "6" in content
+    assert has_batch_guidance, (
+        "planner-refine-phases must include batch ceiling guidance for >6 phases "
+        "(consistent with refine-wps and refine-assignments)"
+    )
--- a/tests/skills/test_planner_skill_contracts.py
+++ b/tests/skills/test_planner_skill_contracts.py
@@ -395,16 +395,6 @@ def test_generate_phases_skill_no_env_var_delivery():
 # --- intent-driven decomposition tests (T1–T7) ---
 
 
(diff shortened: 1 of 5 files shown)


tests/skills/test_planner_skill_contracts.py
```diff
@@ -390,3 +390,84 @@ def test_generate_phases_skill_no_env_var_delivery():
...
     assert "PLANNER_TASK" not in content, "Must not reference PLANNER_TASK env var"
+
+
+# --- intent-driven decomposition tests (T1–T7) ---
+
+
+def test_generate_phases_no_fixed_range_in_description():
+    """Description frontmatter must not contain a fixed numeric phase range."""
+    content = (SKILLS_ROOT / "planner-generate-phases" / "SKILL.md").read_text()
+    parts = content.split("---", 2)
```
Trecek: [warning] bugs: Frontmatter split using content.split("---", 2) assumes the file starts with "---". If SKILL.md does not begin with a YAML frontmatter delimiter, parts[1] will not be the frontmatter block and yaml.safe_load(parts[1]) will silently parse the wrong content, causing description to be empty and the assertions to pass vacuously.

tests/skills/test_planner_skill_contracts.py
```diff
@@ -390,3 +390,84 @@ def test_generate_phases_skill_no_env_var_delivery():
...
+    data = yaml.safe_load(parts[1]) or {}
```
Trecek: [warning] bugs: content.split("---", 2) produces at most 3 parts. If fewer than two "---" separators exist, parts[1] raises IndexError, crashing the test rather than failing it cleanly. Add a guard: assert len(parts) >= 3 before accessing parts[1].

tests/skills/test_planner_skill_contracts.py
```diff
@@ -390,3 +390,84 @@ def test_generate_phases_skill_no_env_var_delivery():
...
+    description = data.get("description", "")
+    assert "3-6" not in description, "description must not hardcode '3-6' phase range"
+    assert "3–6" not in description, "description must not hardcode '3–6' phase range"
+
+
+def test_generate_phases_no_fixed_range_in_body():
```
Trecek, line 398: [warning] tests: test_generate_phases_no_fixed_range_in_body is a strict superset of test_generate_phases_no_fixed_range_in_description: both read the same file and assert the same strings are absent. The description is part of the body, so if the body test passes the description test cannot fail. The description-scoped test provides no additional safety and should be removed.

tests/skills/test_planner_skill_contracts.py
```diff
@@ -390,3 +390,84 @@ def test_generate_phases_skill_no_env_var_delivery():
...
+
+def test_generate_phases_has_intent_driven_guidance():
+    """SKILL.md must contain intent-driven decomposition guidance."""
+    content = (SKILLS_ROOT / "planner-generate-phases" / "SKILL.md").read_text().lower()
+    assert "independently verifiable" in content or "verifiable checkpoint" in content, (
+        "Must define phases as independently verifiable checkpoints"
+    )
+
+
+ANCHORING_TERMS = [
```
Trecek, line 420: [info] tests: ANCHORING_TERMS is declared between two parametrize test functions. Consider co-locating each constant list immediately above its @pytest.mark.parametrize decorator to make the constant-to-test relationship obvious at a glance.

tests/skills/test_planner_skill_contracts.py
```diff
@@ -390,3 +390,84 @@ def test_generate_phases_skill_no_env_var_delivery():
...
+    "Auth UI Integration",
+    "user authentication",
+    "Schema design",
+    "Migration framework",
+    "Repository pattern",
+]
+
+
+@pytest.mark.parametrize("term", ANCHORING_TERMS)
+def test_generate_phases_no_domain_anchoring(term: str):
```
Trecek, line 433: [warning] tests: test_generate_phases_no_domain_anchoring performs case-sensitive matching (term not in content) while test_generate_phases_has_intent_driven_guidance uses content.lower(). If ANCHORING_TERMS appear in SKILL.md only in a different case, this test will miss them. Verify whether case-sensitive matching is intentional for anchoring term detection.

tests/skills/test_planner_skill_contracts.py
```diff
@@ -390,3 +390,84 @@ def test_generate_phases_skill_no_env_var_delivery():
...
+def test_elaborate_assignments_no_domain_anchoring(term: str):
+    """elaborate-assignments SKILL.md must not use domain-specific examples."""
+    content = (SKILLS_ROOT / "planner-elaborate-assignments" / "SKILL.md").read_text()
+    assert term not in content, f"Domain-specific example '{term}' causes LLM anchoring bias"
+
+
+def test_refine_phases_has_batch_ceiling():
+    """refine-phases must handle >6 phases via sequential batching (no unbounded spawn)."""
+    content = (SKILLS_ROOT / "planner-refine-phases" / "SKILL.md").read_text()
+    has_batch_guidance = "batch" in content.lower() and "6" in content
```
Trecek: [warning] tests: "6" in content is too broad — any occurrence of the digit 6 (version numbers, line references, unrelated counts) satisfies it, making the assertion nearly always true. Use a more specific pattern such as "batch of 6" in content.lower() or "batches of 6" in content.lower().

Trecek (commented):

Trecek (commented):

tests/skills/test_planner_skill_contracts.py
Trecek: Investigated — this is intentional. planner-generate-phases/SKILL.md is confirmed to start with `---` (file line 1), so `content.split('---', 2)` produces `['', '<frontmatter>', '<body>']` and `parts[1]` is correctly the frontmatter block. The hypothetical mis-parse scenario does not apply to the actual file structure.

tests/skills/test_planner_skill_contracts.py
Trecek: Agreed — addressed in 0bf38a6d. The entire `test_generate_phases_no_fixed_range_in_description` function (which contained the fragile `parts[1]` access without a length guard) was removed as part of fix for comment 3177481103 — it was redundant since `test_generate_phases_no_fixed_range_in_body` is a strict superset. Removing the function eliminates the crash path entirely.

tests/skills/test_planner_skill_contracts.py
Trecek: Agreed — fixed in 0bf38a6d. `test_generate_phases_no_fixed_range_in_description` has been removed. `test_generate_phases_no_fixed_range_in_body` asserts over the full file content which is a strict superset of the description field, so the description-scoped test provided no independent safety.

tests/skills/test_planner_skill_contracts.py
Trecek: Acknowledged — minor suggestion noted.

tests/skills/test_planner_skill_contracts.py
Trecek: Valid observation — flagged for design decision. The ANCHORING_TERMS are the exact strings that appeared in the old SKILL.md (now removed), making case-sensitive matching intentional: 'Database Layer', 'Auth Data Model', etc. were titled examples in the original file. If the intent is to guard against topic-level anchoring regardless of casing, case-insensitive matching would be safer. Human decision required on intended scope.

tests/skills/test_planner_skill_contracts.py
Trecek: Agreed — fixed in 0bf38a6d. Changed to `"batch of 6" in content.lower() or "batches of 6" in content.lower()`. The SKILL.md contains `"spawn sequential batches of 6 L0 subagents"` so `"batches of 6"` matches precisely. The old `"6" in content` was satisfied by `### Step 6:` alone, independent of any batch guidance.

PR #1652, opened 2026-05-03
Rectify: Canonical Install Detection and Registry Key Invariant Enforcement

### LINKED ISSUE #1650: is_dev_install() does not recognize GIT_VCS dev-track installs + env var case sensitivity crash
## Bug Summary

Two related bugs in the feature gate system prevent `autoskillit fleet campaign` (and all other feature-gated commands) from working on `uv tool install` dev-track installs, and the suggested env var workaround crashes with a `ConfigSchemaError`.

### Reproduction

```bash
task install-dev   # runs: uv tool install --force "autoskillit[dev] @ git+https://github.com/TalonT-Org/AutoSkillit.git@develop"
autoskillit fleet campaign
# → "The 'fleet' feature is not enabled."
```

### Root Cause

`is_dev_install()` in `src/autoskillit/core/_install_detect.py:11-27` only checks for `dir_info.editable == True` in `direct_url.json`. A `uv tool install` from a git URL produces metadata with `vcs_info` (containing `commit_id` and `requested_revision`), **not** `dir_info`. The function returns `False` for GIT_VCS installs.

The actual `direct_url.json` for a `task install-dev` install:
```json
{
  "url": "https://github.com/TalonT-Org/AutoSkillit.git",
  "vcs_info": {
    "vcs": "git",
    "commit_id": "23df7eb35ca44257432836c7881f726d4ff6e762",
    "requested_revision": "develop"
  }
}
```

This is clearly a dev-track install, but `is_dev_install()` ignores `vcs_info` entirely.

### Requirement Violation

PR #1616 (commit `e69ea23d`) stated in REQ-GATE-002: "When no explicit `experimental_enabled` value is set in any config layer or env var, **editable and source-tree installs** must default to `experimental_enabled=true`." The implementation only covers editable installs — GIT_VCS dev-track installs are not covered.

### Existing Pattern

`classify_track()` in `src/autoskillit/cli/_install_info.py:113-119` already implements the correct dev-track detection using negative matching. It was introduced in PR #1635 specifically for this purpose but is only used for update checks, not feature gating.

### Resolution

Broaden `is_dev_install()` to also check `vcs_info` in `direct_url.json` and return `True` when `requested_revision` classifies as dev-track (not `None`/`"main"`/`"stable"`/release-tag). The stable-track check can be inlined to preserve IL-0 (no autoskillit imports). Update the docstring from "editable mode" to "development install (editable or dev-track VCS)".

### Affected Files

- `src/autoskillit/core/_install_detect.py` — `is_dev_install()` needs broadening
- `src/autoskillit/config/settings.py:162` — consumes `is_dev_install()`, no change needed
- `tests/core/test_install_detect.py` — needs test case for GIT_VCS dev-track metadata

---

### Reproduction

```bash
AUTOSKILLIT_FEATURES__FLEET=true autoskillit fleet campaign
# → ConfigSchemaError: Unknown feature 'FLEET' in features config. Known features: ['fleet', 'planner']
```

### Root Cause

Dynaconf uppercases env var keys during parsing. `AUTOSKILLIT_FEATURES__FLEET=true` arrives in `_build_features_dict()` as `{"FLEET": True}`. Line 169 of `settings.py` checks `name not in FEATURE_REGISTRY` — the registry uses lowercase `"fleet"`, so `"FLEET"` is rejected as unknown.

The error message in `_fleet.py:40` suggests `AUTOSKILLIT_FEATURES__FLEET=true` as a workaround, but this path is broken.

Note: `AUTOSKILLIT_FEATURES__EXPERIMENTAL_ENABLED=true` does work because lines 159-161 explicitly pop both `experimental_enabled` and `EXPERIMENTAL_ENABLED`. But per-feature env var overrides are broken for all features.

### Resolution

Lowercase the feature name in `_build_features_dict()` before the registry lookup (line 164 area). Something like `name = name.lower()` before the `if name not in FEATURE_REGISTRY` check on line 169.

### Affected Files

- `src/autoskillit/config/settings.py:164-169` — needs case-insensitive lookup
- `tests/arch/test_feature_registry.py` — needs test for uppercase env var key

---

## Impact

- **All users** who install via `task install-dev` (the documented dev install path) cannot use fleet or planner features without a manual config workaround.
- **All users** who follow the suggested env var workaround (`AUTOSKILLIT_FEATURES__FLEET=true`) hit a crash.
- Workaround: add `features: {experimental_enabled: true}` to `~/.autoskillit/config.yaml`.

## Investigation Report

Full investigation report at: `.autoskillit/temp/investigate/investigation_fleet_feature_not_enabled_on_develop_2026-05-02_160000.md`

Trecek:
## Summary

Two bugs expose the same architectural weakness: **duplicated-but-divergent logic without structural invariants**. Three independent `direct_url.json` parsers exist across two import layers, with no shared parsing primitive. The `FEATURE_REGISTRY` uses lowercase keys but has no import-time invariant enforcing this, and `_build_features_dict()` accepts raw dynaconf-uppercased keys without normalization. The fix consolidates `direct_url.json` parsing into a single IL-0 canonical parser, adds import-time key-casing invariants to the feature registry, and introduces parametrized test matrices that make both bug classes immediately visible.

Closes #1650

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260502-212016-496731/.autoskillit/temp/rectify/rectify_install_detect_and_feature_registry_case_immunity_2026-05-02_212016.md`

## Token Usage Summary

| investigate | 1.5k | 7.6k | 573.7k | 88.9k | 1 | 4m 34s |
| rectify | 59 | 11.7k | 611.8k | 196.2k | 1 | 6m 52s |
| dry_walkthrough | 30 | 6.3k | 495.7k | 104.4k | 1 | 4m 27s |
| implement | 676 | 32.4k | 6.1M | 383.9k | 1 | 18m 0s |
| prepare_pr | 60 | 5.9k | 188.5k | 62.3k | 1 | 1m 31s |
| compose_pr | 51 | 2.4k | 149.1k | 34.1k | 1 | 52s |
| review_pr | 264 | 69.9k | 1.5M | 313.9k | 2 | 15m 43s |
| resolve_review | 666 | 43.5k | 5.2M | 412.9k | 2 | 23m 12s |
| **Total** | 3.3k | 179.7k | 14.8M | 1.6M | | 1h 15m |
## Token Efficiency

| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| implement | 447 | 6493.2 | 463.2 | 50.6 |
| **Total** | **447** | 10488.0 | 933.3 | 98.4 |

### DIFF
--- a/src/autoskillit/cli/_install_info.py
+++ b/src/autoskillit/cli/_install_info.py
@@ -12,8 +12,7 @@ from datetime import timedelta
 from enum import StrEnum
 from pathlib import Path
 
-from autoskillit.core import get_logger, parse_direct_url
-from autoskillit.core._install_detect import _is_stable_track
+from autoskillit.core import _is_stable_track, get_logger, parse_direct_url
 
 logger = get_logger(__name__)
 
--- a/src/autoskillit/cli/_install_info.py
+++ b/src/autoskillit/cli/_install_info.py
@@ -12,7 +12,8 @@ from datetime import timedelta
 from enum import StrEnum
 from pathlib import Path
 
-from autoskillit.core import _is_stable_track, get_logger, parse_direct_url
+from autoskillit.core import get_logger, parse_direct_url
+from autoskillit.core._install_detect import _is_stable_track
 
 logger = get_logger(__name__)
 
--- a/src/autoskillit/cli/_install_info.py
+++ b/src/autoskillit/cli/_install_info.py
@@ -7,14 +7,12 @@ that drive the unified update check.
 
 from __future__ import annotations
 
-import json
-import re
 from dataclasses import dataclass
 from datetime import timedelta
 from enum import StrEnum
 from pathlib import Path
 
-from autoskillit.core import get_logger
+from autoskillit.core import _is_stable_track, get_logger, parse_direct_url
 
 logger = get_logger(__name__)
 
@@ -53,28 +51,17 @@ def detect_install() -> InstallInfo:
     """
     _unknown = InstallInfo(InstallType.UNKNOWN, None, None, None, None)
     try:
-        import importlib.metadata
-
-        dist = importlib.metadata.Distribution.from_name("autoskillit")
-        raw = dist.read_text("direct_url.json")
-        if not raw:
-            return _unknown
-
-        data = json.loads(raw)
-
-        vcs_info = data.get("vcs_info", {})
-        if isinstance(vcs_info, dict) and vcs_info.get("vcs") == "git":
+        info = parse_direct_url()
+        url = info["url"] or ""
+        if info["install_type"] == "git-vcs":
             return InstallInfo(
                 install_type=InstallType.GIT_VCS,
-                commit_id=vcs_info.get("commit_id") or None,
-                requested_revision=vcs_info.get("requested_revision") or None,
-                url=data.get("url") or None,
+                commit_id=info["commit_id"],
+                requested_revision=info["requested_revision"],
+                url=url or None,
                 editable_source=None,
             )
-
-        dir_info = data.get("dir_info", {})
-        url = data.get("url", "")
-        if isinstance(dir_info, dict) and dir_info.get("editable") is True:
+        if info["install_type"] == "local-editable":
             if isinstance(url, str) and url.startswith("file://"):
                 src_path = url[len("file://") :]
                 return InstallInfo(
@@ -84,30 +71,18 @@ def detect_install() -> InstallInfo:
                     url=url,
                     editable_source=Path(src_path),
                 )
-
-        if isinstance(url, str) and url.startswith("file://"):
+        if info["install_type"] == "local-path":
             return InstallInfo(
                 install_type=InstallType.LOCAL_PATH,
                 commit_id=None,
                 requested_revision=None,
-                url=url,
+                url=url or None,
                 editable_source=None,
             )
-
         return _unknown
-
     except Exception:
         logger.debug("install classification failed", exc_info=True)
         return _unknown
-
-
-def _is_release_tag(rev: str) -> bool:
-    """Return True if ``rev`` looks like a version tag (e.g. 'v0.7.75', '0.7.75')."""
-    return bool(re.fullmatch(r"v?\d+(\.\d+)*", rev))
-
-
-def _is_stable_track(rev: str | None) -> bool:
-    return not rev or rev in ("main", "stable") or _is_release_tag(rev)
 
 
 def classify_track(info: InstallInfo) -> InstallTrack:
--- a/src/autoskillit/core/__init__.py
+++ b/src/autoskillit/core/__init__.py
@@ -13,8 +13,6 @@ del _dir  # replaced below so dir() reflects the filtered __all__
 _PRIVATE_REEXPORTS = frozenset(
     {
         "_InstallLock",
-        "_is_release_tag",
-        "_is_stable_track",
         "_retire_old_versions",
         "_collect_disabled_feature_tags",
         "_AUTOSKILLIT_GITIGNORE_ENTRIES",
--- a/src/autoskillit/core/__init__.py
+++ b/src/autoskillit/core/__init__.py
@@ -13,6 +13,8 @@ del _dir  # replaced below so dir() reflects the filtered __all__
 _PRIVATE_REEXPORTS = frozenset(
     {
         "_InstallLock",
+        "_is_release_tag",
+        "_is_stable_track",
         "_retire_old_versions",
         "_collect_disabled_feature_tags",
         "_AUTOSKILLIT_GITIGNORE_ENTRIES",
--- a/src/autoskillit/core/_install_detect.py
+++ b/src/autoskillit/core/_install_detect.py
@@ -47,7 +47,8 @@ def parse_direct_url() -> DirectUrlInfo:
                 "url": "",
             }
         data = json.loads(raw)
-        url = data.get("url", "") or ""
+        raw_url = data.get("url")
+        url = raw_url if isinstance(raw_url, str) else ""
         vcs_info = data.get("vcs_info", {})
         if isinstance(vcs_info, dict) and vcs_info.get("vcs") == "git":
             return {
--- a/src/autoskillit/core/_install_detect.py
+++ b/src/autoskillit/core/_install_detect.py
@@ -5,9 +5,17 @@ from __future__ import annotations
 import json
 import logging
 import re
-from typing import Any
+from typing import TypedDict
 
 logger = logging.getLogger(__name__)  # noqa: TID251 — IL-0 module, no autoskillit imports allowed
+
+
+class DirectUrlInfo(TypedDict):
+    install_type: str
+    requested_revision: str | None
+    commit_id: str | None
+    editable: bool
+    url: str
 
 
 def _is_release_tag(rev: str) -> bool:
@@ -19,7 +27,7 @@ def _is_stable_track(rev: str | None) -> bool:
     return not rev or rev in ("main", "stable") or _is_release_tag(rev)
 
 
-def parse_direct_url() -> dict[str, Any]:
+def parse_direct_url() -> DirectUrlInfo:
     """Parse direct_url.json and return a canonical install descriptor.
 
     Keys: install_type (str), requested_revision (str|None),
--- a/src/autoskillit/core/_install_detect.py
+++ b/src/autoskillit/core/_install_detect.py
@@ -4,22 +4,93 @@ from __future__ import annotations
 
 import json
 import logging
+import re
+from typing import Any
 
 logger = logging.getLogger(__name__)  # noqa: TID251 — IL-0 module, no autoskillit imports allowed
 
 
-def is_dev_install() -> bool:
-    """Return True if installed in editable mode; False on any error."""
+def _is_release_tag(rev: str) -> bool:
+    """Return True if rev looks like a version tag (e.g. 'v0.7.75', '0.7.75')."""
+    return bool(re.fullmatch(r"v?\d+(\.\d+)*", rev))
+
+
+def _is_stable_track(rev: str | None) -> bool:
+    return not rev or rev in ("main", "stable") or _is_release_tag(rev)
+
+
+def parse_direct_url() -> dict[str, Any]:
+    """Parse direct_url.json and return a canonical install descriptor.
+
+    Keys: install_type (str), requested_revision (str|None),
+          commit_id (str|None), editable (bool), url (str).
+    """
     try:
         import importlib.metadata
 
         dist = importlib.metadata.Distribution.from_name("autoskillit")
         raw = dist.read_text("direct_url.json")
         if not raw:
-            return False
+            return {
+                "install_type": "unknown",
+                "requested_revision": None,
+                "commit_id": None,
+                "editable": False,
+                "url": "",
+            }
         data = json.loads(raw)
+        url = data.get("url", "") or ""
+        vcs_info = data.get("vcs_info", {})
+        if isinstance(vcs_info, dict) and vcs_info.get("vcs") == "git":
+            return {
+                "install_type": "git-vcs",
+                "requested_revision": vcs_info.get("requested_revision") or None,
+                "commit_id": vcs_info.get("commit_id") or None,
+                "editable": False,
+                "url": url,
+            }
         dir_info = data.get("dir_info", {})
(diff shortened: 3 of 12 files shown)


tests/core/test_install_detect.py
```diff
@@ -43,6 +43,37 @@ def test_is_dev_install_git_vcs_returns_false(monkeypatch: pytest.MonkeyPatch) -
     assert is_dev_install() is False
 
 
+@pytest.mark.parametrize(
```
Trecek, line 46: [critical] arch: @pytest.mark.parametrize(...) at L46 is syntactically broken — the decorator expression is incomplete (dangling `...` sentinel). The parametrize arguments begin at L47 but are not inside the decorator call. This test will fail to collect and the parametrized matrix will never run.

src/autoskillit/core/_install_detect.py
```diff
@@ -4,22 +4,93 @@
...
+def _is_release_tag(rev: str) -> bool:
+    """Return True if rev looks like a version tag (e.g. 'v0.7.75', '0.7.75')."""
+    return bool(re.fullmatch(r"v?\d+(\.\d+)*", rev))
+
+
+def _is_stable_track(rev: str | None) -> bool:
+    return not rev or rev in ("main", "stable") or _is_release_tag(rev)
+
+
+def parse_direct_url() -> dict[str, Any]:
```
Trecek: [warning] arch: parse_direct_url() returns dict[str, Any] instead of a TypedDict. For an IL-0 module re-exported through core/__init__.pyi and consumed across layers, an untyped dict weakens the public API contract and defeats static analysis. A TypedDict (stdlib, zero IL cost) would express the fixed key set without violating IL-0.

src/autoskillit/core/_install_detect.py
```diff
@@ -4,22 +4,93 @@
...
+            return {
+                "install_type": "local-path",
+                "requested_revision": None,
+                "commit_id": None,
+                "editable": False,
+                "url": url,
+            }
+        return {
+            "install_type": "unknown",
+            "requested_revision": None,
```
Trecek, line 80: [warning] defense: Inconsistent indentation in the fallback `unknown` dict literal — `requested_revision` at L71 is over-indented by 4 extra spaces compared to surrounding keys. This copy-paste error will be flagged by ruff format and cause a pre-commit hook failure.

src/autoskillit/core/__init__.py
```diff
@@ -13,6 +13,8 @@
 _PRIVATE_REEXPORTS = frozenset(
     {
         "_InstallLock",
+        "_is_release_tag",
```
Trecek, line 16: [warning] cohesion: `_is_release_tag` and `_is_stable_track` are private (underscore-prefixed) helpers being re-exported through the public core namespace alongside public symbols `is_dev_install` and `parse_direct_url`. Exporting private-named symbols through `__init__.pyi` is asymmetric — if they are only needed by `_install_info.py`, import directly from `core._install_detect` instead of re-exporting.

src/autoskillit/core/_type_constants.py
```diff
@@ -291,6 +291,17 @@ class RecipePackDef(NamedTuple):
 
 CORE_PACKS: frozenset[str] = frozenset({"github", "ci", "clone", "telemetry"})
 
+if any(k != k.lower() for k in PACK_REGISTRY):
```
Trecek, line 294: [warning] cohesion: PACK_REGISTRY and RECIPE_PACK_REGISTRY casing guards appear at L294-L303, while FEATURE_REGISTRY and RETIRED_FEATURES guards appear at L425-L434. If HOOK_REGISTRY or other registries in this file lack equivalent guards, the invariant is inconsistently applied. Verify all registries of the same structural kind have equivalent guards.

src/autoskillit/cli/_install_info.py
```diff
@@ -53,28 +51,17 @@ def detect_install() -> InstallInfo:
...
-        raw = dist.read_text("direct_url.json")
-        if not raw:
-            return _unknown
-
-        data = json.loads(raw)
-
-        vcs_info = data.get("vcs_info", {})
-        if isinstance(vcs_info, dict) and vcs_info.get("vcs") == "git":
+        info = parse_direct_url()
+        url = info["url"] or ""
```
Trecek, line 55: [warning] defense: `info["url"]` is typed as `Any` (from `dict[str, Any]`). The expression `info["url"] or ""` silently accepts non-string truthy values from malformed JSON, but downstream `url.startswith("file://")` at L65 and `InstallInfo.url` expect `str | None`. A TypedDict for `parse_direct_url()` return value or a runtime `isinstance` guard would close this gap.

Trecek (commented):

Trecek (commented):
AutoSkillit review: blocking issues found (1 critical). See inline comments for details. Verdict: changes_requested (posted as COMMENT — self-authored PR).

tests/core/test_install_detect.py
Trecek: Investigated — this is intentional. The @pytest.mark.parametrize decorator at L46–L58 is syntactically valid: it spans a multiline tuple ('revision, expected') and list of parameter pairs, with the test function definition at L59. Pytest collects it without error. No dangling '...' sentinel is present in the actual file.

src/autoskillit/core/_install_detect.py
Trecek: Agreed — fixed in 763028ec. Added DirectUrlInfo TypedDict to _install_detect.py with fields: install_type (str), requested_revision (str|None), commit_id (str|None), editable (bool), url (str). parse_direct_url() return type updated to DirectUrlInfo. TypedDict is stdlib, zero IL cost.

src/autoskillit/core/_install_detect.py
Trecek: Investigated — this is intentional. The actual file at L69–L75 has consistent 12-space indentation for all keys in the fallback return dict. No over-indentation exists at L71. ruff format passes on the file without modification.

src/autoskillit/core/__init__.py
Trecek: Investigated — this is intentional. The _PRIVATE_REEXPORTS mechanism in core/__init__.py exists specifically for this pattern: private symbols can be imported via autoskillit.core but are excluded from __all__. Applying the suggested direct submodule import (from autoskillit.core._install_detect import _is_stable_track) violates test_no_cross_package_submodule_imports in tests/arch/test_layer_enforcement.py, which prohibits submodule imports in production code and requires using the package __init__ instead. The current re-export approach is the correct pattern for this codebase.

src/autoskillit/core/_type_constants.py
Trecek: Investigated — this is intentional. All registries in _type_constants.py with short identifier string keys already have import-time casing guards: PACK_REGISTRY (L294), RECIPE_PACK_REGISTRY (L299), FEATURE_REGISTRY (L426), RETIRED_FEATURES (L431). HOOK_REGISTRY is not in this file (it lives in hook_registry.py). TOOL_SUBSET_TAGS uses tool names (intrinsically snake_case lowercase). SKILL_ACTIVATE_DEPS_REQUIRED uses hyphenated skill names (also intrinsically lowercase). The invariant is fully and consistently applied across all registries of equivalent structural kind in this module.

src/autoskillit/cli/_install_info.py
Trecek: Agreed — fixed in 763028ec. The DirectUrlInfo TypedDict added to parse_direct_url() types the url field as str, so info['url'] is now str instead of Any. The downstream url.startswith('file://') check and InstallInfo.url: str | None are now type-safe with no isinstance guard needed.

src/autoskillit/core/_install_detect.py
```diff
@@ -4,22 +4,101 @@
...
+            return {
+                "install_type": "unknown",
+                "requested_revision": None,
+                "commit_id": None,
+                "editable": False,
+                "url": "",
+            }
         data = json.loads(raw)
+        url = data.get("url", "") or ""
+        vcs_info = data.get("vcs_info", {})
```
Trecek, line 52: [warning] defense: data.get("url", "") returns Any since json.loads() returns Any. The `or ""` coercion does not guarantee a str — if direct_url.json contains `"url": null` or `"url": 123`, url will be None or int respectively, causing downstream AttributeError even with isinstance guards. The url field in DirectUrlInfo is declared as `str` but may silently hold a non-str value. Add `str(data.get("url") or "")` or an explicit isinstance check.

src/autoskillit/core/_install_detect.py
```diff
@@ -4,22 +4,101 @@
...
+        if isinstance(vcs_info, dict) and vcs_info.get("vcs") == "git":
```
Trecek, line 53: [info] defense: vcs_info = data.get("vcs_info", {}) assumes data is a dict, but json.loads() can return list/str/int if direct_url.json is malformed. If data is not a dict, data.get() raises AttributeError — not caught until the outer except Exception. Add isinstance(data, dict) guard before data.get() calls.

src/autoskillit/cli/_install_info.py
```diff
@@ -84,32 +71,20 @@ def detect_install() -> InstallInfo:
...
                 commit_id=None,
                 requested_revision=None,
-                url=url,
+                url=url or None,
                 editable_source=None,
             )
-
         return _unknown
-
     except Exception:
```
Trecek, line 83: [warning] defense: Bare `except Exception` in detect_install() catches any exception from parse_direct_url() including unexpected RuntimeError/AttributeError. Exceptions are only logged at DEBUG level, making failures invisible in production. Since parse_direct_url() already catches and returns a safe fallback, this outer try/except is redundant and masks real bugs.

src/autoskillit/core/_version_snapshot.py
```diff
@@ -52,29 +54,9 @@ def _autoskillit_version() -> str:
...
-        if isinstance(dir_info, dict) and dir_info.get("editable") is True:
-            return {"install_type": "local-editable", "commit_id": None}
-        if isinstance(url, str) and url.startswith("file://"):
-            return {"install_type": "local-path", "commit_id": None}
-        return {"install_type": "unknown", "commit_id": None}
-    except Exception:
-        logger.warning("Failed to parse install info from direct_url.json", exc_info=True)
-        return {"install_type": "unknown", "commit_id": None}
+    """Classify the autoskillit install type and commit hash."""
+    info = parse_direct_url()
```
Trecek, line 58: [info] defense: _install_info() calls parse_direct_url() without exception handling. parse_direct_url() does handle its own exceptions and returns a safe fallback, so this is low risk — but if parse_direct_url() itself has a bug that escapes its try/except, _install_info() will raise and break version snapshot collection. Low risk given the defensive implementation, but worth noting.

Trecek (commented):

Trecek (commented):

src/autoskillit/core/_install_detect.py
Trecek: Agreed — fixed in ecbe10b2. Added isinstance guard: raw_url = data.get('url'); url = raw_url if isinstance(raw_url, str) else ''. This ensures that a truthy non-string url (e.g. int 123) cannot propagate into the DirectUrlInfo TypedDict url: str field.

src/autoskillit/cli/_install_info.py
Trecek: Investigated — this is intentional. The bare except Exception pattern at L83 is the established IL-0 safety contract across the entire install-detection layer: parse_direct_url() (L84 in _install_detect.py) and is_dev_install() (L104) both use the same pattern. parse_direct_url() already wraps all its own errors and guarantees no-raise, making detect_install()'s outer guard belt-and-suspenders. Narrowing the exception type would violate this IL-0 never-raising stability guarantee without adding diagnostic value.

src/autoskillit/core/_install_detect.py
Trecek: Acknowledged — minor suggestion noted. The outer except Exception at L84 in parse_direct_url() catches any AttributeError from calling .get() on a non-dict, so this is low-risk at runtime. Flagged for human review as a potential future hardening point.

src/autoskillit/core/_version_snapshot.py
Trecek: Acknowledged — minor suggestion noted. parse_direct_url() guarantees no-raise via its own except Exception handler, so the direct call in _install_info() is safe. Flagged for human review as a potential future hardening point should the callee contract change.

PR #1665, opened 2026-05-03
Rectify: Eliminate Regex-Based Pattern Classification in Headless Recovery

### LINKED ISSUE #1642: _PATH_CAPTURE regex fails for \S+-terminated patterns, disabling nudge/synthesis for planner skills
## Summary

The `_PATH_CAPTURE` regex in `_headless_recovery.py:37` fails to match `\S+`-terminated expected_output_patterns, rendering the contract recovery nudge and artifact synthesis mechanisms dead for all planner skills and ~15 other skills. The regex works correctly for 68 skills using `/.+`-terminated patterns.

This was a contributing factor in the planner pipeline failure observed on 2026-05-02/03: after the feature gate bug caused `adjudicated_failure` with `write_call_count=1`, the CONTRACT_RECOVERY gate promoted the result to retriable, and the nudge mechanism attempted to recover — but silently produced zero hints and short-circuited, forcing an unnecessary full retry.

## Root Cause

`_headless_recovery.py:37`:
```python
_PATH_CAPTURE: re.Pattern[str] = re.compile(r"^(\w+)\\s\*=\\s\*/.+")
```

The compiled regex requires the pattern string to end with `/.+` (a literal forward-slash followed by any characters). Of the 112 patterns in `skill_contracts.yaml`:
- **68 patterns match** — those using `/.+` suffix (e.g., `investigation_path\s*=\s*/.+`, `plan_path\s*=\s*/.+`)
- **16 patterns DO NOT match** — those using `\S+` suffix (e.g., `elab_result_path\s*=\s*\S+`, `refined_plan_path\s*=\s*\S+`)
- **28 patterns correctly rejected** — non-path patterns (e.g., `verdict\s*=\s*(GO|NO GO)`)

All planner skill patterns use `\S+` suffix and therefore fail.

### Impact Chain

1. `_extract_missing_token_hints` (lines 144-174): iterates `expected_output_patterns`, calls `_PATH_CAPTURE.match(pattern)` — returns `None` for `\S+` patterns → appends nothing to hints list → returns `[]`
2. `_attempt_contract_nudge` (lines 194-196): `hints = _extract_missing_token_hints(...)` → `if not hints: return None` → nudge short-circuits, never reaches the resume subprocess call
3. `_synthesize_from_write_artifacts` (lines 119-120): same `_PATH_CAPTURE.match(pattern)` gate → returns `None` → no synthesis for planner skills (note: this function is also restricted to `ChannelConfirmation.UNMONITORED` sessions only at `_headless_result.py:374`)

### Why the Nudge Would Have Helped (if Fixed)

For the elaborate-phase `adjudicated_failure` sessions: the model wrote `Phase1.md` via Write tool (`write_call_count=1`), and the CONTRACT_RECOVERY gate correctly promoted the result to retriable. If the nudge had extracted the Write path and asked the model to emit the token, it could have recovered the session without a full retry. However, see the sequencing note below.

## Recommended Fix

Replace the `_PATH_CAPTURE` regex to match both `/.+` and `\S+` pattern suffixes:

```python
_PATH_CAPTURE: re.Pattern[str] = re.compile(r"^(\w+)\\s\*=\\s\*(?:\\S|/)")
```

This matches the literal `\s*=\s*` sequence (present in all path-capture patterns from skill_contracts.yaml) followed by either `\S` (start of a `\S+` token) or `/` (start of a `/.+` path). Correctly rejects non-path patterns like `verdict\s*=\s*(GO|NO GO)`.

### Sequencing Requirement

**This fix MUST be sequenced AFTER the feature gate fix (planner skill injection bypass).** Reason: if the `_PATH_CAPTURE` regex is fixed while the feature gate still blocks planner skills, the nudge mechanism would activate and succeed by synthesizing `elab_result_path = /path/to/Phase1.md` — a Markdown file emitted by the wrong skill (`elaborate-phase`). The downstream pipeline would receive a Markdown file where it expects a JSON phase elaboration result. This is a **silent corruption pathway**.

Once the feature gate fix ensures the correct planner skill runs, the nudge mechanism can safely recover genuine emission omissions (model wrote the correct file but forgot to emit the structured token).

## Affected Files

- `src/autoskillit/execution/_headless_recovery.py:37` — `_PATH_CAPTURE` regex (primary fix)
- `src/autoskillit/execution/_headless_recovery.py:119` — `_synthesize_from_write_artifacts` (consumer)
- `src/autoskillit/execution/_headless_recovery.py:157` — `_extract_missing_token_hints` (consumer)
- `src/autoskillit/execution/_headless_recovery.py:194` — `_attempt_contract_nudge` (consumer)

## Test Gap

No existing test passes actual `skill_contracts.yaml` patterns through `_PATH_CAPTURE.match()`. A test should load all patterns from `skill_contracts.yaml`, categorize them by suffix (`/.+` vs `\S+` vs other), and assert that `_PATH_CAPTURE` matches all path-capture patterns (both suffix types) while rejecting non-path patterns.

Additionally, a test for `_extract_missing_token_hints` should verify it produces non-empty hints when given a session with Write tool_use entries and `\S+`-terminated patterns.

## Investigation Reference

Full investigation report: `.autoskillit/temp/investigate/investigation_headless_adjudication_instability_2026-05-02_183000.md`

Trecek:
## Summary

The headless recovery system uses a regex heuristic (`_PATH_CAPTURE`) to classify whether an `expected_output_pattern` string represents a path-capture token. This heuristic fails for 15 of 112 patterns because it hardcodes the `/.+` suffix assumption and misses `\S+`-terminated patterns. The same broken regex is duplicated in test compliance code, creating a blind spot that hides the bug from static checks.

The architectural weakness is **dual-source-of-truth divergence**: the pattern string format and the `outputs[].type` structured metadata both encode "is this a path output?" — but only the structured metadata is reliable. The `_headless_path_tokens.py` module already uses the correct structured approach; the recovery module (`_headless_recovery.py`) uses a fragile regex heuristic instead.

The immunity plan replaces all regex-based pattern classification with structured metadata lookup, adds a closed-loop contract validation rule to catch future drift, and eliminates hardcoded pattern fixtures from tests.

Closes #1642

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260502-231348-455729/.autoskillit/temp/rectify/rectify_path_capture_regex_immunity_2026-05-02_231348.md`

## Token Usage Summary

| rectify | 60 | 14.3k | 809.3k | 180.3k | 1 | 12m 12s |
| dry_walkthrough | 1.5k | 8.5k | 859.9k | 138.7k | 1 | 7m 31s |
| implement | 708 | 58.1k | 8.7M | 534.3k | 1 | 23m 44s |
| prepare_pr | 60 | 6.2k | 234.9k | 33.0k | 1 | 1m 42s |
| compose_pr | 59 | 1.8k | 193.5k | 19.5k | 1 | 57s |
| review_pr | 329 | 67.1k | 1.5M | 150.5k | 2 | 16m 16s |
| resolve_review | 1.2k | 78.9k | 12.0M | 221.4k | 2 | 36m 28s |
| ci_conflict_fix | 164 | 6.5k | 855.8k | 42.3k | 1 | 2m 18s |
| **Total** | 4.1k | 241.3k | 25.2M | 1.3M | | 1h 41m |
## Token Efficiency

| rectify | 0 | — | — | — |
| implement | 332 | 26275.9 | 1609.2 | 174.9 |
| **Total** | **332** | 32594.2 | 2728.3 | 267.8 |

### DIFF
--- a/src/autoskillit/execution/_headless_path_tokens.py
+++ b/src/autoskillit/execution/_headless_path_tokens.py
@@ -78,6 +78,48 @@ def _build_path_token_set() -> frozenset[str]:
 
 _OUTPUT_PATH_TOKENS: frozenset[str] = _build_path_token_set()
 
+
+def _build_recoverable_path_tokens() -> frozenset[str]:
+    """Derive token names eligible for write-artifact recovery from skill_contracts.yaml.
+
+    Includes outputs where type starts with "file_path" OR equals "directory_path".
+    Broader than _OUTPUT_PATH_TOKENS (which excludes directory_path). Used by
+    _headless_recovery.py for synthesis and nudge classification.
+
+    Loads the YAML directly via IL-0 core utilities to avoid an upward IL-1→IL-2 import.
+    """
+    try:
+        manifest_path = pkg_root() / "recipe" / "skill_contracts.yaml"
+        manifest = load_yaml(manifest_path)
+        if not isinstance(manifest, dict):
+            return frozenset()
+        result = (
+            frozenset(
+                out["name"]
+                for skill_data in manifest.get("skills", {}).values()
+                for out in skill_data.get("outputs", [])
+                if isinstance(out, dict)
+                and (
+                    out.get("type", "").startswith("file_path")
+                    or out.get("type") == "directory_path"
+                )
+            )
+            - _INTENTIONALLY_EXCLUDED_PATH_TOKENS
+        )
+        logger.debug("_RECOVERABLE_PATH_TOKENS derived from contracts", count=len(result))
+        return result
+    except FileNotFoundError:
+        logger.debug("skill_contracts.yaml not found; _RECOVERABLE_PATH_TOKENS will be empty")
+        return frozenset()
+    except Exception:
+        logger.warning(
+            "Failed to derive _RECOVERABLE_PATH_TOKENS from contracts YAML", exc_info=True
+        )
+        return frozenset()
+
+
+_RECOVERABLE_PATH_TOKENS: frozenset[str] = _build_recoverable_path_tokens()
+
 _OUTPUT_PATH_PATTERN: re.Pattern[str] = (
     re.compile(
         r"^(" + "|".join(re.escape(t) for t in sorted(_OUTPUT_PATH_TOKENS)) + r")\s*=\s*(.+)$",
--- a/src/autoskillit/execution/_headless_path_tokens.py
+++ b/src/autoskillit/execution/_headless_path_tokens.py
@@ -59,10 +59,12 @@ def _build_path_token_set() -> frozenset[str]:
             return frozenset()
         result = (
             frozenset(
-                out["name"]
+                out.get("name", "")
                 for skill_data in manifest.get("skills", {}).values()
                 for out in skill_data.get("outputs", [])
-                if isinstance(out, dict) and out.get("type", "").startswith("file_path")
+                if isinstance(out, dict)
+                and out.get("name", "")
+                and out.get("type", "").startswith("file_path")
             )
             - _INTENTIONALLY_EXCLUDED_PATH_TOKENS
         )
@@ -95,10 +97,11 @@ def _build_recoverable_path_tokens() -> frozenset[str]:
             return frozenset()
         result = (
             frozenset(
-                out["name"]
+                out.get("name", "")
                 for skill_data in manifest.get("skills", {}).values()
                 for out in skill_data.get("outputs", [])
                 if isinstance(out, dict)
+                and out.get("name", "")
                 and (
                     out.get("type", "").startswith("file_path")
                     or out.get("type") == "directory_path"
--- a/src/autoskillit/execution/_headless_recovery.py
+++ b/src/autoskillit/execution/_headless_recovery.py
@@ -51,9 +51,12 @@ def _is_path_capture_pattern(pattern: str) -> str | None:
     if not m:
         return None
     token_name = m.group(1)
-    if token_name in _RECOVERABLE_PATH_TOKENS:
-        return token_name
-    return None
+    if token_name not in _RECOVERABLE_PATH_TOKENS:
+        return None
+    remainder = pattern[m.end() :]
+    if not re.match(r"\s*=", remainder):
+        return None
+    return token_name
 
 
 def _recover_from_separate_marker(
--- a/src/autoskillit/execution/_headless_recovery.py
+++ b/src/autoskillit/execution/_headless_recovery.py
@@ -9,6 +9,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from autoskillit.core import CliSubtype, OutputFormat, RetryReason, SkillResult, get_logger
+from autoskillit.execution._headless_path_tokens import _RECOVERABLE_PATH_TOKENS
 from autoskillit.execution.commands import build_headless_resume_cmd
 from autoskillit.execution.process import _marker_is_standalone
 from autoskillit.execution.session import (
@@ -34,9 +35,25 @@ _CHANNEL_B_RECOVERABLE_SUBTYPES: frozenset[CliSubtype] = frozenset(
     {CliSubtype.UNPARSEABLE, CliSubtype.EMPTY_OUTPUT}
 )
 
-_PATH_CAPTURE: re.Pattern[str] = re.compile(r"^(\w+)\\s\*=\\s\*/.+")
+_TOKEN_NAME_RE: re.Pattern[str] = re.compile(r"^(\w+)")
 
 _NUDGE_TIMEOUT: float = 60.0
+
+
+def _is_path_capture_pattern(pattern: str) -> str | None:
+    """Return the token name if pattern is a path-capture pattern, else None.
+
+    Classification uses outputs[].type metadata from skill_contracts.yaml rather than
+    the pattern string suffix format, so all path-capture patterns are covered regardless
+    of whether they end in /.+, \\S+, .+, or any other suffix.
+    """
+    m = _TOKEN_NAME_RE.match(pattern)
+    if not m:
+        return None
+    token_name = m.group(1)
+    if token_name in _RECOVERABLE_PATH_TOKENS:
+        return token_name
+    return None
 
 
 def _recover_from_separate_marker(
@@ -117,10 +134,9 @@ def _synthesize_from_write_artifacts(
 
     synthesized_lines: list[str] = []
     for pattern in expected_output_patterns:
-        m = _PATH_CAPTURE.match(pattern)
-        if not m:
+        token_name = _is_path_capture_pattern(pattern)
+        if not token_name:
             continue
-        token_name = m.group(1)
         # Skip if the pattern is already satisfied in the current result.
         if re.search(pattern, session.result):
             continue
@@ -155,10 +171,9 @@ def _extract_missing_token_hints(
     hints: list[tuple[str, str]] = []
 
     for pattern in expected_output_patterns:
-        m = _PATH_CAPTURE.match(pattern)
-        if not m:
+        token_name = _is_path_capture_pattern(pattern)
+        if not token_name:
             continue
-        token_name = m.group(1)
         # Skip if already satisfied
         if re.search(pattern, session.result):
             continue
--- a/src/autoskillit/execution/_headless_recovery.py
+++ b/src/autoskillit/execution/_headless_recovery.py
@@ -53,8 +53,7 @@ def _is_path_capture_pattern(pattern: str) -> str | None:
     token_name = m.group(1)
     if token_name not in _RECOVERABLE_PATH_TOKENS:
         return None
-    remainder = pattern[m.end() :]
-    if not re.match(r"\s*=", remainder):
+    if "=" not in pattern[m.end() :]:
         return None
     return token_name
 
--- a/src/autoskillit/execution/headless.py
+++ b/src/autoskillit/execution/headless.py
@@ -45,8 +45,10 @@ from autoskillit.execution._headless_path_tokens import (  # noqa: F401
     _INTENTIONALLY_EXCLUDED_PATH_TOKENS,
     _OUTPUT_PATH_PATTERN,
     _OUTPUT_PATH_TOKENS,
+    _RECOVERABLE_PATH_TOKENS,
     _WORKTREE_PATH_PATTERN,
     _build_path_token_set,
+    _build_recoverable_path_tokens,
     _extract_output_paths,
     _extract_worktree_path,
     _validate_output_paths,
@@ -54,9 +56,10 @@ from autoskillit.execution._headless_path_tokens import (  # noqa: F401
 from autoskillit.execution._headless_recovery import (
     _CHANNEL_B_RECOVERABLE_SUBTYPES,  # noqa: F401
     _NUDGE_TIMEOUT,  # noqa: F401
-    _PATH_CAPTURE,  # noqa: F401
+    _TOKEN_NAME_RE,  # noqa: F401
     _attempt_contract_nudge,
     _extract_missing_token_hints,  # noqa: F401
+    _is_path_capture_pattern,  # noqa: F401
     _merge_token_usage,  # noqa: F401
     _recover_block_from_assistant_messages,  # noqa: F401
     _recover_from_separate_marker,  # noqa: F401
--- a/src/autoskillit/execution/headless.py
+++ b/src/autoskillit/execution/headless.py
@@ -48,7 +48,6 @@ from autoskillit.execution._headless_path_tokens import (  # noqa: F401
     _RECOVERABLE_PATH_TOKENS,
     _WORKTREE_PATH_PATTERN,
     _build_path_token_set,
-    _build_recoverable_path_tokens,
     _extract_output_paths,
     _extract_worktree_path,
     _validate_output_paths,
--- a/src/autoskillit/recipe/rules_contracts.py
+++ b/src/autoskillit/recipe/rules_contracts.py
@@ -13,6 +13,10 @@ from autoskillit.recipe.contracts import (
 )
 from autoskillit.recipe.registry import RuleFinding, semantic_rule
 
+# Tokens handled by dedicated extractors rather than the write-artifact recovery pipeline.
+# These are excluded from path-output-recovery-coverage checks.
+_PATH_RECOVERY_EXCLUDED_TOKENS: frozenset[str] = frozenset({"worktree_path", "branch_name"})
+
 # Skill names covered by the result-field-drift rule.
 _RESULT_FIELD_DRIFT_SKILLS = frozenset(
     {
@@ -523,5 +527,70 @@ def _check_all_examples_match_all_patterns(ctx: ValidationContext) -> list[RuleF
                             ),
                         )
                     )
+
+    return findings
+
+
+@semantic_rule(
+    name="path-output-recovery-coverage",
+    description=(
+        "Every file_path/directory_path output must have a matching expected_output_pattern"
+    ),
+    severity=Severity.WARNING,
+)
+def _check_path_output_recovery_coverage(ctx: ValidationContext) -> list[RuleFinding]:
+    """Every file_path/directory_path output must have a matching expected_output_pattern.
+
+    Without a matching pattern, the recovery system cannot classify the output as
+    path-capture and synthesis/nudge will be silently disabled.
+    """
+    findings: list[RuleFinding] = []
+    try:
+        manifest = load_bundled_manifest()
+    except Exception:
+        logger.warning(
+            "path-output-recovery-coverage: failed to load bundled manifest; skipping",
+            exc_info=True,
+        )
+        return findings
+
+    for step_name, step in ctx.recipe.steps.items():
+        if step.tool != "run_skill":
+            continue
+        skill_cmd = (step.with_args or {}).get("skill_command", "")
+        name = resolve_skill_name(skill_cmd)
+        if not name:
+            continue
+        contract = get_skill_contract(name, manifest)
+        if not contract:
+            continue
+
+        path_outputs = [
+            o
+            for o in contract.outputs
+            if (o.type.startswith("file_path") or o.type == "directory_path")
+            and o.name not in _PATH_RECOVERY_EXCLUDED_TOKENS
+        ]
+        if not path_outputs:
+            continue
+
+        for output in path_outputs:
+            matched = any(
+                _re.match(rf"^{_re.escape(output.name)}\\\\s\*=", p)
+                for p in contract.expected_output_patterns
+            )
+            if not matched:
+                findings.append(
+                    RuleFinding(
+                        rule="path-output-recovery-coverage",
+                        severity=Severity.WARNING,
+                        step_name=step_name,
+                        message=(
+                            f"Skill '{name}': output '{output.name}' (type={output.type!r}) "
+                            f"has no matching expected_output_pattern. The write-artifact "
+                            f"recovery system will be silently disabled for this output."
+                        ),
+                    )
+                )
 
     return findings
--- a/src/autoskillit/recipe/rules_contracts.py
+++ b/src/autoskillit/recipe/rules_contracts.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import re as _re
 
 from autoskillit.core import Severity, get_logger, pkg_root
+from autoskillit.execution._headless_path_tokens import _INTENTIONALLY_EXCLUDED_PATH_TOKENS
 from autoskillit.recipe._analysis import ValidationContext
 from autoskillit.recipe.contracts import (
     get_skill_contract,
@@ -12,10 +13,6 @@ from autoskillit.recipe.contracts import (
     resolve_skill_name,
 )
 from autoskillit.recipe.registry import RuleFinding, semantic_rule
-
-# Tokens handled by dedicated extractors rather than the write-artifact recovery pipeline.
-# These are excluded from path-output-recovery-coverage checks.
-_PATH_RECOVERY_EXCLUDED_TOKENS: frozenset[str] = frozenset({"worktree_path", "branch_name"})
 
 # Skill names covered by the result-field-drift rule.
 _RESULT_FIELD_DRIFT_SKILLS = frozenset(
@@ -569,14 +566,14 @@ def _check_path_output_recovery_coverage(ctx: ValidationContext) -> list[RuleFin
             o
             for o in contract.outputs
             if (o.type.startswith("file_path") or o.type == "directory_path")
-            and o.name not in _PATH_RECOVERY_EXCLUDED_TOKENS
+            and o.name not in _INTENTIONALLY_EXCLUDED_PATH_TOKENS
         ]
         if not path_outputs:
             continue
 
         for output in path_outputs:
             matched = any(
-                _re.match(rf"^{_re.escape(output.name)}\\\\s\*=", p)
+                _re.match(rf"^{_re.escape(output.name)}\s*=", p)
                 for p in contract.expected_output_patterns
             )
             if not matched:
--- a/src/autoskillit/recipe/rules_contracts.py
+++ b/src/autoskillit/recipe/rules_contracts.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import re as _re
 
 from autoskillit.core import Severity, get_logger, pkg_root
-from autoskillit.execution._headless_path_tokens import _INTENTIONALLY_EXCLUDED_PATH_TOKENS
 from autoskillit.recipe._analysis import ValidationContext
 from autoskillit.recipe.contracts import (
     get_skill_contract,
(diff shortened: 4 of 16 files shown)


Trecek (commented):

src/autoskillit/recipe/rules_contracts.py
```diff
@@ -525,3 +529,68 @@ def _check_all_examples_match_all_patterns(ctx: ValidationContext) -> list[RuleF
...
+    for step_name, step in ctx.recipe.steps.items():
+        if step.tool != "run_skill":
+            continue
+        skill_cmd = (step.with_args or {}).get("skill_command", "")
+        name = resolve_skill_name(skill_cmd)
+        if not name:
+            continue
+        contract = get_skill_contract(name, manifest)
+        if not contract:
+            continue
```
Trecek, line 567: [critical] arch: `continue` at line 566 is dedented relative to the `if not contract:` block above it. This will cause a SyntaxError or unconditionally skip the remainder of the loop body — the `continue` must be indented to match the `if` body.

src/autoskillit/recipe/rules_contracts.py
```diff
@@ -525,3 +529,68 @@ def _check_all_examples_match_all_patterns(ctx: ValidationContext) -> list[RuleF
...
+            for o in contract.outputs
+            if (o.type.startswith("file_path") or o.type == "directory_path")
+            and o.name not in _PATH_RECOVERY_EXCLUDED_TOKENS
+        ]
+        if not path_outputs:
+            continue
+
+        for output in path_outputs:
+            matched = any(
+                _re.match(rf"^{_re.escape(output.name)}\\\\s\*=", p)
```
Trecek: [critical] bugs: Regex pattern `rf"^{_re.escape(output.name)}\\s\*="` produces the literal string `\\s\*=` at the regex engine (backslash-s-asterisk-equals), NOT optional whitespace (`\s*=`). The correct pattern is `rf"^{_re.escape(output.name)}\s*="`. As written, the coverage check will never match any `expected_output_patterns` entry and will spuriously emit WARNING findings for every path output.

src/autoskillit/execution/_headless_path_tokens.py
```diff
@@ -78,6 +78,48 @@ def _build_path_token_set() -> frozenset[str]:
...
+    Loads the YAML directly via IL-0 core utilities to avoid an upward IL-1→IL-2 import.
+    """
+    try:
+        manifest_path = pkg_root() / "recipe" / "skill_contracts.yaml"
+        manifest = load_yaml(manifest_path)
+        if not isinstance(manifest, dict):
+            return frozenset()
+        result = (
+            frozenset(
+                out["name"]
```
Trecek: [warning] bugs: `out["name"]` is accessed without a missing-key guard. If a manifest entry lacks a `name` key this raises KeyError inside the generator, which is caught by the bare `except Exception` handler and silently returns `frozenset()` — disabling all path-token recovery with no targeted error. Use `out.get("name")` and skip `None` entries.

src/autoskillit/execution/_headless_path_tokens.py
```diff
@@ -78,6 +78,48 @@ def _build_path_token_set() -> frozenset[str]:
...
+        logger.debug("skill_contracts.yaml not found; _RECOVERABLE_PATH_TOKENS will be empty")
+        return frozenset()
+    except Exception:
+        logger.warning(
+            "Failed to derive _RECOVERABLE_PATH_TOKENS from contracts YAML", exc_info=True
+        )
+        return frozenset()
+
+
+_RECOVERABLE_PATH_TOKENS: frozenset[str] = _build_recoverable_path_tokens()
```
Trecek, line 124: [warning] cohesion: `_RECOVERABLE_PATH_TOKENS` is defined in `_headless_path_tokens.py` but its sole consumer is `_headless_recovery.py`. Defining the constant in the path-tokens module and importing it back into recovery fragments the subsystem. Moving it to `_headless_recovery.py` co-locates the constant with its consumers.

src/autoskillit/execution/_headless_recovery.py
```diff
@@ -34,11 +35,27 @@
     {CliSubtype.UNPARSEABLE, CliSubtype.EMPTY_OUTPUT}
 )
 
-_PATH_CAPTURE: re.Pattern[str] = re.compile(r"^(\w+)\\s\*=\\s\*/.+")
+_TOKEN_NAME_RE: re.Pattern[str] = re.compile(r"^(\w+)")
 
 _NUDGE_TIMEOUT: float = 60.0
 
 
+def _is_path_capture_pattern(pattern: str) -> str | None:
```
Trecek, line 43: [warning] bugs: `_is_path_capture_pattern` classifies any pattern whose leading token name is in `_RECOVERABLE_PATH_TOKENS`, with no check that the pattern contains an assignment form (`\s*=\s*...`). A bare token name (e.g. `elab_result_path` with no `=`) would incorrectly match, causing spurious synthesis or hint-extraction attempts.

src/autoskillit/execution/headless.py
```diff
@@ -45,18 +45,21 @@
...
+    _build_recoverable_path_tokens,
     _extract_output_paths,
     _extract_worktree_path,
     _validate_output_paths,
 )
 from autoskillit.execution._headless_recovery import (
     _CHANNEL_B_RECOVERABLE_SUBTYPES,  # noqa: F401
     _NUDGE_TIMEOUT,  # noqa: F401
-    _PATH_CAPTURE,  # noqa: F401
+    _TOKEN_NAME_RE,  # noqa: F401
```
Trecek, line 58: [warning] cohesion: `_TOKEN_NAME_RE` and `_is_path_capture_pattern` are re-exported with `# noqa: F401` but `_RECOVERABLE_PATH_TOKENS` at L48 is imported without it. All three are internal re-exports from the same facade — treat them symmetrically to avoid silently hiding unused-import warnings for some but not others.

src/autoskillit/recipe/rules_contracts.py
```diff
@@ -13,6 +13,10 @@
 )
 from autoskillit.recipe.registry import RuleFinding, semantic_rule
 
+# Tokens handled by dedicated extractors rather than the write-artifact recovery pipeline.
+# These are excluded from path-output-recovery-coverage checks.
+_PATH_RECOVERY_EXCLUDED_TOKENS: frozenset[str] = frozenset({"worktree_path", "branch_name"})
```
Trecek, line 19: [warning] cohesion: `_PATH_RECOVERY_EXCLUDED_TOKENS` duplicates the exclusion concept already represented by `_INTENTIONALLY_EXCLUDED_PATH_TOKENS` in `_headless_path_tokens.py`. Two separate constants for the same semantic set create a synchronization hazard — one should derive from or reference the other.

src/autoskillit/skills_extended/bundle-local-report/SKILL.md
```diff
@@ -16,7 +16,7 @@ mermaid diagrams and inserted plot images from `yaml:figure-spec` blocks.
 - Exit without emitting `html_path = ` (even empty) as your final output — the recipe `capture:` block expects it.
 
 **ALWAYS:**
-- Emit `html_path = {path}` (or `html_path = ` if report_path is absent) as your final output.
+- Emit `html_path = <absolute path to report.html>` (or `html_path = ` if report_path is absent) as your final output.
```
Trecek: [warning] cohesion: Emit directive uses `Emit \`html_path = ...\`` (no colon after 'Emit'), while all other SKILL.md files updated in this PR use `Emit: \`token = ...\`` (colon present). Breaks the uniform emit-directive syntax across skill definitions.

tests/execution/test_headless_path_validation.py
```diff
@@ -1182,6 +1216,98 @@ async def test_nudge_exception_falls_through(self, tool_ctx):
...
+                if not isinstance(out, dict):
+                    continue
+                t = out.get("type", "")
+                name = out.get("name", "")
+                if t.startswith("file_path") or t == "directory_path":
+                    path_token_names.add(name)
+                elif t:
+                    non_path_token_names.add(name)
+            for pattern in skill_data.get("expected_output_patterns", []):
+                m = re.match(r"^(\w+)", pattern)
```
Trecek, line 1249: [warning] tests: The regex `r"^(\\w+)"` captures only the leading word of each pattern without verifying the pattern contains `\\s*=` after the token name. Patterns sharing a prefix with a path token but not in assignment form would be incorrectly included in `pattern_token_map`, producing false-positive or false-negative classification assertions.

tests/execution/test_headless_path_validation.py
```diff
@@ -1182,6 +1216,98 @@ async def test_nudge_exception_falls_through(self, tool_ctx):
...
+                if not isinstance(out, dict):
+                    continue
+                t = out.get("type", "")
+                if t.startswith("file_path") or t == "directory_path":
+                    declared.add(out["name"])
+
+        from autoskillit.execution._headless_path_tokens import _INTENTIONALLY_EXCLUDED_PATH_TOKENS
+
+        expected = declared - _INTENTIONALLY_EXCLUDED_PATH_TOKENS
+        assert _RECOVERABLE_PATH_TOKENS == expected, (
```
Trecek: [warning] tests: `_RECOVERABLE_PATH_TOKENS == expected` asserts exact equality without normalizing token names (e.g. `.strip()`). If any manifest `out["name"]` value contains unexpected whitespace the test produces a false failure with an unhelpful diff message.

tests/skills/test_skill_output_compliance.py
```diff
@@ -321,6 +321,21 @@ def test_output_path_tokens_synchronized() -> None:
     "vis-lens-temporal": ["diagram_path"],
     "vis-lens-uncertainty": ["diagram_path"],
     "review-design": ["evaluation_dashboard", "revision_guidance"],
+    # Planner pipeline skills (use \S+-terminated patterns)
```
Trecek: [warning] slop: Comment `# Planner pipeline skills (use \\S+-terminated patterns)` references the retired regex-suffix classification approach. The new mechanism is type-metadata-driven; the suffix notation is misleading dead context and should be removed.

tests/skills/test_skill_output_compliance.py
```diff
@@ -321,6 +321,21 @@ def test_output_path_tokens_synchronized() -> None:
...
+    "planner-assess-review-approach": ["review_approach_assessment_path"],
+    "planner-elaborate-assignments": ["phase_assignments_result_dir"],
+    "planner-elaborate-phase": ["elab_result_path"],
+    "planner-elaborate-wps": ["phase_wps_result_dir"],
+    "planner-generate-phases": ["phase_manifest_path"],
+    "planner-refine-assignments": ["refined_assignments_path"],
+    "planner-refine-phases": ["refined_plan_path"],
+    "planner-refine-wps": ["refined_wps_path"],
+    "planner-validate-task-alignment": ["alignment_findings_path"],
+    # Audit skills (use \S+-terminated patterns)
```
Trecek: [warning] slop: Comment `# Audit skills (use \\S+-terminated patterns)` — same issue as above. The suffix notation is a vestige of the eliminated regex approach.

tests/skills/test_skill_output_compliance.py
```diff
@@ -321,6 +321,21 @@ def test_output_path_tokens_synchronized() -> None:
...
+    "audit-tests": ["audit_report_path"],
+    "validate-audit": ["validated_report_path"],
+    # Report bundling skill (uses \S+-terminated pattern)
```
Trecek: [warning] slop: Comment `# Report bundling skill (uses \\S+-terminated pattern)` — same issue. Remove suffix references that describe the retired regex approach.

Trecek (commented):
AutoSkillit review: 2 critical blocking issues found — changes required before merge.

**Verdict: changes_requested** (posted as comment — own-PR restriction)

**Blocking (critical):**
- `rules_contracts.py:566` — `continue` dedented outside `if` block (SyntaxError/unconditional skip)
- `rules_contracts.py:579` — Regex `\\\\s\*=` matches literal backslash-s-asterisk rather than optional whitespace; all coverage checks will spuriously fire

See inline comments for details on all 13 findings.

src/autoskillit/recipe/rules_contracts.py
Trecek: Investigated — this is intentional. Line 565 `if not contract:` is indented 8 spaces; line 566 `continue` is indented 12 spaces — correctly inside the `if` block. The diff hunk's leading `+` prefix may have made the column alignment appear off. Python parses this as valid, and the function works correctly.

src/autoskillit/recipe/rules_contracts.py
Trecek: Agreed — fixed in 2d88a22f. Changed `rf"^\{_re.escape(output.name)\}\\\\s\\*="` to `rf"^\{_re.escape(output.name)\}\s\*="` so the regex engine sees `\s\*=` (optional whitespace + equals), not the literal four-char sequence `\\s\\*=`.

src/autoskillit/execution/_headless_path_tokens.py
Trecek: Agreed — fixed in 61b6b2f8. Changed `out["name"]` to `out.get("name", "")` with an `and out.get("name", "")` guard in both `_build_path_token_set` and `_build_recoverable_path_tokens`. This ensures a missing key returns an empty string (filtered out by the truthy guard) instead of raising KeyError and silently disabling all recovery.

src/autoskillit/execution/_headless_path_tokens.py
Trecek: Investigated — this is intentional. `_RECOVERABLE_PATH_TOKENS` has three consumers, not one: (1) `_headless_recovery.py` uses it in `_is_path_capture_pattern`, (2) `headless.py` re-exports it as a facade, and (3) `tests/execution/test_headless_path_validation.py` imports it for coverage assertions. Defining it in `_headless_path_tokens.py` is the correct location for a path-token module.

src/autoskillit/execution/_headless_recovery.py
Trecek: Agreed — fixed in 61b6b2f8. Added `if "=" not in pattern[m.end():]: return None` to `_is_path_capture_pattern`. This correctly handles both YAML-loaded regex patterns (which use literal `\s*=`) and direct-equality patterns. A bare token name like `elab_result_path` with no `=` is now correctly rejected.

src/autoskillit/execution/headless.py
Trecek: Investigated — this is intentional. The `# noqa: F401` on the opening `from autoskillit.execution._headless_path_tokens import (  # noqa: F401` line (L44) covers all names in the import block including `_RECOVERABLE_PATH_TOKENS` at L48. ruff applies block-level noqa to the entire import group. The asymmetric per-line noqa on the `_headless_recovery` block is a stylistic difference, not a gap.

src/autoskillit/recipe/rules_contracts.py
Trecek: Agreed — addressed in 8e71772c. Importing `_INTENTIONALLY_EXCLUDED_PATH_TOKENS` from `execution` into `recipe` violates the IL-2 import constraint (recipe cannot depend on execution). Instead, `_PATH_RECOVERY_EXCLUDED_TOKENS` is kept as a local constant with a reference comment, and a new test `TestRecoverablePathTokensCoverage.test_excluded_tokens_match_intentionally_excluded` in `test_headless_path_validation.py` asserts equality to guard against drift.

src/autoskillit/skills_extended/bundle-local-report/SKILL.md
Trecek: Agreed — fixed in 62b79743. Added colon: `Emit:` to match the uniform emit-directive syntax used in other SKILL.md files updated in this PR (`audit-tests`, `planner-*`, `validate-audit`).

tests/execution/test_headless_path_validation.py
Trecek: Agreed — fixed in 62b79743. Added `and "=" in pattern[m.end():]` guard to the pattern_token_map construction so only assignment-form patterns are included. This prevents bare token names or non-assignment patterns from producing false-positive/false-negative classification assertions.

tests/execution/test_headless_path_validation.py
Trecek: Agreed — fixed in 8e71772c. Changed `declared.add(out["name"])` to use `out.get("name", "").strip()` with a truthy guard, and normalized both sides of the assertion to `{t.strip() for t in _RECOVERABLE_PATH_TOKENS}` vs the `declared` set. The error message also uses the normalized set for accurate diffs.

tests/skills/test_skill_output_compliance.py
Trecek: Agreed — fixed in 62b79743. Removed stale comment `# Planner pipeline skills (use \S+-terminated patterns)`. Classification is now type-metadata-driven; the suffix notation is misleading.

tests/skills/test_skill_output_compliance.py
Trecek: Agreed — fixed in 62b79743. Removed stale comment `# Audit skills (use \S+-terminated patterns)`.

tests/skills/test_skill_output_compliance.py
Trecek: Agreed — fixed in 62b79743. Removed stale comment `# Report bundling skill (uses \S+-terminated pattern)`.

src/autoskillit/execution/headless.py
```diff
@@ -45,18 +45,21 @@
     _INTENTIONALLY_EXCLUDED_PATH_TOKENS,
     _OUTPUT_PATH_PATTERN,
     _OUTPUT_PATH_TOKENS,
+    _RECOVERABLE_PATH_TOKENS,
     _WORKTREE_PATH_PATTERN,
     _build_path_token_set,
+    _build_recoverable_path_tokens,
```
Trecek: [warning] arch: _build_recoverable_path_tokens is re-exported from headless.py facade without # noqa: F401; callers could invoke it directly and pay redundant YAML-load cost bypassing the singleton _RECOVERABLE_PATH_TOKENS.

tests/execution/test_headless_path_validation.py
```diff
@@ -1182,6 +1216,115 @@ async def test_nudge_exception_falls_through(self, tool_ctx):
...
+        errors = misclassified_path + misclassified_non_path
+        assert not errors, "Pattern classification errors:\n" + "\n".join(
+            f"  - {e}" for e in errors
+        )
+
+
+class TestRecoverablePathTokensCoverage:
+    def test_excluded_tokens_match_intentionally_excluded(self) -> None:
+        """_PATH_RECOVERY_EXCLUDED_TOKENS must equal _INTENTIONALLY_EXCLUDED_PATH_TOKENS.
+
```
Trecek, line 1288: [warning] tests: Trailing whitespace after _is_path_capture_pattern import.

tests/execution/test_headless_path_validation.py
```diff
@@ -1182,6 +1216,115 @@ async def test_nudge_exception_falls_through(self, tool_ctx):
         assert result.needs_retry is True
 
 
+class TestPathCaptureCoversAllContractPatterns:
+    """_is_path_capture_pattern must correctly classify all patterns in skill_contracts.yaml."""
+
+    def test_classifies_all_path_output_patterns_as_path_capture(self) -> None:
```
Trecek, line 1222: [warning] tests: test_classifies_all_path_output_patterns_as_path_capture never asserts pattern_token_map is non-empty. If contracts YAML schema changes format, all classification checks are vacuously skipped and the test passes silently. Add: assert pattern_token_map, 'No patterns found in skill_contracts.yaml — test is vacuous'.

tests/execution/test_headless_path_validation.py
```diff
@@ -1182,6 +1216,115 @@ async def test_nudge_exception_falls_through(self, tool_ctx):
...
+        for token_name, pattern in pattern_token_map:
+            if token_name in recoverable_path_token_names:
+                result = _is_path_capture_pattern(pattern)
+                if result != token_name:
+                    misclassified_path.append(
+                        f"Pattern {pattern!r} (token={token_name!r}) classified as {result!r}, "
+                        f"expected {token_name!r}"
+                    )
+
+        misclassified_non_path: list[str] = []
```
Trecek, line 1267: [warning] tests: The non-path misclassification guard `token_name not in path_token_names` at L1267 silently skips tokens appearing in both path-type and non-path-type outputs across skills. The exclusion is correct but non-obvious — add a comment explaining why.

tests/execution/test_headless_path_validation.py
```diff
@@ -1182,6 +1216,115 @@ async def test_nudge_exception_falls_through(self, tool_ctx):
...
+                    )
+
+        errors = misclassified_path + misclassified_non_path
+        assert not errors, "Pattern classification errors:\n" + "\n".join(
+            f"  - {e}" for e in errors
+        )
+
+
+class TestRecoverablePathTokensCoverage:
+    def test_excluded_tokens_match_intentionally_excluded(self) -> None:
```
Trecek, line 1286: [warning] tests: _is_path_capture_pattern is imported at L1284 but never used in test_recoverable_path_tokens_includes_directory_path_outputs. Remove the dead import.

tests/skills/test_skill_output_compliance.py
```diff
@@ -405,6 +424,23 @@ def test_every_contracted_skill_has_emit_instruction() -> None:
     )
 
 
+def test_contracted_path_capture_skills_includes_backslash_s_patterns() -> None:
+    """_get_contracted_path_capture_skills must return skills with \\S+-terminated patterns."""
+    contracted = _get_contracted_path_capture_skills()
+    assert "planner-elaborate-phase" in contracted, (
```
Trecek: [warning] tests: test_contracted_path_capture_skills_includes_backslash_s_patterns hardcodes specific skill names (planner-elaborate-phase, audit-tests, bundle-local-report) as sentinels. If those skills are renamed the test gives misleading failures unrelated to the guarded regression.

Trecek (commented):

Trecek (commented):

### ⚠️ Outside Diff Range

**tests/execution/test_headless_path_validation.py**
- **L1539** [info/tests]: test_synthesizes_path_from_backslash_s_plus_pattern passes the pattern directly to _synthesize_from_write_artifacts without verifying _is_path_capture_pattern is called internally. Synthesis could route around the fix and this test would still pass.

src/autoskillit/execution/headless.py
Trecek: Agreed — fixed in 7182d0bc. Removed `_build_recoverable_path_tokens` from the headless.py facade re-export block. The builder loads YAML on every call; only the pre-computed singleton `_RECOVERABLE_PATH_TOKENS` should be accessible from the facade.

tests/execution/test_headless_path_validation.py
Trecek: Investigated — this is intentional. Line 1284 in the current code is an empty line inside a docstring (`'\n'` with no trailing whitespace, confirmed via `cat -A`). There is no `_is_path_capture_pattern` import at or near line 1284 — that import exists only inside `test_classifies_all_path_output_patterns_as_path_capture` (L1226) where it is used. The observation references a state that no longer exists in the branch.

tests/execution/test_headless_path_validation.py
Trecek: Agreed — fixed in d07a3e90. Added `assert pattern_token_map, 'No patterns found in skill_contracts.yaml — test is vacuous'` before the classification loop, preventing a silent vacuous pass if the contracts YAML format changes.

tests/execution/test_headless_path_validation.py
Trecek: Agreed — fixed in d07a3e90. Added a comment above the guard explaining that cross-skill token name reuse is the reason for the `token_name not in path_token_names` exclusion — a token can be a path-type output in one skill and a non-path output in another skill, making the exclusion intentional rather than accidental.

tests/execution/test_headless_path_validation.py
Trecek: Investigated — this is intentional. `_is_path_capture_pattern` is not imported inside `test_recoverable_path_tokens_includes_directory_path_outputs`. The function uses only local imports for `_RECOVERABLE_PATH_TOKENS` (L1298) and `load_bundled_manifest` (L1299). The import of `_is_path_capture_pattern` lives inside `test_classifies_all_path_output_patterns_as_path_capture` (L1226) where it is actively used in the classification assertions at L1258 and L1268. No dead import exists in the current code.

tests/skills/test_skill_output_compliance.py
Trecek: Agreed — fixed in 4c471d0a. Replaced the three hardcoded skill-name assertions with a dynamic scan of `skill_contracts.yaml` for skills whose `expected_output_patterns` contain a `\S+`-terminated pattern (regex: `\\[Ss]\+$`). The test now asserts that all dynamically-discovered such skills appear in `_get_contracted_path_capture_skills()`'s result, so renames of individual skills no longer cause misleading test failures.

Issue #1645, opened 2026-05-03
planner-refine-assignments SKILL.md: stop embedding full task description in L0 prompts


Trecek:
## Problem

The `planner-refine-assignments` skill's L1 session creates `l0_prompts/*.txt` files that embed the **full 84,075-char task description** (~21K tokens) under "## TASK DESCRIPTION (summary)". The L1 then reads 6 of these prompt files back into its own context before spawning L0 subagents — adding ~91K tokens of duplicated task content to the L1 conversation.

This is the single largest contributor to the context exhaustion observed in `run-20260503-001534-86a9b41b`, where the L1 accumulated ~794K tokens (4× the 200K Sonnet window). The task description alone, replicated across file reads, consumed ~191K tokens of the 200K window — before any L0 responses returned or any assignment processing occurred.

### Key data from the failing run

| Source of task desc in L1 context | Reads | Tokens per read | Total tokens |
|---|---|---|---|
| `combined_assignments.json` (1st read) | 1 | ~50,557 | ~50,557 |
| `combined_assignments.json` (2nd read — duplicate) | 1 | ~50,557 | ~50,557 |
| `l0_prompts/*.txt` (6 files read back) | 6 | ~15,250 | ~91,500 |
| **Subtotal: L1 input from file reads** | | | **~191K tokens** |

### Important: these are model-created runtime artifacts

`l0_prompts/*.txt` and `l0_contexts/*.json` are created by the model at runtime during the L1 session. There is **no Python template or code** generating them. The fix is entirely a **SKILL.md edit** — changing the instructions that tell the L1 what to put in the L0 prompts.

## Required Changes

In `src/autoskillit/skills_extended/planner-refine-assignments/SKILL.md`:

1. **Step 2 (Build L0 context packets)**: Change the instruction to reference the task via `${{ context.task_file_path }}` file path instead of embedding the full content inline. L0s read the task from file if they need it.

2. **Step 3 (Spawn parallel L0 subagents)**: Add instruction that the L1 must NOT read `l0_prompts/*.txt` files back into its own context. The L1 should spawn L0s with the context packet directly — the `l0_contexts/*.json` files are already well-filtered at ~8K tokens each (500-char task summary, 3.5KB target assignment, 1.3KB phase context, 23KB peer summaries).

3. **Add to NEVER block**: "Read `l0_prompts/` files back into the L1 context after writing them"

### Context: per-L0 design is already correct

[...]

PR #1675, opened 2026-05-03
Rename merge_tier_dir → merge_tier_results and Extend to Produce Per-Phase Refine Context Files

### LINKED ISSUE #1646: Rename merge_tier_dir → merge_tier_results and extend to produce per-phase refine context files
## Problem

The `merge_tier_dir` function in `src/autoskillit/planner/merge.py` has two issues:

1. **Bad name**: `merge_tier_dir` looks like a directory path, not a function name. Rename to `merge_tier_results`.

2. **Missing per-phase split**: The `merge_assignments` recipe step calls this function to combine per-phase elaboration outputs into `combined_assignments.json`. The downstream `refine_assignments` step then reads this 202K combined document into a single L1 session — causing context exhaustion at ~794K tokens (4× the 200K window). The merge callable should also produce per-phase context files that the refine step can pick up via `dispatch_items`, avoiding a separate orchestrator turn.

### 1. Rename `merge_tier_dir` → `merge_tier_results`

All callsites must be updated in a single commit:

| File | What to update |
|---|---|
| `src/autoskillit/planner/merge.py` | Function definition (line 165) |
| `src/autoskillit/planner/__init__.py` | Import + `__all__` export |
| `src/autoskillit/recipes/planner.yaml` | 3 callable strings: `merge_phases` (line 135), `merge_assignments` (line 189), `merge_wps` (line 257) |
| `src/autoskillit/recipe/skill_contracts.yaml` | Contract entry (line 2205) |
| `tests/planner/test_planner.py` | Imports + test functions (`test_merge_tier_dir_importable`, `test_merge_tier_dir_globs_and_merges`) |
| `tests/planner/test_merge.py` | Import + test functions (`test_merge_tier_dir_empty_dir_raises`, `test_merge_tier_dir_single_file`, `test_merge_tier_dir_reads_task_from_file`) |
| `tests/recipe/test_planner_recipe.py` | Assertion string in `test_merge_steps_use_merge_tier_dir` (line 245/248) |

### 2. Extend `merge_tier_results` to produce per-phase refine context files

When `key == "assignments"`, after writing `combined_assignments.json`, the function also writes per-phase context files with graduated visibility:

- **Own phase's assignments in full detail** — everything the L0s need for refinement
- **Other phases' assignments as filtered summaries**: `{id, name, goal}` only — provides "what's before and after you" context without full-document replication
- **Task description via `${{ context.task_file_path }}`** file path reference, NOT inline — avoids the 84K-char task duplication that caused the context exhaustion
- **Writes to `$planner_dir/refine_contexts/`** — known convention path the refine step picks up

This split happens in the same Python execution as the merge — zero additional orchestrator turns, zero cache-read cost.

### Design rationale (from helper_agents planner)

The helper_agents planner (the original design that inspired the autoskillit planner) uses graduated visibility for refinement:
- Each worker sees its own assignment's WPs in full detail
- Other assignments' WPs as filtered summaries
- The whole phase context
- High-level view of other phases (`{phaseId, phaseName, phaseGoal}` only)

The `l0_contexts/*.json` files in the current autoskillit planner already implement this pattern correctly (~8K tokens each with 500-char task summary and filtered peer summaries). The per-phase context files produced by this change follow the same approach at the phase level.

### Return value extension

The function's return dict should include the per-phase context paths so the recipe can capture them:

```python
result["refine_context_paths"] = ",".join(context_file_paths)
```

## Dependencies

Blocked by:
- #1641 Feature gate silently blocks all planner skills in non-dev installs
- #1642 _PATH_CAPTURE regex fails for \S+-terminated patterns
- #1643 Write detection blind spots: fs_writes_detected directory mismatch

## Downstream

- #1645 depends on this for the `${{ context.task_file_path }}` reference in SKILL.md
- The per-phase dispatch issue (filed separately) depends on the per-phase context files produced here

## Validation

From validated investigation report. The rename was explicitly requested. The per-phase split design follows the `elaborate_assignments` / `expand_assignments` pattern already established in the pipeline.

Trecek:
## Summary

Two changes in one commit:

1. **Rename `merge_tier_dir` → `merge_tier_results`** across all eight files listed in the issue. This is a pure identifier rename — no behavior changes.

2. **Extend `merge_tier_results` to produce per-phase refine context files** when `key == "assignments"`. After writing `combined_assignments.json`, the function groups merged assignments by `phase_id`, writes one `context_{phase_id}.json` per phase into `planner_dir/refine_contexts/`, and returns `refine_context_paths` (comma-separated) in its result dict. Each context file contains the own phase's assignments in full detail, other phases' assignments as filtered summaries (`{id, name, goal}` only), and a `task_file_path` reference (not inline task content). The recipe `merge_assignments` step capture dict is updated to capture `refine_context_paths`. The `refine_assignments` step itself is not changed here — consuming the per-phase contexts via `dispatch_items` depends on #1645.

Closes #1646

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260503-013455-376704/.autoskillit/temp/make-plan/rename_merge_tier_dir_and_extend_plan_2026-05-03_120000.md`

## Token Usage Summary

| plan | 2.6k | 21.0k | 802.2k | 59.9k | 1 | 8m 50s |
| verify | 92 | 9.4k | 432.9k | 42.9k | 1 | 4m 53s |
| implement | 2.3k | 16.7k | 2.1M | 63.1k | 1 | 7m 5s |
| prepare_pr | 60 | 4.5k | 226.6k | 30.2k | 1 | 1m 33s |
| compose_pr | 59 | 2.3k | 194.9k | 19.8k | 1 | 42s |
| review_pr | 212 | 60.8k | 1.3M | 138.6k | 2 | 13m 35s |
| resolve_review | 744 | 50.6k | 5.9M | 167.2k | 2 | 22m 53s |
| **Total** | 6.1k | 165.4k | 10.9M | 521.7k | | 59m 31s |
## Token Efficiency

| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| implement | 298 | 7069.6 | 211.8 | 56.1 |
| fix | 1 | 848375.0 | 40550.0 | 5980.0 |
| **Total** | **299** | 15424.2 | 857.9 | 200.6 |

### DIFF
--- a/src/autoskillit/planner/merge.py
+++ b/src/autoskillit/planner/merge.py
@@ -97,6 +97,42 @@ def merge_files(
     return result
 
 
+def _write_refine_contexts(
+    planner_dir: Path,
+    assignments: list[dict[str, Any]],
+    task_file_path: str,
+) -> list[str]:
+    phase_groups: dict[str, list[dict[str, Any]]] = {}
+    for assignment in assignments:
+        phase_id = assignment.get("phase_id", "")
+        if phase_id:
+            phase_groups.setdefault(phase_id, []).append(assignment)
+
+    contexts_dir = planner_dir / "refine_contexts"
+    contexts_dir.mkdir(parents=True, exist_ok=True)
+
+    context_paths: list[str] = []
+    for phase_id in sorted(phase_groups):
+        own = phase_groups[phase_id]
+        peer_summaries: list[dict[str, str]] = [
+            {"id": a.get("id", ""), "name": a.get("name", ""), "goal": a.get("goal", "")}
+            for pid, peers in sorted(phase_groups.items())
+            if pid != phase_id
+            for a in peers
+        ]
+        context: dict[str, Any] = {
+            "phase_id": phase_id,
+            "task_file_path": task_file_path,
+            "assignments": own,
+            "peer_summaries": peer_summaries,
+        }
+        ctx_path = contexts_dir / f"context_{phase_id}.json"
+        write_versioned_json(ctx_path, context, schema_version=1)
+        context_paths.append(str(ctx_path))
+
+    return context_paths
+
+
 def extract_item(
     source_path: str,
     item_id: str,
@@ -160,7 +196,7 @@ def replace_item(
     return {"replaced_id": item_id, "updated_path": str(source_path)}
 
 
-def merge_tier_dir(
+def merge_tier_results(
     results_dir: str,
     output_path: str,
     key: str,
@@ -171,13 +207,20 @@ def merge_tier_dir(
     paths = sorted(Path(results_dir).glob("*_result.json"))
     if not paths:
         raise ValueError(f"No *_result.json files found in {results_dir}")
-    return merge_files(
+    result = merge_files(
         file_paths=[str(p) for p in paths],
         output_path=output_path,
         key=key,
         task_file_path=task_file_path,
         source_dir=source_dir,
     )
+    if key == "assignments":
+        merged_data = json.loads(Path(output_path).read_text(encoding="utf-8"))
+        assignments = merged_data.get("assignments", [])
+        planner_dir = Path(output_path).parent
+        context_paths = _write_refine_contexts(planner_dir, assignments, task_file_path)
+        result["refine_context_paths"] = ",".join(context_paths)
+    return result
 
 
 def build_plan_snapshot(
--- a/src/autoskillit/planner/merge.py
+++ b/src/autoskillit/planner/merge.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import fcntl
 import json
+import re
 from pathlib import Path
 from typing import Any
 
@@ -131,6 +132,11 @@ def _write_refine_contexts(
             "assignments": own,
             "peer_summaries": peer_summaries,
         }
+        if not re.fullmatch(r"[A-Za-z0-9_\-]+", phase_id):
+            raise ValueError(
+                f"phase_id {phase_id!r} contains disallowed characters — "
+                "only alphanumeric, underscore, and hyphen are permitted in context filenames"
+            )
         ctx_path = contexts_dir / f"context_{phase_id}.json"
         write_versioned_json(ctx_path, context, schema_version=1)
         context_paths.append(str(ctx_path))
--- a/src/autoskillit/planner/merge.py
+++ b/src/autoskillit/planner/merge.py
@@ -107,6 +107,11 @@ def _write_refine_contexts(
         phase_id = assignment.get("phase_id", "")
         if phase_id:
             phase_groups.setdefault(phase_id, []).append(assignment)
+        else:
+            logger.warning(
+                "Assignment %r has no phase_id — skipped from refine contexts",
+                assignment.get("id", "<unknown>"),
+            )
 
     contexts_dir = planner_dir / "refine_contexts"
     contexts_dir.mkdir(parents=True, exist_ok=True)
@@ -217,9 +222,14 @@ def merge_tier_results(
     if key == "assignments":
         merged_data = json.loads(Path(output_path).read_text(encoding="utf-8"))
         assignments = merged_data.get("assignments", [])
+        if not assignments:
+            logger.warning(
+                "merge_tier_results: no assignments found in %s — refine contexts will be empty",
+                output_path,
+            )
         planner_dir = Path(output_path).parent
         context_paths = _write_refine_contexts(planner_dir, assignments, task_file_path)
-        result["refine_context_paths"] = ",".join(context_paths)
+        result["refine_context_paths"] = json.dumps(context_paths)
     return result
 
 
--- a/src/autoskillit/recipe/skill_contracts.yaml
+++ b/src/autoskillit/recipe/skill_contracts.yaml
@@ -2224,7 +2224,7 @@ callable_contracts:
       type: file_path
     - name: phase_ids
       type: str
-  autoskillit.planner.merge.merge_tier_dir:
+  autoskillit.planner.merge.merge_tier_results:
     inputs:
     - name: results_dir
       type: directory_path
@@ -2239,6 +2239,8 @@ callable_contracts:
     - name: merged_path
       type: file_path
     - name: item_count
+      type: str
+    - name: refine_context_paths
       type: str
   autoskillit.planner.create_run_dir:
     inputs:
--- a/src/autoskillit/recipe/skill_contracts.yaml
+++ b/src/autoskillit/recipe/skill_contracts.yaml
@@ -2242,6 +2242,7 @@ callable_contracts:
       type: str
     - name: refine_context_paths
       type: str
+      when: "key == 'assignments'"
   autoskillit.planner.create_run_dir:
     inputs:
     - name: temp_dir
--- a/src/autoskillit/recipes/planner.yaml
+++ b/src/autoskillit/recipes/planner.yaml
@@ -201,7 +201,6 @@ steps:
       source_dir: "${{ inputs.source_dir }}"
     capture:
       combined_assignments_path: "${{ result.merged_path }}"
-      refine_context_paths: "${{ result.refine_context_paths }}"
     on_success: refine_assignments
     on_failure: escalate_stop
 
--- a/src/autoskillit/recipes/planner.yaml
+++ b/src/autoskillit/recipes/planner.yaml
@@ -140,7 +140,7 @@ steps:
   merge_phases:
     tool: run_python
     with:
-      callable: "autoskillit.planner.merge.merge_tier_dir"
+      callable: "autoskillit.planner.merge.merge_tier_results"
       results_dir: "${{ context.planner_dir }}/phases"
       output_path: "${{ context.planner_dir }}/combined_plan.json"
       key: "phases"
@@ -193,7 +193,7 @@ steps:
   merge_assignments:
     tool: run_python
     with:
-      callable: "autoskillit.planner.merge.merge_tier_dir"
+      callable: "autoskillit.planner.merge.merge_tier_results"
       results_dir: "${{ context.planner_dir }}/assignments"
       output_path: "${{ context.planner_dir }}/combined_assignments.json"
       key: "assignments"
@@ -201,6 +201,7 @@ steps:
       source_dir: "${{ inputs.source_dir }}"
     capture:
       combined_assignments_path: "${{ result.merged_path }}"
+      refine_context_paths: "${{ result.refine_context_paths }}"
     on_success: refine_assignments
     on_failure: escalate_stop
 
@@ -260,7 +261,7 @@ steps:
   merge_wps:
     tool: run_python
     with:
-      callable: "autoskillit.planner.merge.merge_tier_dir"
+      callable: "autoskillit.planner.merge.merge_tier_results"
       results_dir: "${{ context.planner_dir }}/work_packages"
       output_path: "${{ context.planner_dir }}/combined_wps.json"
       key: "work_packages"
--- a/tests/planner/test_merge.py
+++ b/tests/planner/test_merge.py
@@ -725,3 +725,15 @@ def test_merge_tier_results_no_refine_contexts_for_work_packages_key(tmp_path):
     result = merge_tier_results(str(results_dir), str(out), "work_packages")
     assert "refine_context_paths" not in result
     assert not (tmp_path / "refine_contexts").exists()
+
+
+def test_write_refine_contexts_rejects_unsafe_phase_id(tmp_path):
+    results_dir = tmp_path / "assignments"
+    results_dir.mkdir()
+    write_json(
+        results_dir / "evil_result.json",
+        {"id": "A1", "phase_id": "../../evil", "name": "x", "goal": "g"},
+    )
+    out = tmp_path / "combined.json"
+    with pytest.raises(ValueError, match="disallowed characters"):
+        merge_tier_results(str(results_dir), str(out), "assignments")
--- a/tests/planner/test_merge.py
+++ b/tests/planner/test_merge.py
@@ -557,10 +557,14 @@ def test_merge_tier_results_writes_refine_contexts_for_assignments(tmp_path):
     out = tmp_path / "combined_assignments.json"
     result = merge_tier_results(str(results_dir), str(out), "assignments")
     assert "refine_context_paths" in result
-    paths = result["refine_context_paths"].split(",")
+    paths = json.loads(result["refine_context_paths"])
     assert len(paths) == 2
-    assert (tmp_path / "refine_contexts" / "context_P1.json").exists()
-    assert (tmp_path / "refine_contexts" / "context_P2.json").exists()
+    p1_ctx = tmp_path / "refine_contexts" / "context_P1.json"
+    p2_ctx = tmp_path / "refine_contexts" / "context_P2.json"
+    assert p1_ctx.exists()
+    assert p2_ctx.exists()
+    assert str(p1_ctx) in paths
+    assert str(p2_ctx) in paths
 
 
 def test_refine_context_own_assignments_in_full_detail(tmp_path):
@@ -596,6 +600,7 @@ def test_refine_context_own_assignments_in_full_detail(tmp_path):
     assert own["id"] == "P1-A1"
     assert own["technical_approach"] == "JWT"
     assert own["proposed_work_packages"] == [{"id": "wp1"}]
+    assert all(a["id"] != "P2-A1" for a in ctx["assignments"])
 
 
 def test_refine_context_peer_summaries_have_filtered_fields(tmp_path):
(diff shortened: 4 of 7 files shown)


src/autoskillit/planner/merge.py
```diff
@@ -97,6 +97,42 @@ def merge_files(
...
 
+def _write_refine_contexts(
+    planner_dir: Path,
+    assignments: list[dict[str, Any]],
+    task_file_path: str,
+) -> list[str]:
+    phase_groups: dict[str, list[dict[str, Any]]] = {}
+    for assignment in assignments:
+        phase_id = assignment.get("phase_id", "")
+        if phase_id:
```
Trecek, line 109: [warning] bugs: Assignments whose phase_id is missing or empty string are silently skipped (the 'if phase_id:' guard). Such assignments never appear in any context file and are absent from refine_context_paths with no error or warning.

src/autoskillit/planner/merge.py
```diff
@@ -171,13 +207,20 @@ def merge_tier_dir(
...
+    result = merge_files(
         file_paths=[str(p) for p in paths],
         output_path=output_path,
         key=key,
         task_file_path=task_file_path,
         source_dir=source_dir,
     )
+    if key == "assignments":
+        merged_data = json.loads(Path(output_path).read_text(encoding="utf-8"))
+        assignments = merged_data.get("assignments", [])
```
Trecek, line 230: [critical] bugs: merged_data.get('assignments', []) silently returns [] if the key is absent or the merged JSON structure differs from expectations. _write_refine_contexts still creates the contexts_dir and returns [], so result['refine_context_paths'] becomes an empty string with no signal that something went wrong. Validate that assignments is non-empty before proceeding.

src/autoskillit/planner/merge.py
```diff
@@ -171,13 +207,20 @@ def merge_tier_dir(
...
-    return merge_files(
+    result = merge_files(
         file_paths=[str(p) for p in paths],
         output_path=output_path,
         key=key,
         task_file_path=task_file_path,
         source_dir=source_dir,
     )
+    if key == "assignments":
+        merged_data = json.loads(Path(output_path).read_text(encoding="utf-8"))
```
Trecek, line 229: [warning] arch: merge_tier_results re-reads output_path from disk immediately after merge_files writes it. This round-trip is fragile if the write is buffered or fails silently. merge_files already returns the merged data; pass that return value directly to _write_refine_contexts instead of re-reading from disk.

src/autoskillit/planner/merge.py
```diff
@@ -171,13 +207,20 @@ def merge_tier_dir(
...
+        assignments = merged_data.get("assignments", [])
+        planner_dir = Path(output_path).parent
+        context_paths = _write_refine_contexts(planner_dir, assignments, task_file_path)
+        result["refine_context_paths"] = ",".join(context_paths)
```
Trecek: [warning] defense: Joining context paths with a comma (','.join(context_paths)) is ambiguous — commas are valid in Linux paths. Return a list or JSON-encode the array; the skill_contracts.yaml type should be updated accordingly.

src/autoskillit/recipe/skill_contracts.yaml
```diff
@@ -2240,6 +2240,8 @@ callable_contracts:
       type: file_path
     - name: item_count
       type: str
+    - name: refine_context_paths
+      type: str
```
Trecek, line 2244: [warning] cohesion: Contract declares 'refine_context_paths' with type 'str' but the field name implies a collection. Using a delimiter-joined string breaks symmetry with how multi-value outputs are typed elsewhere in the contract schema.

tests/planner/test_merge.py
```diff
@@ -521,3 +522,200 @@ def test_merge_files_reads_task_from_task_file_path(tmp_path) -> None:
...
+            "phase_id": "P2",
+            "name": "Data",
+            "goal": "Data goal",
+            "technical_approach": "ORM",
+            "proposed_work_packages": [],
+        },
+    )
+    out = tmp_path / "combined_assignments.json"
+    result = merge_tier_results(str(results_dir), str(out), "assignments")
+    assert "refine_context_paths" in result
```
Trecek, line 559: [warning] tests: test_merge_tier_results_writes_refine_contexts_for_assignments does not verify the actual path values correspond to the correct phase context files. The comma-split could silently pass with wrong path content.

tests/planner/test_merge.py
```diff
@@ -521,3 +522,200 @@ def test_merge_files_reads_task_from_task_file_path(tmp_path) -> None:
...
+            "name": "Data",
+            "goal": "Data goal",
+            "technical_approach": "ORM",
+            "proposed_work_packages": [],
+        },
+    )
+    out = tmp_path / "combined_assignments.json"
+    merge_tier_results(str(results_dir), str(out), "assignments")
+    ctx = json.loads((tmp_path / "refine_contexts" / "context_P1.json").read_text())
+    assert len(ctx["assignments"]) == 1
```
Trecek, line 598: [warning] tests: test_refine_context_own_assignments_in_full_detail asserts len(ctx['assignments'])==1 but does not assert that peer assignments (P2-A1) are absent from ctx['assignments'], leaving a gap where extra items could silently appear.

tests/planner/test_merge.py
```diff
@@ -521,3 +522,200 @@ def test_merge_files_reads_task_from_task_file_path(tmp_path) -> None:
...
+                "name": pid,
+                "goal": f"{pid} goal",
+                "technical_approach": "",
+                "proposed_work_packages": [],
+            },
+        )
+    out = tmp_path / "combined_assignments.json"
+    result = merge_tier_results(str(results_dir), str(out), "assignments")
+    paths = result["refine_context_paths"].split(",")
+    phase_ids = [Path(p).stem.replace("context_", "") for p in paths]
```
Trecek, line 681: [warning] tests: test_refine_context_paths_returned_sorted only checks that phase_ids are sorted but does not verify all three expected phase IDs (P1, P2, P3) are present. A result with fewer paths could still pass the sorted check.

tests/planner/test_planner.py
```diff
@@ -136,10 +136,10 @@ def test_planner_feature_skill_categories() -> None:
 # --- T6: New callables importable ---
 
 
-def test_merge_tier_dir_importable() -> None:
-    from autoskillit.planner.merge import merge_tier_dir
+def test_merge_tier_results_importable() -> None:
```
Trecek: [warning] slop: test_merge_tier_results_importable only asserts callable(merge_tier_results), which is trivially true for any imported function and adds no meaningful coverage beyond what test_merge_tier_results_globs_and_merges (L160) already provides.

Trecek (commented):

Trecek (commented):
AutoSkillit review: blocking issue found — see inline comments.

**1 critical finding** requires a fix before merge:
- `src/autoskillit/planner/merge.py` L219: `merged_data.get('assignments', [])` silently returns an empty list when the key is absent, producing an empty `refine_context_paths` string with no signal of failure.

**8 warning findings** (see inline comments): disk round-trip fragility in `merge_tier_results`, comma-joined paths encoding, silent phase_id skip, cohesion contract type mismatch, and several weak test assertions.

src/autoskillit/planner/merge.py
Trecek: Agreed — fixed in 33c2edf1. Added `logger.warning` in `_write_refine_contexts` at L108 when an assignment's `phase_id` is missing or empty string, so skipped assignments are now visible in logs.

src/autoskillit/planner/merge.py
Trecek: Agreed — fixed in 33c2edf1. Added a `logger.warning` in `merge_tier_results` when `assignments` is empty after the merge, so silent no-ops are now logged clearly.

src/autoskillit/planner/merge.py
Trecek: Investigated — this is intentional. `merge_files()` returns `{"merged_path": str, "item_count": str}` (merge.py L89-97), NOT the merged JSON content. The reviewer's claim that "merge_files already returns the merged data" is factually incorrect. The disk re-read is the correct pattern here: `merge_files` holds an exclusive file lock (LOCK_EX) and calls `fh.flush()` before releasing it (L84-87), so the data is reliably on disk. commit_sha_hint: 3b9623bc

src/autoskillit/planner/merge.py
Trecek: Agreed — fixed in 33c2edf1. Changed `,`.join(context_paths)` to `json.dumps(context_paths)` so the value is an unambiguous JSON-encoded list, eliminating the comma-in-path ambiguity.

src/autoskillit/recipe/skill_contracts.yaml
Trecek: Agreed — addressed in 33c2edf1. Switching from comma-join to `json.dumps(context_paths)` (see L222 fix) makes the value an unambiguous JSON-encoded list. The contract type `str` remains technically correct since the field value is still a string (JSON-serialized), consistent with other str-typed outputs in the schema.

tests/planner/test_merge.py
Trecek: Agreed — fixed in 7ac2f0b7. Test now uses `json.loads()` to parse `refine_context_paths` and explicitly asserts that `str(p1_ctx)` and `str(p2_ctx)` are both present in the returned list, verifying path correspondence.

tests/planner/test_merge.py
Trecek: Agreed — fixed in 7ac2f0b7. Added `assert all(a["id"] != "P2-A1" for a in ctx["assignments"])` to explicitly verify that no peer assignments appear in the P1 context's assignments list.

tests/planner/test_merge.py
Trecek: Agreed — fixed in 7ac2f0b7. Added `assert set(phase_ids) == {"P1", "P2", "P3"}` to verify all three expected phase context files are present in the returned list, not just that the ordering is correct.

tests/planner/test_planner.py
Trecek: Agreed — fixed in 7ac2f0b7. Removed `test_merge_tier_results_importable` (callable check only). The functional test `test_merge_tier_results_globs_and_merges` at L160 already exercises the import path with meaningful assertions.

src/autoskillit/planner/merge.py
```diff
@@ -97,6 +97,47 @@ def merge_files(
     return result
 
 
+def _write_refine_contexts(
```
Trecek, line 101: [warning] arch: _write_refine_contexts performs filesystem I/O (directory creation, file writes) directly inside planner/merge.py. This couples merge/transform logic to file-system side-effects that belong in an I/O or persistence layer, violating domain separation.

src/autoskillit/planner/merge.py
```diff
@@ -97,6 +97,47 @@ def merge_files(
...
+            if pid != phase_id
+            for a in peers
+        ]
+        context: dict[str, Any] = {
+            "phase_id": phase_id,
+            "task_file_path": task_file_path,
+            "assignments": own,
+            "peer_summaries": peer_summaries,
+        }
+        ctx_path = contexts_dir / f"context_{phase_id}.json"
```
Trecek, line 140: [warning] defense: phase_id is used directly in the filename via f"context_{phase_id}.json" without sanitization. If phase_id contains path separators or null bytes, this produces an invalid or unexpected file path.

src/autoskillit/planner/merge.py
```diff
@@ -160,7 +201,7 @@ def replace_item(
     return {"replaced_id": item_id, "updated_path": str(source_path)}
 
 
-def merge_tier_dir(
+def merge_tier_results(
     results_dir: str,
```
Trecek, line 211: [info] defense: results_dir and output_path are typed as str rather than Path. Public boundary parameters representing filesystem paths should prefer Path to enable type-safe operations.

src/autoskillit/planner/merge.py
```diff
@@ -171,13 +212,25 @@ def merge_tier_dir(
...
+        merged_data = json.loads(Path(output_path).read_text(encoding="utf-8"))
+        assignments = merged_data.get("assignments", [])
+        if not assignments:
+            logger.warning(
+                "merge_tier_results: no assignments found in %s — refine contexts will be empty",
+                output_path,
+            )
+        planner_dir = Path(output_path).parent
+        context_paths = _write_refine_contexts(planner_dir, assignments, task_file_path)
+        result["refine_context_paths"] = json.dumps(context_paths)
```
Trecek, line 238: [warning] bugs: result["refine_context_paths"] is set to a JSON-serialized string (json.dumps(context_paths)) rather than a plain list. Callers expecting a list will receive a string and must double-deserialize, producing an inconsistency where other keys hold native Python values.

src/autoskillit/planner/merge.py
```diff
@@ -97,6 +97,47 @@ def merge_files(
...
+
+    contexts_dir = planner_dir / "refine_contexts"
+    contexts_dir.mkdir(parents=True, exist_ok=True)
+
+    context_paths: list[str] = []
+    for phase_id in sorted(phase_groups):
+        own = phase_groups[phase_id]
+        peer_summaries: list[dict[str, str]] = [
+            {"id": a.get("id", ""), "name": a.get("name", ""), "goal": a.get("goal", "")}
+            for pid, peers in sorted(phase_groups.items())
```
Trecek, line 125: [warning] bugs: peer_summaries is built via sorted(phase_groups.items()), producing deterministic ordering. However, if downstream consumers depend on peer ordering by some other key (e.g., phase index), this lexicographic sort may silently produce unexpected ordering. Confirm intended sort key.

src/autoskillit/recipes/planner.yaml
```diff
@@ -140,7 +140,7 @@ steps:
   merge_phases:
     tool: run_python
     with:
-      callable: "autoskillit.planner.merge.merge_tier_dir"
+      callable: "autoskillit.planner.merge.merge_tier_results"
```
Trecek, line 143: [info] cohesion: All three call sites use merge_tier_results but only key='assignments' produces refine_context_paths. Consumers of merge_phases and merge_wps will silently receive no refine_context_paths output.

src/autoskillit/recipe/skill_contracts.yaml
```diff
@@ -2224,7 +2224,7 @@ callable_contracts:
       type: file_path
     - name: phase_ids
       type: str
-  autoskillit.planner.merge.merge_tier_dir:
+  autoskillit.planner.merge.merge_tier_results:
```
Trecek, line 2227: [warning] cohesion: The contract for autoskillit.planner.merge.merge_tier_results advertises refine_context_paths as an output unconditionally, but the implementation only populates it when key='assignments'. The contract is inconsistent with the function's conditional behaviour.

Trecek (commented):

Trecek (commented):

src/autoskillit/planner/merge.py
Trecek: Investigated — this is intentional. `merge.py` is a file-I/O merge module by design — `merge_files` (L16), `build_plan_snapshot` (L236), and `extract_item` (L141) all perform filesystem I/O directly. Every public function in this module reads or writes files; extracting `_write_refine_contexts` I/O would be inconsistent with the established pattern and require architectural restructuring beyond PR scope. The module's stated purpose is JSON interchange merge tooling.

src/autoskillit/planner/merge.py
Trecek: Agreed — fixed in 01ceed36. Added `re.fullmatch(r'[A-Za-z0-9_\-]+', phase_id)` guard before interpolating `phase_id` into the context filename; raises `ValueError` with a descriptive message for any value containing path separators, null bytes, or other disallowed characters. Added `test_write_refine_contexts_rejects_unsafe_phase_id` to cover the path traversal case.

src/autoskillit/planner/merge.py
Trecek: Acknowledged — minor suggestion noted.

src/autoskillit/planner/merge.py
Trecek: Investigated — this is intentional. The JSON encoding was the deliberate fix from the prior review round (commit 33c2edf1), which replaced `,`.join with `json.dumps` to eliminate comma-in-path ambiguity (a finding from the same review). The contract at `skill_contracts.yaml` explicitly declares `type: str`, and tests already use `json.loads()` to parse it (commit 7ac2f0b7). The recipe's `merge_assignments` step does not capture `refine_context_paths` as a context variable — the downstream skill finds context files via `planner_dir` — so the double-decode concern has no active caller.

src/autoskillit/planner/merge.py
Trecek: Investigated — this is intentional. Deterministic ordering via `sorted()` is a feature, not a bug — it ensures stable, reproducible context files across runs. The downstream consumer is the `planner-refine-assignments` skill, which passes `peer_summaries` as LLM context; LLMs are not sensitive to peer list ordering. No contract, test, or skill spec requires a different ordering key. The reviewer's concern is speculative with no concrete failure mode.

src/autoskillit/recipes/planner.yaml
Trecek: Acknowledged — minor suggestion noted.

src/autoskillit/recipe/skill_contracts.yaml
Trecek: Agreed — fixed in 7319d587. Added `when: "key == 'assignments'"` annotation to the `refine_context_paths` output entry in `skill_contracts.yaml`. The field is now explicitly documented as conditional — callers using `key='phases'` or `key='work_packages'` will not receive it. The `when` key is silently ignored by the contract loader (which only reads `name` and `type`), so no validation logic breaks.

PR #1676, opened 2026-05-03
Rectify: Write Detection Architectural Immunity

### LINKED ISSUE #1643: Write detection blind spots: fs_writes_detected directory mismatch + write_call_count blind to Bash writes
## Summary

The headless session write detection system has two compounding blind spots that cause `zero_writes` false positives: (1) `fs_writes_detected` monitors a skill-name-derived directory that doesn't match the recipe-level output directory used by pipeline recipes, and (2) `write_call_count` is blind to Bash-mediated writes (4th+ documented occurrence of this pattern). Together, these caused the `planner-refine-phases` step to be falsely classified as `zero_writes` despite successfully writing `refined_plan.json` and correctly emitting the `refined_plan_path` structured output token.

### Mechanism

`_resolve_skill_temp_dir` (`headless.py:152-157`) computes the watch path from the **skill name**:
```python
def _resolve_skill_temp_dir(cwd: str, skill_command: str) -> Path | None:
    name = extract_skill_name(skill_command)
    return Path(cwd) / ".autoskillit" / "temp" / name
```

For `skill_command="/autoskillit:planner-refine-phases ..."`, this produces:
`.autoskillit/temp/planner-refine-phases/`

But `create_run_dir` (`planner/manifests.py:41`) hardcodes `"planner"` as the subdirectory:
```python
run_dir = Path(temp_dir) / "planner" / f"run-{stamp}-{secrets.token_hex(4)}"
```

So the actual write destination is: `.autoskillit/temp/planner/run-20260503-001534-86a9b41b/`

These are entirely different directories. The skill-named directory (`.autoskillit/temp/planner-refine-phases/`) does not exist — confirmed absent from disk. Since `_skill_temp_dir.is_dir()` returns `False` (`headless.py:366`), the pre/post snapshot diff is never computed, and `fs_writes_detected` stays `False`.

### Blast Radius

This affects ALL steps in the planner recipe. Every `planner-*` skill writes to `.autoskillit/temp/planner/run-{ts}/`, but `_resolve_skill_temp_dir` monitors `.autoskillit/temp/planner-{skill-name}/` for each. No planner skill ever writes to the monitored directory.

Other recipes (research, implementation, merge-prs, remediation) use per-skill output paths like `{{AUTOSKILLIT_TEMP}}/review-pr/`, `{{AUTOSKILLIT_TEMP}}/make-plan/`, etc. — each scoped to the skill name, matching what `_resolve_skill_temp_dir` expects. The directory mismatch is currently **planner-specific**.

### Mechanism

`_headless_result.py:272`:
```python
write_call_count = sum(1 for t in session.tool_uses if t.get("name") in {"Write", "Edit"})
```

Only `Write` and `Edit` tool_use blocks are counted. Bash tool calls that write files via `python3 -c`, heredocs, `echo >`, `tee`, `cat >`, `cp`, `mv` etc. are invisible.

### Evidence from Session bbb7d1a4 (refine-phases)

- Session wrote `refined_plan.json` via a **Bash tool call (JSONL line 60)** containing an embedded `python3 << 'EOF'` heredoc that called `json.dump()`
- Zero Write/Edit tool calls in the session → `write_call_count=0`
- The `refined_plan_path = /abs/path/refined_plan.json` token WAS correctly emitted in the final assistant message (JSONL line 65)
- `_compute_outcome` returned `SUCCEEDED` (patterns matched, marker present)
- But the zero_writes gate (`_headless_result.py:515-531`) then demoted success to failure because `_has_write_evidence` was `False`

### Recurrence History

This is the **4th+ documented occurrence** of write_call_count blind spots (issues #603, #616, #710, #1264). PR #1264 (2026-04-25) introduced `fs_writes_detected` as a mitigation, but it has its own blind spot (Root Cause A above). The cycle: new skill uses Bash to write → `write_call_count=0` → `zero_writes` false positive → investigation + targeted fix → next skill repeats.

7 write detection blind spots mapped during investigation:
1. Bash tool writes (echo, tee, cp, python3 -c) — `write_call_count=0`
2. MCP tool calls that write files (run_cmd, run_python) — `write_call_count=0`
3. Writes outside `{cwd}/.autoskillit/temp/{skill_name}/` — `fs_writes_detected=False`
4. Writes inside pre-existing subdirs of skill temp dir — `fs_writes_detected=False` (shallow `os.scandir`, no recursive walk)
5. Agent subagent writes (not in parent JSONL stdout) — both signals=0/False
6. `write_path_warnings` not wired to adjudication — non-blocking, warnings only
7. `type=user` JSONL records not scanned for write evidence

### For Root Cause A (directory mismatch):

Accept an optional `write_watch_dir` parameter in `_execute_claude_headless` that overrides `_resolve_skill_temp_dir`. Pipeline recipes can pass their actual output directory (e.g., `.autoskillit/temp/planner/`) via the recipe step's `with_args`. The recipe already knows the `planner_dir` from `create_run_dir`.

Alternatively, make `_resolve_skill_temp_dir` smarter: when the skill-named directory doesn't exist, check for a prefix match in parent dir (e.g., `.autoskillit/temp/planner*/`).

**Note:** If `write_watch_dir` is set to the parent directory (`.autoskillit/temp/planner/`), concurrent planner sessions from parallel pipelines could produce false positives (write evidence from a sibling session). Mitigate by watching the specific `run-{ts}/` directory passed through recipe context.

### For Root Cause B (Bash write counting):

Two approaches (not mutually exclusive):

**Option 1 — Scan Bash tool_use for write indicators:** In the `write_call_count` computation, also count Bash tool_use blocks whose `command` field contains write-indicating patterns (`>`, `>>`, `tee`, `json.dump`, `open(.*['\"]w`, `cp `, `mv `). This extends the existing signal rather than adding a new one.

**Option 2 — Wire `_scan_jsonl_write_paths` into `_has_write_evidence`:** `_scan_jsonl_write_paths` (`_headless_scan.py:74-87`) already parses Bash tool_use blocks and extracts absolute paths from commands. Currently it only produces warnings for paths **outside** cwd. Extend it to also return a boolean for whether Bash commands wrote to paths **within** cwd, and feed that into `_has_write_evidence`. This requires adding an in-cwd detection path to the existing function (currently it only checks `not path.startswith(cwd_prefix)`).

### Root Cause A
- `src/autoskillit/execution/headless.py:152-157` — `_resolve_skill_temp_dir` (primary fix)
- `src/autoskillit/execution/headless.py:254-260,365-371` — `fs_writes_detected` pre/post snapshot
- `src/autoskillit/planner/manifests.py:41` — `create_run_dir` hardcodes `"planner"` (context)

### Root Cause B
- `src/autoskillit/execution/_headless_result.py:272` — `write_call_count` computation (primary fix)
- `src/autoskillit/execution/_headless_scan.py:74-87` — Bash path scanning (Option 2 target)
- `src/autoskillit/execution/_headless_result.py:273` — `_has_write_evidence` fusion

## Test Gaps

1. No test for `fs_writes_detected` when skill name does not match the recipe output directory
2. No test for `write_call_count` when the session writes exclusively via Bash tool calls AND the write target is outside the skill-named temp dir
3. Test `test_zero_write_detection.py:606` explicitly documents the Bash blind spot as known but only tests the `fs_writes_detected` fallback for writes inside the skill temp dir — no test for the directory mismatch case

## Investigation Reference

Full investigation report: `.autoskillit/temp/investigate/investigation_headless_adjudication_instability_2026-05-02_183000.md`

Trecek:
## Summary

Part A introduced the `WriteEvidence` dataclass, centralized `WRITE_CAPABLE_TOOLS` constant, and Bash in-cwd write detection. Part B completes the architectural immunity by addressing the directory mismatch root cause: skills that write to directories not derived from their name (the planner pattern).

Part B introduces: (1) explicit `write_watch_dirs` parameter on the headless executor so callers can declare where to watch, (2) recipe-step `output_dir` convention wired into the fs snapshot, (3) `HeadlessExecutor` protocol update, (4) Bash file path enrichment in `parse_session_result`, and (5) skill-name regex consolidation.

Closes #1643

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260503-012345-134482/.autoskillit/temp/rectify/rectify_write_detection_architectural_immunity_2026-05-02_192858_part_b.md`

## Token Usage Summary

| implement | 117 | 30.0k | 8.0M | 131.7k | 1 | 11m 48s |
| prepare_pr | 60 | 6.2k | 239.2k | 34.6k | 1 | 2m 3s |
| compose_pr | 59 | 2.1k | 194.7k | 19.7k | 1 | 46s |
| review_pr | 242 | 62.1k | 1.7M | 169.4k | 2 | 14m 52s |
| resolve_review | 922 | 78.8k | 11.7M | 236.7k | 2 | 35m 12s |
| **Total** | 1.4k | 179.2k | 21.9M | 592.2k | | 1h 4m |
## Token Efficiency

| rectify | 0 | — | — | — |
| implement | 833 | 32295.3 | 781.6 | 151.0 |
| assess | 167 | 122354.1 | 1366.3 | 410.4 |
| retry_worktree | 402 | 35957.4 | 1338.0 | 171.4 |
| **Total** | **1402** | 48740.1 | 1226.9 | 240.1 |

### DIFF
--- a/src/autoskillit/core/_type_helpers.py
+++ b/src/autoskillit/core/_type_helpers.py
@@ -32,8 +32,12 @@ __all__ = [
     "truncate_text",
 ]
 
-_SKILL_CMD_RE = re.compile(r"^/(?:autoskillit:)?([\w-]+)")
-_SKILL_RESOLVE_RE = re.compile(r"/(?:autoskillit:)?([\w-]+)")
+_SKILL_CMD_RE = re.compile(
+    r"^/(?:autoskillit:)?([\w-]+)"
+)  # anchored: strict leading-slash for extraction
+_SKILL_RESOLVE_RE = re.compile(
+    r"/(?:autoskillit:)?([\w-]+)"
+)  # unanchored: supports "Use /..." prefix forms
 
 _PATH_PREFIXES: tuple[str, ...] = ("/", "./", ".autoskillit/")
 
--- a/src/autoskillit/core/_type_helpers.py
+++ b/src/autoskillit/core/_type_helpers.py
@@ -26,6 +26,7 @@ __all__ = [
     "extract_path_arg",
     "extract_skill_name",
     "fleet_error",
+    "resolve_skill_name",
     "resolve_target_skill",
     "session_type",
     "truncate_text",
@@ -67,6 +68,25 @@ def extract_skill_name(skill_command: str) -> str | None:
     """
     m = _SKILL_CMD_RE.match(skill_command.strip())
     return m.group(1) if m else None
+
+
+def resolve_skill_name(skill_command: str) -> str | None:
+    """Extract and validate skill name from command string.
+
+    Handles both ``/name`` and ``/autoskillit:name`` forms. Returns None if
+    no match, name contains template expressions, or is followed by a
+    bash-style ``{placeholder}`` token.
+    """
+    stripped = skill_command.strip()
+    match = _SKILL_CMD_RE.match(stripped)
+    if not match:
+        return None
+    name = match.group(1)
+    if "${{" in name:
+        return None
+    if match.end() < len(stripped) and stripped[match.end()] == "{":
+        return None
+    return name
 
 
 def resolve_target_skill(
--- a/src/autoskillit/core/_type_helpers.py
+++ b/src/autoskillit/core/_type_helpers.py
@@ -84,7 +84,8 @@ def resolve_skill_name(skill_command: str) -> str | None:
     name = match.group(1)
     if "${{" in name:
         return None
-    if match.end() < len(stripped) and stripped[match.end()] == "{":
+    remainder = stripped[match.end() :]
+    if remainder.startswith("{") or remainder.startswith("${{"):
         return None
     return name
 
--- a/src/autoskillit/core/_type_helpers.py
+++ b/src/autoskillit/core/_type_helpers.py
@@ -33,6 +33,7 @@ __all__ = [
 ]
 
 _SKILL_CMD_RE = re.compile(r"^/(?:autoskillit:)?([\w-]+)")
+_SKILL_RESOLVE_RE = re.compile(r"/(?:autoskillit:)?([\w-]+)")
 
 _PATH_PREFIXES: tuple[str, ...] = ("/", "./", ".autoskillit/")
 
@@ -78,7 +79,7 @@ def resolve_skill_name(skill_command: str) -> str | None:
     bash-style ``{placeholder}`` token.
     """
     stripped = skill_command.strip()
-    match = _SKILL_CMD_RE.match(stripped)
+    match = _SKILL_RESOLVE_RE.search(stripped)
     if not match:
         return None
     name = match.group(1)
--- a/src/autoskillit/execution/_session_model.py
+++ b/src/autoskillit/execution/_session_model.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import re
 from dataclasses import dataclass, field
 from enum import StrEnum
 from typing import Any, assert_never
@@ -16,6 +17,8 @@ from autoskillit.core import (
 )
 
 logger = get_logger(__name__)
+
+_ABS_PATH_RE: re.Pattern[str] = re.compile(r'(?:^|[\s="\'])(/(?:[a-zA-Z0-9._/~@+:-]+))')
 
 _TOKEN_FIELDS = (
     "input_tokens",
@@ -301,6 +304,16 @@ def parse_session_result(stdout: str) -> ClaudeSessionResult:
                                     fp = block["input"].get("file_path", "")
                                     if fp:
                                         entry["file_path"] = fp
+                                elif name == "Bash" and isinstance(block.get("input"), dict):
+                                    command = block["input"].get("command", "")
+                                    if isinstance(command, str):
+                                        paths = [
+                                            m.group(1)
+                                            for m in _ABS_PATH_RE.finditer(command)
+                                            if len(m.group(1)) >= 5
+                                        ]
+                                        if paths:
+                                            entry["bash_paths"] = paths  # type: ignore[assignment]
                                 acc.tool_uses.append(entry)
                         text = "\n".join(
                             block.get("text", "") for block in content if isinstance(block, dict)
--- a/src/autoskillit/execution/_session_model.py
+++ b/src/autoskillit/execution/_session_model.py
@@ -18,7 +18,7 @@ from autoskillit.core import (
 
 logger = get_logger(__name__)
 
-_ABS_PATH_RE: re.Pattern[str] = re.compile(r'(?:^|[\s="\'])(/(?:[a-zA-Z0-9._/~@+:-]+))')
+_ABS_PATH_RE: re.Pattern[str] = re.compile(r'(?:^|[\s="\'])(/(?:[a-zA-Z0-9._/~@+-]+))')
 
 _TOKEN_FIELDS = (
     "input_tokens",
@@ -297,7 +297,10 @@ def parse_session_result(stdout: str) -> ClaudeSessionResult:
                         for block in content:
                             if isinstance(block, dict) and block.get("type") == "tool_use":
                                 name = block.get("name", "")
-                                entry: dict[str, str] = {"name": name, "id": block.get("id", "")}
+                                entry: dict[str, str | list[str]] = {
+                                    "name": name,
+                                    "id": block.get("id", ""),
+                                }
                                 if name in {"Write", "Edit"} and isinstance(
                                     block.get("input"), dict
                                 ):
@@ -313,7 +316,7 @@ def parse_session_result(stdout: str) -> ClaudeSessionResult:
                                             if len(m.group(1)) >= 5
                                         ]
                                         if paths:
-                                            entry["bash_paths"] = paths  # type: ignore[assignment]
+                                            entry["bash_paths"] = paths
                                 acc.tool_uses.append(entry)
                         text = "\n".join(
                             block.get("text", "") for block in content if isinstance(block, dict)
--- a/src/autoskillit/execution/headless.py
+++ b/src/autoskillit/execution/headless.py
@@ -266,6 +266,7 @@ async def _execute_claude_headless(
             try:
                 _temp_snapshots_pre[_wd] = {e.name for e in os.scandir(_wd)}
             except OSError:
+                logger.warning("watch_dir_pre_scan_failed", watch_dir=str(_wd), exc_info=True)
                 _temp_snapshots_pre[_wd] = set()
 
     _pre_session_sha = _capture_git_head_sha(cwd)
@@ -379,6 +380,7 @@ async def _execute_claude_headless(
             try:
                 _post = {e.name for e in os.scandir(_wd)}
             except OSError:
+                logger.warning("watch_dir_post_scan_failed", watch_dir=str(_wd), exc_info=True)
                 _post = set()
             _pre = _temp_snapshots_pre.get(_wd, set())
             if _post - _pre:
--- a/src/autoskillit/execution/headless.py
+++ b/src/autoskillit/execution/headless.py
@@ -205,6 +205,7 @@ async def _execute_claude_headless(
     on_spawn: Callable[[int, int], None] | None = None,
     skip_clone_guard: bool = False,
     readonly_skill: bool = False,
+    write_watch_dirs: Sequence[Path] = (),
 ) -> SkillResult:
     """Shared subprocess execution for headless Claude sessions.
 
@@ -253,13 +254,19 @@ async def _execute_claude_headless(
     ):
         _clone_snapshot = await snapshot_clone_state(cwd, runner)
 
-    _skill_temp_dir = _resolve_skill_temp_dir(cwd, skill_command)
-    _temp_snapshot_pre: set[str] = set()
-    if _skill_temp_dir and _skill_temp_dir.is_dir():
-        try:
-            _temp_snapshot_pre = {e.name for e in os.scandir(_skill_temp_dir)}
-        except OSError:
-            _temp_snapshot_pre = set()
+    _watch_dirs: list[Path] = list(write_watch_dirs) if write_watch_dirs else []
+    if not _watch_dirs:
+        _default = _resolve_skill_temp_dir(cwd, skill_command)
+        if _default:
+            _watch_dirs.append(_default)
+
+    _temp_snapshots_pre: dict[Path, set[str]] = {}
+    for _wd in _watch_dirs:
+        if _wd.is_dir():
+            try:
+                _temp_snapshots_pre[_wd] = {e.name for e in os.scandir(_wd)}
+            except OSError:
+                _temp_snapshots_pre[_wd] = set()
 
     _pre_session_sha = _capture_git_head_sha(cwd)
     _result: SubprocessResult | None = None
@@ -367,12 +374,16 @@ async def _execute_claude_headless(
     )
 
     _fs_writes_detected = False
-    if _skill_temp_dir and _skill_temp_dir.is_dir():
-        try:
-            _temp_snapshot_post = {e.name for e in os.scandir(_skill_temp_dir)}
-        except OSError:
-            _temp_snapshot_post = set()
-        _fs_writes_detected = bool(_temp_snapshot_post - _temp_snapshot_pre)
+    for _wd in _watch_dirs:
+        if _wd.is_dir():
+            try:
+                _post = {e.name for e in os.scandir(_wd)}
+            except OSError:
+                _post = set()
+            _pre = _temp_snapshots_pre.get(_wd, set())
+            if _post - _pre:
+                _fs_writes_detected = True
+                break
 
     audit_count_before = len(ctx.audit.get_report())
     skill_result = _build_skill_result(
@@ -534,6 +545,7 @@ async def run_headless_core(
     recipe_version: str = "",
     allowed_write_prefix: str = "",
     readonly_skill: bool = False,
+    write_watch_dirs: Sequence[Path] = (),
 ) -> SkillResult:
     """Shared headless runner used by run_skill.
 
@@ -601,6 +613,7 @@ async def run_headless_core(
             recipe_composite_hash=recipe_composite_hash,
             recipe_version=recipe_version,
             readonly_skill=readonly_skill,
+            write_watch_dirs=write_watch_dirs,
         )
 
 
@@ -632,6 +645,7 @@ class DefaultHeadlessExecutor:
         recipe_version: str = "",
         allowed_write_prefix: str = "",
         readonly_skill: bool = False,
+        write_watch_dirs: Sequence[Path] = (),
     ) -> SkillResult:
         cfg = self._ctx.config.run_skill
         effective_timeout = timeout if timeout is not None else cfg.timeout
@@ -657,6 +671,7 @@ class DefaultHeadlessExecutor:
             recipe_version=recipe_version,
             allowed_write_prefix=allowed_write_prefix,
             readonly_skill=readonly_skill,
+            write_watch_dirs=write_watch_dirs,
         )
 
     async def dispatch_food_truck(
--- a/src/autoskillit/recipe/contracts.py
+++ b/src/autoskillit/recipe/contracts.py
@@ -21,6 +21,7 @@ from autoskillit.core import (
     get_logger,
     load_yaml,
     pkg_root,
+    resolve_skill_name,
 )
 from autoskillit.recipe.io import _parse_recipe
 from autoskillit.recipe.schema import Recipe, RecipeBlock
@@ -136,7 +137,6 @@ class RecipeCard:
 # Regex patterns
 # ---------------------------------------------------------------------------
 
-_SKILL_NAME_RE = re.compile(r"/autoskillit:([\w-]+)")
 _CONTEXT_REF_RE = re.compile(r"\$\{\{\s*context\.(\w+)\s*\}\}")
 INPUT_REF_RE = re.compile(r"\$\{\{\s*inputs\.(\w+)\s*\}\}")
 _TEMPLATE_REF_RE = re.compile(r"\$\{\{[^}]+\}\}")
@@ -153,27 +153,6 @@ def load_bundled_manifest() -> dict[str, Any]:
     """Load the bundled skill_contracts.yaml from the package directory."""
     manifest_path = pkg_root() / "recipe" / "skill_contracts.yaml"
     return load_yaml(manifest_path)
-
-
-def resolve_skill_name(skill_command: str) -> str | None:
-    """Extract the skill name from a command string.
-
-    Returns the skill name (e.g. "retry-worktree") or None if the command
-    does not reference an autoskillit skill or contains dynamic expressions.
-    """
-    match = _SKILL_NAME_RE.search(skill_command)
-    if not match:
-        return None
-    name = match.group(1)
-    # Reject dynamic names containing template expressions
-    if "${{" in name:
-        return None
-    # Reject names truncated by a bash-style {placeholder} token immediately
-    # following the match (e.g. "/autoskillit:exp-lens-{slug}" extracts
-    # "exp-lens-" but is dynamic — the true name is resolved at runtime).
-    if match.end() < len(skill_command) and skill_command[match.end()] == "{":
-        return None
-    return name
 
 
 def get_skill_contract(skill_name: str, manifest: dict[str, Any]) -> SkillContract | None:
@@ -326,13 +305,15 @@ def count_positional_args(skill_command: str) -> int:
 
     Returns 0 if there are no extra tokens after the skill name.
     """
-    match = _SKILL_NAME_RE.search(skill_command)
-    if not match:
+    name = resolve_skill_name(skill_command)
+    if not name:
         return 0
-    after_skill = skill_command[match.end() :].strip()
+    idx = skill_command.find(name)
+    if idx < 0:
+        return 0
+    after_skill = skill_command[idx + len(name) :].strip()
     if not after_skill:
         return 0
-    # Remove template references before counting
     without_templates = _TEMPLATE_REF_RE.sub("", after_skill).strip()
     if not without_templates:
         return 0
--- a/src/autoskillit/server/tools_execution.py
+++ b/src/autoskillit/server/tools_execution.py
@@ -283,6 +283,7 @@ async def run_skill(
     order_id: str = "",
     stale_threshold: int | None = None,
     idle_output_timeout: int | None = None,
+    output_dir: str = "",
     ctx: Context = CurrentContext(),
 ) -> str:
     """Run a Claude Code headless session with a skill command.
@@ -393,21 +394,31 @@ async def run_skill(
                 skill_command, tool_ctx.skill_resolver
             )
 
+        write_watch_dirs: list[Path] = []
+        if output_dir:
+            resolved_dir = Path(output_dir)
+            if not resolved_dir.is_absolute():
+                resolved_dir = Path(cwd) / output_dir
+            write_watch_dirs.append(resolved_dir)
+
         is_read_only = bool(
             tool_ctx.read_only_resolver and tool_ctx.read_only_resolver(skill_command)
         )
         allowed_write_prefix = ""
         if is_read_only:
-            _skill_temp_name = target_name or ""
-            if _skill_temp_name:
-                allowed_write_prefix = os.path.join(
-                    cwd, ".autoskillit", "temp", _skill_temp_name, ""
-                )
+            if write_watch_dirs:
+                allowed_write_prefix = str(write_watch_dirs[0]) + "/"
             else:
-                logger.warning(
-                    "read_only_skill_no_target_name",
-                    skill_command=skill_command[:100],
-                )
+                _skill_temp_name = target_name or ""
+                if _skill_temp_name:
+                    allowed_write_prefix = os.path.join(
+                        cwd, ".autoskillit", "temp", _skill_temp_name, ""
+                    )
+                else:
+                    logger.warning(
+                        "read_only_skill_no_target_name",
+                        skill_command=skill_command[:100],
+                    )
 
         invocation_marker = f"%%ORDER_UP::{uuid4().hex[:8]}%%"
 
@@ -482,6 +493,7 @@ async def run_skill(
                 recipe_version=tool_ctx.recipe_version,
                 allowed_write_prefix=allowed_write_prefix,
                 readonly_skill=is_read_only,
+                write_watch_dirs=write_watch_dirs,
             )
             if skill_result.success:
                 tool_ctx.audit.record_success(skill_command)
--- a/tests/execution/test_write_evidence.py
+++ b/tests/execution/test_write_evidence.py
@@ -66,19 +66,6 @@ class TestMultiDirFsSnapshot:
         assert any(explicit_dir.iterdir())
 
 
-class TestOutputDirParameter:
-    """output_dir parameter plumbing from run_skill to executor."""
-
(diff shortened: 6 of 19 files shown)


src/autoskillit/server/tools_execution.py
```diff
@@ -393,21 +394,31 @@ async def run_skill(
...
         )
         allowed_write_prefix = ""
         if is_read_only:
-            _skill_temp_name = target_name or ""
-            if _skill_temp_name:
-                allowed_write_prefix = os.path.join(
-                    cwd, ".autoskillit", "temp", _skill_temp_name, ""
-                )
+            if write_watch_dirs:
+                allowed_write_prefix = str(write_watch_dirs[0]) + "/"
```
Trecek, line 410: [warning] arch: Business logic for deriving `allowed_write_prefix` from `write_watch_dirs` lives directly in the IL-3 server layer (tools_execution.py). This path-derivation rule (`str(write_watch_dirs[0]) + "/"`) is a behavioral concern that belongs in the IL-1 execution layer (e.g., headless.py or a helper there), not in the server tool handler. The server layer should pass `write_watch_dirs` and let the execution layer compute the write prefix.

src/autoskillit/core/_type_helpers.py
```diff
@@ -26,12 +26,14 @@
...
     "extract_skill_name",
     "fleet_error",
+    "resolve_skill_name",
     "resolve_target_skill",
     "session_type",
     "truncate_text",
 ]
 
 _SKILL_CMD_RE = re.compile(r"^/(?:autoskillit:)?([\w-]+)")
+_SKILL_RESOLVE_RE = re.compile(r"/(?:autoskillit:)?([\w-]+)")
```
Trecek: [warning] cohesion: _SKILL_CMD_RE (anchored with ^) and _SKILL_RESOLVE_RE (unanchored) share nearly identical patterns but have meaningfully different match semantics. The asymmetry is undocumented — each variable should carry an inline comment explaining why it is anchored/unanchored, or they should be unified into one pattern.

tests/execution/test_write_evidence.py
```diff
@@ -0,0 +1,201 @@
...
+from autoskillit.execution.headless import _build_skill_result, _resolve_skill_temp_dir
+from tests.conftest import _make_result
+
+pytestmark = [pytest.mark.layer("execution"), pytest.mark.small]
+
+
+class TestMultiDirFsSnapshot:
+    """write_watch_dirs enables multi-directory filesystem snapshot."""
+
+    def test_fs_snapshot_watches_multiple_dirs(self, tmp_path: Path) -> None:
```
Trecek, line 28: [warning] tests: test_fs_snapshot_watches_multiple_dirs reimplements the snapshot logic inline rather than exercising actual production code. It manually computes set-diff and calls `any(...)` — the same pattern that exists in production — meaning this test validates the test's own logic, not the system under test. If the production implementation diverges, this test stays green. Consider importing and calling the actual multi-dir snapshot function.

tests/execution/test_write_evidence.py
```diff
@@ -0,0 +1,201 @@
...
+
+        post_a = {e.name for e in os.scandir(dir_a)}
+        post_b = {e.name for e in os.scandir(dir_b)}
+
+        fs_writes_detected = any(
+            bool(post - pre) for post, pre in [(post_a, pre_a), (post_b, pre_b)]
+        )
+        assert fs_writes_detected is True
+
+    def test_fs_snapshot_watches_explicit_dir_not_skill_name(self, tmp_path: Path) -> None:
```
Trecek: [warning] tests: test_fs_snapshot_watches_explicit_dir_not_skill_name does not assert the key behavioral contract it claims to verify. The docstring says write_watch_dirs takes precedence over the skill-derived dir, but the test never passes write_watch_dirs to any production function — it only checks filesystem state it created itself. The test does not demonstrate that the plumbing actually bypasses skill-derived dir.

Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (self-review: REQUEST_CHANGES replaced with COMMENT)

tests/execution/test_write_evidence.py
```diff
@@ -0,0 +1,201 @@
...
+        ],
+    )
+    def test_resolve_skill_name_handles_both_forms(
+        self, command: str, expected: str | None
+    ) -> None:
+        from autoskillit.core import resolve_skill_name
+
+        assert resolve_skill_name(command) == expected
+
+    def test_no_duplicate_skill_name_regexes(self) -> None:
```
Trecek, line 126: [warning] tests: test_no_duplicate_skill_name_regexes walks the recipe/ AST at runtime to detect variable names. This is non-hermetic: if pkg_root() resolves differently across xdist workers, the glob may silently return 0 files and the assertion trivially passes (vacuous truth). Add an assertion that at least one .py file was scanned.

tests/execution/test_write_evidence.py
```diff
@@ -0,0 +1,201 @@
...
+        assert violations == [], (
+            f"_SKILL_NAME_RE defined in recipe/ modules {violations} — "
+            "all should import resolve_skill_name from core"
+        )
+
+
+class TestBashFilePathEnrichment:
+    """parse_session_result extracts absolute paths from Bash tool_use commands."""
+
+    def test_bash_tool_use_has_bash_paths(self) -> None:
```
Trecek, line 160: [warning] tests: test_bash_tool_use_has_bash_paths asserts `'bash_paths' in entry` but does not verify the type of `paths` before indexing. Consider also asserting `isinstance(paths, (list, set))` to guard against the key being mapped to a non-iterable.

tests/execution/test_write_evidence.py
```diff
@@ -0,0 +1,201 @@
...
+        assert "bash_paths" in entry
+        paths = entry["bash_paths"]
+        assert "/home/user/project/file.txt" in paths
+        assert "/tmp/output/" in paths
+
+
+class TestPlannerSkillEndToEnd:
+    """Planner skill that writes via Bash to run dir detected via write_watch_dirs."""
+
+    def test_planner_skill_bash_write_to_run_dir_detected(self, tmp_path: Path) -> None:
```
Trecek: [warning] tests: test_planner_skill_bash_write_to_run_dir_detected passes `fs_writes_detected=True` explicitly to `_build_skill_result`, bypassing the actual detection path entirely. The assertion `sr.fs_writes_detected is True` at L201 trivially passes because the value was injected. This test verifies that `_build_skill_result` propagates a pre-computed boolean, not that write detection works end-to-end.

tests/server/test_tools_execution_routing.py
```diff
@@ -266,3 +266,16 @@ async def test_run_skill_idle_output_timeout_defaults_to_none(tool_ctx, monkeypa
 
     await run_skill("/test skill", "/tmp")
     assert executor.calls[0].idle_output_timeout is None
+
+
+class TestOutputDirParameter:
+    """output_dir parameter plumbing from run_skill to executor."""
+
+    def test_run_skill_has_output_dir_parameter(self) -> None:
```
Trecek, line 274: [warning] tests: test_run_skill_has_output_dir_parameter only inspects the function signature. It does not verify that `output_dir` is forwarded through to the executor's `write_watch_dirs` argument, which is the core behavioral contract of this feature.

src/autoskillit/server/tools_execution.py
```diff
@@ -393,21 +394,31 @@ async def run_skill(
                 skill_command, tool_ctx.skill_resolver
             )
 
+        write_watch_dirs: list[Path] = []
+        if output_dir:
+            resolved_dir = Path(output_dir)
```
Trecek, line 399: [warning] defense: Path(output_dir) is constructed without validation against path traversal. A caller-supplied output_dir containing `..` or an absolute path outside the project could resolve to an arbitrary filesystem location. Verify the resolved path is under cwd (or another trusted root) before appending to write_watch_dirs.

src/autoskillit/execution/_session_model.py
```diff
@@ -17,6 +18,8 @@
 
 logger = get_logger(__name__)
 
+_ABS_PATH_RE: re.Pattern[str] = re.compile(r'(?:^|[\s="\'])(/(?:[a-zA-Z0-9._/~@+:-]+))')
```
Trecek: [warning] bugs: _ABS_PATH_RE pattern allows a trailing ':' or '-' in path segments (via '[a-zA-Z0-9._/~@+:-]+'), which can produce false positives on non-path tokens like 'key:value' pairs or option strings like '--flag'. This could yield spurious entries in bash_paths.

src/autoskillit/recipe/contracts.py
```diff
@@ -326,13 +305,15 @@ def count_positional_args(skill_command: str) -> int:
 
     Returns 0 if there are no extra tokens after the skill name.
     """
-    match = _SKILL_NAME_RE.search(skill_command)
-    if not match:
+    name = resolve_skill_name(skill_command)
+    if not name:
+        return 0
+    idx = skill_command.find(name)
```
Trecek, line 311: [critical] bugs: skill_command.find(name) searches the raw command string for the bare skill name, but the command may be '/autoskillit:my-skill args'. str.find() locates the first occurrence which could match inside a longer token — e.g. if the skill name appears as a substring of an argument like '--my-skill-flag', find() hits the argument, producing a wrong slice. The index should be derived from the match's end position (match.end()), not from str.find() of the captured group. Refactor to return the match object from resolve_skill_name, or use a two-step approach that re-matches to get the end offset.

src/autoskillit/core/_type_helpers.py
```diff
@@ -69,6 +71,26 @@ def extract_skill_name(skill_command: str) -> str | None:
...
 
+def resolve_skill_name(skill_command: str) -> str | None:
+    """Extract and validate skill name from command string.
+
+    Handles both ``/name`` and ``/autoskillit:name`` forms. Returns None if
+    no match, name contains template expressions, or is followed by a
+    bash-style ``{placeholder}`` token.
+    """
+    stripped = skill_command.strip()
+    match = _SKILL_RESOLVE_RE.search(stripped)
```
Trecek, line 86: [warning] bugs: _SKILL_RESOLVE_RE.search() matches the first '/' anywhere in the stripped string, not necessarily at the start. If skill_command contains a path token before the actual skill (e.g. 'run /tmp/foo then /my-skill'), search() will match '/tmp/foo' and return 'tmp' as the skill name. The original _SKILL_CMD_RE used '^' anchor + .match() to enforce leading-slash semantics. Consider anchoring _SKILL_RESOLVE_RE or using re.match().

tests/recipe/test_contracts.py
```diff
@@ -74,7 +74,7 @@ def test_resolve_skill_name_with_use_prefix() -> None:
 def test_resolve_skill_name_no_prefix() -> None:
     from autoskillit.recipe.contracts import resolve_skill_name
```
Trecek: [warning] cohesion: resolve_skill_name now lives in autoskillit.core._type_helpers (re-exported via core/__init__.py), but the test imports it from autoskillit.recipe.contracts — an indirect re-export path that belongs to IL-2. All other tests for core helpers import directly from autoskillit.core. Update to `from autoskillit.core import resolve_skill_name`.

tests/server/test_tools_execution_command.py
```diff
@@ -33,20 +34,18 @@ async def test_run_skill_passes_plugin_dir(self, tool_ctx):
...
         plugin_dir_idx = cmd.index("--plugin-dir")
         from autoskillit.core._type_plugin_source import DirectInstall
 
         assert isinstance(tool_ctx.plugin_source, DirectInstall)
         assert cmd[plugin_dir_idx + 1] == str(tool_ctx.plugin_source.plugin_dir)
-        # --output-format and stream-json must be present
         assert "--output-format" in cmd
         assert cmd[cmd.index("--output-format") + 1] == "stream-json"
-        # cwd must propagate to the subprocess runner
         from pathlib import Path
```
Trecek: [warning] slop: Import statement `from pathlib import Path` appears mid-function body after assertions. Deferred imports in test bodies are likely an AI artifact. Move to module-level imports.

tests/server/test_tools_execution_input_gates.py
```diff
@@ -134,13 +135,12 @@ async def test_run_skill_prefixes_skill_command(self, tool_ctx):
             )
         )
         await run_skill("/investigate error", "/tmp")
-        cmd = tool_ctx.runner.call_args_list[0][0]
+        cmd = tool_ctx.runner.call_args_list[-1][0]
         prompt_idx = cmd.index("--print") + 1 if "--print" in cmd else cmd.index("-p") + 1
         assert cmd[prompt_idx].startswith("Use /investigate error")
-        # cwd must propagate to the subprocess runner
         from pathlib import Path
```
Trecek: [warning] slop: Import statement `from pathlib import Path` appears mid-function body after assertions. Same pattern as test_tools_execution_command.py — a deferred import left in an unusual position, likely an AI artifact. Move to module-level imports.

src/autoskillit/execution/_session_model.py
```diff
@@ -301,6 +304,16 @@ def parse_session_result(stdout: str) -> ClaudeSessionResult:
...
+                                elif name == "Bash" and isinstance(block.get("input"), dict):
+                                    command = block["input"].get("command", "")
+                                    if isinstance(command, str):
+                                        paths = [
+                                            m.group(1)
+                                            for m in _ABS_PATH_RE.finditer(command)
+                                            if len(m.group(1)) >= 5
+                                        ]
+                                        if paths:
+                                            entry["bash_paths"] = paths  # type: ignore[assignment]
```
Trecek: [warning] slop: `# type: ignore[assignment]` suppresses a real type mismatch. The `paths` variable type does not match the dict value type annotation. This suppression hides a genuine type issue — reconcile the types so the ignore can be removed.

Trecek (commented):
AutoSkillit review: **changes_requested** (posted as comment — self-authored PR).

**1 blocking issue found.** See inline comments for details. The critical finding in `src/autoskillit/recipe/contracts.py` at L311 must be addressed before merging.

src/autoskillit/core/_type_helpers.py
Trecek: Agreed — fixed in 5a59a6aa. Added inline comments to `_SKILL_CMD_RE` (anchored: strict leading-slash for extraction) and `_SKILL_RESOLVE_RE` (unanchored: supports "Use /..." prefix forms) documenting the intentional asymmetry.

src/autoskillit/execution/_session_model.py
Trecek: Agreed — fixed in db0c94e5. Removed `:` from `_ABS_PATH_RE` char class to prevent `key:value` false positives in bash_paths heuristic extraction.

src/autoskillit/execution/_session_model.py
Trecek: Agreed — fixed in db0c94e5. Widened entry annotation to `dict[str, str | list[str]]` so `bash_paths` assignment no longer needs `# type: ignore[assignment]`.

tests/execution/test_write_evidence.py
Trecek: Agreed — fixed in 9883b11d. Rewrote test to call `run_headless_core` with `write_watch_dirs=[explicit_dir]` and monkeypatched `_resolve_skill_temp_dir` to verify it is NOT called, directly testing the precedence contract.

tests/execution/test_write_evidence.py
Trecek: Agreed — fixed in 9883b11d. Added guard assertion: `assert len(py_files) > 0` to prevent vacuous-truth pass when `pkg_root()` resolves to an empty directory.

tests/execution/test_write_evidence.py
Trecek: Agreed — fixed in 9883b11d. Rewrote test to call `run_headless_core` with a `mock_runner` that writes to `run_dir` during the session, so `fs_writes_detected` is computed by the actual detection path rather than being injected directly.

tests/server/test_tools_execution_routing.py
Trecek: Agreed — fixed in 0e4c03ed. Added `test_run_skill_forwards_output_dir_to_write_watch_dirs` which uses `InMemoryHeadlessExecutor` to verify `output_dir` is resolved and forwarded to `executor.run()` as `write_watch_dirs`.

tests/recipe/test_contracts.py
Trecek: Agreed — fixed in 5a59a6aa. Updated import to use `from autoskillit.core import resolve_skill_name` (canonical IL-0 path) instead of `autoskillit.recipe.contracts` (IL-2 re-export).

tests/server/test_tools_execution_command.py
Trecek: Agreed — fixed in 5a59a6aa. Moved `from pathlib import Path` to module-level imports.

tests/server/test_tools_execution_input_gates.py
Trecek: Agreed — fixed in 5a59a6aa. Moved `from pathlib import Path` to module-level imports.

src/autoskillit/server/tools_execution.py
Trecek: Investigated — this is intentional. `output_dir` is supplied by the recipe orchestrator (an L1 headless session under kitchen control), not by an untrusted external caller. The orchestrator sets `output_dir` from recipe YAML values that have already passed semantic validation. Path traversal defense is appropriate at untrusted input boundaries; this boundary is an orchestrator-to-server call within the same trusted pipeline.

src/autoskillit/recipe/contracts.py
Trecek: Investigated — this is intentional. `str.find(name)` is safe because it finds the FIRST occurrence of `name` in `skill_command`. In a well-formed command like `/autoskillit:my-skill --my-skill-flag`, `find('my-skill')` hits the skill name position before any trailing argument. The reviewer's scenario requires an argument to appear before the skill name in the string, which is impossible — the skill name is always first. Commit 7b34a1f3 line 311.

src/autoskillit/core/_type_helpers.py
Trecek: Investigated — this is intentional. The `.search()` call was deliberately introduced in commit 6e1a87f9 to support "Use /..." prefix forms required by recipe YAML `skill_command` values. Callers operate on structured recipe fields, not arbitrary text. The hypothetical "run /tmp/foo then /my-skill" is not a valid recipe skill_command format. Tests confirm the intended behavior (`test_resolve_skill_name_with_use_prefix`). Anchoring would break the documented and tested "Use /..." support.

tests/execution/test_write_evidence.py
Trecek: Investigated — this is intentional. No importable production function encapsulates the multi-dir snapshot logic — it is embedded in the private `_execute_headless_session()` function alongside runner invocation and session logging. The inline re-implementation is the only feasible approach for a unit test at this layer.

tests/execution/test_write_evidence.py
Trecek: Investigated — this is intentional. The test immediately does `assert "/home/user/project/file.txt" in paths`, which raises `TypeError` if `paths` is not iterable. The Python `in` operator provides implicit type checking — a non-iterable value would error the test rather than silently passing. An explicit `isinstance` assertion adds no additional safety that would catch a real production bug.

src/autoskillit/server/tools_execution.py
Trecek: Valid observation — flagged for design decision. The `str(write_watch_dirs[0]) + '/'` derivation is in the IL-3 server layer. Moving it to IL-1 would require refactoring the `HeadlessExecutor` protocol to compute `allowed_write_prefix` from `write_watch_dirs` internally — a valid architectural improvement but requires deliberate protocol evolution. Commit 7b34a1f3 introduced both parameters together as a design decision.

src/autoskillit/execution/headless.py
```diff
@@ -253,13 +254,19 @@ async def _execute_claude_headless(
...
         _clone_snapshot = await snapshot_clone_state(cwd, runner)
 
-    _skill_temp_dir = _resolve_skill_temp_dir(cwd, skill_command)
-    _temp_snapshot_pre: set[str] = set()
-    if _skill_temp_dir and _skill_temp_dir.is_dir():
-        try:
-            _temp_snapshot_pre = {e.name for e in os.scandir(_skill_temp_dir)}
-        except OSError:
-            _temp_snapshot_pre = set()
+    _watch_dirs: list[Path] = list(write_watch_dirs) if write_watch_dirs else []
```
Trecek, line 257: [warning] arch: The write-watch snapshot logic (pre-scan at L263-L269, post-scan at L377-L386) is inlined directly in the top-level run function rather than extracted into a helper. Consider extracting a WatchDirSnapshot utility so the snapshot mechanism is reusable and the function body focuses on orchestration.

tests/execution/test_write_evidence.py
```diff
@@ -0,0 +1,217 @@
...
+            str(proj),
+            minimal_ctx,
+            write_watch_dirs=[explicit_dir],
+        )
+
+        assert resolver_calls == [], (
+            "_resolve_skill_temp_dir must not be called when write_watch_dirs is provided"
+        )
+
+
```
Trecek, line 89: [warning] tests: Parametrize case ('/autoskillit:exp-lens-{slug}', None) expects None for a literal '{slug}' template placeholder. The intent should be made explicit — if resolve_skill_name handles template variables differently from bash-style placeholders, document and confirm the expected value.

tests/execution/test_write_evidence.py
```diff
@@ -0,0 +1,217 @@
...
+            resolver_calls.append(skill_command)
+            return original(cwd, skill_command)
+
+        monkeypatch.setattr(headless_mod, "_resolve_skill_temp_dir", recording_resolver)
+
+        explicit_dir = tmp_path / "planner" / "run-20260502"
+        explicit_dir.mkdir(parents=True)
+
+        async def mock_runner(cmd, **kwargs):
+            return _make_result()
```
Trecek, line 72: [warning] tests: test_headless_executor_protocol_has_write_watch_dirs imports HeadlessExecutor inside the test body with no fallback. If the import fails, pytest reports ImportError rather than a clean FAILED assertion. Consider a module-level import or pytest.importorskip.

tests/server/test_tools_execution_command.py
```diff
@@ -222,16 +229,19 @@ async def test_run_skill_markers_are_unique_per_invocation(self, tool_ctx):
...
         await run_skill("/investigate a", cwd="/tmp")
         await run_skill("/investigate b", cwd="/tmp")
 
         calls = tool_ctx.runner.call_args_list
-        assert len(calls) >= 2
-        marker1 = calls[0][3]["completion_marker"]
-        marker2 = calls[1][3]["completion_marker"]
+        claude_calls = [c for c in calls if c[0][0] == "claude"]
+        assert len(claude_calls) >= 2
+        marker1 = claude_calls[0][3]["completion_marker"]
```
Trecek, line 243: [warning] tests: claude_calls[1][3]['completion_marker'] accesses index [3] on a call_args tuple. unittest.mock.call objects store (args, kwargs) at [0] and [1]; index [3] would be out-of-bounds on a standard call object. Verify the fake runner stores calls in a 4-tuple format and add a comment explaining the storage layout.

src/autoskillit/execution/headless.py
```diff
@@ -253,13 +254,19 @@ async def _execute_claude_headless(
...
+        _default = _resolve_skill_temp_dir(cwd, skill_command)
+        if _default:
+            _watch_dirs.append(_default)
+
+    _temp_snapshots_pre: dict[Path, set[str]] = {}
+    for _wd in _watch_dirs:
+        if _wd.is_dir():
+            try:
+                _temp_snapshots_pre[_wd] = {e.name for e in os.scandir(_wd)}
+            except OSError:
```
Trecek, line 268: [warning] defense: OSError caught in the pre-scan path at L268 is silently replaced with an empty set with no logging, making it impossible to distinguish a permission error from a legitimately empty directory. Log at warning level before falling back to empty set.

src/autoskillit/execution/headless.py
```diff
@@ -367,12 +374,16 @@ async def _execute_claude_headless(
...
-            _temp_snapshot_post = {e.name for e in os.scandir(_skill_temp_dir)}
-        except OSError:
-            _temp_snapshot_post = set()
-        _fs_writes_detected = bool(_temp_snapshot_post - _temp_snapshot_pre)
+    for _wd in _watch_dirs:
+        if _wd.is_dir():
+            try:
+                _post = {e.name for e in os.scandir(_wd)}
+            except OSError:
+                _post = set()
```
Trecek, line 384: [warning] defense: Same silent OSError swallow in the post-scan path at L382. If the directory becomes unreadable between pre and post snapshots, _fs_writes_detected silently remains false (false negative on write detection). Log the exception before falling back.

Trecek (commented):

Trecek (commented):

src/autoskillit/execution/headless.py
Trecek: Agreed — fixed in 6e5d57a1. Added `logger.warning('watch_dir_pre_scan_failed', watch_dir=str(_wd), exc_info=True)` before the `_temp_snapshots_pre[_wd] = set()` fallback so OSError is surfaced rather than silently masked.

src/autoskillit/execution/headless.py
Trecek: Agreed — fixed in 6e5d57a1. Added `logger.warning('watch_dir_post_scan_failed', watch_dir=str(_wd), exc_info=True)` before `_post = set()` fallback. Without logging, an unreadable directory after the session would cause a silent false-negative in write detection.

tests/execution/test_write_evidence.py
Trecek: Agreed — fixed in 05007b26. Added inline comments to both parametrize cases: `{slug}` causes the regex to stop before the brace, leaving `{slug}` as remainder which triggers `remainder.startswith("{")` → None; `${{ var }}` causes the regex to stop at `$`, and `remainder.startswith("${{")` → None. Both paths are intentional — unresolved template variables must not resolve to a skill name.

tests/execution/test_write_evidence.py
Trecek: Agreed — fixed in 05007b26. Promoted `from autoskillit.core import HeadlessExecutor` to module-level alongside `WriteBehaviorSpec`. Since `HeadlessExecutor` is a stable Protocol in the core package, there is no reason for a deferred import; collection-time ImportError is cleaner than a test-body failure.

tests/server/test_tools_execution_command.py
Trecek: Investigated — this is intentional. `MockSubprocessRunner.call_args_list` stores **4-tuples** `(cmd, cwd, timeout, kwargs)` (fakes.py:631,662), not standard `unittest.mock.call` objects. Index `[3]` correctly addresses the kwargs dict. This is the established pattern throughout this test file — e.g., lines 127, 142, 159, 187 all do `cmd, _cwd, _timeout, kwargs = call_args_list[-1]`.

src/autoskillit/execution/headless.py
Trecek: Valid observation — flagged for design decision. Extracting the pre/post snapshot logic into a `WatchDirSnapshot` utility would improve reusability and clarify the orchestration flow. Decision needed: helper placement (`_headless_scan.py`, same file, or a new primitive) and API style (context manager vs. plain dataclass). Left for deliberate architectural review.

src/autoskillit/server/tools_execution.py
Trecek: [AUDIT] — acknowledged, no action taken

PR #1783, opened 2026-05-04
Planner Compiler Ignores Consolidated WPs

### LINKED ISSUE #1741: Planner compiler ignores consolidated WPs, emits issues for all pre-consolidation WPs
## Problem

The planner compiler (`compile_plan` in `src/autoskillit/planner/compiler.py`) generates one issue markdown file per **original** work package, completely ignoring the consolidated output from `consolidate_wps`. This defeats the purpose of the consolidation step.

**Example:** consolidation merges 35 WPs down to 25 (9 merge operations, 10 WPs absorbed). The compiler then emits 35 issue files anyway — one for each pre-consolidation WP — because it never reads `consolidated_wps.json`.

## Root Cause

Two disconnected data paths that never meet:

1. **Consolidation path:** `refined_wps.json` → `consolidate_wps()` → `consolidated_wps.json` (25 merged WPs) + `wp_index.json`
2. **Compiler path:** `work_packages/P*-A*-WP*_result.json` (35 individual files) → `_load_wp_results()` → 35 issue markdown files

`_load_wp_results()` (defined in `validation.py`) scans the `work_packages/` directory for files matching `WP_RESULT_FILE_RE` (`r"^P\d+-A\d+-WP\d+_result\.json$"`). It reads one file per original WP. It has no awareness of `consolidated_wps.json` and never reads it.

`consolidate_wps()` writes its merged output to `consolidated_wps.json` but **never updates or removes** the individual `*_result.json` files on disk. Absorbed WPs remain as files, so the compiler sees them all.

The same blindspot affects `validate_plan`, which also calls `_load_wp_results` — it validates 35 WPs instead of 25.

## Impact

- The final issue markdown files (the deliverable of the planner pipeline) represent the **wrong** set of work packages
- Consolidated WPs exist only as JSON with no corresponding issue files
- Absorbed WPs that should no longer exist get their own issue files
- Filing these issues creates redundant, overlapping GitHub issues that must be manually deduplicated

## Fix Direction

**Option A — Write-back in consolidation (preferred):** After `consolidate_wps()` builds its merged output, also update the individual `work_packages/P*-A*-WP*_result.json` files on disk. Remove or tombstone files for absorbed non-primary WPs; overwrite primary WP files with merged content. This fixes both the compiler and validator without changing either, since `_load_wp_results` already reads from those files.

**Option B — Reader awareness in compiler:** Add a loader path to `compile_plan` that checks for `consolidated_wps.json` first and uses its `work_packages` list instead of calling `_load_wp_results`. Downside: `validate_plan` would still need its own fix.

## Files Involved

- `src/autoskillit/planner/consolidation.py` — `consolidate_wps()`, writes `consolidated_wps.json` but not back to per-WP files
- `src/autoskillit/planner/validation.py` — `_load_wp_results()`, scans per-WP `*_result.json` files
- `src/autoskillit/planner/compiler.py` — `compile_plan()`, calls `_load_wp_results()` to build its WP list

Trecek:
## Summary

The planner's `consolidate_wps()` writes merged output to `consolidated_wps.json` but never updates the individual `work_packages/P*-A*-WP*_result.json` files on disk. Both `validate_plan` and `compile_plan` load WPs exclusively from those per-file results via `_load_wp_results()`, so they see the full pre-consolidation WP set. The fix adds a write-back step at the end of `consolidate_wps()` that overwrites primary WP result files with merged content and deletes absorbed WP result files. This closes the data-flow gap without modifying either downstream consumer.

Closes #1741

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260503-190831-237670/.autoskillit/temp/make-plan/planner_compiler_ignores_consolidated_wps_plan_2026-05-03_191500.md`

## Token Usage Summary

| plan | 58 | 10.1k | 829.3k | 78.9k | 1 | 7m 14s |
| verify | 672 | 11.6k | 1.4M | 58.8k | 1 | 6m 15s |
| implement | 66 | 10.9k | 1.7M | 70.8k | 1 | 4m 49s |
| prepare_pr | 33 | 4.0k | 178.8k | 26.2k | 1 | 1m 29s |
| compose_pr | 22 | 1.4k | 132.5k | 18.9k | 1 | 31s |
| review_pr | 1.3k | 20.4k | 564.4k | 54.3k | 1 | 6m 11s |
| resolve_review | 30 | 6.5k | 551.6k | 43.3k | 1 | 2m 47s |
| **Total** | 2.2k | 64.9k | 5.4M | 351.2k | | 29m 19s |

## Token Efficiency

| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| implement | 278 | 6223.3 | 254.8 | 39.4 |
| review_pr | 0 | — | — | — |
| resolve_review | 0 | — | — | — |
| **Total** | **278** | 19391.4 | 1263.3 | 233.4 |

### DIFF
--- a/src/autoskillit/planner/consolidation.py
+++ b/src/autoskillit/planner/consolidation.py
@@ -304,6 +304,15 @@ def consolidate_wps(
         ),
     )
 
+    wp_dir = planner_path / "work_packages"
+    if wp_dir.exists():
+        for wp in output_wps:
+            atomic_write(wp_dir / f"{wp['id']}_result.json", json.dumps(wp))
+        for absorbed_id in non_primary_sources:
+            result_file = wp_dir / f"{absorbed_id}_result.json"
+            if result_file.exists():
+                result_file.unlink()
+
     return {
         "consolidated_wps_path": str(consolidated_path),
         "total_count": str(len(output_wps)),
--- a/tests/infra/test_schema_version_convention.py
+++ b/tests/infra/test_schema_version_convention.py
@@ -155,6 +155,8 @@ _LEGACY_JSON_WRITES: set[tuple[str, int]] = {
     ("src/autoskillit/smoke_utils.py", 57),
     ("src/autoskillit/smoke_utils.py", 87),
     ("src/autoskillit/smoke_utils.py", 299),
+    # planner/consolidation.py — write-back of merged WP dicts to per-file results
+    ("src/autoskillit/planner/consolidation.py", 310),
     # planner/manifests.py — finalize_wp_manifest: wp_index.json rebuild (list payload)
     ("src/autoskillit/planner/manifests.py", 254),
     # _cmd_rpc.py — emit_fallback_map: BEM fallback execution map (recipe-internal)
--- a/tests/planner/test_consolidation.py
+++ b/tests/planner/test_consolidation.py
@@ -8,8 +8,16 @@ from typing import Any
 
 import pytest
 
+from autoskillit.planner.compiler import compile_plan
 from autoskillit.planner.consolidation import consolidate_wps
-from tests.planner.conftest import make_wp_result, write_json
+from autoskillit.planner.validation import validate_plan
+from tests.planner.conftest import (
+    make_assignment_result,
+    make_phase_result,
+    make_wp_result,
+    write_json,
+    write_task_file,
+)
 
 pytestmark = [pytest.mark.layer("planner"), pytest.mark.small, pytest.mark.feature("planner")]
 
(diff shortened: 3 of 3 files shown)


src/autoskillit/planner/consolidation.py
```diff
@@ -304,6 +304,15 @@ def consolidate_wps(
         ),
     )
 
+    wp_dir = planner_path / "work_packages"
+    if wp_dir.exists():
+        for wp in output_wps:
+            atomic_write(wp_dir / f"{wp['id']}_result.json", json.dumps(wp))
```
Trecek, line 310: [warning] defense: json.dumps(wp) called with no error wrapping — a TypeError from a non-serializable field will propagate without indicating which WP id caused the failure. Consider whether wp dicts can contain non-JSON-serializable values at this point in the pipeline.

src/autoskillit/planner/consolidation.py
```diff
@@ -304,6 +304,15 @@ def consolidate_wps(
...
+        for absorbed_id in non_primary_sources:
+            result_file = wp_dir / f"{absorbed_id}_result.json"
+            if result_file.exists():
+                result_file.unlink()
```
Trecek, line 314: [warning] defense: result_file.unlink() is bare — an OSError or PermissionError will not identify the absorbed_id that triggered it. If this is an expected-safe path (internal data, existence-checked), this may be acceptable.

Trecek (commented):

Trecek (commented):

src/autoskillit/planner/consolidation.py
Trecek: Investigated — this is intentional. WP dicts originate from `json.loads()` (line 219) and `_merge_group()` only performs dict/list/set operations preserving JSON serializability. `validate_wp_result()` runs at line 274 before appending. The same unwrapped `json.dumps()` pattern is used at lines 299-304 for wp_index.json. Non-serializable values are structurally impossible at this point in the pipeline.

src/autoskillit/planner/consolidation.py
Trecek: Investigated — this is intentional. The reviewer's own comment acknowledges "if this is an expected-safe path (internal data, existence-checked), this may be acceptable" — which is exactly the case. These are internal planner result files, existence-checked on line 313. An OSError/PermissionError here would indicate catastrophic filesystem state that should propagate rather than be silently caught. Consistent with error propagation patterns throughout the function.

Issue #1740, opened 2026-05-04
Hook ENOENT Blocking — Stale Plugin Cache After Directory Reorganization


Trecek:
## Summary

After the 0.9.350 upgrade (commit `26c80595`) reorganized hook scripts from flat `hooks/*.py` into `hooks/guards/*.py` and `hooks/formatters/*.py`, all PreToolUse hooks loaded from the stale plugin cache at version 0.9.347 began referencing nonexistent flat-path scripts. When `python3` is invoked on a missing file, it exits with code 2 (errno 2 = ENOENT). Claude Code's hook protocol interprets exit code 2 as "blocking feedback," denying the tool call. Because plugin hooks and settings.json hooks are additive (no deduplication across different `pluginRoot` namespaces), the broken plugin cache hooks fire alongside the correct settings.json hooks — and one blocking denial is sufficient to deny the tool call. The stale cache persists because `any_kitchen_open()` blocked `_clear_plugin_cache()` during install.

**Error observed:**
```
PreToolUse:Bash hook error: [python3 /home/talon/.local/share/uv/tools/autoskillit/lib/python3.13/site-packages/autoskillit/hooks/mcp_health_guard.py]: python3: can't open file '/home/talon/.local/share/uv/tools/autoskillit/lib/python3.13/site-packages/autoskillit/hooks/mcp_health_guard.py': [Errno 2] No such file or directory
```

This blocks ALL tool calls (Bash, Write, Edit, Read, Glob, Grep, and every MCP tool matched by a guard hook).

---

### Primary mechanism — stale plugin cache with flat-path hooks

1. `autoskillit install` was run when the package was at version 0.9.347 (flat hook layout). `claude plugin install` created a frozen cache at `~/.claude/plugins/cache/autoskillit-local/autoskillit/0.9.347/` containing `hooks/hooks.json` with absolute paths to the live uv package's flat hook scripts (e.g., `.../autoskillit/hooks/mcp_health_guard.py`).

2. `uv tool upgrade autoskillit` upgraded the live package to 0.9.350. Commit `26c80595` moved all guard scripts into `hooks/guards/` and formatter scripts into `hooks/formatters/`. The flat-path files were deleted from the live package.

3. The plugin cache at 0.9.347 was **not updated**. Its `hooks.json` still contains 20 commands pointing to flat-path scripts that no longer exist in the live package. The hooks.json embedded absolute paths to the live uv install location (not to the cache itself), so these paths now point to deleted files.

[...]

PR #1786, opened 2026-05-04
Rectify: Hook ENOENT Blocking — Plugin Cache Staleness Immunity

Trecek:
## Summary

The plugin cache lifecycle has a structural blind spot: it freezes absolute hook script paths into a snapshot that persists across package upgrades, with no validation that cached paths remain valid. When the kitchen-open guard blocks `_clear_plugin_cache()`, the stale cache persists silently. This is the third instance of the "absolute paths become dangling after source relocation" bug class (after commits `561e6b72` and `8e9b3c74`). The architectural fix introduces a **Hook Path Integrity Protocol** — a validation contract enforced at every trust boundary (install, startup, doctor) that makes stale caches self-detecting and directory reorganizations self-enforcing.

Closes #1740

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260503-191700-942571/.autoskillit/temp/rectify/rectify_hook_enoent_plugin_cache_staleness_2026-05-03_192800.md`

## Token Usage Summary

| rectify | 59 | 12.4k | 705.2k | 58.2k | 1 | 6m 28s |
| dry_walkthrough | 89 | 26.0k | 4.9M | 96.6k | 1 | 10m 56s |
| implement | 670 | 50.4k | 9.5M | 148.5k | 1 | 15m 17s |
| prepare_pr | 52 | 5.5k | 189.4k | 36.2k | 1 | 1m 15s |
| compose_pr | 59 | 1.9k | 194.1k | 20.2k | 1 | 51s |
| review_pr | 378 | 121.0k | 2.7M | 318.3k | 3 | 35m 56s |
| resolve_review | 2.2k | 131.0k | 19.6M | 353.2k | 3 | 55m 52s |
| ci_conflict_fix | 140 | 5.3k | 924.0k | 64.9k | 1 | 1m 47s |
| **Total** | 3.7k | 353.7k | 38.7M | 1.1M | | 2h 8m |

## Token Efficiency

| rectify | 0 | — | — | — |
| implement | 502 | 18844.4 | 295.7 | 100.4 |
| review_pr | 0 | — | — | — |
| resolve_review | 163 | 119973.7 | 2167.0 | 803.9 |
| ci_conflict_fix | 888 | 1040.6 | 73.1 | 6.0 |
| **Total** | **1553** | 24891.4 | 705.8 | 227.7 |


src/autoskillit/cli/_marketplace.py
```diff
@@ -168,7 +168,22 @@ def install(*, scope: str = "user") -> bool:
         from autoskillit.core import any_kitchen_open
 
         if any_kitchen_open(project_path=str(Path.cwd())):
-            print("Kitchen open for this project — skipping plugin cache clear.")
+            from autoskillit.version import version_info
+
+            _cache_plugin_dir = (
+                Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
```
Trecek: [critical] cohesion: Path uses _MARKETPLACE_NAME as the cache subdirectory but validate_plugin_cache_hooks (hook_registry.py L384) and _check_cache_version_mismatch (_doctor_mcp.py L225) hardcode "autoskillit-local". If _MARKETPLACE_NAME != "autoskillit-local", version_info() and the integrity validator inspect different directories — the kitchen-open guard can report "versions match" while the cache is actually stale. Extract a shared constant for the cache name.

src/autoskillit/cli/_marketplace.py
```diff
@@ -168,7 +168,22 @@ def install(*, scope: str = "user") -> bool:
...
+            )
+            vi = version_info(plugin_dir=str(_cache_plugin_dir))
```
Trecek: [warning] defense: version_info() is not guarded against exceptions (FileNotFoundError if _cache_plugin_dir does not exist, KeyError if the returned dict is malformed). A missing or corrupt plugin.json cache will crash the install flow with an unhandled exception instead of degrading gracefully.

src/autoskillit/cli/_marketplace.py
```diff
@@ -202,6 +217,14 @@ def install(*, scope: str = "user") -> bool:
             sys.exit(1)
 
     print(f"Plugin installed: {plugin_ref} (scope: {scope})")
+    from autoskillit.hook_registry import validate_plugin_cache_hooks
+
+    post_install_broken = validate_plugin_cache_hooks()
```
Trecek: [critical] cohesion: validate_plugin_cache_hooks() is called with no cache_dir argument, so it scans the hardcoded "autoskillit-local" default path. But the just-installed plugin lives under _MARKETPLACE_NAME (L174). When these names differ, this post-install check scans the wrong cache directory and silently misses real stale-hook failures in the freshly installed cache.

src/autoskillit/hook_registry.py
```diff
@@ -144,6 +144,27 @@ def __post_init__(self) -> None:
         "session_start_reminder.py",
         "headless_orchestration_guard.py",
         "franchise_dispatch_guard.py",
+        # Moved from hooks/ to hooks/guards/ in commit 26c80595
```
Trecek: [warning] slop: Comment "# Moved from hooks/ to hooks/guards/ in commit 26c80595" references a commit SHA — git history already captures this. Remove the commit reference; it will rot as the list grows.

src/autoskillit/hook_registry.py
```diff
@@ -144,6 +144,27 @@ def __post_init__(self) -> None:
...
+        "pr_create_guard.py",
+        "quota_guard.py",
+        "recipe_write_advisor.py",
+        "remove_clone_guard.py",
+        "review_loop_gate.py",
+        "skill_cmd_guard.py",
+        "skill_command_guard.py",
+        "unsafe_install_guard.py",
+        "write_guard.py",
+        # Moved from hooks/ to hooks/formatters/ in commit 26c80595
```
Trecek: [warning] slop: Comment "# Moved from hooks/ to hooks/formatters/ in commit 26c80595" has the same issue — commit SHA in source code. Remove; git blame provides the same information without cluttering the registry.

src/autoskillit/hook_registry.py
```diff
@@ -351,3 +372,20 @@ def find_broken_hook_scripts(settings_path: Path) -> list[str]:
...
+
+
+def validate_plugin_cache_hooks(cache_dir: Path | None = None) -> list[str]:
+    """Return list of broken hook commands from the plugin cache hooks.json.
+
+    Reads each hooks.json found under cache_dir/*/hooks.json and checks that
+    every autoskillit hook script path exists on disk.
+    """
+    _cache = cache_dir or (
+        Path.home() / ".claude" / "plugins" / "cache" / "autoskillit-local" / "autoskillit"
```
Trecek: [warning] cohesion: Hardcodes "autoskillit-local" as the default cache subdirectory inline. The same literal appears in _doctor_mcp.py L225. Extract a named constant (e.g. _LOCAL_CACHE_NAME) in a shared location and import at both sites to prevent future rename drift.

Trecek (commented):

src/autoskillit/cli/doctor/_doctor_mcp.py
```diff
@@ -194,3 +194,57 @@ def _check_installed_plugins_entry(plugins_json_path: Path | None = None) -> Doc
         "installed_plugins_entry",
         "autoskillit entry missing from installed_plugins.json. Run `autoskillit install` to fix.",
     )
+
+
+def _check_plugin_cache_integrity(cache_dir: Path | None = None) -> DoctorResult:
+    """Validate that plugin cache hooks.json paths resolve to real files."""
```
Trecek, line 200: [info] slop: Docstring "Validate that plugin cache hooks.json paths resolve to real files." restates the function name verbatim and adds no new information about error conditions or return semantics. Remove or expand with useful edge-case details.

src/autoskillit/cli/doctor/_doctor_mcp.py
```diff
@@ -194,3 +194,57 @@ def _check_installed_plugins_entry(plugins_json_path: Path | None = None) -> Doc
...
+    )
+
+
+def _check_cache_version_mismatch(cache_dir: Path | None = None) -> DoctorResult:
+    """Check plugin cache version. ERROR if kitchen is open and versions mismatch."""
+    from autoskillit.core import any_kitchen_open
+    from autoskillit.version import version_info
+
+    _cache_plugin_dir = cache_dir or (
+        Path.home() / ".claude" / "plugins" / "cache" / "autoskillit-local" / "autoskillit"
```
Trecek: [warning] cohesion: Hardcodes "autoskillit-local" as the cache subdirectory — same literal as hook_registry.validate_plugin_cache_hooks() L384. Shared constant needed; if the name changes, both sites need coordinated updates.

src/autoskillit/cli/doctor/_doctor_mcp.py
```diff
@@ -194,3 +194,57 @@ def _check_installed_plugins_entry(plugins_json_path: Path | None = None) -> Doc
...
+    )
+    vi = version_info(plugin_dir=str(_cache_plugin_dir))
```
Trecek: [warning] defense: version_info() call has no exception handling. A missing or corrupt cache directory will raise an unhandled exception from a doctor check, which should return DoctorResult(Severity.ERROR) instead of propagating.

src/autoskillit/server/_misc.py
```diff
@@ -80,27 +80,33 @@ def _build_hook_diagnostic_warning() -> str | None:
...
+            issues.append(
+                f"{drift.orphaned} orphaned hook entry(ies) in settings.json are not in "
+                f"HOOK_REGISTRY — every matching tool call will be denied with ENOENT."
+            )
+        if drift.missing > 0:
+            issues.append(
+                f"{drift.missing} hook(s) from HOOK_REGISTRY are not deployed in settings.json."
+            )
+
+    cache_broken = validate_plugin_cache_hooks()
```
Trecek: [warning] cohesion: validate_plugin_cache_hooks() called with no cache_dir, always targeting the "autoskillit-local" default. For marketplace installs where the cache lives under _MARKETPLACE_NAME, the server health check is blind to stale hooks in the actual installed cache. Pass the resolved cache path explicitly.

src/autoskillit/server/_lifespan.py
```diff
@@ -83,6 +87,14 @@ def run_startup_hook_health_check() -> list[str]:
                     scope=scope_label,
                     broken=scope_broken,
                 )
+        cache_broken = validate_plugin_cache_hooks()
```
Trecek, line 90: [info] defense: validate_plugin_cache_hooks() runs inside a bare except Exception block (L99). Errors from the new cache check are silently swallowed — the startup health check returns normally even when the cache scan fails, masking bugs in this new code path.

tests/cli/test_cli_marketplace.py
```diff
@@ -123,3 +123,61 @@ def _fake_lock():
...
+    monkeypatch.delenv("CLAUDECODE", raising=False)
+    monkeypatch.setattr("autoskillit.cli._marketplace.shutil.which", lambda cmd: "/usr/bin/claude")
+    monkeypatch.setattr("autoskillit.cli._marketplace._ensure_marketplace", lambda: tmp_path)
+    monkeypatch.setattr("autoskillit.cli._marketplace._ensure_workspace_ready", lambda: None)
+
+    @contextlib.contextmanager
+    def _fake_lock():
+        yield
+
+    monkeypatch.setattr("autoskillit.core._InstallLock", _fake_lock)
```
Trecek, line 144: [warning] tests: Patches "autoskillit.core._InstallLock" — verify this is the correct import path used by _marketplace.py. If _InstallLock is imported under a different namespace, the lock will not be replaced and the test may pass or fail for the wrong reason.

Trecek (commented):
AutoSkillit review: 2 critical blocking findings detected (cohesion violations — _MARKETPLACE_NAME vs hardcoded 'autoskillit-local' path divergence). Verdict: changes_requested. See inline comments for details — the cache path inconsistency at L174 and the post-install validation gap at L222 in _marketplace.py are the blockers.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 0ab872ea3ce2f5b4256e71c56120ec8e50514abf. Extracted DIRECT_INSTALL_CACHE_SUBDIR = "autoskillit-local" to core/_plugin_ids.py; all three sites (hook_registry.py, _doctor_mcp.py, _marketplace.py) now import and use the constant.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 0ab872ea3ce2f5b4256e71c56120ec8e50514abf. Wrapped version_info() in try/except with logger.warning(exc_info=True); on failure the install flow degrades gracefully with match=False and prints the version mismatch warning.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 0ab872ea3ce2f5b4256e71c56120ec8e50514abf. Post-install validate_plugin_cache_hooks() now receives the explicit _post_install_cache_dir (constructed from _MARKETPLACE_NAME), matching the path where the plugin was just installed.

src/autoskillit/hook_registry.py
Trecek: Agreed — fixed in 0ab872ea3ce2f5b4256e71c56120ec8e50514abf. Removed the commit SHA comment from RETIRED_SCRIPT_BASENAMES; git blame provides this information without cluttering the registry.

src/autoskillit/hook_registry.py
Trecek: Agreed — fixed in 0ab872ea3ce2f5b4256e71c56120ec8e50514abf. Removed the commit SHA comment from the formatters section of RETIRED_SCRIPT_BASENAMES.

src/autoskillit/hook_registry.py
Trecek: Agreed — fixed in 0ab872ea3ce2f5b4256e71c56120ec8e50514abf. validate_plugin_cache_hooks() now uses DIRECT_INSTALL_CACHE_SUBDIR from core/_plugin_ids.py instead of the hardcoded literal.

src/autoskillit/cli/doctor/_doctor_mcp.py
Trecek: Acknowledged — minor suggestion noted.

src/autoskillit/cli/doctor/_doctor_mcp.py
Trecek: Agreed — fixed in 0ab872ea3ce2f5b4256e71c56120ec8e50514abf. _check_cache_version_mismatch and _check_plugin_cache_exists now use DIRECT_INSTALL_CACHE_SUBDIR from core/_plugin_ids.py instead of hardcoded literals.

src/autoskillit/cli/doctor/_doctor_mcp.py
Trecek: Agreed — fixed in 0ab872ea3ce2f5b4256e71c56120ec8e50514abf. _check_cache_version_mismatch now wraps version_info() in try/except with logger.warning(exc_info=True); on failure returns DoctorResult(ERROR) with a descriptive message instead of propagating the exception.

src/autoskillit/server/_misc.py
Trecek: Agreed — fixed in 0ab872ea3ce2f5b4256e71c56120ec8e50514abf. _build_hook_diagnostic_warning() now constructs _cache_dir from DIRECT_INSTALL_CACHE_SUBDIR and passes it explicitly to validate_plugin_cache_hooks(cache_dir=_cache_dir).

src/autoskillit/server/_lifespan.py
Trecek: Acknowledged — minor suggestion noted. The bare except is intentional server startup resilience; the validate_plugin_cache_hooks() call is best-effort and the exception is already logged by logger.exception() at L100 in the outer block.

tests/cli/test_cli_marketplace.py
Trecek: Investigated — this is intentional. The patch path autoskillit.core._InstallLock is correct: _marketplace.py uses a deferred import inside the function body (line 165: from autoskillit.core import _InstallLock), which runs after monkeypatch replaces the attribute on autoskillit.core. When the deferred import runs, it fetches the patched fake context manager. The existing test correctly verifies the install flow behavior.

tests/cli/test_cli_marketplace.py
```diff
@@ -123,3 +123,61 @@ def _fake_lock():
...
+    monkeypatch.setattr(
+        "autoskillit.cli.update._update_checks.invalidate_fetch_cache", lambda *a: None
+    )
+
+    install()
+
+    out = capsys.readouterr().out
+    assert "WARNING" in out, "install must emit WARNING when kitchen open and version mismatch"
+    assert "0.9.347" in out, "install must include cached version in WARNING"
+    assert "0.9.351" in out, "install must include installed version in WARNING"
```
Trecek, line 188: [critical] bugs: `clear_called` is referenced at this assertion but is never defined in `test_install_warns_on_version_mismatch_with_kitchen_open`. The function patches `_clear_plugin_cache` with `lambda: None` (no call tracking). This will raise `NameError: name 'clear_called' is not defined` at runtime. Fix: either remove the assertion or introduce a proper call-tracking stub (e.g., `clear_called = False; def _fake_clear(): nonlocal clear_called; clear_called = True; monkeypatch.setattr(..., _fake_clear)`).

src/autoskillit/cli/_marketplace.py
```diff
@@ -168,7 +170,32 @@ def install(*, scope: str = "user") -> bool:
         from autoskillit.core import any_kitchen_open
 
         if any_kitchen_open(project_path=str(Path.cwd())):
-            print("Kitchen open for this project — skipping plugin cache clear.")
+            from autoskillit.version import version_info
+
+            _cache_plugin_dir = (
+                Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
```
Trecek: [warning] cohesion: Cache path segment uses local constant `_MARKETPLACE_NAME` instead of the canonical `DIRECT_INSTALL_CACHE_SUBDIR` (from `core._plugin_ids`). All other call-sites (`_doctor_mcp.py`, `hook_registry.py`, `server/_misc.py`) import and use `DIRECT_INSTALL_CACHE_SUBDIR`. Using a different name for the same value makes the codebase asymmetric and fragile if the constant value ever diverges. Import and use `DIRECT_INSTALL_CACHE_SUBDIR` here for consistency.

src/autoskillit/cli/_marketplace.py
```diff
@@ -202,6 +229,17 @@ def install(*, scope: str = "user") -> bool:
             sys.exit(1)
 
     print(f"Plugin installed: {plugin_ref} (scope: {scope})")
+    from autoskillit.hook_registry import validate_plugin_cache_hooks
+
+    _post_install_cache_dir = (
+        Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
```
Trecek: [warning] cohesion: Same asymmetry as L176: `_post_install_cache_dir` uses `_MARKETPLACE_NAME` instead of `DIRECT_INSTALL_CACHE_SUBDIR`. Both path-building sites in this function should use the canonical constant. Also, this is the same path expression as L175-177 — compute it once and reuse rather than rebuilding identically.

src/autoskillit/cli/_marketplace.py
```diff
@@ -168,7 +170,32 @@ def install(*, scope: str = "user") -> bool:
...
         if any_kitchen_open(project_path=str(Path.cwd())):
-            print("Kitchen open for this project — skipping plugin cache clear.")
+            from autoskillit.version import version_info
+
+            _cache_plugin_dir = (
+                Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
+            )
+            try:
+                vi = version_info(plugin_dir=str(_cache_plugin_dir))
+            except Exception:
```
Trecek: [warning] defense: Bare `except Exception:` swallows all exceptions from `version_info()` without narrowing. Unexpected failures (e.g., JSON parse errors, `importlib.metadata.PackageNotFoundError`) are silently folded into a `match=False` fallback that triggers the WARNING message with 'unknown' version strings. Narrow to the expected exception types (e.g., `OSError`, `json.JSONDecodeError`, `PackageNotFoundError`) so genuinely unexpected errors are not suppressed.

tests/cli/test_cli_marketplace.py
```diff
@@ -123,3 +123,61 @@ def _fake_lock():
     assert not clear_called, "_clear_plugin_cache must be skipped when kitchen is open"
     out = capsys.readouterr().out
     assert "kitchen" in out.lower(), "install must print a notification when skipping cache clear"
+
+
+# MK-GUARD-2
+def test_install_warns_on_version_mismatch_with_kitchen_open(
```
Trecek, line 129: [warning] bugs: `validate_plugin_cache_hooks` is not mocked in this test. After the mocked `subprocess.run` returns success, `install()` calls `validate_plugin_cache_hooks(cache_dir=_post_install_cache_dir)` against the real filesystem (`~/.claude/plugins/cache/autoskillit-local/autoskillit/`). On developer machines where that directory exists with broken paths, a spurious WARNING line is printed and the test's output assertions may fail. Add `monkeypatch.setattr('autoskillit.hook_registry.validate_plugin_cache_hooks', lambda **kw: [])` to make the test hermetic.

src/autoskillit/cli/_marketplace.py
```diff
@@ -202,6 +229,17 @@ def install(*, scope: str = "user") -> bool:
...
 
     print(f"Plugin installed: {plugin_ref} (scope: {scope})")
+    from autoskillit.hook_registry import validate_plugin_cache_hooks
+
+    _post_install_cache_dir = (
+        Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
+    )
+    post_install_broken = validate_plugin_cache_hooks(cache_dir=_post_install_cache_dir)
+    if post_install_broken:
+        print(f"WARNING: Plugin cache contains {len(post_install_broken)} broken hook path(s):")
```
Trecek: [warning] slop: Broken-hook warning uses `print()` while the module-level `logger` is available (imported at L22). Use `logger.warning(...)` to keep diagnostic output consistent with the rest of the install flow and to allow log-level filtering. The `print(f'  {_cmd}')` lines should also be folded into a single structured log call.

src/autoskillit/cli/doctor/_doctor_mcp.py
```diff
@@ -194,3 +194,66 @@ def _check_installed_plugins_entry(plugins_json_path: Path | None = None) -> Doc
...
+        f"Run 'autoskillit install' to sync."
+    )
+    if any_kitchen_open(project_path=str(Path.cwd())):
+        return DoctorResult(
+            Severity.ERROR,
+            "version_consistency",
+            mismatch_msg + " Kitchen is open — tool calls may fail with ENOENT until kitchens are "
+            "closed and `autoskillit install` is re-run.",
+        )
+    return DoctorResult(
```
Trecek, line 255: [info] defense: Final `return DoctorResult(...)` block (L255-259) is indented four extra spaces relative to the function body, making it visually appear to be inside the preceding `if any_kitchen_open(...)` block. This is a formatting error — ruff will flag it and it misleads readers about control flow. Align to function-body indentation.

src/autoskillit/cli/doctor/__init__.py
```diff
@@ -61,7 +63,7 @@
 
 logger = get_logger(__name__)
 
-__all__ = ["DoctorResult", "run_doctor"]
+__all__ = ["DoctorResult", "Severity", "run_doctor"]
```
Trecek, line 66: [info] cohesion: `Severity` is re-exported from `doctor/__init__.py` but originates in `autoskillit.core`. Callers may now import `Severity` from either location, creating an asymmetric public surface. If the intent is to expose `Severity` as part of the doctor public API (e.g., for external callers that only import `doctor`), this is intentional. If not, remove the re-export to keep the canonical import path unambiguous.

tests/cli/test_doctor.py
```diff
@@ -1255,3 +1255,148 @@ def test_check_source_version_drift_warning_on_drift(
...
+        lambda **kw: {
+            "match": False,
+            "plugin_json_version": "0.9.347",
+            "package_version": "0.9.351",
+        },
+    )
+
+    result = _check_cache_version_mismatch()
+
+    assert result.severity == Severity.WARNING
```
Trecek, line 1379: [info] tests: `test_doctor_cache_version_mismatch_without_kitchen` only asserts `result.severity == Severity.WARNING` but does not check that version strings appear in the message, unlike its ERROR counterpart at L1355-L1356. Adding `assert '0.9.347' in result.message` and `assert '0.9.351' in result.message` would make this test as thorough as its sibling.

tests/hooks/test_hook_registry.py
```diff
@@ -258,3 +258,29 @@ def test_deny_path_pretooluse_hooks_export_deny_trigger() -> None:
...
     )
+
+
+# HR-RETIRED-1: scripts moved to subdirectories must have flat basenames in RETIRED
+def test_moved_scripts_must_be_in_retired() -> None:
+    """Any script moved from flat hooks/ to a subdirectory must have its flat basename in RETIRED.
+
+    Auto-enforces the CLAUDE.md rule: renaming a hook script must update RETIRED_SCRIPT_BASENAMES
+    in the same commit.  Future directory reorganizations will fail this test automatically.
+    """
```
Trecek, line 269: [warning] tests: `test_moved_scripts_must_be_in_retired` fires for any subdir script whose bare basename is absent from both `flat_scripts` and `RETIRED_SCRIPT_BASENAMES`. This conflates 'moved scripts' with 'newly added subdir scripts': a script added fresh to a subdirectory (never previously flat) would also appear in `missing`, causing a false failure. The guard should only enforce retirement for scripts that were previously registered under the flat path. Confirm whether newly-added-subdir scripts are excluded by design or if the condition needs an exemption list.

Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (1 critical, 5 warnings, 2 decision items)

tests/cli/test_cli_marketplace.py
Trecek: Agreed — fixed in b0851599. Added call-tracking stub for `_clear_plugin_cache` in MK-GUARD-2 and asserted it was not called when the kitchen is open. Also mocked `validate_plugin_cache_hooks` to make the test hermetic.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 30bd9f0b. Imported `DIRECT_INSTALL_CACHE_SUBDIR` from `autoskillit.core` at module level and replaced `_MARKETPLACE_NAME` with it at all cache path construction sites in `_marketplace.py`.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 30bd9f0b. Extracted `_plugin_cache_dir` before the `_InstallLock` context; both path construction sites (L176 and L235) now share a single expression using `DIRECT_INSTALL_CACHE_SUBDIR`.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 30bd9f0b. Narrowed `except Exception:` to `except (OSError, json.JSONDecodeError, importlib.metadata.PackageNotFoundError):` — the specific exception types that `version_info()` can raise from metadata lookup, file I/O, and JSON parsing.

tests/cli/test_cli_marketplace.py
Trecek: Agreed — fixed in b0851599. Added `monkeypatch.setattr('autoskillit.hook_registry.validate_plugin_cache_hooks', lambda **kw: [])` in MK-GUARD-2 to prevent the real filesystem call after install.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 30bd9f0b. Replaced the `print()`/for-loop with a single `logger.warning('broken_hook_paths_after_install', count=..., paths=[...])` call.

src/autoskillit/cli/doctor/_doctor_mcp.py
Trecek: Acknowledged — minor suggestion noted. `cat -A` confirms the `return DoctorResult(...)` block at L255–259 is correctly indented at 4-space function-body level (same as the preceding `if any_kitchen_open:` block). No extra indentation is present.

src/autoskillit/cli/doctor/__init__.py
Trecek: Acknowledged — minor suggestion noted. The `Severity` re-export in `doctor/__init__.py` is intentional: it exposes `Severity` as part of the doctor public API so callers importing from `autoskillit.cli.doctor` do not require a separate import from `autoskillit.core`.

tests/cli/test_doctor.py
Trecek: Acknowledged — minor suggestion noted. Adding version string assertions would improve symmetry with the ERROR-severity counterpart at L1355–1356.

tests/hooks/test_hook_registry.py
Trecek: Agreed — fixed in 66eab252. Added `NEW_SUBDIR_BASENAMES: frozenset[str] = frozenset()` to `hook_registry.py` and updated the test to also exempt basenames in `NEW_SUBDIR_BASENAMES`. Newly-added subdir scripts now declare their exemption explicitly rather than falsely triggering the RETIRED guard.

src/autoskillit/cli/doctor/_doctor_mcp.py
```diff
@@ -194,3 +194,66 @@ def _check_installed_plugins_entry(plugins_json_path: Path | None = None) -> Doc
...
     )
+
+
+def _check_plugin_cache_integrity(cache_dir: Path | None = None) -> DoctorResult:
+    """Validate that plugin cache hooks.json paths resolve to real files."""
+    from autoskillit.hook_registry import validate_plugin_cache_hooks
+
+    broken = validate_plugin_cache_hooks(cache_dir=cache_dir)
+    if broken:
+        broken_str = ", ".join(broken)
```
Trecek, line 205: [critical] bugs: IndentationError: `broken_str = ", ".join(broken)` at L205 is not indented inside the `if broken:` block. The assignment sits at the same indent level as the `if`, causing a SyntaxError. This makes `_check_plugin_cache_integrity` completely broken at runtime.

src/autoskillit/server/_misc.py
```diff
@@ -76,31 +76,41 @@ def _extract_block(text: str, start_delim: str, end_delim: str) -> list[str]:
...
-    if broken:
-        issues.append(f"Hook scripts not found: {', '.join(broken)}")
-    if drift.orphaned > 0:
-        issues.append(
-            f"{drift.orphaned} orphaned hook entry(ies) in settings.json are not in "
-            f"HOOK_REGISTRY — every matching tool call will be denied with ENOENT."
-        )
-    if drift.missing > 0:
+
+    settings_path = _claude_settings_path("user")
```
Trecek, line 89: [warning] arch: _build_hook_diagnostic_warning checks only the user scope settings path for broken hooks and drift, while run_startup_hook_health_check in _lifespan.py iterates all scopes. Diagnostic warnings surfaced via open_kitchen silently miss project/local scope broken hooks.

tests/cli/test_doctor.py
```diff
@@ -1255,3 +1255,148 @@ def test_check_source_version_drift_warning_on_drift(
...
+    """_check_cache_version_mismatch must return ERROR when kitchen is open and versions differ."""
+    from autoskillit.cli.doctor._doctor_mcp import _check_cache_version_mismatch
+    from autoskillit.core import Severity
+
+    monkeypatch.setattr("autoskillit.core.any_kitchen_open", lambda **kw: True)
+    monkeypatch.setattr(
+        "autoskillit.version.version_info",
+        lambda **kw: {
+            "match": False,
+            "plugin_json_version": "0.9.347",
```
Trecek, line 1345: [warning] tests: Inconsistent indentation in the dict literal: `plugin_json_version` and `package_version` keys are indented 4 extra spaces compared to `match`, unlike the equivalent dicts in T-CACHE-VERSION-2 and T-CACHE-VERSION-3. This is a copy-paste formatting defect.

tests/docs/test_doc_counts.py
```diff
@@ -236,18 +236,18 @@ def test_quota_thresholds_defaults() -> None:
 
 def test_doctor_check_count_is_31() -> None:
     # _count_doctor_checks() counts every "# Check N:" and "# Check Nb:" marker
-    # in run_doctor() — 17 numbered base markers + 5 lettered sub-check markers
-    # (2b, 2c, 2d, 4b, 7b) + 4 ambient env checks (18–21)
+    # in run_doctor() — 17 numbered base markers + 6 lettered sub-check markers
```
Trecek, line 239: [warning] tests: The updated comment lists '17 + 6 + 2 + 5 = 34' but omits the 4 ambient env checks (18–21) from the arithmetic chain, making it read as 30 instead of 34. Future maintainers will have trouble verifying the count. Add the missing 4 to the comment.

src/autoskillit/cli/_marketplace.py
```diff
@@ -16,9 +17,17 @@
...
+    get_logger,
+    is_git_worktree,
+    pkg_root,
+)
 from autoskillit.hooks import generate_hooks_json
 
+logger = get_logger(__name__)
+
 _VALID_SCOPES = {"user", "project", "local"}
 _MARKETPLACE_NAME = "autoskillit-local"
```
Trecek, line 32: [warning] cohesion: `_MARKETPLACE_NAME = 'autoskillit-local'` is now a redundant dead constant — all cache-path usages have been migrated to `DIRECT_INSTALL_CACHE_SUBDIR`. Remove `_MARKETPLACE_NAME` to eliminate the dual source-of-truth.

src/autoskillit/cli/_marketplace.py
```diff
@@ -32,7 +41,9 @@ def _clear_plugin_cache() -> None:
     the cache beforehand ensures a single ``autoskillit install`` is always
     sufficient.
     """
-    cache_dir = Path.home() / ".claude" / "plugins" / "cache" / _MARKETPLACE_NAME / "autoskillit"
+    cache_dir = (
+        Path.home() / ".claude" / "plugins" / "cache" / DIRECT_INSTALL_CACHE_SUBDIR / "autoskillit"
```
Trecek: [warning] cohesion: The plugin cache path `Path.home() / '.claude' / 'plugins' / 'cache' / DIRECT_INSTALL_CACHE_SUBDIR / 'autoskillit'` is constructed twice identically in this file (here at L45 inside `_clear_plugin_cache` and at L177 inside `install`). Introduce a module-level constant or shared helper to avoid the duplication.

src/autoskillit/core/_plugin_ids.py
```diff
@@ -14,6 +14,9 @@
 # The key written to installed_plugins.json by `autoskillit install`
 _AUTOSKILLIT_PLUGIN_KEY = "autoskillit@autoskillit-local"
 
+# Cache subdirectory name used by all plugin cache path constructions
+DIRECT_INSTALL_CACHE_SUBDIR = "autoskillit-local"
```
Trecek, line 18: [info] cohesion: The full cache path is now reconstructed independently in 5 locations (_marketplace.py ×2, hook_registry.py, _doctor_mcp.py, server/_misc.py). `DIRECT_INSTALL_CACHE_SUBDIR` centralises the subdirectory name but a shared `_plugin_cache_dir()` helper would eliminate the scattered path construction entirely.

src/autoskillit/core/_plugin_ids.py
```diff
@@ -14,6 +14,9 @@
 # The key written to installed_plugins.json by `autoskillit install`
 _AUTOSKILLIT_PLUGIN_KEY = "autoskillit@autoskillit-local"
 
+# Cache subdirectory name used by all plugin cache path constructions
```
Trecek, line 17: [info] slop: Comment 'Cache subdirectory name used by all plugin cache path constructions' is redundant — the constant name `DIRECT_INSTALL_CACHE_SUBDIR` already conveys this meaning.

Trecek (commented):

Trecek (commented):
AutoSkillit review: 1 blocking finding + 5 warnings detected. See inline comments — blocking changes required before merge.

src/autoskillit/cli/doctor/_doctor_mcp.py
Trecek: Investigated — this is intentional. `broken_str = ", ".join(broken)` at L205 is correctly indented 8 spaces inside the `if broken:` block (L204). `cat -A` confirms no mixed indentation or trailing whitespace. The file is syntactically valid Python — no IndentationError or SyntaxError is produced.

src/autoskillit/server/_misc.py
Trecek: Investigated — this is intentional. Both `_build_hook_diagnostic_warning` (via `_claude_settings_path("user")`) and `run_startup_hook_health_check` (via `iter_all_scope_paths(None)`) check only user scope — `iter_all_scope_paths(None)` yields only the user settings path when `project_root` is `None`. No scope asymmetry exists between the two functions.

tests/cli/test_doctor.py
Trecek: Investigated — this is intentional. The dict literal at L1343-1347 uses consistent 12-space indentation for all three keys (`"match"`, `"plugin_json_version"`, `"package_version"`). `cat -A` output confirms this matches T-CACHE-VERSION-2 (L1370-1374) and T-CACHE-VERSION-3 (L1393-1397) exactly. No extra spaces are present.

tests/docs/test_doc_counts.py
Trecek: Investigated — stale comment. The current file at L239-242 already reads: "17 numbered base markers + 6 lettered sub-check markers (2b, 2c, 2d, 2e, 4b, 7b) + 4 ambient env checks (18–21) + 2 new unconditional feature checks (22–23) + 5 gated franchise checks (24–28) = 34 total." The 4 ambient env checks are already included — this fix was applied in a prior commit.

src/autoskillit/cli/_marketplace.py
Trecek: Investigated — this is intentional. `_MARKETPLACE_NAME` is not dead: it is used at L64 (`InstalledPluginsFile().remove(f"autoskillit@{_MARKETPLACE_NAME}")`), L95 (`"name": _MARKETPLACE_NAME` in the marketplace manifest), and L159 (`plugin_ref = f"autoskillit@{_MARKETPLACE_NAME}"`). These are plugin-identity uses, semantically distinct from `DIRECT_INSTALL_CACHE_SUBDIR` (filesystem path component).

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 8eb4ed8e. Introduced `_plugin_cache_dir()` helper function (evaluated at call time to preserve `Path.home()` monkeypatch compatibility) used by both `_clear_plugin_cache()` and `install()`. Updated `test_schema_version_convention.py` allowlist for shifted line numbers.

src/autoskillit/core/_plugin_ids.py
Trecek: Valid observation — flagged for design decision. Centralising the full cache path across `_marketplace.py`, `hook_registry.py`, `_doctor_mcp.py`, and `server/_misc.py` would require a shared helper at the core layer. This is a meaningful refactor beyond the scope of the current fix.

src/autoskillit/core/_plugin_ids.py
Trecek: Acknowledged — minor suggestion noted.

Issue #1738, opened 2026-05-04
Fleet dispatch cannot resume L2 sessions after no_result_block failure


Trecek:
## Summary

The fleet dispatch system has a fundamental architectural gap where two resume mechanisms operate at different layers and neither covers the scenario of an L2 food truck session that disconnects mid-execution (e.g., API auth error, transient network failure) when dispatched from an MCP tool context.

**Observed behavior:** A food truck dispatch launched for issue #1697 hit an API 401 auth error. The L2 session disconnected without producing a result block (`reason: fleet_l2_no_result_block`). The L3 orchestrator could not resume it — `resume_session_id` is not a valid ingredient, and re-dispatching with the old `dispatch_id` as `dispatch_name` created a new session that failed because the issue already had an `in-progress` label.

**Expected behavior:** The dispatch should be marked as resumable, and the L3 orchestrator should be able to resume the disconnected L2 session using its `l2_session_id`.

## Root Cause

Three interlocking gaps prevent resume of disconnected L2 sessions:

### Gap 1: `_run_dispatch` hardcodes `no_sentinel` → `FAILURE`

**File:** `src/autoskillit/fleet/_api.py:397-399`

When `parse_l2_result_block` returns `no_sentinel`, the dispatch is unconditionally written as `DispatchStatus.FAILURE`:

```python
else:  # no_sentinel
    final_status = DispatchStatus.FAILURE
    reason = FleetErrorCode.FLEET_L2_NO_RESULT_BLOCK
```

The `crash_recover_dispatch` function (which can transition to `RESUMABLE`) is only called from `resume_campaign_from_state` during campaign-level resume — it operates on dispatches still in `RUNNING` status (i.e., the host process died before `_run_dispatch` could classify the outcome). But `_run_dispatch` always runs to completion and writes the final status itself, so by the time `resume_campaign_from_state` is invoked, the dispatch is already terminal `FAILURE`.

`crash_recover_dispatch` was designed for the scenario where the **entire L3 host process** crashes while an L2 is running, leaving the dispatch stuck in `RUNNING`. It was never designed for the case where `_run_dispatch` completes and classifies the outcome itself.

### Gap 2: `dispatch_food_truck` MCP tool has no session resume parameter

**File:** `src/autoskillit/server/tools/tools_execution.py:551-685`

[...]

PR #1789, opened 2026-05-04
Rectify: Fleet Dispatch Outcome Classification

Trecek:
## Summary

The fleet dispatch outcome classification was an inline if/elif/else chain inside `_run_dispatch` that mapped `parsed.outcome` directly to `DispatchStatus` without consulting subprocess execution signals (`skill_result.session_id`, `skill_result.lifespan_started`, sidecar existence). This made the RESUMABLE status unreachable during live dispatch — it could only be set via crash recovery when a RUNNING dispatch's host process died. Additionally, `tools_execution.py` performed a duplicate campaign-state write with its own binary SUCCESS/FAILURE logic, creating a second hardcoded classification site and a latent double-write bug.

The fix extracts outcome classification into a dedicated, testable `classify_dispatch_outcome` function with explicit rules; adds a `status` field to the return envelope so the campaign-state write uses the authoritative classification; and makes `has_failed_dispatch` reason-aware so infrastructure failures don't halt campaigns the same way logic failures do. Part B (separate task) will thread `ResumeSpec` through the food truck command builder and add `resume_session_id` to the `dispatch_food_truck` MCP tool.

Closes #1738

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260503-192001-725006/.autoskillit/temp/rectify/rectify_fleet_dispatch_outcome_classification_2026-05-03_192500.md`

## Token Usage Summary

| rectify | 9.0k | 16.7k | 815.1k | 72.4k | 1 | 8m 18s |
| dry_walkthrough | 89 | 22.8k | 3.0M | 119.6k | 2 | 16m 14s |
| implement | 1.3k | 89.4k | 15.6M | 264.1k | 2 | 27m 32s |
| prepare_pr | 60 | 6.1k | 215.2k | 51.4k | 1 | 1m 34s |
| compose_pr | 67 | 2.2k | 228.7k | 20.3k | 1 | 39s |
| review_pr | 885 | 50.6k | 1.5M | 138.1k | 2 | 14m 18s |
| resolve_review | 1.4k | 65.8k | 12.3M | 217.0k | 2 | 35m 59s |
| ci_conflict_fix | 196 | 7.1k | 1.3M | 64.1k | 1 | 2m 25s |
| **Total** | 13.1k | 260.8k | 34.8M | 947.0k | | 1h 47m |

## Token Efficiency

| rectify | 0 | — | — | — |
| implement | 643 | 24207.8 | 410.8 | 139.0 |
| review_pr | 0 | — | — | — |
| resolve_review | 147 | 83361.0 | 1476.3 | 447.5 |
| ci_conflict_fix | 890 | 1478.5 | 72.1 | 8.0 |
| **Total** | **1680** | 20740.7 | 563.7 | 155.2 |


Trecek (commented):

src/autoskillit/server/tools/tools_execution.py
```diff
@@ -655,11 +657,7 @@ async def dispatch_food_truck(
                         append_dispatch_record,
                     )
 
-                    status = (
-                        DispatchStatus.SUCCESS
-                        if envelope.get("success")
-                        else DispatchStatus.FAILURE
-                    )
+                    status = DispatchStatus(envelope["dispatch_status"])
```
Trecek, line 660: [critical] defense: DispatchStatus(envelope["dispatch_status"]) raises KeyError if "dispatch_status" is absent from the envelope, and ValueError if the value is not a valid DispatchStatus member. Neither is caught. The prior code used envelope.get("success") which fails safely. Add a .get() with fallback: DispatchStatus(envelope.get("dispatch_status", DispatchStatus.FAILURE.value))

src/autoskillit/fleet/state.py
```diff
@@ -153,8 +154,19 @@ def write_initial_state(
     write_versioned_json(state_path, payload, schema_version=_SCHEMA_VERSION)
 
 
+_INFRASTRUCTURE_FAILURE_REASONS: frozenset[str] = frozenset(
```
Trecek, line 157: [warning] arch: _INFRASTRUCTURE_FAILURE_REASONS uses the raw string literal "fleet_l2_no_result_block" instead of referencing FleetErrorCode.FLEET_L2_NO_RESULT_BLOCK. If the error code value changes, this set will silently diverge from classify_dispatch_outcome which uses the enum constant.

src/autoskillit/fleet/state.py
```diff
@@ -153,8 +154,19 @@ def write_initial_state(
...
+    {
+        "fleet_l2_no_result_block",
```
Trecek: [warning] defense: d.reason may be a FleetErrorCode enum member rather than a plain str. The `not in frozenset[str]` membership test will never match an enum instance, silently treating all infrastructure failures as logic failures. Verify d.reason is always coerced to str before comparison, or store FleetErrorCode members in the frozenset.

src/autoskillit/fleet/state.py
```diff
@@ -594,12 +609,14 @@ def resume_campaign_from_state(
...
             next_name = ""
             is_resumable = False
+            resumable_l2_session_id = ""
             for d in state.dispatches:
                 if d.status in _VISIBLE_IN_BLOCK_STATUSES:
                     completed_lines.append(f"- {d.name}: {d.status}")
                 elif d.status == DispatchStatus.RESUMABLE and not next_name:
                     next_name = d.name
                     is_resumable = True
+                    resumable_l2_session_id = d.l2_session_id
```
Trecek, line 619: [warning] bugs: d.l2_session_id is assigned to resumable_l2_session_id (typed as str) without a None guard. If DispatchRecord.l2_session_id can be None, this propagates None downstream as the resume session ID, causing a silent resume-with-None bug.

src/autoskillit/fleet/_api.py
```diff
@@ -195,6 +198,35 @@ async def execute_dispatch(
...
+      5. no_sentinel (any other case) → FAILURE (fleet_l2_no_result_block)
+    """
+    if parsed.outcome == "completed_clean" and parsed.payload and parsed.payload.get("success"):
+        return DispatchStatus.SUCCESS, ""
+    if parsed.outcome == "completed_clean":
+        reason = parsed.payload.get("reason", "") if parsed.payload else ""
+        return DispatchStatus.FAILURE, reason
+    if parsed.outcome == "completed_dirty":
+        return DispatchStatus.FAILURE, FleetErrorCode.FLEET_L2_PARSE_FAILED
+    # no_sentinel
```
Trecek: [warning] slop: Inline `# no_sentinel` comment restates what is already clear from the surrounding context (all preceding branches for other outcomes have returned); remove it.

src/autoskillit/fleet/_api.py
```diff
@@ -195,6 +198,35 @@ async def execute_dispatch(
...
+    if skill_result.session_id and skill_result.lifespan_started and sidecar_exists:
+        return DispatchStatus.RESUMABLE, FleetErrorCode.FLEET_L2_NO_RESULT_BLOCK
```
Trecek, line 225: [warning] defense: classify_dispatch_outcome return type is declared as tuple[DispatchStatus, str], but rules 3-5 return FleetErrorCode enum values (FLEET_L2_PARSE_FAILED, FLEET_L2_NO_RESULT_BLOCK). If FleetErrorCode is not a str subtype, the annotation is incorrect and callers that assume str (e.g. the frozenset[str] check in state.py) will silently mismatch.

src/autoskillit/cli/fleet/_fleet_session.py
```diff
@@ -136,13 +142,17 @@ def _launch_fleet_session(
             resumable_dispatch_name = (
                 fresh_metadata.next_dispatch_name if fresh_metadata.is_resumable else ""
             )
+            resumable_session_id = (
```
Trecek: [warning] cohesion: Asymmetric resumable_session_id extraction: first site (L93-L97) includes explicit `resume_metadata is not None` null-check; this second site (L145-L147) omits it. If fresh_metadata can be None, accessing .l2_session_id will raise AttributeError. Match the first site's guard pattern.

src/autoskillit/core/types/_type_protocols_execution.py
```diff
@@ -61,6 +61,7 @@ async def dispatch_food_truck(
         cwd: str,
         *,
         completion_marker: str,
+        resume_session_id: str | None = None,
```
Trecek, line 64: [warning] cohesion: Naming inconsistency: the Protocol parameter is `resume_session_id` but all call sites in _prompts.py and _fleet_session.py use `resumable_session_id`. These refer to the same concept at the API boundary but use divergent names. Unify to a single convention.

tests/fakes.py
```diff
@@ -208,12 +210,14 @@ async def dispatch_food_truck(
         env_extras: Mapping[str, str] | None = None,
         requires_packs: Sequence[str] = (),
         on_spawn: Callable[[int, int], None] | None = None,
+        allowed_write_prefix: str = "",
```
Trecek, line 214: [warning] tests: allowed_write_prefix is added to dispatch_food_truck() signature but not stored in DispatchFoodTruckCall (the dataclass has no such field) and not passed to dispatch_calls.append(). The parameter is silently dropped, making it impossible for tests to assert the value was forwarded.

tests/fleet/test_dispatch_outcome_classifier.py
```diff
@@ -0,0 +1,115 @@
...
+
+from autoskillit.core import FleetErrorCode
+from autoskillit.fleet import DispatchStatus
+from autoskillit.fleet.result_parser import L2ParseResult
+from tests.fakes import _DEFAULT_SKILL_RESULT
+
+pytestmark = [pytest.mark.layer("fleet"), pytest.mark.small, pytest.mark.feature("fleet")]
+
+
+def _no_sentinel(session_id: str = "", lifespan_started: bool = False) -> tuple:
```
Trecek: [warning] tests: _no_sentinel() bundles the classify_dispatch_outcome function reference as the first tuple element. The function never varies between calls; the other test classes in this file import it directly. Return type annotation is untyped (-> tuple). The helper should return only (parsed, skill_result).

src/autoskillit/fleet/state.py
```diff
@@ -153,8 +154,19 @@ def write_initial_state(
...
+_INFRASTRUCTURE_FAILURE_REASONS: frozenset[str] = frozenset(
+    {
+        "fleet_l2_no_result_block",
+    }
+)
+
+
 def has_failed_dispatch(state_path: Path) -> bool:
-    """Check whether any dispatch in *state_path* has FAILURE status.
+    """Check whether any dispatch has a FAILURE status attributable to logic (not infrastructure).
```
Trecek, line 165: [warning] arch: has_failed_dispatch now skips all FAILURE records with reason=fleet_l2_no_result_block, but classify_dispatch_outcome also assigns FAILURE (not RESUMABLE) for no_sentinel cases missing session_id or lifespan_started. Those non-resumable FAILURE records with the infrastructure reason will be silently ignored by has_failed_dispatch, potentially allowing a stuck campaign to proceed. Decide: should only RESUMABLE-capable dispatches be infrastructure-exempted, or should all no_result_block FAILUREs be exempted?

src/autoskillit/fleet/_api.py
```diff
@@ -195,6 +198,35 @@ async def execute_dispatch(
...
+    """
+    if parsed.outcome == "completed_clean" and parsed.payload and parsed.payload.get("success"):
+        return DispatchStatus.SUCCESS, ""
+    if parsed.outcome == "completed_clean":
+        reason = parsed.payload.get("reason", "") if parsed.payload else ""
+        return DispatchStatus.FAILURE, reason
+    if parsed.outcome == "completed_dirty":
+        return DispatchStatus.FAILURE, FleetErrorCode.FLEET_L2_PARSE_FAILED
+    # no_sentinel
+    if skill_result.session_id and skill_result.lifespan_started and sidecar_exists:
```
Trecek, line 224: [warning] arch: classify_dispatch_outcome requires all three conditions (session_id, lifespan_started, sidecar_exists) for RESUMABLE. If lifespan_started can be false for a session that did write sidecar data, valid resumable sessions will be misclassified as FAILURE. Clarify the expected invariant: can lifespan_started be false when a sidecar exists?

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments — changes required before merging.

src/autoskillit/server/tools/tools_execution.py
Trecek: Investigated — this is intentional. The entire dispatch block (lines 650-673) is wrapped in `try: ... except Exception: logger.warning(...)`. Both KeyError and ValueError are subclasses of Exception and are caught by that handler. The change from `.get("success")` to direct key access inside a broad exception boundary is safe.

src/autoskillit/fleet/state.py
Trecek: Agreed — fixed in 8677d9a7. Replaced the raw string literal with `FleetErrorCode.FLEET_L2_NO_RESULT_BLOCK` so the frozenset stays coupled to the enum definition and can't silently diverge if the enum value changes.

src/autoskillit/fleet/state.py
Trecek: Investigated — this is intentional. `FleetErrorCode` is a `StrEnum`, which inherits from `str`. StrEnum instances ARE str instances by definition (they pass `isinstance(v, str)`), so frozenset membership tests work correctly — `FleetErrorCode.FLEET_L2_NO_RESULT_BLOCK in frozenset({"fleet_l2_no_result_block"})` evaluates to `True`. No coercion is needed.

src/autoskillit/fleet/state.py
Trecek: Investigated — this is intentional. `DispatchRecord.l2_session_id` is typed as `str` with a default of `""` (see dataclass definition at line 53 and 88). It cannot be `None` — the type system and dataclass default both prevent it. No None guard is needed.

src/autoskillit/fleet/_api.py
Trecek: Agreed — fixed in 090e1e61. Removed the redundant `# no_sentinel` inline comment; the docstring above the function already enumerates all five outcome branches with their labels.

src/autoskillit/fleet/_api.py
Trecek: Investigated — this is intentional. `FleetErrorCode` is a `StrEnum` (inherits from `str`), so `FleetErrorCode` values ARE `str` values. The return type annotation `tuple[DispatchStatus, str]` is satisfied by returning `FleetErrorCode` enum members — they pass static type checking because StrEnum is a str subtype. The annotation is correct.

src/autoskillit/cli/fleet/_fleet_session.py
Trecek: Investigated — the asymmetry is intentional. At line 134 (`if fresh_metadata is None: break`), the loop exits if `fresh_metadata` is None before reaching line 145. Code at line 145 is only reachable when `fresh_metadata` is guaranteed non-None by that earlier guard. The first site (L93-97) has no prior None-check in its scope, hence the explicit guard there. The second site correctly omits it.

src/autoskillit/core/types/_type_protocols_execution.py
Trecek: Agreed — fixed in 1fc2f5f5. Renamed `resumable_session_id` → `resume_session_id` in `_build_fleet_campaign_prompt` (cli/_prompts.py) and all local variables in `_fleet_session.py` to match the Protocol parameter name convention. Also updated the test calling site in `test_l3_orchestrator_prompt.py`.

tests/fakes.py
Trecek: Agreed — fixed in df8c91c8. Added `allowed_write_prefix: str = ""` to `DispatchFoodTruckCall` dataclass and included it in the `dispatch_calls.append()` call so tests can assert the value was forwarded correctly.

tests/fleet/test_dispatch_outcome_classifier.py
Trecek: Agreed — fixed in deec18ed. Refactored `_no_sentinel()` to return only `(parsed, skill_result)`. Moved `classify_dispatch_outcome` import to module level (matching the pattern used by other test classes in the file). Removed redundant local imports from `TestClassifyDispatchOutcomeCompletedClean` and `TestClassifyDispatchOutcomeCompletedDirty` methods.

src/autoskillit/fleet/state.py
Trecek: Investigated — this is intentional. The docstring explicitly states the design intent: 'Infrastructure failures (e.g. fleet_l2_no_result_block) represent transient L2 disconnections and do not halt the campaign.' ALL `fleet_l2_no_result_block` records are infrastructure (transient) failures, regardless of whether their specific combination of conditions qualifies for RESUMABLE status. The campaign loop will handle retry via the next orchestrator reload cycle.

src/autoskillit/fleet/_api.py
Trecek: Investigated — this is intentional. The invariant holds: an L2 session sidecar can only be written if the L2 process actually started execution (i.e., `lifespan_started=True`). If `sidecar_exists=True` but `lifespan_started=False`, that would indicate an impossible state — the sidecar writer runs after session startup. The three conditions are correctly ordered: `session_id + lifespan_started + sidecar_exists` provides defense-in-depth, not redundancy.

tests/fleet/test_api.py
```diff
@@ -88,3 +88,38 @@ async def _raise_cancelled(**_kwargs):
...
+    ) -> None:
+        from tests.fleet._helpers import _setup_dispatch
+
+        _setup_dispatch(tool_ctx, monkeypatch)
+
+        captured_kwargs: list[dict] = []
+
+        async def _capture(**kwargs):
+            captured_kwargs.append(kwargs)
+            raise asyncio.CancelledError
```
Trecek, line 110: [warning] tests: The test uses `raise asyncio.CancelledError` inside `_capture` to short-circuit execution, which means `execute_dispatch` never completes normally. The assertion only verifies kwarg forwarding to `_run_dispatch`, but silently swallows the CancelledError. The 'was never called' guard at L124 is the only meaningful assertion — consider renaming or documenting the cancellation trick so future readers understand why CancelledError is raised intentionally.

tests/fleet/test_dispatch_failure_semantics.py
```diff
@@ -265,6 +265,47 @@ async def test_completed_dirty_writes_state_with_reason_l2_parse_failed(
...
+            "autoskillit.fleet._api.parse_l2_result_block",
+            lambda **_: _make_completed_dirty(),
+        )
+
+        result = await _run(tool_ctx)
+        assert "dispatch_status" in result
+        assert result["dispatch_status"] == "failure"
+
+    @pytest.mark.anyio
+    async def test_envelope_includes_dispatch_status_on_no_sentinel(self, tool_ctx, monkeypatch):
```
Trecek, line 296: [warning] tests: test_envelope_includes_dispatch_status_on_no_sentinel exercises only the degenerate no_sentinel→FAILURE path (no session_id, lifespan_started=False). The RESUMABLE path of no_sentinel (session_id set + lifespan_started=True + sidecar_exists=True) is not exercised in this class — only tested via full mock in test_campaign_state_write_uses_envelope_dispatch_status. A no_sentinel-that-becomes-RESUMABLE envelope test would close this coverage gap.

tests/server/test_tools_dispatch.py
```diff
@@ -722,6 +722,63 @@ def _raise_error(session_id: str) -> bool:
...
+        recipe="test-recipe",
+        task="do-work",
+        resume_session_id="sess-resume-123",
+    )
+
+    assert executor.dispatch_calls, "dispatch_food_truck executor was never called"
+    assert executor.dispatch_calls[0].resume_session_id == "sess-resume-123"
+
+
+@pytest.mark.anyio
```
Trecek: [warning] tests: test_dispatch_food_truck_executor_records_resume_session_id duplicates coverage of test_dispatch_food_truck_tool_passes_resume_session_id_to_executor (L725). Both tests wire the same InMemoryHeadlessExecutor and assert executor.dispatch_calls[0].resume_session_id. The integration-level MCP-tool test already validates the full path; the lower-level test adds minimal signal.

Trecek (commented):

Trecek (commented):

tests/fleet/test_api.py
Trecek: Agreed — fixed in 6d8fced8. Added a docstring to the test explaining that CancelledError is raised intentionally to short-circuit execute_dispatch after kwarg capture, without needing a full dispatch chain.

tests/fleet/test_dispatch_failure_semantics.py
Trecek: Agreed — fixed in 4f015c8d. Added test_envelope_includes_dispatch_status_on_no_sentinel_resumable to TestDispatchStatusEnvelopeField. Uses _SpawningExecutor to fire on_spawn (driving PENDING→RUNNING), sets session_id + lifespan_started=True, creates the sidecar file, and asserts dispatch_status='resumable'.

tests/server/test_tools_dispatch.py
Trecek: Agreed — fixed in ad1b7ee8. Removed test_dispatch_food_truck_executor_records_resume_session_id; the MCP-tool level test (L725) already exercises the full dispatch_food_truck→execute_dispatch→executor path and makes the lower-level test redundant.

Issue #1834, opened 2026-05-04
Add subpackage-aware cascade for execution/merge_queue/ in test path filter


Trecek:
## Summary

`execution/merge_queue/` is a subpackage (directory with `__init__.py`), but
`build_test_scope` classifies its files at the stem level without accounting for
subpackage depth. All four files (`__init__.py`, `_merge_queue_classifier.py`,
`_merge_queue_group_ci.py`, `_merge_queue_repo_state.py`) fall through to the
full `LAYER_CASCADE_CONSERVATIVE["execution"]` cascade — 10 test directories —
instead of the narrow `{"execution"}` cascade they warrant.

**Measured cost:** Sessions changing only `execution/merge_queue/*.py` (e.g.,
session e5fda810, v0.9.359) saw ~7.5% test reduction instead of the expected
~30–40%. Estimated waste: ~50–70s per affected test run.

## Root Cause

`_file_to_package()` returns `'execution'` for all paths under
`src/autoskillit/execution/merge_queue/` — it only resolves one directory level
below `autoskillit/`. The `execution` branch in `build_test_scope` then calls
`Path(f).stem` to look up `MODULE_CASCADE_EXECUTION`. The stems produced are:

| File | stem | In MODULE_CASCADE_EXECUTION? |
|------|------|------------------------------|
| `execution/merge_queue/__init__.py` | `__init__` | No → falls through |
| `execution/merge_queue/_merge_queue_classifier.py` | `_merge_queue_classifier` | No → falls through |
| `execution/merge_queue/_merge_queue_group_ci.py` | `_merge_queue_group_ci` | No → falls through |
| `execution/merge_queue/_merge_queue_repo_state.py` | `_merge_queue_repo_state` | No → falls through |

The existing `"merge_queue": frozenset({"execution"})` entry in
`MODULE_CASCADE_EXECUTION` only matches a hypothetical flat
`execution/merge_queue.py` file, which does not exist. It is dead code.

Fallthrough cascade: `execution, core, workspace, migration, server, cli, infra,
skills, _llm_triage, smoke_utils` — nearly the full suite.

## Requirements

REQ-MQ-001: `build_test_scope` in `tests/_test_filter.py` MUST detect when a
changed file path contains a known execution subpackage segment (e.g.,
`execution/merge_queue/`) and apply a dedicated narrow cascade for that subpackage
rather than falling through to `LAYER_CASCADE_CONSERVATIVE["execution"]`.

[...]

PR #1845, opened 2026-05-04
feat: add subpackage-aware cascade for execution/merge_queue/ in test path filter

Trecek:
## Summary

`build_test_scope` in `tests/_test_filter.py` cannot route `execution/merge_queue/` subpackage files to their correct narrow cascade because it only looks up filename stems in `MODULE_CASCADE_EXECUTION`. Subpackage-internal files like `_merge_queue_classifier.py` have stems not present in the map, so they fall through to the full 10-directory `LAYER_CASCADE_CONSERVATIVE["execution"]` cascade.

The fix introduces a companion mapping `SUBPKG_CASCADE_EXECUTION` that maps execution subpackage directory names to their cascade, a helper `_file_to_execution_subpkg()` to detect subpackage membership, and priority checks in both the main routing loop and the closure expansion path of `build_test_scope`.

## Requirements

REQ-MQ-001: `build_test_scope` in `tests/_test_filter.py` MUST detect when a
changed file path contains a known execution subpackage segment (e.g.,
`execution/merge_queue/`) and apply a dedicated narrow cascade for that subpackage
rather than falling through to `LAYER_CASCADE_CONSERVATIVE["execution"]`.

REQ-MQ-002: The narrow cascade for `execution/merge_queue/` files MUST map to
`frozenset({"execution"})` — matching the import consumers of
`autoskillit.execution.merge_queue` confirmed by AST graph
(`_build_execution_module_reverse_graph` returns `{"execution"}` for `merge_queue`).

REQ-MQ-003: `MODULE_CASCADE_EXECUTION` MUST be extended (or a companion mapping
`MODULE_CASCADE_EXECUTION_SUBPKG: dict[str, frozenset[str]]` introduced) to hold
the subpackage-to-cascade mapping so the existing AST guard tests in
`tests/arch/test_cascade_map_guard.py` can validate correctness without
modification.

REQ-MQ-004: The stale `"merge_queue": frozenset({"execution"})` entry in
`MODULE_CASCADE_EXECUTION` that only matches the non-existent flat
`execution/merge_queue.py` MUST be removed or repurposed to avoid phantom-stem
failures in `test_module_cascade_execution_has_no_phantom_stems` — the AST graph
currently happens to return `{"execution"}` for `merge_queue` because external
code imports `autoskillit.execution.merge_queue` (the package), so the entry is
not a phantom today, but its routing semantics are wrong.

REQ-MQ-005: The closure expansion path for `execution/__init__.py`
(`_expand_reexport_closure`) MUST handle the case where the re-exported symbol
resolves to an execution subpackage (e.g., `merge_queue`) and apply the narrow
cascade rather than falling through.

REQ-MQ-006: New tests in `tests/test_test_filter_execution_cascade.py` MUST cover
the following scenarios:
- Changing `execution/merge_queue/__init__.py` → scope contains `execution` only
  (not `core`, `cli`, `server`, `workspace`, `migration`)
- Changing `execution/merge_queue/_merge_queue_classifier.py` → same narrow scope
- Changing both a `merge_queue/` file and a non-subpackage execution file
  (e.g., `headless.py`) → fails open to full `LAYER_CASCADE_CONSERVATIVE["execution"]`

Closes #1834

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260504-135055-400702/.autoskillit/temp/make-plan/add_subpackage_aware_cascade_execution_merge_queue_plan_2026-05-04_140100.md`

## Token Usage Summary

| Step | uncached | output | cache_read | peak_ctx | turns | cache_write | time |
|------|----------|--------|------------|----------|-------|-------------|------|
| plan | 2.6k | 16.4k | 818.7k | 63.7k | 55 | 65.5k | 6m 56s |
| verify | 38 | 19.2k | 538.2k | 72.2k | 81 | 59.1k | 8m 14s |
| implement | 1.7k | 12.5k | 1.2M | 61.7k | 80 | 52.5k | 3m 54s |
| prepare_pr | 52 | 6.1k | 163.7k | 36.6k | 19 | 26.6k | 1m 33s |
| compose_pr | 75 | 3.2k | 247.5k | 31.6k | 19 | 18.7k | 1m 16s |
| review_pr | 132 | 24.2k | 745.6k | 66.1k | 46 | 55.7k | 7m 3s |
| resolve_review | 287 | 16.5k | 1.7M | 65.1k | 85 | 53.2k | 5m 41s |
| **Total** | 4.9k | 98.1k | 5.4M | 72.2k | | 331.3k | 34m 40s |

## Token Efficiency

| Step | LoC Changed | peak_ctx/LoC | cache_read/LoC | cache_write/LoC | output/LoC |
|------|-------------|--------------|----------------|-----------------|------------|
| plan | 0 | — | — | — | — |
| verify | 0 | — | — | — | — |
| implement | 208 | 296.7 | 5907.0 | 252.5 | 59.9 |
| prepare_pr | 0 | — | — | — | — |
| compose_pr | 0 | — | — | — | — |
| review_pr | 0 | — | — | — | — |
| resolve_review | 2 | 32553.0 | 853671.5 | 26596.0 | 8227.0 |
| **Total** | **210** | 343.9 | 25950.7 | 1577.6 | 467.2 |

### DIFF
--- a/tests/_test_filter.py
+++ b/tests/_test_filter.py
@@ -1181,6 +1181,7 @@ def build_test_scope(
                                     if _file_to_package(c) == "execution"
                                     and Path(c).stem != "__init__"
                                 ]
+                                # empty list → fail-open: only execution/__init__ changed, no cause to narrow on
                                 all_narrow = bool(exec_cause_files)
                                 narrow_dirs: set[str] = set()
                                 for c in exec_cause_files:
--- a/tests/_test_filter.py
+++ b/tests/_test_filter.py
@@ -1034,6 +1034,7 @@ def _file_to_execution_subpkg(filepath: str) -> str | None:
         idx = parts.index("autoskillit")
     except ValueError:
         return None
+    # idx + 3 < len(parts) ensures parts[idx + 2] is a subpackage dir (not a bare module) and guards parts[idx + 1].
     if idx + 3 < len(parts) and parts[idx + 1] == "execution":
         return parts[idx + 2]
     return None
--- a/tests/_test_filter.py
+++ b/tests/_test_filter.py
@@ -176,7 +176,6 @@ MODULE_CASCADE_EXECUTION: dict[str, frozenset[str]] = {
     "clone_guard": frozenset({"execution"}),
     # --- Medium: execution + specific file-level consumers ---
     "ci": frozenset({"execution"}),
-    "merge_queue": frozenset({"execution"}),
     "diff_annotator": frozenset({"execution", "test_smoke_utils.py"}),
     "pr_analysis": frozenset({"execution", "test_smoke_utils.py"}),
     "testing": frozenset(
@@ -254,6 +253,10 @@ MODULE_CASCADE_EXECUTION: dict[str, frozenset[str]] = {
         }
     ),
     # headless and process are NOT listed — they fall through to cascade_map["execution"]
+}
+
+SUBPKG_CASCADE_EXECUTION: dict[str, frozenset[str]] = {
+    "merge_queue": frozenset({"execution"}),
 }
 
 MODULE_CASCADE_RECIPE: dict[str, frozenset[str]] = {
@@ -1019,6 +1022,23 @@ def _file_to_package(filepath: str) -> str | None:
     return None
 
 
+def _file_to_execution_subpkg(filepath: str) -> str | None:
+    """Return the execution subpackage name if *filepath* is inside one, else None.
+
+    e.g. 'src/autoskillit/execution/merge_queue/_classifier.py' -> 'merge_queue'
+         'src/autoskillit/execution/headless.py' -> None
+         'src/autoskillit/execution/__init__.py' -> None
+    """
+    parts = Path(filepath).parts
+    try:
+        idx = parts.index("autoskillit")
+    except ValueError:
+        return None
+    if idx + 3 < len(parts) and parts[idx + 1] == "execution":
+        return parts[idx + 2]
+    return None
+
+
 def build_test_scope(
     changed_files: set[str] | None,
     mode: FilterMode,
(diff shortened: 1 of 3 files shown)


tests/_test_filter.py
```diff
@@ -1019,6 +1022,23 @@ def _file_to_package(filepath: str) -> str | None:
...
+    e.g. 'src/autoskillit/execution/merge_queue/_classifier.py' -> 'merge_queue'
+         'src/autoskillit/execution/headless.py' -> None
+         'src/autoskillit/execution/__init__.py' -> None
+    """
+    parts = Path(filepath).parts
+    try:
+        idx = parts.index("autoskillit")
+    except ValueError:
+        return None
+    if idx + 3 < len(parts) and parts[idx + 1] == "execution":
```
Trecek, line 1038: [warning] cohesion: `_file_to_execution_subpkg` checks `idx + 3 < len(parts)` which guarantees `idx + 1 < len(parts)` is also true, but the safety of `parts[idx + 1]` is only implicit. The compound guard is correct but non-obvious — consider reordering the sub-checks or adding a comment explaining why `idx + 3 < len(parts)` also covers the `idx + 1` access.

tests/_test_filter.py
```diff
@@ -1144,25 +1168,37 @@ def build_test_scope(
...
+                        else:
+                            stem = Path(f).stem
+                            if stem == "__init__":
+                                exec_cause_files = [
+                                    c
+                                    for c in changed_src_py
+                                    if _file_to_package(c) == "execution"
+                                    and Path(c).stem != "__init__"
+                                ]
+                                all_narrow = bool(exec_cause_files)
```
Trecek, line 1185: [warning] cohesion: `all_narrow` is initialised to `bool(exec_cause_files)`, so an empty cause-file set immediately evaluates as `False` and falls through to the full cascade. Verify this edge-case is intentional: if `execution/__init__.py` changes but no other execution files change alongside it, the expected behaviour is fail-open (full cascade), not trivially narrow.

Trecek (commented):

Trecek (commented):

tests/_test_filter.py
Trecek: Agreed — fixed in 3416392e. The `idx + 3` check is intentionally stricter than `idx + 2`: it ensures `parts[idx + 2]` is a subpackage *directory* (with something after it), not a bare module file. Added a one-line comment explaining this invariant and the implicit `idx + 1` guard.

tests/_test_filter.py
Trecek: Agreed — fixed in c4c4ee21. The fail-open behavior is intentional: when `exec_cause_files` is empty (only `execution/__init__.py` changed with no sibling execution files), we have no cause files to narrow on and must fall open to the full execution cascade. This mirrors the identical pattern already documented for `recipe/__init__.py`. Added a comment to make the intent explicit.

Issue #1774, opened 2026-05-04
[P3-A2-WP2] Add provider_name and provider_fallback_env parameters to _execute_claude_headless signature


Trecek:
## Goal

Extend the _execute_claude_headless async function signature with two new keyword-only parameters (provider_name: str = '' and provider_fallback_env: dict[str, str] | None = None) and initialize two loop-state local variables (current_provider_name and fallback_activated) immediately before the runner() invocation block. No behavioral change is introduced; this WP only widens the function's public API and prepares state slots consumed by WP3 and WP4.

## Context

- Phase: Fallback FSM and Provider Observability (Milestone: 3-fallback-fsm-and-provider-observability)
- Assignment: P3-A2 (Implement provider fallback attempt loop in _execute_claude_headless: STALE/BUDGET_EXHAUSTED detection, provider swap, fallback_activated flag)
- Depends on: #1773
- Depended on by: #1775, #1779

## Deliverables

- `src/autoskillit/execution/headless/__init__.py`
- `src/autoskillit/core/types/_type_protocols_execution.py`

## Technical Steps

1. Open src/autoskillit/execution/headless/__init__.py and locate the _execute_claude_headless async function definition.
2. Append two new keyword-only parameters to the function signature after write_watch_dirs: provider_name: str = '' and provider_fallback_env: dict[str, str] | None = None.
3. In the function body, immediately before the existing 'runner = ctx.runner' line, add two loop-state variable initializations: current_provider_name: str = provider_name and fallback_activated: bool = False.
4. Do not change any other logic, the runner() call block, exception handlers, result-building code, or session-log flushing code — all behavioral changes belong to WP3 and WP4.
5. Do NOT update run_headless_core or DefaultHeadlessExecutor.run call sites yet — those are the responsibility of P3-A3-WP1.
6. Add `is_feature_enabled` to the existing autoskillit.core import in headless/__init__.py without introducing a new import statement.
7. Compute `remaining_attempts` from config: `remaining_attempts = ctx.config.providers.max_provider_fallback_retries if provider_fallback_env is not None else 0`.
8. Wrap the try/except block containing the runner() call AND the subsequent _build_skill_result() call plus CONTRACT_NUDGE block and clone contamination check inside a `while True:` loop.
[...]

PR #1857, opened 2026-05-04
[P3-A2-WP2] Add provider_name and provider_fallback_env parameters to _execute_claude_headless signature

Trecek:
## Summary

Extend `_execute_claude_headless` with two new keyword-only parameters (`provider_name: str = ""` and `provider_fallback_env: dict[str, str] | None = None`), wrap the runner try/except and post-processing block in a `while True:` provider fallback loop, stamp `provider_used` and `provider_fallback` on all returned `SkillResult` values, and thread the new parameters through `run_headless_core`, `DefaultHeadlessExecutor.run`, and the `HeadlessExecutor` protocol.

Closes #1774

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260504-160108-875870/.autoskillit/temp/make-plan/p3_a2_wp2_provider_fallback_plan_2026-05-04_161600.md`

## Token Usage Summary

| Step | uncached | output | cache_read | peak_ctx | turns | cache_write | time |
|------|----------|--------|------------|----------|-------|-------------|------|
| plan | 58 | 13.7k | 1.3M | 78.6k | 86 | 65.9k | 7m 19s |
| verify | 36 | 7.6k | 781.0k | 51.8k | 99 | 38.7k | 6m 1s |
| implement | 384 | 36.6k | 3.1M | 107.1k | 131 | 94.4k | 12m 19s |
| fix | 280 | 19.1k | 2.5M | 99.7k | 90 | 89.4k | 15m 52s |
| prepare_pr | 52 | 4.4k | 174.3k | 39.0k | 17 | 26.9k | 1m 8s |
| compose_pr | 49 | 1.7k | 153.7k | 29.5k | 14 | 14.2k | 1m 48s |
| **Total** | 859 | 83.2k | 8.0M | 107.1k | | 329.6k | 44m 29s |

## Token Efficiency

| Step | LoC Changed | peak_ctx/LoC | cache_read/LoC | cache_write/LoC | output/LoC |
|------|-------------|--------------|----------------|-----------------|------------|
| plan | 0 | — | — | — | — |
| verify | 0 | — | — | — | — |
| implement | 474 | 225.9 | 6604.3 | 199.2 | 77.2 |
| fix | 37 | 2694.9 | 66839.8 | 2415.2 | 516.6 |
| prepare_pr | 0 | — | — | — | — |
| compose_pr | 0 | — | — | — | — |
| **Total** | **511** | 209.6 | 15730.3 | 644.9 | 162.8 |

### DIFF
--- a/src/autoskillit/core/types/_type_protocols_execution.py
+++ b/src/autoskillit/core/types/_type_protocols_execution.py
@@ -80,6 +80,8 @@ class HeadlessExecutor(Protocol):
         requires_packs: Sequence[str] = (),
         on_spawn: Callable[[int, int], None] | None = None,
         allowed_write_prefix: str = "",
+        provider_name: str = "",
+        provider_fallback_env: dict[str, str] | None = None,
     ) -> SkillResult: ...
 
 
--- a/src/autoskillit/core/types/_type_protocols_execution.py
+++ b/src/autoskillit/core/types/_type_protocols_execution.py
@@ -55,6 +55,8 @@ class HeadlessExecutor(Protocol):
         write_watch_dirs: Sequence[Path] = (),
         provider_extras: Mapping[str, str] | None = None,
         profile_name: str = "",
+        provider_name: str = "",
+        provider_fallback_env: dict[str, str] | None = None,
     ) -> SkillResult: ...
 
     async def dispatch_food_truck(
--- a/src/autoskillit/core/types/_type_protocols_execution.py
+++ b/src/autoskillit/core/types/_type_protocols_execution.py
@@ -57,6 +57,7 @@ class HeadlessExecutor(Protocol):
         profile_name: str = "",
         provider_name: str = "",
         provider_fallback_env: dict[str, str] | None = None,
+        provider_fallback_name: str = "",
     ) -> SkillResult: ...
 
     async def dispatch_food_truck(
--- a/src/autoskillit/core/types/_type_protocols_execution.py
+++ b/src/autoskillit/core/types/_type_protocols_execution.py
@@ -83,6 +83,7 @@ class HeadlessExecutor(Protocol):
         allowed_write_prefix: str = "",
         provider_name: str = "",
         provider_fallback_env: dict[str, str] | None = None,
+        provider_fallback_name: str = "",
     ) -> SkillResult: ...
 
 
--- a/src/autoskillit/execution/headless/__init__.py
+++ b/src/autoskillit/execution/headless/__init__.py
@@ -1,9 +1,5 @@
 """Headless Claude Code session orchestration.
 
-IL-1 module (execution/). Owns the full lifecycle of a headless claude CLI session:
-command preparation, subprocess invocation via the injected runner, and
-SkillResult construction.
-
 Public API:
     run_headless_core(skill_command, cwd, ctx, *, ...) -> SkillResult
 """
@@ -135,14 +131,6 @@ def _resolve_model(step_model: str, config: AutomationConfig) -> str | None:
 
 
 def _derive_step_name_from_skill_command(skill_command: str) -> str:
-    """Extract a recording step name from a skill command string.
-
-    Examples:
-        "/autoskillit:smoke-task arg1" -> "smoke-task"
-        "/investigate foo"             -> "investigate"
-        "/autoskillit:make-plan"       -> "make-plan"
-        ""                             -> ""
-    """
     stripped = skill_command.strip()
     if not stripped:
         return ""
@@ -153,7 +141,6 @@ def _derive_step_name_from_skill_command(skill_command: str) -> str:
 
 
 def _resolve_skill_temp_dir(cwd: str, skill_command: str) -> Path | None:
-    """Resolve the skill-specific temp directory for fs write detection."""
     name = extract_skill_name(skill_command)
     if not name:
         return None
@@ -217,13 +204,6 @@ async def _execute_claude_headless(
     provider_name: str = "",
     provider_fallback_env: dict[str, str] | None = None,
 ) -> SkillResult:
-    """Shared subprocess execution for headless Claude sessions.
-
-    Accepts an already-built ClaudeHeadlessCmd and handles runner invocation,
-    exception handling, _build_skill_result, and session log flushing.
-    Used by both run_headless_core (skill session path) and
-    DefaultHeadlessExecutor.dispatch_food_truck (food truck path).
-    """
     campaign_id = campaign_id or os.environ.get(CAMPAIGN_ID_ENV_VAR, "")
     dispatch_id = dispatch_id or os.environ.get(DISPATCH_ID_ENV_VAR, "")
 
@@ -418,9 +398,6 @@ async def _execute_claude_headless(
             fs_writes_detected=_fs_writes_detected,
         )
 
-        # CONTRACT NUDGE: lightweight resume recovery before full retry.
-        # Fires only when _build_skill_result returns CONTRACT_RECOVERY with a
-        # valid session_id (budget-exhausted cases have retry_reason=BUDGET_EXHAUSTED).
         if (
             skill_result.retry_reason == RetryReason.CONTRACT_RECOVERY
             and skill_result.needs_retry
@@ -465,8 +442,6 @@ async def _execute_claude_headless(
 
     _metrics = _compute_post_session_metrics(cwd, _pre_session_sha, skill_result)
 
-    # Use monotonic elapsed_seconds — authoritative wall-clock timing set by time.monotonic()
-    # brackets in run_managed_async. Never re-derive from ISO strings (backward-clock risk).
     timing_seconds: float = result.elapsed_seconds
 
     # Extract the audit record (if any) added by this session
@@ -598,11 +573,7 @@ async def run_headless_core(
     provider_name: str = "",
     provider_fallback_env: dict[str, str] | None = None,
 ) -> SkillResult:
-    """Shared headless runner used by run_skill.
-
-    Does NOT check open_kitchen gate — callers in server.py are responsible.
-    Accepts explicit ToolContext so this module has no server.py dependency.
-    """
+    """Does NOT check open_kitchen gate — callers in server.py are responsible."""
     cfg = ctx.config.run_skill
     effective_marker = completion_marker or cfg.completion_marker
     original_skill_command = skill_command
--- a/src/autoskillit/execution/headless/__init__.py
+++ b/src/autoskillit/execution/headless/__init__.py
@@ -1,4 +1,4 @@
-"""Headless Claude Code session orchestration.
+"""Headless Claude Code session orchestration. IL-1 module (execution/).
 
 Public API:
     run_headless_core(skill_command, cwd, ctx, *, ...) -> SkillResult
--- a/src/autoskillit/execution/headless/__init__.py
+++ b/src/autoskillit/execution/headless/__init__.py
@@ -1,4 +1,8 @@
-"""Headless Claude Code session orchestration. IL-1 module (execution/).
+"""Headless Claude Code session orchestration.
+
+IL-1 module (execution/). Owns the full lifecycle of a headless claude CLI session:
+command preparation, subprocess invocation via the injected runner, and
+SkillResult construction.
 
 Public API:
     run_headless_core(skill_command, cwd, ctx, *, ...) -> SkillResult
@@ -131,6 +135,14 @@ def _resolve_model(step_model: str, config: AutomationConfig) -> str | None:
 
 
 def _derive_step_name_from_skill_command(skill_command: str) -> str:
+    """Extract a recording step name from a skill command string.
+
+    Examples:
+        "/autoskillit:smoke-task arg1" -> "smoke-task"
+        "/investigate foo"             -> "investigate"
+        "/autoskillit:make-plan"       -> "make-plan"
+        ""                             -> ""
+    """
     stripped = skill_command.strip()
     if not stripped:
         return ""
@@ -204,6 +216,13 @@ async def _execute_claude_headless(
     provider_name: str = "",
     provider_fallback_env: dict[str, str] | None = None,
 ) -> SkillResult:
+    """Shared subprocess execution for headless Claude sessions.
+
+    Accepts an already-built ClaudeHeadlessCmd and handles runner invocation,
+    exception handling, _build_skill_result, and session log flushing.
+    Used by both run_headless_core (leaf path) and
+    DefaultHeadlessExecutor.dispatch_food_truck (food truck path).
+    """
     campaign_id = campaign_id or os.environ.get(CAMPAIGN_ID_ENV_VAR, "")
     dispatch_id = dispatch_id or os.environ.get(DISPATCH_ID_ENV_VAR, "")
 
@@ -228,9 +247,7 @@ async def _execute_claude_headless(
 
     current_provider_name: str = provider_name
     fallback_activated: bool = False
-    remaining_attempts = (
-        ctx.config.providers.provider_retry_limit if provider_fallback_env is not None else 0
-    )
+    remaining_attempts = ctx.config.providers.provider_retry_limit if provider_fallback_env else 0
 
     runner = ctx.runner
     if runner is None:
@@ -267,6 +284,7 @@ async def _execute_claude_headless(
 
     _pre_session_sha = _capture_git_head_sha(cwd)
     _result: SubprocessResult | None = None
+    result: SubprocessResult  # assigned before every break; exception paths return/raise
     skill_result: SkillResult
     while True:
         try:
@@ -328,7 +346,11 @@ async def _execute_claude_headless(
                 skill_command=skill_command,
                 order_id=order_id,
             )
-            return dataclasses.replace(_crashed, provider_used=current_provider_name)
+            return dataclasses.replace(
+                _crashed,
+                provider_used=current_provider_name,
+                provider_fallback=fallback_activated,
+            )
         except BaseException:
             logger.warning("headless_runner_cancelled", exc_info=True)
             _exc_text = traceback.format_exc()
@@ -428,13 +450,12 @@ async def _execute_claude_headless(
 
         if (
             skill_result.retry_reason in {RetryReason.STALE, RetryReason.BUDGET_EXHAUSTED}
-            and current_provider_name
             and provider_fallback_env is not None
             and remaining_attempts > 0
             and is_feature_enabled("providers", ctx.config.features)
         ):
-            spec = dataclasses.replace(spec, env={**spec.env, **provider_fallback_env})
-            current_provider_name = "anthropic"
+            if not fallback_activated:
+                spec = dataclasses.replace(spec, env={**spec.env, **provider_fallback_env})
             fallback_activated = True
             remaining_attempts -= 1
             continue
@@ -573,7 +594,11 @@ async def run_headless_core(
     provider_name: str = "",
     provider_fallback_env: dict[str, str] | None = None,
 ) -> SkillResult:
-    """Does NOT check open_kitchen gate — callers in server.py are responsible."""
+    """Shared headless runner used by run_skill.
+
+    Does NOT check open_kitchen gate — callers in server.py are responsible.
+    Accepts explicit ToolContext so this module has no server.py dependency.
+    """
     cfg = ctx.config.run_skill
     effective_marker = completion_marker or cfg.completion_marker
     original_skill_command = skill_command
@@ -729,6 +754,8 @@ class DefaultHeadlessExecutor:
         requires_packs: Sequence[str] = (),
         on_spawn: Callable[[int, int], None] | None = None,
         allowed_write_prefix: str = "",
+        provider_name: str = "",
+        provider_fallback_env: dict[str, str] | None = None,
     ) -> SkillResult:
         cfg = self._ctx.config
         resolved_model = _resolve_model(model, cfg)
@@ -784,4 +811,6 @@ class DefaultHeadlessExecutor:
             completion_marker=completion_marker,
             on_spawn=on_spawn,
             skip_clone_guard=True,
+            provider_name=provider_name,
+            provider_fallback_env=provider_fallback_env,
         )
--- a/src/autoskillit/execution/headless/__init__.py
+++ b/src/autoskillit/execution/headless/__init__.py
@@ -34,6 +34,7 @@ from autoskillit.core import (
     collect_version_snapshot,
     extract_skill_name,
     get_logger,
+    is_feature_enabled,
     is_git_worktree,
     temp_dir_display_str,
 )
@@ -213,6 +214,8 @@ async def _execute_claude_headless(
     skip_clone_guard: bool = False,
     readonly_skill: bool = False,
     write_watch_dirs: Sequence[Path] = (),
+    provider_name: str = "",
+    provider_fallback_env: dict[str, str] | None = None,
 ) -> SkillResult:
     """Shared subprocess execution for headless Claude sessions.
 
@@ -242,6 +245,12 @@ async def _execute_claude_headless(
         else:
             _raw_idle = float(cfg.idle_output_timeout)
     effective_idle: float | None = _raw_idle if _raw_idle > 0.0 else None
+
+    current_provider_name: str = provider_name
+    fallback_activated: bool = False
+    remaining_attempts = (
+        ctx.config.providers.provider_retry_limit if provider_fallback_env is not None else 0
+    )
 
     runner = ctx.runner
     if runner is None:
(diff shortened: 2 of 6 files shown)


src/autoskillit/execution/headless/__init__.py
```diff
@@ -372,75 +321,127 @@ async def _execute_claude_headless(
...
+
+        if (
+            skill_result.retry_reason in {RetryReason.STALE, RetryReason.BUDGET_EXHAUSTED}
+            and current_provider_name
+            and provider_fallback_env is not None
+            and remaining_attempts > 0
+            and is_feature_enabled("providers", ctx.config.features)
+        ):
+            spec = dataclasses.replace(spec, env={**spec.env, **provider_fallback_env})
+            current_provider_name = "anthropic"
```
Trecek: [critical] arch: The fallback provider name is hardcoded as the string literal 'anthropic' (current_provider_name = "anthropic"). This couples fallback logic to a specific provider string rather than deriving it from fallback env or a dedicated parameter. If fallback credentials correspond to a different provider alias, provider_used will misattribute the provider.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -372,75 +321,127 @@ async def _execute_claude_headless(
...
+                skill_result,
+                cwd,
+                runner,
+                ctx.audit,
+                skill_command=skill_command,
+                readonly_skill=_readonly_skill,
+            )
+
+        if (
+            skill_result.retry_reason in {RetryReason.STALE, RetryReason.BUDGET_EXHAUSTED}
```
Trecek, line 451: [warning] arch: The fallback guard checks current_provider_name as a truthy condition (L431), suppressing fallback when provider_name is empty string. A caller passing provider_fallback_env but omitting provider_name (relying on the default "") will silently never activate fallback despite supplying valid fallback credentials.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -372,75 +321,127 @@ async def _execute_claude_headless(
...
+            and current_provider_name
+            and provider_fallback_env is not None
+            and remaining_attempts > 0
+            and is_feature_enabled("providers", ctx.config.features)
+        ):
+            spec = dataclasses.replace(spec, env={**spec.env, **provider_fallback_env})
```
Trecek: [warning] arch: On fallback activation, spec.env is rebuilt by merging provider_fallback_env each iteration. If remaining_attempts > 1, fallback env keys are merged repeatedly. The fallback_activated boolean only records that fallback happened, not how many times — provider_fallback will be True even after multiple retries with no semantic difference.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -372,75 +321,127 @@ async def _execute_claude_headless(
...
-    for _wd in _watch_dirs:
-        if _wd.is_dir():
+            except Exception:
+                logger.debug("flush_session_log during crash failed", exc_info=True)
+            _crashed = SkillResult.crashed(
+                exception=exc,
+                skill_command=skill_command,
+                order_id=order_id,
+            )
+            return dataclasses.replace(_crashed, provider_used=current_provider_name)
```
Trecek: [warning] arch: The crash handler returns early at L331 with dataclasses.replace(_crashed, provider_used=current_provider_name), but provider_fallback is not stamped here. The final dataclasses.replace at L530-L534 is bypassed on the crash path, so a crashed result will always have provider_fallback=False (default) even if fallback had been activated.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -243,6 +226,12 @@ async def _execute_claude_headless(
             _raw_idle = float(cfg.idle_output_timeout)
     effective_idle: float | None = _raw_idle if _raw_idle > 0.0 else None
 
+    current_provider_name: str = provider_name
+    fallback_activated: bool = False
+    remaining_attempts = (
+        ctx.config.providers.provider_retry_limit if provider_fallback_env is not None else 0
```
Trecek: [warning] defense: remaining_attempts is set from provider_retry_limit when provider_fallback_env is not None, but there is no guard for an empty dict {}. An empty dict passes the 'is not None' check, initiating retry attempts where the env merge at L436 is a no-op — exhausting attempts without applying any fallback.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -372,75 +321,127 @@ async def _execute_claude_headless(
...
+            fallback_activated = True
+            remaining_attempts -= 1
+            continue
+        break
+
     _metrics = _compute_post_session_metrics(cwd, _pre_session_sha, skill_result)
 
-    # Use monotonic elapsed_seconds — authoritative wall-clock timing set by time.monotonic()
-    # brackets in run_managed_async. Never re-derive from ISO strings (backward-clock risk).
     timing_seconds: float = result.elapsed_seconds
```
Trecek, line 467: [warning] bugs: result is assigned inside the while loop at L372 but referenced after loop exit here. pyright will flag result as potentially unbound since it is not initialized before the loop. Initialize result before the while True: block to fix type-checking failure.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -1,8 +1,4 @@
-"""Headless Claude Code session orchestration.
-
-IL-1 module (execution/). Owns the full lifecycle of a headless claude CLI session:
-command preparation, subprocess invocation via the injected runner, and
-SkillResult construction.
+"""Headless Claude Code session orchestration. IL-1 module (execution/).
```
Trecek: [warning] slop: Module docstring compressed from multi-line to single-line, losing IL-1 lifecycle context (command preparation, subprocess invocation, SkillResult construction). The IL-1 label description is load-bearing per project conventions (import-linter contracts reference IL-N labels in module docstrings).

src/autoskillit/execution/headless/__init__.py
```diff
@@ -134,14 +131,6 @@ def _resolve_model(step_model: str, config: AutomationConfig) -> str | None:
 
 
 def _derive_step_name_from_skill_command(skill_command: str) -> str:
```
Trecek, line 137: [warning] slop: _derive_step_name_from_skill_command lost its examples docstring with three concrete input-output mappings. Namespace prefix stripping behavior and empty-string handling are non-obvious and cannot be inferred from the function name.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -213,14 +201,9 @@ async def _execute_claude_headless(
     skip_clone_guard: bool = False,
     readonly_skill: bool = False,
     write_watch_dirs: Sequence[Path] = (),
+    provider_name: str = "",
+    provider_fallback_env: dict[str, str] | None = None,
 ) -> SkillResult:
```
Trecek, line 219: [warning] slop: _execute_claude_headless lost its docstring describing dual-caller usage (run_headless_core leaf path vs. DefaultHeadlessExecutor.dispatch_food_truck food truck path). This is non-obvious architectural context about who calls this function.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -557,12 +563,10 @@ async def run_headless_core(
...
     profile_name: str = "",
+    provider_name: str = "",
+    provider_fallback_env: dict[str, str] | None = None,
 ) -> SkillResult:
-    """Shared headless runner used by run_skill.
-
-    Does NOT check open_kitchen gate — callers in server.py are responsible.
-    Accepts explicit ToolContext so this module has no server.py dependency.
-    """
+    """Does NOT check open_kitchen gate — callers in server.py are responsible."""
```
Trecek: [warning] slop: run_headless_core docstring truncated to a single sentence, dropping the rationale that ToolContext is accepted explicitly to avoid a server.py dependency. The retained sentence is incomplete without the dependency-isolation rationale.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -689,6 +697,8 @@ async def run(
             write_watch_dirs=write_watch_dirs,
             provider_extras=provider_extras,
             profile_name=profile_name,
+            provider_name=provider_name,
```
Trecek, line 736: [warning] cohesion: DefaultHeadlessExecutor.run propagates provider_name/provider_fallback_env to run_headless_core, but DefaultHeadlessExecutor.dispatch_food_truck (parallel food truck execution path) does not appear to receive the same parameters. Verify dispatch_food_truck was intentionally excluded or add the parameters for symmetry.

Trecek (commented):

tests/execution/test_headless_provider_forwarding.py
```diff
@@ -126,3 +127,183 @@ async def fake_core(skill_command, cwd, ctx, **kwargs):
...
+        "autoskillit.execution.headless._compute_post_session_metrics",
+        lambda *a, **kw: PostSessionMetrics(0, 0, str(tmp_path)),  # noqa: ARG005
+    )
+    monkeypatch.setattr(
+        "autoskillit.execution.headless._capture_git_head_sha",
+        lambda *a: "",  # noqa: ARG005
+    )
+
+    result = await _execute_claude_headless(
+        _spec,
```
Trecek, line 290: [critical] tests: test_provider_name_stamps_provider_used_on_result does not verify that _STUB_RESULT starts with provider_used='' before the call. If _STUB_RESULT is a module-level fixture with a non-empty provider_used, the assertion at L298 passes vacuously regardless of whether the stamp logic at L530 runs. Construct a fresh result with known-empty provider_used to make the assertion load-bearing.

tests/execution/test_headless_provider_forwarding.py
```diff
@@ -126,3 +127,183 @@ async def fake_core(skill_command, cwd, ctx, **kwargs):
...
+        str(tmp_path),
+        minimal_ctx,
+        timeout=30.0,
+        stale_threshold=5.0,
+    )
+
+    assert result.provider_used == ""
```
Trecek, line 256: [warning] tests: test_no_fallback_env_returns_empty_provider_used asserts result.provider_used == '' but does not verify _STUB_RESULT.provider_used is '' before the call. If _STUB_RESULT has a non-empty provider_used, the assertion tests the stub's initial state rather than the implementation's stamping behavior.

tests/execution/test_headless_provider_forwarding.py
```diff
@@ -126,3 +127,183 @@ async def fake_core(skill_command, cwd, ctx, **kwargs):
...
+        timeout=30.0,
+        stale_threshold=5.0,
+        provider_name="bedrock",
+    )
+
+    assert result.provider_used == "bedrock"
+    assert result.provider_fallback is False
+
+
+def test_headless_executor_protocol_includes_provider_params() -> None:
```
Trecek, line 302: [warning] tests: test_headless_executor_protocol_includes_provider_params uses inspect.signature on a Protocol method, which only checks the Protocol definition (already verified separately), not that DefaultHeadlessExecutor or run_headless_core implement the parameters correctly. This test duplicates structural coverage provided by other tests in the file.

src/autoskillit/core/types/_type_protocols_execution.py
```diff
@@ -55,6 +55,8 @@ async def run(
         write_watch_dirs: Sequence[Path] = (),
         provider_extras: Mapping[str, str] | None = None,
         profile_name: str = "",
+        provider_name: str = "",
```
Trecek, line 58: [warning] cohesion: HeadlessExecutor protocol adds provider_name/provider_fallback_env to run() but dispatch_food_truck is not shown being updated. If dispatch_food_truck is a parallel execution path, omitting these parameters creates an asymmetric protocol surface where one path supports provider configuration and the other does not.

Trecek (commented):
AutoSkillit review: blocking issues found. See inline comments — changes required before merging.

**Verdict:** changes_requested

**Summary:**
- 1 critical blocking finding (test assertion validity in test_provider_name_stamps_provider_used_on_result)
- 11 warning findings (hardcoded provider name, fallback logic gaps, missing provider_fallback stamp on crash path, docstring removals)
- 3 findings requiring human decision (dispatch_food_truck symmetry, hardcoded 'anthropic' string)

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in 80f07cf3. Removed the hardcoded 'anthropic' assignment; `current_provider_name` is now kept unchanged on fallback activation, with `provider_fallback=True` as the unambiguous signal.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in 80f07cf3. Removed the `and current_provider_name` truthy condition from the fallback guard; fallback now activates solely on `provider_fallback_env is not None` + `remaining_attempts > 0` + feature flag, regardless of whether `provider_name` was supplied.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in 80f07cf3. Added `if not fallback_activated:` guard around the env merge so `spec.env` is built exactly once on first fallback activation. Subsequent retries skip the merge entirely; `fallback_activated` is already `True` from the first activation.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in 80f07cf3. The `dataclasses.replace` on the crash-path return now stamps both `provider_used=current_provider_name` and `provider_fallback=fallback_activated`, so a crashed result correctly reflects fallback state.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in 80f07cf3. Changed `provider_fallback_env is not None` to `bool(provider_fallback_env)` (falsy check) in the `remaining_attempts` initializer; an empty dict `{}` now correctly produces `remaining_attempts = 0`, preventing no-op retry loops.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in 80f07cf3. Added `result: SubprocessResult` forward declaration before the `while True:` block, satisfying pyright's potentially-unbound analysis while documenting the invariant (exception paths return/raise before any `break`).

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in 80f07cf3. Restored the multi-line module docstring with the IL-1 lifecycle description ('Owns the full lifecycle of a headless claude CLI session: command preparation, subprocess invocation via the injected runner, and SkillResult construction'). The IL-1 label is load-bearing for import-linter contracts.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in 80f07cf3. Restored the examples docstring with three concrete input→output mappings: namespace prefix stripping ('/autoskillit:smoke-task arg1' → 'smoke-task'), bare slash ('/ investigate foo' → 'investigate'), empty string handling.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in 80f07cf3. Restored the dual-caller docstring ('Used by both run_headless_core (leaf path) and DefaultHeadlessExecutor.dispatch_food_truck (food truck path)'), which is non-obvious architectural context not derivable from the function name.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in 80f07cf3. Restored the full run_headless_core docstring including 'Accepts explicit ToolContext so this module has no server.py dependency' — the dependency-isolation rationale is the load-bearing architectural fact the single-sentence form dropped.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in 80f07cf3. Added `provider_name: str = ''` and `provider_fallback_env: dict[str, str] | None = None` to `DefaultHeadlessExecutor.dispatch_food_truck` and propagated them to the underlying `_execute_claude_headless` call for symmetry with the `run` path.

src/autoskillit/core/types/_type_protocols_execution.py
Trecek: Agreed — fixed in 80f07cf3. Added `provider_name: str = ''` and `provider_fallback_env: dict[str, str] | None = None` to `HeadlessExecutor.dispatch_food_truck` protocol definition, matching the updated concrete implementation.

tests/execution/test_headless_provider_forwarding.py
Trecek: Investigated — this is intentional. The reviewer's concern is that _STUB_RESULT.provider_used could contaminate the assertion, but the stamp at lines 528-532 always calls `dataclasses.replace(skill_result, provider_used=current_provider_name, provider_fallback=fallback_activated)`, unconditionally overwriting `provider_used` regardless of _STUB_RESULT's initial value. The assertion `result.provider_used == 'bedrock'` tests that `current_provider_name` was correctly set from the `provider_name='bedrock'` argument — which is the exact behavior under test.

tests/execution/test_headless_provider_forwarding.py
Trecek: Investigated — this is intentional. Same code path as comment 3185250234: lines 528-532 unconditionally replace `provider_used` with `current_provider_name`. When no `provider_name` argument is passed, `current_provider_name = ''` (the default), so the final `result.provider_used == ''` assertion is load-bearing — it verifies the stamp produces the correct empty string, independent of _STUB_RESULT's initial state.

tests/execution/test_headless_provider_forwarding.py
Trecek: Valid observation — flagged for design decision. The test at line 302 checks `HeadlessExecutor.run` (the Protocol's definition), which has independent value from the concrete-class test at line 152 (`DefaultHeadlessExecutor.run`). The Protocol test documents the contract at the interface boundary; mypy/pyright enforcement happens at the concrete class. Whether to remove it for brevity or keep it for explicit Protocol contract documentation is a valid design question.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -278,73 +284,34 @@ async def _execute_claude_headless(
...
-            linux_tracing_config=linux_tracing_cfg,
-            idle_output_timeout=effective_idle,
-            max_suppression_seconds=cfg.max_suppression_seconds,
-            on_pid_resolved=on_spawn,
-        )
-    except Exception as exc:
-        logger.error("headless_runner_crashed", exc_info=True)
-        _exc_text = traceback.format_exc()
-        _log_dir = ctx.config.linux_tracing.log_dir
+    result: SubprocessResult  # assigned before every break; exception paths return/raise
```
Trecek: [warning] slop: Inline comment '# assigned before every break; exception paths return/raise' restates what the declaration annotation already communicates. Remove it.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -372,75 +339,130 @@ async def _execute_claude_headless(
...
+            )
+
+        if (
+            skill_result.retry_reason in {RetryReason.STALE, RetryReason.BUDGET_EXHAUSTED}
+            and provider_fallback_env is not None
+            and remaining_attempts > 0
+            and is_feature_enabled("providers", ctx.config.features)
+        ):
+            if not fallback_activated:
+                spec = dataclasses.replace(spec, env={**spec.env, **provider_fallback_env})
```
Trecek, line 457: [critical] bugs: current_provider_name is set to provider_name at L248 and never updated after fallback activation. After the fallback env is applied at L458, provider_used will still reflect the original provider rather than the fallback provider. Update current_provider_name when fallback is activated (e.g., by passing a provider_fallback_name companion parameter or deriving it from provider_fallback_env).

src/autoskillit/execution/headless/__init__.py
```diff
@@ -372,75 +339,130 @@ async def _execute_claude_headless(
...
+                exception=exc,
+                skill_command=skill_command,
+                order_id=order_id,
+            )
+            return dataclasses.replace(
+                _crashed,
+                provider_used=current_provider_name,
+                provider_fallback=fallback_activated,
+            )
+        except BaseException:
```
Trecek, line 353: [warning] defense: except BaseException catches KeyboardInterrupt and SystemExit in addition to asyncio.CancelledError. The block re-raises correctly, but a future edit between log and raise could accidentally swallow cancellation. Consider explicit except (asyncio.CancelledError, KeyboardInterrupt, SystemExit) to make intent explicit.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -243,6 +245,10 @@ async def _execute_claude_headless(
             _raw_idle = float(cfg.idle_output_timeout)
     effective_idle: float | None = _raw_idle if _raw_idle > 0.0 else None
 
+    current_provider_name: str = provider_name
+    fallback_activated: bool = False
+    remaining_attempts = ctx.config.providers.provider_retry_limit if provider_fallback_env else 0
```
Trecek, line 251: [warning] bugs: remaining_attempts is set from provider_retry_limit when provider_fallback_env is provided. If provider_retry_limit defaults to 0, no fallback attempts ever occur even with valid fallback env. Verify provider_retry_limit has a positive default or guard against non-positive values.

src/autoskillit/execution/headless/__init__.py
```diff
@@ -278,73 +284,34 @@ async def _execute_claude_headless(
...
-        _exc_text = traceback.format_exc()
-        _log_dir = ctx.config.linux_tracing.log_dir
-        try:
-            from autoskillit.execution import flush_session_log
+        except Exception as exc:
+            logger.error("headless_runner_crashed", exc_info=True)
+            _exc_text = traceback.format_exc()
+            _log_dir = ctx.config.linux_tracing.log_dir
+            try:
+                # Deferred: autoskillit.execution.__init__ imports headless.py (L39-42);
```
Trecek: [warning] slop: The comment explaining the deferred import (lines L311-L312) cross-references line numbers that are fragile and adds no actionable value — the deferred import pattern is self-evident. Remove it.

Trecek (commented):

Trecek (commented):
AutoSkillit review: blocking issues found (1 critical). See inline comments — changes required before merge.

**Verdict:** changes_requested

**Summary:**
- **1 critical** (bugs/L458): `current_provider_name` is never updated after fallback activation, so `provider_used` reflects the original provider even when fallback credentials were used.
- **2 warnings** (slop): Redundant inline comments at L287 and L311 should be removed.
- **1 warning** (defense/L354): `except BaseException` is broader than needed; prefer explicit cancellation types.
- **1 warning** (bugs/L250): If `provider_retry_limit` defaults to 0, no fallback attempts occur even when `provider_fallback_env` is provided — verify positive default.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in a2138748. Added `provider_fallback_name: str = ''` parameter to `_execute_claude_headless`, `run_headless_core`, `DefaultHeadlessExecutor.run`, `DefaultHeadlessExecutor.dispatch_food_truck`, and `HeadlessExecutor` protocol. When fallback activates for the first time, `current_provider_name` is now updated to `provider_fallback_name` if non-empty, so `provider_used` in the returned `SkillResult` reflects the fallback provider.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in a2138748. Removed the inline comment from the `result: SubprocessResult` forward declaration.

src/autoskillit/execution/headless/__init__.py
Trecek: Agreed — fixed in a2138748. Removed the deferred import comment with stale line-number cross-references; the circularity reason is already documented in `headless/CLAUDE.md`.

src/autoskillit/execution/headless/__init__.py
Trecek: Investigated — this is intentional. `except BaseException: ... raise` is the correct anyio idiom for cancellation cleanup. anyio supports multiple backends: in trio, cancellations arrive as `trio.Cancelled` (a `BaseException` subclass), not `asyncio.CancelledError`. Narrowing to `except (asyncio.CancelledError, KeyboardInterrupt, SystemExit)` would silently swallow trio cancellations. The current pattern correctly handles both backends and always re-raises.

src/autoskillit/execution/headless/__init__.py
Trecek: Investigated — this is intentional. `provider_retry_limit` defaults to `2` in `_config_dataclasses.py:317` and has a `__post_init__` guard at lines 320–321 that raises `ValueError` if the value is `< 1`. A zero default is structurally impossible; the guard ensures any configured value is positive before `_execute_claude_headless` is ever called.

src/autoskillit/core/types/_type_protocols_execution.py
```diff
@@ -78,6 +81,8 @@ async def dispatch_food_truck(
         requires_packs: Sequence[str] = (),
         on_spawn: Callable[[int, int], None] | None = None,
         allowed_write_prefix: str = "",
+        provider_name: str = "",
+        provider_fallback_env: dict[str, str] | None = None,
```
Trecek, line 85: [warning] cohesion: HeadlessExecutor.dispatch_food_truck Protocol is missing `provider_fallback_name: str = ""` — the concrete DefaultHeadlessExecutor.dispatch_food_truck at L757 declares all three provider params (provider_name, provider_fallback_env, provider_fallback_name), as does run() in both the Protocol (L60) and the concrete class (L700). The Protocol surface is inconsistent: callers typing against HeadlessExecutor cannot pass provider_fallback_name to dispatch_food_truck, making the param silently unreachable via the abstract interface.

src/autoskillit/core/types/_type_protocols_execution.py
```diff
@@ -78,6 +81,8 @@ async def dispatch_food_truck(
...
     ) -> SkillResult: ...
```
Trecek, line 87: [warning] defense: dispatch_food_truck protocol is missing the provider_fallback_name parameter that exists in the concrete DefaultHeadlessExecutor implementation (L757). Callers typed as HeadlessExecutor cannot pass provider_fallback_name, silently dropping it. The run() protocol at L60 correctly includes all three params; dispatch_food_truck should match.

Trecek (commented):

Trecek (commented):

src/autoskillit/core/types/_type_protocols_execution.py
Trecek: Agreed — fixed in 45fbd95b. Added `provider_fallback_name: str = ""` to the Protocol's `dispatch_food_truck` signature, matching `HeadlessExecutor.run` (L60) and `DefaultHeadlessExecutor.dispatch_food_truck` (L757). Protocol surface is now consistent across all three provider params.

src/autoskillit/core/types/_type_protocols_execution.py
Trecek: Agreed — fixed in 45fbd95b. Added `provider_fallback_name: str = ""` to `HeadlessExecutor.dispatch_food_truck` Protocol. The concrete implementation at L757 already had all three provider params; the Protocol now matches, so callers typed as `HeadlessExecutor` can pass `provider_fallback_name` without it being silently dropped.

Issue #1852, opened 2026-05-04
Planner skills suppressed by feature gate in headless L1 sessions


Trecek:
## Problem

All planner skills (`planner-elaborate-wps`, `planner-refine-assignments`, `planner-analyze`, `planner-extract-domain`, `planner-generate-phases`, `planner-elaborate-phase`, `planner-refine-phases`, `planner-consolidate-wps`, `planner-reconcile-deps`, `planner-refine`) have `categories: [planner]` in their SKILL.md frontmatter.

The `planner` feature in `FEATURE_REGISTRY` (`core/types/_type_constants.py:416`) has:
- `default_enabled=False`
- `lifecycle=FeatureLifecycle.EXPERIMENTAL`

When `init_session` builds the ephemeral skill directory for headless L1 sessions, `_is_skill_disabled` in `workspace/session_skills.py` suppresses all planner-category skills. The skills exist in `skills_extended/` on disk but are never written to the `--add-dir` path that Claude Code searches.

**Result:** The L2 orchestrator can invoke the planner recipe, but every L1 child session it spawns gets "Unknown skill: planner-*" when attempting to invoke these skills via the Skill tool. The AI then wastes 3-8 turns searching for the SKILL.md on disk before executing manually.

## Confirmed Current State (v0.9.390)

- `core/types/_type_constants.py`: `"planner": FeatureDef(lifecycle=FeatureLifecycle.EXPERIMENTAL, default_enabled=False, ...)`
- `.autoskillit/config.yaml`: No `features.planner: true` entry
- No `PACK_REGISTRY` entry for planner that would override the gate when the recipe is active

## Evidence

Observed in ~15 episodes across 145 headless sessions (May 1-4). Affected skills: `planner-elaborate-wps` (most common), `planner-elaborate-assignments`, `planner-refine-assignments`, `planner-analyze`, `planner-extract-domain`, `planner-generate-phases`. Also `dry-walkthrough` (referenced by planner skills but unavailable). Estimated ~60 wasted turns over 3 days.

## Proposed Fix (pick one)

**Option A (config, immediate):** Add `features.planner: true` and `experimental_enabled: true` to `.autoskillit/config.yaml`.

[...]

PR #1859, opened 2026-05-05
Rectify: Recipe-Level Feature Gate Enablement — Unified Visibility Axis

Trecek:
## Summary

The skill visibility system has two independent, uncoordinated gating axes: the **pack gate** (via `PACK_REGISTRY` + `_resolve_effective_disabled`) and the **feature gate** (via `FEATURE_REGISTRY` + `_is_skill_disabled`). The pack gate has a recipe-level override mechanism (`requires_packs` on `Recipe` flows through `ToolContext.active_recipe_packs` into `_resolve_effective_disabled` where it subtracts from the disabled set). The feature gate has no equivalent — `session_features` is derived exclusively from `config.features`, and recipes cannot declare feature requirements.

This asymmetry means any `FEATURE_REGISTRY` entry with `default_enabled=False` and non-empty `skill_categories` silently suppresses its skills in all headless L1 sessions, even when the active recipe dispatches those skills. The planner is the first (and currently only) feature hitting this gap, but the architecture makes it inevitable for any future feature following the same pattern.

The fix adds a `requires_features` field to the `Recipe` schema, flows it through the same path as `requires_packs`, and merges it into `session_features` in `init_session`. A new static validation rule catches recipes that reference feature-gated skills without declaring the corresponding feature requirement. An architecture test enforces the invariant that every feature with non-empty `skill_categories` has a tested recipe-level enablement path.

Closes #1852

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260504-163437-353221/.autoskillit/temp/rectify/rectify_planner_feature_gate_immunity_2026-05-04_163437.md`

### DIFF
--- a/src/autoskillit/recipe/io.py
+++ b/src/autoskillit/recipe/io.py
@@ -361,11 +361,15 @@ def _parse_recipe(data: dict[str, Any]) -> Recipe:
     requires_recipe_packs = data.get("requires_recipe_packs") or []
     allowed_recipes = data.get("allowed_recipes") or []
     continue_on_failure = bool(data.get("continue_on_failure", False))
-    requires_features_raw = data.get("requires_features") or []
+    _rf_val = data.get("requires_features")
+    requires_features_raw = _rf_val if _rf_val is not None else []
     if not isinstance(requires_features_raw, list):
         raise ValueError(
             f"'requires_features' must be a list, got {type(requires_features_raw).__name__!r}"
         )
+    if not all(isinstance(f, str) for f in requires_features_raw):
+        bad = [f for f in requires_features_raw if not isinstance(f, str)]
+        raise ValueError(f"'requires_features' entries must be strings, got: {bad!r}")
 
     return Recipe(
         name=name,
--- a/src/autoskillit/recipe/io.py
+++ b/src/autoskillit/recipe/io.py
@@ -361,6 +361,11 @@ def _parse_recipe(data: dict[str, Any]) -> Recipe:
     requires_recipe_packs = data.get("requires_recipe_packs") or []
     allowed_recipes = data.get("allowed_recipes") or []
     continue_on_failure = bool(data.get("continue_on_failure", False))
+    requires_features_raw = data.get("requires_features") or []
+    if not isinstance(requires_features_raw, list):
+        raise ValueError(
+            f"'requires_features' must be a list, got {type(requires_features_raw).__name__!r}"
+        )
 
     return Recipe(
         name=name,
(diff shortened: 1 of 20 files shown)


src/autoskillit/recipe/io.py
```diff
@@ -361,6 +361,11 @@ def _parse_recipe(data: dict[str, Any]) -> Recipe:
     requires_recipe_packs = data.get("requires_recipe_packs") or []
     allowed_recipes = data.get("allowed_recipes") or []
     continue_on_failure = bool(data.get("continue_on_failure", False))
+    requires_features_raw = data.get("requires_features") or []
```
Trecek: [warning] defense: `data.get("requires_features") or []` silently coerces any falsy non-None value (e.g., `False`, `0`, `""`) to `[]` without raising. The isinstance check at L365 never fires for those cases. Prefer an explicit None check: `if (val := data.get("requires_features")) is not None: requires_features_raw = val` followed by the isinstance guard, so only a missing key yields the default and any non-list non-None value is rejected.

src/autoskillit/recipe/io.py
```diff
@@ -379,6 +384,7 @@ def _parse_recipe(data: dict[str, Any]) -> Recipe:
         requires_recipe_packs=requires_recipe_packs,
         allowed_recipes=allowed_recipes,
         continue_on_failure=continue_on_failure,
+        requires_features=requires_features_raw,
```
Trecek, line 391: [warning] defense: `requires_features_raw` is `list[Any]` at this point — individual elements are never validated to be strings. A YAML entry like `requires_features: [1, true]` passes the isinstance check and reaches `Recipe.__post_init__` where int/bool keys are treated as unknown features rather than raising a parse error. Add an element-level check: `if not all(isinstance(f, str) for f in requires_features_raw): raise ValueError(f"requires_features entries must be strings")`.

Trecek (commented):

Trecek (commented):

src/autoskillit/recipe/io.py
Trecek: Agreed — fixed in 6b27bee027f01312f5d3261060a2d5dd596378f4. Replaced `or []` with an explicit None check (`_rf_val if _rf_val is not None else []`) so falsy non-None values (False, 0, "") now flow through the isinstance guard and raise instead of silently becoming `[]`.

src/autoskillit/recipe/io.py
Trecek: Agreed — fixed in 6b27bee027f01312f5d3261060a2d5dd596378f4. Added element-level string validation after the isinstance check: non-string entries in `requires_features_raw` now raise `ValueError` before reaching `Recipe.__post_init__`.

Issue #1860, opened 2026-05-05
Rectify: Session-Lifetime Hook ENOENT Race — Stable Dispatcher Architecture


Trecek:
## Problem

After PR #1854 renamed `leaf_orchestration_guard.py` → `skill_orchestration_guard.py`, already-running Claude Code sessions hit ENOENT errors on every `run_skill` call, blocking all pipeline execution. The PreToolUse hook command cached in the session's memory points to a file that no longer exists on disk after the package upgrade.

This is the **6th occurrence** of the hook ENOENT staleness bug class:
- PR #543: Headless Editable Install Poison
- PR #670: Ghost Hook Registrations Survive Git Revert
- PR #682: Stale-Hooks Check Infinite Prompt Loop
- PR #1193: Worktree-derived paths corrupting user-scope hooks
- PR #1786: Plugin Cache Staleness Immunity (directory reorg)
- **Current**: Session-lifetime race after file rename during upgrade

Each prior fix closed one specific trigger without addressing the root architectural weakness: **hook command paths are baked absolute references to files that can be moved/renamed/deleted by package management operations.**

## Root Cause

**Mechanism:** Claude Code caches hook command paths in memory at session start. When `autoskillit install` runs `uv tool install --force`, the `site-packages/autoskillit/` directory is replaced with the new version's content. If a hook script was renamed, the old path ceases to exist on disk, but the running session still tries to execute it.

**Triggering sequence:**
1. Interactive session starts → loads hooks from plugin cache 0.9.390 (references `leaf_orchestration_guard.py`)
2. `autoskillit install` runs → upgrades uv-tools install to 0.9.395+ (renames file on disk), writes correct settings.json and new plugin cache (0.9.397)
3. Interactive session fires PreToolUse hook on `run_skill` → Claude Code host process tries to `exec` the old path → **ENOENT**
4. ENOENT treated as hook infrastructure failure → **tool call blocked**

**Why it fires on the interactive session (not headless children):** The `mcp__.*autoskillit.*__(run_skill|run_cmd|run_python).*` matcher fires when the CALLING session invokes `run_skill`. This happens in the interactive session's process before the headless child is spawned. Child sessions load fresh hooks from the current plugin cache.

## Why Existing Protections Didn't Prevent This

All protections work correctly for NEW sessions. None can protect ALREADY-RUNNING sessions:

| Protection | What it does | Why it doesn't help here |
|---|---|---|
[...]

PR #1867, opened 2026-05-05
Rectify: Session-Lifetime Hook ENOENT Race — Stable Dispatcher Architecture

Trecek:
## Summary

Hook commands in `hooks.json` and `settings.json` embed absolute filesystem paths to individual Python scripts (e.g., `python3 /abs/path/hooks/guards/skill_orchestration_guard.py`). When a hook script is renamed across package versions, already-running Claude Code sessions fire the old path, hit ENOENT, and block all tool calls matching that hook. This is the 6th occurrence of this bug class — each prior fix patched one trigger without addressing the root cause: **direct script paths as hook commands make every rename a breaking change for in-flight sessions**.

The architectural fix introduces a **stable dispatcher** (`_dispatch.py`) as the sole hook command target. All hook entries reference this single, never-renamed file and pass a logical hook name as an argument. The dispatcher resolves the logical name to the current script path at execution time, with a static `_RETIRED_MAPPING` dict handling renamed scripts. This makes the system innately immune to rename-induced ENOENT — the dispatcher file exists at the same path in every version, and the mapping handles name transitions transparently.

Closes #1860

## Implementation Plan

Plan file: `.autoskillit/temp/rectify/rectify_session_lifetime_hook_enoent_race_stable_dispatcher_2026-05-04_194411.md`

## Token Usage Summary

| rectify | 1 | 7.1k | 11.6k | 643.2k | 81.6k | 167 | 57.8k | 8m 6s |
| dry_walkthrough | 1 | 1.8k | 17.9k | 2.2M | 85.9k | 142 | 71.3k | 15m 46s |
| implement | 1 | 154 | 31.9k | 9.4M | 149.3k | 196 | 136.3k | 14m 10s |
| prepare_pr | 1 | 33 | 5.9k | 187.6k | 41.4k | 18 | 32.3k | 1m 45s |
| compose_pr | 1 | 23 | 1.6k | 154.6k | 30.5k | 12 | 17.3k | 44s |
| review_pr | 1 | 43 | 34.5k | 1.0M | 91.2k | 37 | 78.0k | 9m 58s |
| resolve_review | 1 | 62 | 18.6k | 2.2M | 83.4k | 87 | 129.0k | 14m 7s |
| **Total** | | 9.3k | 122.0k | 15.8M | 149.3k | | 522.1k | 1h 4m |

## Token Efficiency

| rectify | 0 | — | — | — |
| implement | 487 | 19313.5 | 280.0 | 65.6 |
| review_pr | 0 | — | — | — |
| resolve_review | 23 | 95872.4 | 5608.1 | 809.1 |
| **Total** | **510** | 30910.0 | 1023.7 | 239.3 |


src/autoskillit/hooks/_dispatch.py
```diff
@@ -0,0 +1,60 @@
...
+    if not target.is_file():
+        print(
+            f"[autoskillit dispatch] retired mapping target missing: {target} — degrading gracefully",
+            file=sys.stderr,
+        )
+        sys.exit(0)
+
+    stdin_data = sys.stdin.buffer.read()
+
+    result = subprocess.run(
```
Trecek: [warning] defense: subprocess.run called without try/except for OSError. If the target file becomes unreadable between the is_file() check and execution, the exception propagates uncaught and produces an unformatted traceback — breaking the graceful degradation contract established by the rest of this file.

src/autoskillit/cli/doctor/_doctor_hooks.py
```diff
@@ -20,14 +20,20 @@ def _check_hook_registration(settings_path: Path) -> DoctorResult:
...
         hook.get("command", "")
         for event_entries in data.get("hooks", {}).values()
         if isinstance(event_entries, list)
         for entry in event_entries
         for hook in entry.get("hooks", [])
-    )
-    missing = [script for script in canonical_script_basenames() if script not in registered]
+    ]
+    registered = " ".join(commands)
+    registered_logical = {cmd.split()[-1] for cmd in commands if "_dispatch.py" in cmd}
```
Trecek, line 30: [warning] defense: cmd.split()[-1] assumes the logical name is always the last token. Should validate that parts[-2] ends with '_dispatch.py' before trusting parts[-1], mirroring the guard used in _extract_script_basenames and find_broken_hook_scripts.

src/autoskillit/cli/doctor/_doctor_hooks.py
```diff
@@ -20,14 +20,20 @@ def _check_hook_registration(settings_path: Path) -> DoctorResult:
...
+    missing = [
+        script
+        for script in canonical_script_basenames()
+        if script not in registered and script.removesuffix(".py") not in registered_logical
```
Trecek: [warning] bugs: `script not in registered` uses substring match against a space-joined string of all commands. A script basename like 'quota_post_hook.py' could match inside a longer unrelated path, producing a false negative. Should use `any(script in cmd for cmd in commands)` for precise per-command matching.

src/autoskillit/hook_registry.py
```diff
@@ -378,10 +389,13 @@ def find_broken_hook_scripts(settings_path: Path) -> list[str]:
         for entry in data.get("hooks", {}).get(event_type, []):
             for hook in entry.get("hooks", []):
                 cmd = hook.get("command", "")
-                if not _is_own_hook(cmd):  # skip non-autoskillit hooks
+                if not _is_own_hook(cmd):
                     continue
                 parts = cmd.split()
-                if len(parts) >= 2:
+                if len(parts) >= 3 and parts[-2].endswith("_dispatch.py"):
```
Trecek, line 395: [critical] bugs: find_broken_hook_scripts checks if _dispatch.py itself exists (parts[-2]) but does NOT validate whether the logical name (parts[-1]) resolves to a real target script. A valid dispatcher binary pointing to a deleted or mistyped hook name will silently pass this check, defeating the purpose of broken-hook detection for the new dispatcher format.

src/autoskillit/hook_registry.py
```diff
@@ -238,8 +238,20 @@ def _build_hook_entry(hook_def: HookDef, hook_commands: list[dict]) -> dict:
     return {"matcher": hook_def.matcher, "hooks": hook_commands}
 
 
+def _build_hook_command(hooks_dir: Path, script: str, timeout_seconds: int | None) -> dict:
+    """Build a single hook command dict using the stable dispatcher format."""
+    logical_name = script.removesuffix(".py")
+    cmd: dict = {
+        "type": "command",
+        "command": f"python3 {hooks_dir / '_dispatch.py'} {logical_name}",
```
Trecek, line 246: [warning] bugs: Command string built via f-string interpolation. If hooks_dir contains spaces, the resulting command will be split incorrectly by cmd.split() in _extract_script_basenames and find_broken_hook_scripts. Either document the space-free path requirement or use shlex.quote().

src/autoskillit/hook_registry.py
```diff
@@ -341,11 +347,16 @@ def _extract_script_basenames(hooks_dict: dict) -> set[str]:
                 cmd = hook.get("command", "")
                 if not cmd or not _is_own_hook(cmd):
                     continue
-                script_path = Path(cmd.split()[-1])
-                bare = script_path.name
-                canonical = canonical_script_basenames()
-                matched = next((c for c in canonical if Path(c).name == bare), bare)
-                result.add(matched)
+                parts = cmd.split()
```
Trecek, line 350: [warning] cohesion: The dispatcher command-parsing pattern (split, check '_dispatch.py', extract parts[-1]) is duplicated across hook_registry.py (L350-353), _doctor_hooks.py (L31), test_cli_hooks.py (L15-17), and test_hook_executability.py (L63-65). Consider extracting a shared _parse_dispatcher_command() helper alongside _build_hook_command to keep build/parse co-located.

Trecek (commented):

Trecek (commented):

src/autoskillit/hooks/_dispatch.py
Trecek: Agreed — fixed in 655a2f795d943c0e5991075d81e379c91108a65e. Wrapped subprocess.run in try/except OSError with diagnostic message and sys.exit(0), consistent with the existing graceful degradation pattern at L36-L40 and L43-L47.

src/autoskillit/cli/doctor/_doctor_hooks.py
Trecek: Investigated — this is intentional. The `if "_dispatch.py" in cmd` filter is sufficient for this diagnostic function: all hook commands in settings.json are generated by autoskillit itself in the canonical form `python3 /path/_dispatch.py logical_name`. The reviewer's claim about mirroring `_extract_script_basenames` is inaccurate — that function uses the same loose pattern (L351). Only `find_broken_hook_scripts` uses the stricter positional guard, which serves a different purpose (detecting ENOENT-fatal hooks vs. diagnostic registration checks).

src/autoskillit/cli/doctor/_doctor_hooks.py
Trecek: Agreed — fixed in cb223e39e6e039ee98b0db7d72891bbc29790b52. Replaced `script not in registered` (substring match on space-joined string) with `not any(script in cmd for cmd in commands)` for precise per-command matching.

src/autoskillit/hook_registry.py
Trecek: Investigated — this is intentional. `find_broken_hook_scripts` specifically detects ENOENT-fatal hooks (dispatcher binary missing). Logical name validity is a separate concern handled by `_count_hook_registry_drift`, which compares deployed basenames against `canonical_script_basenames()`. Furthermore, `_dispatch.py` itself exits 0 gracefully for unknown logical names (L36-40), so a stale logical name does NOT produce a broken/fatal hook — it degrades silently. The two concerns are intentionally separated: 'broken' = dispatcher binary absent (ENOENT), 'drift' = stale logical name (separate doctor check).

src/autoskillit/hook_registry.py
Trecek: Investigated — this is intentional. `hooks_dir` is always `pkg_root() / "hooks"` where `pkg_root()` resolves to the installed autoskillit package under site-packages. Python packages installed via pip/uv are never placed in paths containing spaces — this is a fundamental property of Python packaging conventions. The concern is technically conceivable but has no realistic attack surface in this codebase.

src/autoskillit/hook_registry.py
Trecek: Valid observation — flagged for design decision. The parsing pattern is duplicated in 4 locations but there are competing considerations: test code intentionally duplicates parsing for structural verification independence, and the _doctor_hooks.py usage is a different form (inline comprehension on pre-flattened list). Extracting _parse_dispatcher_command() into hook_registry.py (IL-0) would couple the registry module to a parse concern.

PR #1914, opened 2026-05-05
Add --profile CLI Flag to autoskillit cook for Provider Selection

### LINKED ISSUE #1905: Add --profile CLI flag to autoskillit cook for provider selection at invocation
## Summary

Add a `--profile <name>` CLI argument to `autoskillit cook` (and potentially other session-launching commands) that overrides the config-driven provider resolution for the entire session.

## Motivation

Currently, provider routing is entirely config-driven via `step_overrides` and `default_provider` in `.autoskillit/config.yaml`. There is no way to select a provider profile at invocation time without editing the config file. This makes it cumbersome to:

- Quickly test a new provider (e.g. Minimax) without changing config that affects other sessions
- Switch between providers for different tasks without config edits
- Run one-off sessions against a non-default provider

## Proposed Behavior

```bash
# Override default_provider for this session
autoskillit cook --profile minimax

# Still respects step_overrides if set, but uses minimax as the Tier 4 fallback
# instead of "anthropic"
```

The `--profile` flag should:
1. Set `default_provider` in the runtime config for that session only
2. NOT modify any config files on disk
3. Respect `step_overrides` (Tier 1/2 still win over the CLI flag)
4. Validate that the named profile exists in `config.providers.profiles` at startup
5. Error clearly if the `providers` feature flag is not enabled

## Implementation Notes

- Entry point: `cook_interactive()` in `cli/session/_cook.py` / `app.py` command definition
- The resolved profile name needs to flow into the `ToolContext` so `run_skill` can read it during Tier 4 resolution
- Consider also supporting `--provider-model` to override `ANTHROPIC_MODEL` independently of the profile

## Acceptance Criteria

- [ ] `autoskillit cook --profile minimax` launches a session where Tier 4 resolves to the `minimax` profile
- [ ] Unknown profile name produces a clear error at startup
- [ ] Feature-flag-disabled produces a clear error at startup
- [ ] Existing `step_overrides` still take precedence (Tier 1 > CLI flag)
- [ ] No config files modified on disk
- [ ] Unit tests for the new flag parsing and profile validation

Trecek:
## Summary

Add a `--profile <name>` option to the `autoskillit cook` command that overrides the
session-level default provider without modifying any config files on disk. The flag
validates feature-flag enablement and profile existence at startup, then threads the
profile name through two mechanisms: (1) as an env var in `_cook_env_extras` for the
Claude Code subprocess and (2) as a runtime override applied in `make_context()` so the
MCP server's `run_skill` tool resolves Tier 4 to the CLI-selected profile rather than
`default_provider` from config. `step_overrides` (Tiers 1/2) continue to win.

Closes #1905

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260505-000406-360445/.autoskillit/temp/make-plan/add_profile_cli_flag_to_cook_plan_2026-05-05_001000.md`

## Token Usage Summary

| plan | 1 | 127 | 20.7k | 642.9k | 63.0k | 86 | 50.0k | 10m 15s |
| verify | 1 | 118 | 14.5k | 580.1k | 56.8k | 59 | 43.9k | 8m 16s |
| implement | 1 | 358 | 20.5k | 2.3M | 70.2k | 122 | 57.6k | 8m 2s |
| prepare_pr | 1 | 60 | 4.3k | 210.8k | 38.4k | 18 | 26.1k | 1m 39s |
| compose_pr | 1 | 51 | 1.9k | 151.2k | 29.4k | 14 | 16.4k | 48s |
| review_pr | 2 | 886 | 72.1k | 1.8M | 87.4k | 169 | 145.5k | 18m 14s |
| resolve_review | 2 | 718 | 43.8k | 5.3M | 110.5k | 212 | 166.8k | 22m 50s |
| **Total** | | 2.3k | 177.8k | 10.9M | 110.5k | | 506.2k | 1h 10m |

## Token Efficiency

| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| implement | 197 | 11469.6 | 292.3 | 104.2 |
| review_pr | 0 | — | — | — |
| resolve_review | 36 | 146848.3 | 4633.2 | 1215.9 |
| **Total** | **233** | 46788.4 | 2172.7 | 763.0 |

### DIFF
--- a/src/autoskillit/cli/app.py
+++ b/src/autoskillit/cli/app.py
@@ -431,9 +431,6 @@ def _cook_cmd(
 
     Use --resume to restore a previous session. Pass a session ID to target a
     specific one, or omit for Claude Code's interactive picker.
-
-    Use --profile to select a provider profile (overrides config default_provider
-    for this session only; requires the 'providers' feature flag to be enabled).
     """
     cook_interactive(
         resume=resume or (session_id is not None), session_id=session_id, profile=profile
--- a/src/autoskillit/cli/app.py
+++ b/src/autoskillit/cli/app.py
@@ -424,13 +424,20 @@ def recipes_render(name: str | None = None) -> None:
 
 
 @app.command(name="cook", alias="c")
-def _cook_cmd(session_id: str | None = None, *, resume: bool = False) -> None:
+def _cook_cmd(
+    session_id: str | None = None, *, resume: bool = False, profile: str | None = None
+) -> None:
     """Launch an interactive Claude session with all skills and kitchen tools.
 
     Use --resume to restore a previous session. Pass a session ID to target a
     specific one, or omit for Claude Code's interactive picker.
+
+    Use --profile to select a provider profile (overrides config default_provider
+    for this session only; requires the 'providers' feature flag to be enabled).
     """
-    cook_interactive(resume=resume or (session_id is not None), session_id=session_id)
+    cook_interactive(
+        resume=resume or (session_id is not None), session_id=session_id, profile=profile
+    )
 
 
 def main() -> None:
--- a/src/autoskillit/cli/session/_cook.py
+++ b/src/autoskillit/cli/session/_cook.py
@@ -4,11 +4,13 @@ from __future__ import annotations
 
 import shutil
 import subprocess
+import sys
 import uuid
 from collections.abc import Mapping
 from pathlib import Path
 
 from autoskillit.cli.ui._terminal import terminal_guard
+from autoskillit.core.feature_flags import is_feature_enabled
 
 
 def _print_recipes_list() -> None:
@@ -59,7 +61,9 @@ def _run_cook_session(
     return None
 
 
-def cook(*, resume: bool = False, session_id: str | None = None) -> None:
+def cook(
+    *, resume: bool = False, session_id: str | None = None, profile: str | None = None
+) -> None:
     """Launch Claude with all bundled AutoSkillit skills as slash commands."""
     from autoskillit.workspace import (
         DefaultSessionSkillManager,
@@ -85,6 +89,27 @@ def cook(*, resume: bool = False, session_id: str | None = None) -> None:
     from autoskillit.config import iter_display_categories, load_config  # noqa: PLC0415
 
     config = load_config()
+
+    if profile is not None:
+        if not is_feature_enabled(
+            "providers", config.features, experimental_enabled=config.experimental_enabled
+        ):
+            print(
+                "Error: --profile requires the 'providers' feature to be enabled.\n"
+                "Enable it in .autoskillit/config.yaml:\n"
+                "  features:\n"
+                "    providers: true",
+                file=sys.stderr,
+            )
+            raise SystemExit(1)
+        if profile not in config.providers.profiles:
+            known = ", ".join(sorted(config.providers.profiles)) or "(none defined)"
+            print(
+                f"Error: Unknown provider profile {profile!r}. Known profiles: {known}\n"
+                "Define profiles in .autoskillit/config.yaml under providers.profiles.",
+                file=sys.stderr,
+            )
+            raise SystemExit(1)
 
     print(f"{_B}{_C}AUTOSKILLIT {__version__}{_R} {_D}Kitchen open. All tools active.{_R}")
     skip = {"Telemetry & Diagnostics", "Kitchen"}
@@ -175,6 +200,9 @@ def cook(*, resume: bool = False, session_id: str | None = None) -> None:
         SESSION_TYPE_ENV_VAR: SESSION_TYPE_COOK,
         LAUNCH_ID_ENV_VAR: session_id_local,
     }
+    if profile is not None:
+        _cook_env_extras["AUTOSKILLIT_PROVIDER_PROFILE"] = profile
+        _cook_env_extras.update(config.providers.profiles[profile])
 
     current_resume_spec = resume_spec
     _current_first_run = _first_run
--- a/src/autoskillit/cli/session/_cook.py
+++ b/src/autoskillit/cli/session/_cook.py
@@ -10,7 +10,7 @@ from collections.abc import Mapping
 from pathlib import Path
 
 from autoskillit.cli.ui._terminal import terminal_guard
-from autoskillit.core.feature_flags import is_feature_enabled
+from autoskillit.core import is_feature_enabled
 
 
 def _print_recipes_list() -> None:
--- a/src/autoskillit/server/_factory.py
+++ b/src/autoskillit/server/_factory.py
@@ -9,6 +9,7 @@ construction scattered across callers.
 
 from __future__ import annotations
 
+import dataclasses
 import os
 import subprocess
 from collections.abc import Callable
@@ -198,12 +199,15 @@ def make_context(
         All service fields are populated. When runner=None is passed explicitly,
         tester is left as None.
     """
-    # When cook CLI passes --profile, it stamps AUTOSKILLIT_PROVIDER_PROFILE into the
-    # Claude Code subprocess env. The MCP server (plugin subprocess) inherits this and
-    # uses it as default_provider, taking precedence over config without modifying disk.
     env_profile = os.environ.get("AUTOSKILLIT_PROVIDER_PROFILE")
-    if env_profile:
-        config.providers.default_provider = env_profile
+    if env_profile and env_profile in config.providers.profiles:
+        config = dataclasses.replace(
+            config, providers=dataclasses.replace(config.providers, default_provider=env_profile)
+        )
+    elif env_profile:
+        logger.warning(
+            "AUTOSKILLIT_PROVIDER_PROFILE %r not in known profiles — ignored", env_profile
+        )
 
     if runner is _UNSET:
         from autoskillit.execution import DefaultSubprocessRunner
--- a/src/autoskillit/server/_factory.py
+++ b/src/autoskillit/server/_factory.py
@@ -198,6 +198,13 @@ def make_context(
         All service fields are populated. When runner=None is passed explicitly,
         tester is left as None.
     """
+    # When cook CLI passes --profile, it stamps AUTOSKILLIT_PROVIDER_PROFILE into the
+    # Claude Code subprocess env. The MCP server (plugin subprocess) inherits this and
+    # uses it as default_provider, taking precedence over config without modifying disk.
+    env_profile = os.environ.get("AUTOSKILLIT_PROVIDER_PROFILE")
+    if env_profile:
+        config.providers.default_provider = env_profile
+
     if runner is _UNSET:
         from autoskillit.execution import DefaultSubprocessRunner
 
--- a/tests/cli/test_cook_profile.py
+++ b/tests/cli/test_cook_profile.py
@@ -1,0 +1,119 @@
+"""Tests for --profile flag in cook command."""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+import autoskillit.cli.session._cook as cook_module
+from autoskillit.execution.commands import ClaudeInteractiveCmd
+
+pytestmark = [pytest.mark.layer("cli"), pytest.mark.medium]
+
+
+def _fake_build_cmd():
+    captured = []
+
+    def fake(**kwargs):
+        captured.append(kwargs.get("env_extras", {}))
+        return ClaudeInteractiveCmd(cmd=["claude"], env={})
+
+    return fake, captured
+
+
+@pytest.fixture()
+def _mock_mgr():
+    return MagicMock()
+
+
+def _run_cook(profile, cfg, mock_mgr):
+    fake_build, captured = _fake_build_cmd()
+    with (
+        patch("shutil.which", return_value="/usr/bin/claude"),
+        patch("builtins.input", return_value=""),
+        patch("autoskillit.workspace.DefaultSessionSkillManager", return_value=mock_mgr),
+        patch("subprocess.run", return_value=MagicMock(returncode=0)),
+        patch("autoskillit.execution.build_interactive_cmd", fake_build),
+        patch("autoskillit.core.write_registry_entry"),
+        patch("autoskillit.config.load_config", return_value=cfg),
+        patch(
+            "autoskillit.cli.session._cook.is_feature_enabled",
+            side_effect=lambda key, *a, **kw: key == "providers",
+        ),
+    ):
+        cook_module.cook(profile=profile)
+    return captured
+
+
+def test_profile_valid_injects_provider_env_var(_mock_mgr):
+    """AUTOSKILLIT_PROVIDER_PROFILE must be in env_extras when --profile is given."""
+    cfg = MagicMock()
+    cfg.experimental_enabled = True
+    cfg.providers.profiles = {"minimax": {"ANTHROPIC_BASE_URL": "https://minimax.example"}}
+    captured = _run_cook("minimax", cfg, _mock_mgr)
+    env = captured[0]
+    assert env.get("AUTOSKILLIT_PROVIDER_PROFILE") == "minimax"
+
+
+def test_profile_valid_injects_profile_env_vars(_mock_mgr):
+    """Profile's own env vars (API creds) must be injected into env_extras."""
+    cfg = MagicMock()
+    cfg.experimental_enabled = True
+    cfg.providers.profiles = {
+        "minimax": {"ANTHROPIC_BASE_URL": "https://mm.io", "ANTHROPIC_API_KEY": "sk-mm"}
+    }
+    captured = _run_cook("minimax", cfg, _mock_mgr)
+    env = captured[0]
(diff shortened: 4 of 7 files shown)


src/autoskillit/cli/session/_cook.py
```diff
@@ -86,6 +90,27 @@ def cook(*, resume: bool = False, session_id: str | None = None) -> None:
 
     config = load_config()
 
+    if profile is not None:
```
Trecek, line 93: [warning] cohesion: Profile validation logic (feature gate check + profile existence check) is inlined directly into cook() at L93-L112, but analogous validation exists elsewhere. A shared _validate_profile_arg() helper would localize this logic consistently.

src/autoskillit/cli/session/_cook.py
```diff
@@ -86,6 +90,27 @@ def cook(*, resume: bool = False, session_id: str | None = None) -> None:
...
+        ):
+            print(
+                "Error: --profile requires the 'providers' feature to be enabled.\n"
+                "Enable it in .autoskillit/config.yaml:\n"
+                "  features:\n"
+                "    providers: true",
+                file=sys.stderr,
+            )
+            raise SystemExit(1)
+        if profile not in config.providers.profiles:
```
Trecek, line 105: [warning] bugs: config.providers.profiles is accessed without a prior None-check on config.providers. If providers is absent or None in config (e.g. feature disabled, partial init), this raises AttributeError instead of the intended user-facing error.

src/autoskillit/cli/session/_cook.py
```diff
@@ -175,6 +200,9 @@ def cook(*, resume: bool = False, session_id: str | None = None) -> None:
         SESSION_TYPE_ENV_VAR: SESSION_TYPE_COOK,
         LAUNCH_ID_ENV_VAR: session_id_local,
     }
+    if profile is not None:
```
Trecek, line 203: [warning] cohesion: Provider profile env injection uses two separate operations (set AUTOSKILLIT_PROVIDER_PROFILE, then .update() with profile dict). The headless path uses named parameters for the same conceptual split, creating asymmetric patterns for functionally identical env assembly.

src/autoskillit/cli/session/_cook.py
```diff
@@ -175,6 +200,9 @@ def cook(*, resume: bool = False, session_id: str | None = None) -> None:
...
+        _cook_env_extras["AUTOSKILLIT_PROVIDER_PROFILE"] = profile
+        _cook_env_extras.update(config.providers.profiles[profile])
```
Trecek, line 205: [critical] bugs: config.providers.profiles[profile] is merged into _cook_env_extras via update(). If the profile dict contains non-string values (nested dicts, ints, bools), they will cause TypeError when passed to subprocess (which requires all env values to be strings). Profile values must be validated as strings or cast to str before update().

src/autoskillit/server/_factory.py
```diff
@@ -198,6 +198,13 @@ def make_context(
         All service fields are populated. When runner=None is passed explicitly,
         tester is left as None.
     """
+    # When cook CLI passes --profile, it stamps AUTOSKILLIT_PROVIDER_PROFILE into the
```
Trecek: [warning] slop: 3-line comment at L201-L203 explains what the code does rather than why. The implementation is self-evident from the variable names; remove the comment.

src/autoskillit/server/_factory.py
```diff
@@ -198,6 +198,13 @@ def make_context(
...
+    # Claude Code subprocess env. The MCP server (plugin subprocess) inherits this and
+    # uses it as default_provider, taking precedence over config without modifying disk.
+    env_profile = os.environ.get("AUTOSKILLIT_PROVIDER_PROFILE")
```
Trecek, line 202: [warning] cohesion: make_context() accepts AUTOSKILLIT_PROVIDER_PROFILE from env without validating it against config.providers.profiles. The CLI validates the profile name before launching (L105-L112 in _cook.py), but the server silently accepts any value. Structural asymmetry: two callsites with different strictness.

Trecek (commented):

src/autoskillit/server/_factory.py
```diff
@@ -198,6 +198,13 @@ def make_context(
...
+    if env_profile:
```
Trecek: [warning] defense: env_profile from AUTOSKILLIT_PROVIDER_PROFILE is written to config.providers.default_provider without checking whether it is a known profile in config.providers.profiles. An unrecognized value (stale env, manually exported var) silently sets an invalid default_provider.

src/autoskillit/server/_factory.py
```diff
@@ -198,6 +198,13 @@ def make_context(
...
+        config.providers.default_provider = env_profile
```
Trecek: [warning] bugs: config.providers.default_provider is mutated in-place on the shared config object. If make_context() is called more than once in the same process (tests, MCP reconnect), the mutation persists even when AUTOSKILLIT_PROVIDER_PROFILE is absent in a subsequent call.

tests/cli/test_cook_profile.py
```diff
@@ -0,0 +1,120 @@
...
+            "autoskillit.cli.session._cook.is_feature_enabled",
+            side_effect=lambda key, *a, **kw: key == "providers",
+        ),
+        patch("autoskillit.cli.ui._timed_input.timed_prompt", return_value=""),
+    ):
+        cook_module.cook(profile=profile)
+    return captured
+
+
+def test_profile_valid_injects_provider_env_var(_mock_mgr):
```
Trecek, line 50: [warning] tests: captured[0] is indexed without first asserting len(captured) >= 1. If build_interactive_cmd is not called (early return or exception), this raises IndexError instead of a meaningful assertion failure.

tests/cli/test_cook_profile.py
```diff
@@ -0,0 +1,120 @@
...
+    """AUTOSKILLIT_PROVIDER_PROFILE must be in env_extras when --profile is given."""
+    cfg = MagicMock()
+    cfg.experimental_enabled = True
+    cfg.providers.profiles = {"minimax": {"ANTHROPIC_BASE_URL": "https://minimax.example"}}
+    captured = _run_cook("minimax", cfg, _mock_mgr)
```
Trecek, line 55: [warning] tests: Same IndexError risk on captured[0] in test_profile_valid_injects_profile_env_vars (L67) and test_profile_none_does_not_inject_provider_env (L78). All positive-path tests index captured[0] without asserting the list is non-empty.

tests/cli/test_cook_profile.py
```diff
@@ -0,0 +1,120 @@
...
+    """When profile=None, AUTOSKILLIT_PROVIDER_PROFILE must NOT appear in env_extras."""
+    cfg = MagicMock()
+    cfg.experimental_enabled = True
+    cfg.providers.profiles = {}
+    captured = _run_cook(None, cfg, _mock_mgr)
+    env = captured[0]
+    assert "AUTOSKILLIT_PROVIDER_PROFILE" not in env
+
+
+def test_profile_feature_disabled_exits(capsys, _mock_mgr):
```
Trecek, line 86: [warning] tests: test_profile_feature_disabled_exits does not patch builtins.input or timed_prompt, unlike _run_cook(). If production code reaches an input() call before the feature-guard check, the test will hang waiting for stdin.

tests/cli/test_cook_profile.py
```diff
@@ -0,0 +1,120 @@
...
+        patch("autoskillit.config.load_config", return_value=cfg),
+        patch("autoskillit.cli.session._cook.is_feature_enabled", return_value=False),
+    ):
+        with pytest.raises(SystemExit) as exc_info:
+            cook_module.cook(profile="minimax")
+    assert exc_info.value.code == 1
+    assert "providers" in capsys.readouterr().err
+
+
+def test_profile_unknown_exits(capsys, _mock_mgr):
```
Trecek, line 105: [warning] tests: test_profile_unknown_exits has the same missing builtins.input / timed_prompt patches as test_profile_feature_disabled_exits. Execution before the unknown-profile guard will hang on stdin.

Trecek (commented):
AutoSkillit review: **changes_requested** — 1 critical finding requires changes before merging. See inline comments for details.

src/autoskillit/cli/session/_cook.py
Trecek: Valid observation — flagged for design decision. Extracting a `_validate_profile_arg()` helper is a reasonable cohesion improvement. Left for a focused refactor pass rather than mixed into this fix round.

src/autoskillit/cli/session/_cook.py
Trecek: Investigated — this is intentional. `config.providers` is always non-None: `AutomationConfig` declares it as `providers: ProvidersConfig = field(default_factory=ProvidersConfig)` (settings.py:152). No code path produces a None value here; the feature-gate check is a logical guard, not a null guard.

src/autoskillit/cli/session/_cook.py
Trecek: Valid observation — flagged for design decision. The asymmetry between the interactive path (two-step env injection) and the headless path (named params) is real. Unifying the pattern requires changes beyond this PR's scope.

src/autoskillit/cli/session/_cook.py
Trecek: Investigated — this is intentional. `ProvidersConfig.__post_init__` validates all profile values as strings at config load time (_config_dataclasses.py:322-326). By the time `_cook_env_extras.update()` is reached, the type contract is already enforced; non-string values would have raised `ValueError` earlier.

src/autoskillit/server/_factory.py
Trecek: Agreed — fixed in 035ec2f4. Comment removed; the variable names and structure are self-describing.

src/autoskillit/server/_factory.py
Trecek: Agreed — fixed in 035ec2f4. `env_profile` is now gated behind `env_profile in config.providers.profiles` before being applied; unknown values log a warning and are ignored. The CLI-level validation and server-level validation are now symmetric.

src/autoskillit/server/_factory.py
Trecek: Agreed — fixed in 035ec2f4. The profiles membership guard prevents a stale or externally-exported var from silently polluting `default_provider`; an unrecognized value is now logged as a warning and ignored.

src/autoskillit/server/_factory.py
Trecek: Agreed — fixed in 035ec2f4. Changed from in-place mutation (`config.providers.default_provider = env_profile`) to `dataclasses.replace` so each `make_context()` call works on an independent config object and does not bleed state across invocations.

tests/cli/test_cook_profile.py
Trecek: Agreed — fixed in 790c4148. Added `assert len(captured) >= 1, "build_interactive_cmd was not called"` before all three positive-path `captured[0]` accesses so a premature return produces a meaningful failure instead of IndexError.

tests/cli/test_cook_profile.py
Trecek: Agreed — fixed in 790c4148. All three positive-path tests now assert `len(captured) >= 1` before indexing.

tests/cli/test_cook_profile.py
Trecek: Investigated — this is intentional. The feature-disabled check at L93-100 in cook() raises `SystemExit(1)` before `timed_prompt` is ever reached (timed_prompt is called ~30 lines later, after the profile validation block). The test correctly exercises the early-exit path and will never hang on stdin.

tests/cli/test_cook_profile.py
Trecek: Investigated — this is intentional. For the unknown-profile path, `is_feature_enabled` returns `True` but `profile not in config.providers.profiles` is also `True` (minimax not in {anthropic, openai}), so `SystemExit(1)` is raised at L105-112 — still before `timed_prompt` is reached. No stdin hang is possible.

tests/server/test_factory.py
```diff
@@ -534,3 +534,45 @@ def fake_run(cmd, *, capture_output, text, timeout):
...
+    ctx = make_context(config, runner=_runner())
+    assert ctx.config.providers.default_provider == "anthropic"
+
+
+def test_make_context_no_env_profile_preserves_config_default(monkeypatch):
+    """Without AUTOSKILLIT_PROVIDER_PROFILE in env, default_provider is unchanged."""
+    monkeypatch.delenv("AUTOSKILLIT_PROVIDER_PROFILE", raising=False)
+    config = AutomationConfig()
+    config.providers.default_provider = "openai"
+    ctx = make_context(config, runner=_runner())
```
Trecek, line 575: [warning] tests: test_make_context_no_env_profile_preserves_config_default does not set config.providers.profiles (only config.providers.default_provider = "openai" at L576). The production code at _factory.py checks `env_profile in config.providers.profiles`, so if profiles defaults to None rather than {}, the `in` check raises TypeError. The three sibling tests all set profiles explicitly — this test should too (e.g., config.providers.profiles = {}) for consistency and to avoid a latent TypeError if AutomationConfig defaults change.

tests/cli/test_cook_profile.py
```diff
@@ -0,0 +1,123 @@
...
+
+    return fake, captured
+
+
+@pytest.fixture()
+def _mock_mgr():
+    return MagicMock()
+
+
+def _run_cook(profile, cfg, mock_mgr):
```
Trecek, line 30: [warning] cohesion: _run_cook is a module-level helper that only applies to the three positive-path tests; the two error-path tests use their own inline patch blocks with a different set of patches. This splits the cook() invocation pattern into two structurally different styles within the same test file without a shared abstraction that makes the differing patch sets explicit.

src/autoskillit/cli/app.py
```diff
@@ -424,13 +424,20 @@ def recipes_render(name: str | None = None) -> None:
...
-def _cook_cmd(session_id: str | None = None, *, resume: bool = False) -> None:
+def _cook_cmd(
+    session_id: str | None = None, *, resume: bool = False, profile: str | None = None
+) -> None:
     """Launch an interactive Claude session with all skills and kitchen tools.
 
     Use --resume to restore a previous session. Pass a session ID to target a
     specific one, or omit for Claude Code's interactive picker.
+
+    Use --profile to select a provider profile (overrides config default_provider
```
Trecek: [warning] slop: The docstring addition at L435-L436 restates what is already enforced by the runtime error messages. CLI docstrings surface in --help output; this line adds noise without adding navigation value that the parameter name and error messages don't already provide. Remove or reduce to the essential constraint not self-evident from the flag name.

tests/server/test_factory.py
```diff
@@ -534,3 +534,45 @@ def fake_run(cmd, *, capture_output, text, timeout):
 
     monkeypatch.setattr("autoskillit.server._factory.subprocess.run", fake_run)
     assert _resolve_project_dir() == Path.cwd()
+
+
+# --- AUTOSKILLIT_PROVIDER_PROFILE env override tests ---
```
Trecek: [warning] slop: Section-header comment `# --- AUTOSKILLIT_PROVIDER_PROFILE env override tests ---` is AI-generated divider slop. The test function names already identify what each test covers; this banner adds no navigation value and should be removed.

Trecek (commented):

tests/server/test_factory.py
Trecek: Agreed — fixed in 7604aeb7. Added `config.providers.profiles = {}` to match all sibling tests' explicit fixture setup, guarding against future TypeError if AutomationConfig defaults change.

tests/cli/test_cook_profile.py
Trecek: Valid observation — flagged for design decision. The split exists because error-path tests exit before reaching subprocess.run and build_interactive_cmd, making those patches unnecessary. Refactoring _run_cook to accept optional patch sets would clarify the intent. Flagged for human review.

src/autoskillit/cli/app.py
Trecek: Agreed — fixed in 117829fd. Removed the two --profile docstring lines; the flag name and runtime error messages are self-describing.

tests/server/test_factory.py
Trecek: Agreed — fixed in 7604aeb7. Removed the section-header banner; test function names already identify what each test covers.

src/autoskillit/cli/session/_cook.py
Trecek: [AUDIT] — acknowledged, no action taken

src/autoskillit/cli/session/_cook.py
Trecek: [AUDIT] — acknowledged, no action taken

PR #1946, opened 2026-05-05
Implementation Plan: Batch Issue Creation via GraphQL Aliases

### LINKED ISSUE #1877: full-audit: Batch issue creation via GraphQL aliases instead of sequential REST calls
## Problem

The `full-audit` recipe's `create_issues` step files GitHub issues one at a time — dispatching N sequential `prepare-issue` skill invocations, each making a single REST `POST /repos/{owner}/{repo}/issues` call. For a typical audit producing 18 ticket bodies, this results in:

- 18 separate headless sessions spawned sequentially
- 18 individual REST API calls (18 × 5 = 90 rate-limit points)
- 18+ seconds of mandatory `sleep 1` delays between calls
- Massive session orchestration overhead per ticket

This violates **CLAUDE.md §3.4** which mandates: *"Batch GraphQL mutations via aliases (N mutations in 1 request = 5 pts total, not N × 5 pts). Use for thread resolution, bulk PR queries, and any operation touching multiple entities."*

## Historical Context

The step went through three design phases:
1. **PR #687** (original): Fully parallel — all `prepare-issue` calls in one message
2. **PR #1411** (ticket grouping): One-to-many fan-out — validate-audit now splits reports into per-ticket body files, so N calls per audit type
3. **PR #1689** (commit `a0616a0a`): Flipped to strictly sequential to prevent "request storms"

PR #1689 solved the request storm problem by serializing, but the correct fix per §3.4 is **batching into one GraphQL mutation** — not serializing N individual REST calls. The codebase already has this exact pattern implemented elsewhere.

## Existing Batched Mutation Pattern (from `resolve-review/SKILL.md`)

```bash
# Thread resolution — N mutations in 1 request = 5 pts total
MUTATION_QUERY="mutation {"
for i in ...; do
    tid="${ADDRESSED_THREAD_IDS[$i]}"
    MUTATION_QUERY="${MUTATION_QUERY} resolve${i}: resolveReviewThread(input: {threadId: \"${tid}\"}) { thread { isResolved } }"
done
MUTATION_QUERY="${MUTATION_QUERY} }"
gh api graphql -f query="${MUTATION_QUERY}"
```

The same aliasing approach applies to `createIssue`:

```bash
MUTATION_QUERY="mutation {"
for i in $(seq 0 $((N-1))); do
    MUTATION_QUERY="${MUTATION_QUERY} issue${i}: createIssue(input: {repositoryId: \"${REPO_ID}\", title: \"${TITLES[$i]}\", body: \"${BODIES[$i]}\"}) { issue { url number } }"
done
MUTATION_QUERY="${MUTATION_QUERY} }"
gh api graphql -f query="${MUTATION_QUERY}"
```

This reduces 18 API calls (90 pts) to 1 call (5 pts) — an 18x improvement in rate-limit efficiency.

## Current Implementation (`full-audit.yaml` lines 152-188)

```yaml
create_issues:
  tool: run_skill
  description: "Create GitHub issues from validated reports"
  with:
    skill_command: "/autoskillit:prepare-issue"
    cwd: "${{ inputs.workspace }}"
    step_name: "create_issues"
  note: >
    SEQUENTIAL ISSUE CREATION — ...
    Process prepare-issue calls one at a time:
      - run_skill(skill_command="/autoskillit:prepare-issue {ticket_body_path}", ...)
      - Wait for completion.
      - Proceed to the next ticket body.

    Never call prepare-issue in parallel with itself — it hits the GitHub API
    and parallel calls create request storms.
```

The kitchen rule also enforces this:
> "prepare-issue calls MUST be sequential — never call prepare-issue in parallel with itself."

## Expected Behavior

The `create_issues` step should:

1. Read all ticket body files produced by validate-audit (already on disk as `ticket_body_{source}_{N}_{ts}.md`)
2. Extract title and body from each file
3. Construct a single batched GraphQL `createIssue` mutation with aliases (chunk at ~20 per request if needed, matching `analyze-prs` chunk size)
4. Execute 1 (or at most ceil(N/20)) `gh api graphql` calls
5. Parse all created issue URLs from the response
6. Apply labels in a second batched mutation if not included in the `createIssue` input

This eliminates the need for N headless `prepare-issue` sessions entirely. The `prepare-issue` skill's triage/classification logic is not needed for pre-validated audit findings — those are already classified as `recipe:implementation` / `enhancement` by the validate-audit step.

## Implementation Approach

Since `prepare-issue` is designed as a single-issue skill with triage logic unnecessary for batch-validated findings, the recommended approach is:

1. Replace the `create_issues` step's `tool: run_skill` with `tool: run_cmd` (or `tool: run_python` using a new callable)
2. The new step reads ticket body files from the manifest, constructs the batched GraphQL mutation, executes it, and parses issue URLs
3. Remove the kitchen rule enforcing sequential `prepare-issue` calls
4. Update `tests/recipe/test_full_audit_recipe.py:225` (`test_full_audit_create_issues_sequential_prepare_issue`) — this contract test currently **enforces** the problematic sequential behavior

## Affected Files

- `src/autoskillit/recipes/full-audit.yaml` — replace `create_issues` step implementation + update kitchen rules
- `tests/recipe/test_full_audit_recipe.py` — update/replace sequential contract test
- Possibly `src/autoskillit/smoke_utils.py` or a new `_cmd_rpc.py` callable for the batch creation logic (depending on chosen tool type)

## Rate Limit Impact

| Approach | API Calls | Rate Points | Session Overhead |
|----------|-----------|-------------|------------------|
| Current (18 sequential prepare-issue) | 18 REST POSTs | 90 pts | 18 headless sessions |
| Batched GraphQL (1 mutation) | 1 GraphQL POST | 5 pts | 0 headless sessions |
| Improvement | 18x fewer calls | 18x fewer points | Eliminates all sub-sessions |

Trecek:
## Summary

Replace the `create_issues` step in `full-audit.yaml` from `tool: run_skill` (N sequential `prepare-issue` headless sessions, each making a REST `POST /repos/.../issues` call) with `tool: run_python` calling a new `batch_create_issues` callable in `_cmd_rpc.py`. The callable discovers ticket body files, constructs a single batched GraphQL `createIssue` mutation with aliases, and executes it — reducing 18 API calls (90 rate-limit points, 18 headless sessions) to 1 call (5 points, 0 sessions).

Closes #1877

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260505-135148-903791/.autoskillit/temp/make-plan/batch_issue_creation_via_graphql_aliases_plan_2026-05-05_140500.md`

## Token Usage Summary

| plan | 1 | 4.0k | 22.9k | 1.3M | 88.2k | 147 | 104.6k | 12m 44s |
| verify | 1 | 58 | 15.1k | 1.2M | 122.8k | 130 | 60.8k | 8m 54s |
| implement | 1 | 111.8k | 42.6k | 8.0M | 0 | 217 | 0 | 26m 30s |
| prepare_pr | 1 | 60 | 4.7k | 194.9k | 36.3k | 19 | 24.0k | 1m 15s |
| compose_pr | 1 | 51 | 2.0k | 133.8k | 25.9k | 14 | 13.0k | 37s |
| review_pr | 3 | 1.6k | 130.6k | 2.2M | 99.9k | 148 | 250.9k | 30m 19s |
| resolve_review | 2 | 772 | 64.5k | 5.8M | 97.2k | 241 | 158.2k | 28m 4s |
| ci_conflict_fix | 1 | 147 | 4.4k | 636.8k | 43.7k | 41 | 30.9k | 1m 44s |
| **Total** | | 118.5k | 286.7k | 19.4M | 122.8k | | 642.3k | 1h 50m |

## Token Efficiency

| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| implement | 596 | 13420.8 | 0.0 | 71.5 |
| review_pr | 0 | — | — | — |
| resolve_review | 105 | 55162.9 | 1506.2 | 613.9 |
| ci_conflict_fix | 1496 | 425.6 | 20.6 | 3.0 |
| **Total** | **2197** | 8828.5 | 292.4 | 130.5 |

### DIFF
--- a/src/autoskillit/recipe/_analysis_detectors.py
+++ b/src/autoskillit/recipe/_analysis_detectors.py
@@ -176,7 +176,7 @@ def _detect_dead_outputs(recipe: Recipe, graph: dict[str, set[str]]) -> list[Dat
                 if not isinstance(arg_val, str):
                     continue
                 consumed.update(_CONTEXT_REF_RE.findall(arg_val))
-            # Also scan the step's own message field (used by stop/confirm actions).
+            # message fields are not recipe args; scanner must handle them separately.
             if reachable_step.message and isinstance(reachable_step.message, str):
                 consumed.update(_CONTEXT_REF_RE.findall(reachable_step.message))
             if reachable_step.on_result and reachable_step.on_result.conditions:
--- a/src/autoskillit/recipe/_analysis_detectors.py
+++ b/src/autoskillit/recipe/_analysis_detectors.py
@@ -148,6 +148,11 @@ def _is_observability_capture(cap_key: str, step_name: str, step: RecipeStep) ->
     if cap_key == "local_bundle_path" and step_name == "export_local_bundle":
         return True
 
+    # batch_create_issues issue_count: captured output used for telemetry/logging,
+    # not threaded to downstream recipe steps (the done message uses issue_urls only).
+    if cap_key == "issue_count" and step_name == "create_issues" and step.tool == "run_python":
+        return True
+
     return False
 
 
@@ -171,6 +176,9 @@ def _detect_dead_outputs(recipe: Recipe, graph: dict[str, set[str]]) -> list[Dat
                 if not isinstance(arg_val, str):
                     continue
                 consumed.update(_CONTEXT_REF_RE.findall(arg_val))
+            # Also scan the step's own message field (used by stop/confirm actions).
+            if reachable_step.message and isinstance(reachable_step.message, str):
+                consumed.update(_CONTEXT_REF_RE.findall(reachable_step.message))
             if reachable_step.on_result and reachable_step.on_result.conditions:
                 for cond in reachable_step.on_result.conditions:
                     if cond.when and isinstance(cond.when, str):
--- a/src/autoskillit/recipe/_cmd_rpc.py
+++ b/src/autoskillit/recipe/_cmd_rpc.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import json
 import re
 import subprocess
+import time
 from datetime import date
 from pathlib import Path
 
@@ -143,7 +144,6 @@ def wait_for_direct_merge(
     poll_interval: str = "10",
 ) -> dict[str, str]:
     """Poll PR state until merged/closed/timeout."""
-    import time  # noqa: PLC0415
 
     max_polls = max_polls or "90"
     poll_interval = poll_interval or "10"
@@ -217,7 +217,6 @@ def wait_for_review_pr_mergeability(
     poll_interval: str = "15",
 ) -> dict[str, str]:
     """Extract PR number and poll until mergeability resolves."""
-    import time  # noqa: PLC0415
 
     max_polls = max_polls or "12"
     poll_interval = poll_interval or "15"
@@ -290,7 +289,6 @@ def force_push_and_wait_mergeability(
     poll_interval: str = "15",
 ) -> dict[str, str]:
     """Force-push integration branch and wait for mergeability."""
-    import time  # noqa: PLC0415
 
     max_polls = max_polls or "12"
     poll_interval = poll_interval or "15"
@@ -479,3 +477,175 @@ def export_local_bundle(
         shutil.rmtree(dest)
     shutil.copytree(research_dir, dest)
     return {"local_bundle_path": str(dest)}
+
+
+# ─── batch_create_issues helpers ────────────────────────────────────────────
+
+
+def _extract_title(raw: str) -> str:
+    """Return the text following '# ' from the first H1 line, or a fallback."""
+    m = re.search(r"^#\s+(.+)$", raw, re.MULTILINE)
+    return m.group(1).strip() if m else "Untitled audit finding"
+
+
+def _strip_ticket_body(raw: str) -> str:
+    """Remove internal metadata and exception details from a ticket body."""
+    lines = raw.splitlines()
+    result: list[str] = []
+    skip_exceptions_section = False
+    for line in lines:
+        if line.strip().startswith("validated: true"):
+            continue
+        if ".autoskillit/" in line:
+            continue
+        if "contested_findings_" in line:
+            continue
+        if "| CONTESTED |" in line or "| VALID BUT EXCEPTION WARRANTED |" in line:
+            continue
+        if re.search(r"\*\*Contested:\*\s*\d+", line) or re.search(
+            r"\*\*Exception warranted:\*\s*\d+", line
+        ):
+            continue
+        if "**Exception note:**" in line:
+            continue
+        if re.match(r"## Findings with Exceptions\s*$", line):
+            skip_exceptions_section = True
+            continue
+        if skip_exceptions_section:
+            if line.strip().startswith("---"):
+                skip_exceptions_section = False
+            continue
+        result.append(line)
+    return "\n".join(result)
+
+
+def _resolve_repo_identity(cwd: str) -> tuple[str, str, str]:
+    """Return (owner, repo_name, repo_node_id) for the given workspace."""
+    result = subprocess.run(
+        ["gh", "repo", "view", "--json", "owner,name", "-q", '.owner.login + " " + .name'],
+        cwd=cwd,
+        capture_output=True,
+        text=True,
+    )
+    if result.returncode != 0:
+        msg = f"gh repo view failed: {result.stderr}"
+        raise RuntimeError(msg)
+    parts = result.stdout.strip().split()
+    owner, repo_name = parts[0], parts[1]
+    query = '{ repository(owner: "%s", name: "%s") { id } }' % (owner, repo_name)
+    result = subprocess.run(
+        ["gh", "api", "graphql", "-f", f"query={query}"],
+        cwd=cwd,
+        capture_output=True,
+        text=True,
+    )
+    if result.returncode != 0:
+        msg = f"gh graphql repo ID query failed: {result.stderr}"
+        raise RuntimeError(msg)
+    data = json.loads(result.stdout)
+    node_id = data["data"]["repository"]["id"]
+    return owner, repo_name, node_id
+
+
+def _ensure_and_resolve_labels(cwd: str, owner: str, repo_name: str) -> list[str]:
+    """Create labels if absent, resolve and return their node IDs."""
+    label_defs = [
+        ("recipe:implementation", "0E8A16"),
+        ("enhancement", "a2eeef"),
+    ]
+    for name, color in label_defs:
+        subprocess.run(
+            ["gh", "label", "create", name, "--force", "--color", color],
+            cwd=cwd,
+            capture_output=True,
+        )
+        time.sleep(1)
+    query = (
+        '{ repository(owner: "%s", name: "%s") { '
+        'impl: label(name: "recipe:implementation") { id } '
+        'enh: label(name: "enhancement") { id } } }' % (owner, repo_name)
+    )
+    result = subprocess.run(
+        ["gh", "api", "graphql", "-f", f"query={query}"],
+        cwd=cwd,
+        capture_output=True,
+        text=True,
+    )
+    if result.returncode != 0:
+        msg = f"gh graphql label query failed: {result.stderr}"
+        raise RuntimeError(msg)
+    data = json.loads(result.stdout)
+    repo = data["data"]["repository"]
+    return [repo["impl"]["id"], repo["enh"]["id"]]
+
+
+def batch_create_issues(
+    workspace: str,
+    chunk_size: str = "20",
+) -> dict[str, str]:
+    """Batch-create GitHub issues from validated ticket body files via GraphQL."""
+    temp_dir = Path(workspace) / ".autoskillit" / "temp" / "validate-audit"
+    ticket_bodies = sorted(temp_dir.glob("ticket_body_*.md"))
+    if not ticket_bodies:
+        return {"issue_urls": "", "issue_count": "0"}
+
+    parsed: list[tuple[str, str, str]] = []
+    for f in ticket_bodies:
+        raw = f.read_text()
+        m = re.match(r"ticket_body_(\w+)_\d+_(.+)\.md", f.name)
+        source = m.group(1) if m else "unknown"
+        ts = m.group(2) if m else ""
+        title = _extract_title(raw)
+        body = _strip_ticket_body(raw)
+        summary_path = temp_dir / f"validation_summary_{source}_{ts}.md"
+        if summary_path.exists():
+            body += "\n\n---\n\n" + summary_path.read_text()
+        parsed.append((title, body, ts))
+
+    owner, repo_name, repo_id = _resolve_repo_identity(workspace)
+    label_ids = _ensure_and_resolve_labels(workspace, owner, repo_name)
+
+    all_urls: list[str] = []
+    chunk_sz = int(chunk_size) if chunk_size else 20
+    for offset in range(0, len(parsed), chunk_sz):
+        chunk = parsed[offset : offset + chunk_sz]
+        mutation_parts = []
+        variables: dict[str, object] = {"repoId": repo_id, "labelIds": label_ids}
+        for idx, (title, body, _) in enumerate(chunk):
+            alias = f"issue{idx}"
+            mutation_parts.append(
+                f"{alias}: createIssue(input: $i{idx}) {{ issue {{ number url }} }}"
+            )
+            variables[f"i{idx}"] = {
+                "repositoryId": repo_id,
+                "title": title,
+                "body": body,
+                "labelIds": label_ids,
+            }
+        mutation = (
+            "mutation("
+            + ",".join(f"$i{k}: CreateIssueInput!" for k in range(len(chunk)))
+            + ") {"
+            + " ".join(mutation_parts)
+            + "}"
+        )
+        payload = json.dumps({"query": mutation, "variables": variables})
+        result = subprocess.run(
+            ["gh", "api", "graphql", "--input", "-"],
+            input=payload,
+            cwd=workspace,
+            capture_output=True,
+            text=True,
+        )
+        if result.returncode != 0:
+            msg = f"gh graphql createIssue failed: {result.stderr}"
+            raise RuntimeError(msg)
+        data = json.loads(result.stdout)
+        for idx in range(len(chunk)):
+            alias = f"issue{idx}"
+            issue_data = data["data"][alias]["issue"]
+            all_urls.append(issue_data["url"])
+        if offset + chunk_sz < len(parsed):
+            time.sleep(1)
+
+    return {"issue_urls": ",".join(all_urls), "issue_count": str(len(all_urls))}
--- a/src/autoskillit/recipe/_cmd_rpc.py
+++ b/src/autoskillit/recipe/_cmd_rpc.py
@@ -502,8 +502,8 @@ def _strip_ticket_body(raw: str) -> str:
             continue
         if "| CONTESTED |" in line or "| VALID BUT EXCEPTION WARRANTED |" in line:
             continue
-        if re.search(r"\*\*Contested:\*\s*\d+", line) or re.search(
-            r"\*\*Exception warranted:\*\s*\d+", line
+        if re.search(r"\*\*Contested:\*\*\s+\d+", line) or re.search(
+            r"\*\*Exception warranted:\*\*\s+\d+", line
         ):
             continue
         if "**Exception note:**" in line:
@@ -532,7 +532,7 @@ def _resolve_repo_identity(cwd: str) -> tuple[str, str, str]:
         raise RuntimeError(msg)
     parts = result.stdout.strip().split()
     owner, repo_name = parts[0], parts[1]
-    query = '{ repository(owner: "%s", name: "%s") { id } }' % (owner, repo_name)
+    query = f'{{ repository(owner: "{owner}", name: "{repo_name}") {{ id }} }}'
     result = subprocess.run(
         ["gh", "api", "graphql", "-f", f"query={query}"],
         cwd=cwd,
@@ -561,9 +561,9 @@ def _ensure_and_resolve_labels(cwd: str, owner: str, repo_name: str) -> list[str
         )
         time.sleep(1)
     query = (
-        '{ repository(owner: "%s", name: "%s") { '
-        'impl: label(name: "recipe:implementation") { id } '
-        'enh: label(name: "enhancement") { id } } }' % (owner, repo_name)
+        f'{{ repository(owner: "{owner}", name: "{repo_name}") {{'
+        f' impl: label(name: "recipe:implementation") {{ id }}'
+        f' enh: label(name: "enhancement") {{ id }} }} }}'
     )
     result = subprocess.run(
         ["gh", "api", "graphql", "-f", f"query={query}"],
--- a/src/autoskillit/recipe/_cmd_rpc.py
+++ b/src/autoskillit/recipe/_cmd_rpc.py
@@ -531,8 +531,13 @@ def _resolve_repo_identity(cwd: str) -> tuple[str, str, str]:
         msg = f"gh repo view failed: {result.stderr}"
         raise RuntimeError(msg)
     parts = result.stdout.strip().split()
+    if len(parts) < 2:
+        msg = f"Unexpected gh repo view output: {result.stdout!r}"
+        raise RuntimeError(msg)
     owner, repo_name = parts[0], parts[1]
-    query = f'{{ repository(owner: "{owner}", name: "{repo_name}") {{ id }} }}'
+    safe_owner = json.dumps(owner)[1:-1]
+    safe_repo = json.dumps(repo_name)[1:-1]
+    query = f'{{ repository(owner: "{safe_owner}", name: "{safe_repo}") {{ id }} }}'
     result = subprocess.run(
         ["gh", "api", "graphql", "-f", f"query={query}"],
         cwd=cwd,
@@ -542,7 +547,13 @@ def _resolve_repo_identity(cwd: str) -> tuple[str, str, str]:
     if result.returncode != 0:
         msg = f"gh graphql repo ID query failed: {result.stderr}"
         raise RuntimeError(msg)
-    data = json.loads(result.stdout)
+    try:
+        data = json.loads(result.stdout)
+    except json.JSONDecodeError as exc:
+        msg = f"gh graphql repo ID: non-JSON output: {result.stdout!r}"
+        raise RuntimeError(msg) from exc
+    if "errors" in data:
+        raise RuntimeError(f"gh graphql repo ID errors: {data['errors']}")
     node_id = data["data"]["repository"]["id"]
     return owner, repo_name, node_id
 
@@ -560,8 +571,10 @@ def _ensure_and_resolve_labels(cwd: str, owner: str, repo_name: str) -> list[str
             capture_output=True,
         )
         time.sleep(1)
+    safe_owner = json.dumps(owner)[1:-1]
+    safe_repo = json.dumps(repo_name)[1:-1]
     query = (
-        f'{{ repository(owner: "{owner}", name: "{repo_name}") {{'
+        f'{{ repository(owner: "{safe_owner}", name: "{safe_repo}") {{'
         f' impl: label(name: "recipe:implementation") {{ id }}'
         f' enh: label(name: "enhancement") {{ id }} }} }}'
     )
@@ -574,8 +587,18 @@ def _ensure_and_resolve_labels(cwd: str, owner: str, repo_name: str) -> list[str
     if result.returncode != 0:
         msg = f"gh graphql label query failed: {result.stderr}"
         raise RuntimeError(msg)
-    data = json.loads(result.stdout)
+    try:
+        data = json.loads(result.stdout)
+    except json.JSONDecodeError as exc:
+        msg = f"gh graphql label query: non-JSON output: {result.stdout!r}"
+        raise RuntimeError(msg) from exc
+    if "errors" in data:
+        raise RuntimeError(f"gh graphql label query errors: {data['errors']}")
     repo = data["data"]["repository"]
+    if repo["impl"] is None or repo["enh"] is None:
+        raise RuntimeError(
+            f"Label resolution returned null: impl={repo['impl']!r} enh={repo['enh']!r}"
+        )
     return [repo["impl"]["id"], repo["enh"]["id"]]
 
 
@@ -584,6 +607,8 @@ def batch_create_issues(
     chunk_size: str = "20",
 ) -> dict[str, str]:
     """Batch-create GitHub issues from validated ticket body files via GraphQL."""
+    if not workspace or not Path(workspace).is_dir():
+        raise ValueError(f"workspace must be an existing directory, got: {workspace!r}")
     temp_dir = Path(workspace) / ".autoskillit" / "temp" / "validate-audit"
     ticket_bodies = sorted(temp_dir.glob("ticket_body_*.md"))
     if not ticket_bodies:
@@ -606,11 +631,16 @@ def batch_create_issues(
     label_ids = _ensure_and_resolve_labels(workspace, owner, repo_name)
 
     all_urls: list[str] = []
-    chunk_sz = int(chunk_size) if chunk_size else 20
+    try:
+        chunk_sz = int(chunk_size) if chunk_size else 20
+    except ValueError as exc:
+        raise ValueError(f"chunk_size must be a positive integer, got: {chunk_size!r}") from exc
+    if chunk_sz <= 0:
+        raise ValueError(f"chunk_size must be positive, got: {chunk_sz}")
     for offset in range(0, len(parsed), chunk_sz):
         chunk = parsed[offset : offset + chunk_sz]
         mutation_parts = []
-        variables: dict[str, object] = {"repoId": repo_id, "labelIds": label_ids}
+        variables: dict[str, object] = {}
         for idx, (title, body, _) in enumerate(chunk):
             alias = f"issue{idx}"
             mutation_parts.append(
@@ -640,10 +670,22 @@ def batch_create_issues(
         if result.returncode != 0:
             msg = f"gh graphql createIssue failed: {result.stderr}"
             raise RuntimeError(msg)
-        data = json.loads(result.stdout)
+        try:
+            data = json.loads(result.stdout)
+        except json.JSONDecodeError as exc:
+            msg = f"gh graphql createIssue: non-JSON output: {result.stdout!r}"
+            raise RuntimeError(msg) from exc
+        if "errors" in data:
+            raise RuntimeError(f"gh graphql createIssue errors: {data['errors']}")
+        resp_data = data.get("data") or {}
         for idx in range(len(chunk)):
             alias = f"issue{idx}"
-            issue_data = data["data"][alias]["issue"]
+            alias_result = resp_data.get(alias)
+            if alias_result is None:
+                raise RuntimeError(f"createIssue response missing alias {alias!r}: {data}")
+            issue_data = alias_result.get("issue")
+            if issue_data is None:
+                raise RuntimeError(f"createIssue alias {alias!r} returned null issue: {data}")
             all_urls.append(issue_data["url"])
         if offset + chunk_sz < len(parsed):
             time.sleep(1)
--- a/src/autoskillit/recipes/full-audit.yaml
+++ b/src/autoskillit/recipes/full-audit.yaml
@@ -4,7 +4,7 @@ description: >-
   audit-review-decisions in parallel, validate each report individually, and create a
   GitHub issue from each validated report.
 summary: "branch → audit×6 ∥ → validate×6 ∥ → issue×6 ∥ → done"
-recipe_version: "1.1.0"
+recipe_version: "1.2.0"
 requires_packs: [audit, github]
 
 ingredients:
@@ -23,7 +23,7 @@ kitchen_rules:
   - >-
     NEVER use native Claude Code tools (Read, Grep, Glob, Edit, Write, Bash, Agent,
     WebFetch, WebSearch, NotebookEdit) from the orchestrator. All work is delegated
-    through run_skill.
+    through MCP tools (run_skill, run_cmd, run_python).
   - >-
     Each audit chain (tests, cohesion, arch, feature-gates, docs, review-decisions) is
     independent. A failure in one chain MUST NOT block or affect the other five chains.
@@ -33,10 +33,6 @@ kitchen_rules:
     concurrent run_skill calls. When any call completes, immediately dispatch
     the next item from the queue. Do not wait for all slots to finish before
     refilling — fill each freed slot individually.
-  - >-
-    prepare-issue calls MUST be sequential — never call prepare-issue in
-    parallel with itself. It hits the GitHub API and parallel calls create
-    request storms. Other tool types may overlap with prepare-issue.
   - >-
     Report paths are discovered from session output — look for file paths
     in {{AUTOSKILLIT_TEMP}}/ directories. Pass discovered paths as positional
@@ -151,39 +147,24 @@ steps:
     on_exhausted: escalate_stop
 
   create_issues:
-    tool: run_skill
-    model: ""
-    description: "Create GitHub issues from validated reports"
+    tool: run_python
+    description: "Batch-create GitHub issues via GraphQL from validated ticket body files"
     with:
-      skill_command: "/autoskillit:prepare-issue"
-      cwd: "${{ inputs.workspace }}"
-      step_name: "create_issues"
+      callable: "autoskillit.recipe._cmd_rpc.batch_create_issues"
+      workspace: "${{ inputs.workspace }}"
+      timeout: 120
     note: >
-      SEQUENTIAL ISSUE CREATION — validate-audit splits each validated report
-      into per-ticket body files (ticket_body_{source}_{N}_{ts}.md) according to the
-      grouping manifest.
+      BATCHED ISSUE CREATION via GraphQL — The callable discovers all
+      ticket_body_{source}_{N}_{ts}.md files in the validate-audit temp directory,
+      parses titles and bodies, includes validation summaries, resolves repository
+      and label IDs, and executes a single batched createIssue mutation with aliases.
+      Chunks at 20 issues per request if needed.
 
-      For each audit type that succeeded in validation:
-        1. Discover the grouping manifest (grouping_manifest_{source}_{ts}.md) in the
-           validate-audit session output.
-        2. Read the manifest to get the list of ticket body file paths.
-        3. Process prepare-issue calls one at a time:
-           - run_skill(skill_command="/autoskillit:prepare-issue {ticket_body_path}", cwd="${{ inputs.workspace }}", step_name="issue_{source}_{N}")
-           - Wait for completion.
-           - Proceed to the next ticket body.
-
-      Never call prepare-issue in parallel with itself — it hits the GitHub API
-      and parallel calls create request storms.
-
-      After all issues are created, append the validation summary to each issue body
-      (sequentially): fetch the current body, verify it is non-empty, append a separator
-      and the validation summary content, write the combined text to a temp file, then run:
-        gh issue edit {issue_number} --body-file {combined_temp_file}
-
-      Collect all issue URLs. If any ticket body issue creation fails, note it but continue.
-      Route to escalate_stop ONLY if ALL ticket bodies across ALL audit types fail.
-    capture_list:
-      issue_urls: "${{ result.issue_url }}"
+      No headless sessions are spawned. Rate-limit cost: 5 pts per chunk
+      (vs 90 pts for 18 sequential REST calls).
+    capture:
+      issue_urls: "${{ result.issue_urls }}"
+      issue_count: "${{ result.issue_count }}"
     on_success: done
     on_failure: escalate_stop
     retries: 1
@@ -193,11 +174,10 @@ steps:
     action: stop
     message: >-
       Full audit complete. GitHub issues created from validated audit reports.
-      Emit the L3 result sentinel with issue_urls included as a comma-separated
-      string of the created issue URLs. context.issue_urls contains the list of
-      created issue URLs — join them with commas in the sentinel JSON:
+      Emit the L3 result sentinel. context.issue_urls is a comma-separated
+      string of the created issue URLs — include it directly in the sentinel JSON:
       {"success": true, "reason": "completed", "summary": "Full audit complete.
-      GitHub issues created.", "issue_urls": "https://github.com/org/repo/issues/1,https://github.com/org/repo/issues/2"}
+      GitHub issues created.", "issue_urls": "${{ context.issue_urls }}"}
       Artifacts are in {{AUTOSKILLIT_TEMP}}/.
 
   escalate_stop:
--- a/tests/recipe/test_cmd_rpc.py
+++ b/tests/recipe/test_cmd_rpc.py
@@ -10,6 +10,7 @@ from unittest.mock import patch
 import pytest
 
 from autoskillit.recipe._cmd_rpc import (
+    batch_create_issues,
     check_dropped_healthy_loop,
     check_eject_limit,
     commit_guard,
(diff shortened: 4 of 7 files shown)


src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
         shutil.rmtree(dest)
     shutil.copytree(research_dir, dest)
     return {"local_bundle_path": str(dest)}
+
+
+# ─── batch_create_issues helpers ────────────────────────────────────────────
```
Trecek, line 482: [warning] arch: batch_create_issues and its helpers add ~175 lines of GitHub API orchestration directly to _cmd_rpc.py. The module has grown into a catch-all for recipe-layer subprocess calls — consider whether a dedicated module should be split off to prevent this file from becoming a God module.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+            r"\*\*Exception warranted:\*\*\s+\d+", line
+        ):
+            continue
+        if "**Exception note:**" in line:
+            continue
+        if re.match(r"## Findings with Exceptions\s*$", line):
+            skip_exceptions_section = True
+            continue
+        if skip_exceptions_section:
+            if line.strip().startswith("---"):
```
Trecek, line 515: [warning] bugs: _strip_ticket_body sets skip_exceptions_section=False when it encounters '---' but then `continue`s at the same conditional, so the terminator line itself is dropped from the output. If this '---' is a Markdown section divider relied upon by downstream content, it will be silently missing.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+                skip_exceptions_section = False
+            continue
+        result.append(line)
+    return "\n".join(result)
+
+
+def _resolve_repo_identity(cwd: str) -> tuple[str, str, str]:
+    """Return (owner, repo_name, repo_node_id) for the given workspace."""
+    result = subprocess.run(
```
Trecek, line 524: [warning] arch: _resolve_repo_identity and _ensure_and_resolve_labels call subprocess.run(['gh', ...]) directly from the recipe layer. CLAUDE.md §3.4 references GitHubFetcher as the Python API equivalent for GitHub mutations — bypassing it here creates a second mutation path in the recipe layer.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+        ["gh", "repo", "view", "--json", "owner,name", "-q", '.owner.login + " " + .name'],
+        cwd=cwd,
+        capture_output=True,
+        text=True,
+    )
+    if result.returncode != 0:
+        msg = f"gh repo view failed: {result.stderr}"
+        raise RuntimeError(msg)
+    parts = result.stdout.strip().split()
+    owner, repo_name = parts[0], parts[1]
```
Trecek, line 537: [critical] defense: No bounds check on `parts` before unpacking. If `gh repo view` succeeds (returncode==0) but stdout is empty or has only one token, `parts[0]` or `parts[1]` raises IndexError with no context. Add: `if len(parts) < 2: raise RuntimeError(f'Unexpected gh repo view output: {result.stdout!r}')`

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+    query = f'{{ repository(owner: "{owner}", name: "{repo_name}") {{ id }} }}'
```
Trecek: [warning] defense: owner and repo_name are interpolated directly into GraphQL query f-strings at L535 and L564 without sanitization. If a repo owner/name contained '"' or '}' (possible with GitHub Apps), this could corrupt the query. Use json.dumps(owner)[1:-1] when interpolating into GraphQL string literals.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+    result = subprocess.run(
+        ["gh", "api", "graphql", "-f", f"query={query}"],
+        cwd=cwd,
+        capture_output=True,
+        text=True,
+    )
+    if result.returncode != 0:
+        msg = f"gh graphql repo ID query failed: {result.stderr}"
+        raise RuntimeError(msg)
+    data = json.loads(result.stdout)
```
Trecek: [critical] defense: `json.loads(result.stdout)` is unguarded. `gh api graphql` can emit non-JSON output on auth failure or rate limiting (plain-text error). Wrap in try/except and re-raise as RuntimeError with raw stdout included to preserve error context.

Trecek (commented):

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+    node_id = data["data"]["repository"]["id"]
```
Trecek, line 557: [critical] defense: GraphQL errors in the response body (`data['errors']` field) are not checked. A successful returncode==0 response can carry a top-level errors array (e.g. repo not found, insufficient scope). The subsequent KeyError on `data['data']['repository']['id']` will be the only signal, with no GitHub error message surfaced. Check `if 'errors' in data: raise RuntimeError(data['errors'])`.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+    result = subprocess.run(
+        ["gh", "api", "graphql", "-f", f"query={query}"],
+        cwd=cwd,
+        capture_output=True,
+        text=True,
+    )
+    if result.returncode != 0:
+        msg = f"gh graphql label query failed: {result.stderr}"
+        raise RuntimeError(msg)
+    data = json.loads(result.stdout)
```
Trecek: [critical] defense: `json.loads(result.stdout)` in _ensure_and_resolve_labels is unguarded against json.JSONDecodeError. Non-JSON output (auth error, rate limit HTML) will raise with no context about which call failed. Wrap and re-raise as RuntimeError with raw stdout.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+    repo = data["data"]["repository"]
```
Trecek, line 597: [warning] defense: `repo['impl']` or `repo['enh']` can be None if label creation at L557 silently failed (its returncode is ignored). Accessing `['id']` on None raises TypeError. Check for None before indexing: `if repo['impl'] is None or repo['enh'] is None: raise RuntimeError(...)`.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+    return [repo["impl"]["id"], repo["enh"]["id"]]
+
+
+def batch_create_issues(
```
Trecek, line 605: [warning] defense: `workspace` parameter has no validation at construction time. A None or non-existent directory path will produce a confusing FileNotFoundError deep inside subprocess.run. Add upfront guard: `if not workspace or not Path(workspace).is_dir(): raise ValueError(...)`

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+        summary_path = temp_dir / f"validation_summary_{source}_{ts}.md"
+        if summary_path.exists():
+            body += "\n\n---\n\n" + summary_path.read_text()
+        parsed.append((title, body, ts))
+
+    owner, repo_name, repo_id = _resolve_repo_identity(workspace)
+    label_ids = _ensure_and_resolve_labels(workspace, owner, repo_name)
+
+    all_urls: list[str] = []
+    chunk_sz = int(chunk_size) if chunk_size else 20
```
Trecek: [critical] defense: `int(chunk_size)` can raise ValueError on a non-numeric string with no catch. Also `int(chunk_size) if chunk_size else 20` silently uses the default when chunk_size='' but NOT when chunk_size='0' — a zero value causes an infinite loop. Validate: parse with try/except, then `if chunk_sz <= 0: raise ValueError(...)`.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+    for offset in range(0, len(parsed), chunk_sz):
+        chunk = parsed[offset : offset + chunk_sz]
+        mutation_parts = []
+        variables: dict[str, object] = {"repoId": repo_id, "labelIds": label_ids}
```
Trecek: [critical] slop: The `variables` dict is initialized with `repoId` and `labelIds` keys that have no corresponding `$repoId`/`$labelIds` declarations in the mutation signature (which only declares `$i{k}: CreateIssueInput!`). These keys are serialized into the GraphQL payload but unused server-side. Remove `repoId` and `labelIds` from the top-level variables initializer — the data is already embedded inside each `$i{k}` input object.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+            ["gh", "api", "graphql", "--input", "-"],
+            input=payload,
+            cwd=workspace,
+            capture_output=True,
+            text=True,
+        )
+        if result.returncode != 0:
+            msg = f"gh graphql createIssue failed: {result.stderr}"
+            raise RuntimeError(msg)
+        data = json.loads(result.stdout)
```
Trecek: [critical] defense: `json.loads(result.stdout)` in the mutation loop is unguarded against json.JSONDecodeError. Additionally, GraphQL errors in the response body (`data.get('errors')`) are not checked before accessing `data['data'][alias]['issue']` at L646, which will raise KeyError with no GitHub error message surfaced.

src/autoskillit/recipe/_cmd_rpc.py
```diff
@@ -479,3 +477,175 @@ def export_local_bundle(
...
+        for idx in range(len(chunk)):
+            alias = f"issue{idx}"
+            issue_data = data["data"][alias]["issue"]
```
Trecek: [critical] bugs: `data['data'][alias]['issue']` will raise KeyError/TypeError if the GraphQL response carries a top-level `errors` array (with null data), or if a partial-success response returns null for a specific alias. No guard for `data.get('errors')`, no null-check on `data['data']` or `data['data'][alias]`. A single failed mutation in a batch crashes the entire loop.

src/autoskillit/recipe/_analysis_detectors.py
```diff
@@ -167,6 +172,9 @@ def _detect_dead_outputs(recipe: Recipe, graph: dict[str, set[str]]) -> list[Dat
                 if not isinstance(arg_val, str):
                     continue
                 consumed.update(_CONTEXT_REF_RE.findall(arg_val))
+            # Also scan the step's own message field (used by stop/confirm actions).
```
Trecek: [warning] slop: Comment 'Also scan the step's own message field (used by stop/confirm actions)' describes what the code does rather than why. Consider replacing with a rationale comment, e.g. 'message fields are not recipe args, so the context-ref scanner must handle them separately'.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,319 @@ def test_refetch_issues_builds_query():
     assert "repo" in query_arg
     assert "issue(number: 1)" in query_arg
     assert "issue(number: 2)" in query_arg
+
+
+# ─── batch_create_issues tests ─────────────────────────────────────────────
+
+
+def _make_side_effect(repo_id="R_123", label_ids=None, issue_data=None):
```
Trecek, line 179: [warning] tests: _make_side_effect default builds alias_data with 3 issues (issue0, issue1, issue2), but most tests that use it create only 1-2 ticket files. The mock createIssue response contains more aliases than the mutation will request, creating a structural mismatch that can mask bugs in alias-key validation.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,319 @@ def test_refetch_issues_builds_query():
...
+        issue_data = [
+            {"number": 1, "url": "https://github.com/org/repo/issues/1"},
+            {"number": 2, "url": "https://github.com/org/repo/issues/2"},
+            {"number": 3, "url": "https://github.com/org/repo/issues/3"},
+        ]
+    alias_data = {}
+    for idx, issue in enumerate(issue_data):
+        alias_data[f"issue{idx}"] = {"issue": issue}
+    return [
+        # gh repo view
```
Trecek: [warning] slop: Inline comments '# gh repo view', '# gh api graphql (repo ID)', '# gh label create recipe:implementation', etc. narrate the mock call sequence already visible from the CompletedProcess values. Remove these narration comments — they add no non-obvious context.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,319 @@ def test_refetch_issues_builds_query():
...
+        "**Exception note:** this is an exception.\n"
+    )
+    with (
+        patch("autoskillit.recipe._cmd_rpc.subprocess.run") as mock_run,
+        patch("autoskillit.recipe._cmd_rpc.time.sleep"),
+    ):
+        mock_run.side_effect = _make_side_effect()
+        batch_create_issues(workspace=str(tmp_path))
+    # Find the createIssue mutation call
+    mutation_call: dict[str, object] = {}
```
Trecek, line 263: [warning] tests: `mutation_call` is initialized to an empty dict. If no call with '--input' in args is found, the subsequent `mutation_call['variables']['i0']['body']` assertions raise KeyError rather than a meaningful failure. Assert that mutation_call is non-empty after the loop.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,319 @@ def test_refetch_issues_builds_query():
...
+        patch("autoskillit.recipe._cmd_rpc.time.sleep"),
+    ):
+        mock_run.side_effect = _make_side_effect()
+        batch_create_issues(workspace=str(tmp_path))
+    for call in mock_run.call_args_list:
+        kwargs = call[1]
+        if kwargs.get("input"):
+            mutation_call = json.loads(kwargs["input"])
+            assert mutation_call["variables"]["i0"]["title"] == "Audit: Missing test coverage"
+            break
```
Trecek, line 298: [warning] tests: Assertions are inside a for/if block with a break, but there is no post-loop assertion to verify a mutation call was actually found and checked. If no call with input kwarg is found, the test passes vacuously without asserting anything.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,319 @@ def test_refetch_issues_builds_query():
...
+    va_dir.mkdir(parents=True)
+    (va_dir / "ticket_body_tests_1_2026-01-01_120000.md").write_text("# Issue One\n\nBody one.")
+    (va_dir / "ticket_body_tests_2_2026-01-01_120000.md").write_text("# Issue Two\n\nBody two.")
+    with (
+        patch("autoskillit.recipe._cmd_rpc.subprocess.run") as mock_run,
+        patch("autoskillit.recipe._cmd_rpc.time.sleep"),
+    ):
+        mock_run.side_effect = _make_side_effect()
+        batch_create_issues(workspace=str(tmp_path))
+    for call in mock_run.call_args_list:
```
Trecek, line 319: [warning] tests: Same silent-pass risk: assertions are inside a for/if block with a break, no post-loop guard. If no mutation call is found in mock_run.call_args_list, all assertions are skipped and the test passes vacuously.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,319 @@ def test_refetch_issues_builds_query():
...
+        patch("autoskillit.recipe._cmd_rpc.time.sleep"),
+    ):
+
+        def side_effect_factory():
+            call_count = [0]
+
+            def side_effect(_args, **_kwargs):
+                c = call_count[0]
+                call_count[0] += 1
+                # repo view
```
Trecek: [warning] slop: Comments '# repo view', '# repo ID query', '# label create', '# label IDs query', '# createIssue chunks' in side_effect_factory narrate the adjacent if-branch conditions. Remove them — the condition checks (c==0, c==1, etc.) are already readable without commentary.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,319 @@ def test_refetch_issues_builds_query():
...
+                    return subprocess.CompletedProcess(
+                        args=[],
+                        returncode=0,
+                        stdout=json.dumps(
+                            {"data": {"repository": {"impl": {"id": "L_1"}, "enh": {"id": "L_2"}}}}
+                        ),
+                        stderr="",
+                    )
+                # createIssue chunks
+                if c == 5:
```
Trecek, line 376: [critical] tests: Chunking test mock responses for c==5,6,7 use chunk-relative alias keys (issue0..issue9 for each chunk). The production code iterates `for idx in range(len(chunk))` using chunk-relative indices per loop iteration, so this is correct. However, if the implementation ever changes to global indices, this test would silently pass with wrong data. Add an assertion that verifies the total issue_count is 25 after batch_create_issues completes.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,319 @@ def test_refetch_issues_builds_query():
...
+                    return subprocess.CompletedProcess(
+                        args=[], returncode=0, stdout=json.dumps({"data": data7}), stderr=""
+                    )
+                return subprocess.CompletedProcess(args=[], returncode=0, stdout="", stderr="")
+
+            return side_effect
+
+        mock_run.side_effect = side_effect_factory()
+        batch_create_issues(workspace=str(tmp_path), chunk_size="10")
+    mutation_calls = sum(
```
Trecek, line 421: [warning] tests: mutation_calls is counted by string-matching `call[1]['input']` for 'createIssue'. This relies on JSON serialization producing a string containing 'createIssue', which is correct but fragile if the JSON key casing or structure changes. A more robust check would inspect the parsed mutation structure.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,319 @@ def test_refetch_issues_builds_query():
...
+        1
+        for call in mock_run.call_args_list
+        if call[1].get("input") and "createIssue" in call[1]["input"]
+    )
+    assert mutation_calls == 3
+
+
+def test_batch_create_issues_appends_validation_summary(tmp_path):
```
Trecek, line 430: [warning] tests: test_batch_create_issues_appends_validation_summary uses _make_side_effect() with default 3-issue alias_data, but only 1 ticket file is created. The mock createIssue response contains aliases issue0/issue1/issue2 but the mutation only sends issue0 — response/request mismatch that can hide correctness bugs.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,319 @@ def test_refetch_issues_builds_query():
...
+    (va_dir / "validation_summary_tests_2026-01-01_120000.md").write_text(
+        "## Validation Summary\nAll clear."
+    )
+    with (
+        patch("autoskillit.recipe._cmd_rpc.subprocess.run") as mock_run,
+        patch("autoskillit.recipe._cmd_rpc.time.sleep"),
+    ):
+        mock_run.side_effect = _make_side_effect()
+        batch_create_issues(workspace=str(tmp_path))
+    for call in mock_run.call_args_list:
```
Trecek, line 446: [warning] tests: Same silent-pass risk as L299/L313: assertions are inside a for/if block with break, no post-loop guard verifying a mutation call was actually found.

Trecek (commented):
AutoSkillit review: changes_requested — 8 blocking (critical) findings detected. See inline comments for required fixes. No blocking changes can proceed until these are addressed.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Added `if len(parts) < 2: raise RuntimeError(f'Unexpected gh repo view output: {result.stdout!r}')` before unpacking.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Both interpolation sites now use `json.dumps(owner)[1:-1]` and `json.dumps(repo_name)[1:-1]` to sanitize before embedding in GraphQL string literals.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Wrapped in try/except JSONDecodeError that re-raises as RuntimeError with raw stdout included for diagnostics.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Added `if 'errors' in data: raise RuntimeError(data['errors'])` before accessing `data['data']['repository']['id']`.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Wrapped in try/except JSONDecodeError + GraphQL errors check, consistent with the repo-ID call site.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Added `if repo['impl'] is None or repo['enh'] is None: raise RuntimeError(...)` before indexing into the returned objects.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Added `if not workspace or not Path(workspace).is_dir(): raise ValueError(...)` at the top of batch_create_issues.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Wrapped in try/except ValueError and added `if chunk_sz <= 0: raise ValueError(...)` guard. Both the parse error and the infinite-loop cases are now caught upfront.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Removed repoId and labelIds from the top-level variables initializer. They had no corresponding $repoId/$labelIds declarations in the mutation signature; the values are already embedded in each $i{k} input object.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Added try/except JSONDecodeError and `if 'errors' in data: raise RuntimeError(data['errors'])` before iterating aliases in the mutation loop.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Now uses `resp_data.get(alias)` with an explicit None check, and `alias_result.get('issue')` with a None check, raising RuntimeError with the full response included for each failure case.

src/autoskillit/recipe/_analysis_detectors.py
Trecek: Agreed — fixed in 46432ad2. Replaced with rationale comment: 'message fields are not recipe args; scanner must handle them separately.'

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Default issue_data reduced to 1 issue. test_batch_create_issues_discovers_ticket_bodies now passes explicit 3-issue data; test_batch_create_issues_constructs_graphql_mutation passes explicit 2-issue data. Mock/request counts now match exactly.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Removed all narration comments from _make_side_effect return list. The CompletedProcess values are self-documenting.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Added `assert mutation_call, 'no createIssue mutation call found in mock_run calls'` after the search loop, so KeyError on the empty dict is replaced by a clear assertion failure.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Added `found` flag and `assert found, 'no createIssue mutation call found in mock_run calls'` after the loop. Assertions can no longer pass vacuously.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Same `found` flag + post-loop assert pattern applied. Consistent with the L299 fix.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Removed '# repo view', '# repo ID query', '# label create', '# label IDs query' narration comments. The if-branch conditions (c==0, c==1, etc.) are readable without commentary.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Captured the return value of batch_create_issues and added `assert result['issue_count'] == '25'`. This verifies the total count across all 3 chunks and would catch a global-index bug immediately.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. _make_side_effect now defaults to 1-issue alias_data, matching the single ticket file this test creates. Request/response alias counts are now aligned.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Added `found` flag and `assert found, 'no createIssue mutation call found in mock_run calls'` after the loop. Consistent with L299/L313 fixes.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Valid observation — flagged for design decision. The 175-line addition is within the current task scope; extracting to a dedicated module is an architectural refactoring that warrants a separate planning pass to decide on module boundaries and migration strategy.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Valid observation — flagged for design decision. The current behavior intentionally drops the '---' line that terminates the Findings with Exceptions section (it acts as a section boundary marker, not content). Whether downstream consumers rely on its presence in the output depends on the actual ticket body format — this warrants explicit verification before changing the behavior.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Valid observation — flagged for design decision. CLAUDE.md §3.5 references update_issue_body() on GitHubFetcher specifically for issue body mutations; subprocess.run(['gh', ...]) for repo identity and label queries is an established recipe-layer pattern. Migrating to GitHubFetcher would require extending its API for these operations — a separate refactoring task.

tests/recipe/test_cmd_rpc.py
Trecek: Valid observation — flagged for design decision. The string-match approach is correct for current JSON serialization but couples the test to key naming conventions. A more robust approach (parsing and inspecting mutation structure) would require redesigning the side_effect_factory pattern. Flagged for consideration in a test-quality improvement pass.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,326 @@ def test_refetch_issues_builds_query():
...
+            f"validated: true\n\n# Title {n}\n\n| col1 | col2 |\n"
+        )
+    with (
+        patch("autoskillit.recipe._cmd_rpc.subprocess.run") as mock_run,
+        patch("autoskillit.recipe._cmd_rpc.time.sleep"),
+    ):
+        mock_run.side_effect = _make_side_effect(
+            issue_data=[
+                {"number": 1, "url": "https://github.com/org/repo/issues/1"},
+                {"number": 2, "url": "https://github.com/org/repo/issues/2"},
```
Trecek, line 235: [critical] tests: Syntax error: the `_make_side_effect(...)` call is missing a closing parenthesis before `result = batch_create_issues(...)` on L241. This causes a SyntaxError at import time, preventing the entire test file from loading.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,326 @@ def test_refetch_issues_builds_query():
...
+# ─── batch_create_issues tests ─────────────────────────────────────────────
+
+
+def _make_side_effect(repo_id="R_123", label_ids=None, issue_data=None):
+    if label_ids is None:
+        label_ids = ["L_1", "L_2"]
+    if issue_data is None:
+        issue_data = [{"number": 1, "url": "https://github.com/org/repo/issues/1"}]
+    alias_data = {}
+    for idx, issue in enumerate(issue_data):
```
Trecek, line 185: [critical] slop: Lines 185-187 are duplicate dead code: `if issue_data is None` and `alias_data = {}` are already set at L182-184. These blocks never execute and were likely left by an AI edit. Remove lines 185-187.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,326 @@ def test_refetch_issues_builds_query():
...
+            break
+    assert found, "no createIssue mutation call found in mock_run calls"
+
+
+def test_batch_create_issues_constructs_graphql_mutation(tmp_path):
+    va_dir = tmp_path / ".autoskillit" / "temp" / "validate-audit"
+    va_dir.mkdir(parents=True)
+    (va_dir / "ticket_body_tests_1_2026-01-01_120000.md").write_text("# Issue One\n\nBody one.")
+    (va_dir / "ticket_body_tests_2_2026-01-01_120000.md").write_text("# Issue Two\n\nBody two.")
+    with (
```
Trecek, line 307: [warning] tests: test_batch_create_issues_constructs_graphql_mutation writes ticket files without `validated: true` frontmatter. If batch_create_issues ever filters files by that marker, this test silently passes without exercising the mutation path it intends to test.

tests/recipe/test_cmd_rpc.py
```diff
@@ -170,3 +171,326 @@ def test_refetch_issues_builds_query():
...
+    assert found, "no createIssue mutation call found in mock_run calls"
+
+
+def test_batch_create_issues_handles_no_tickets(tmp_path):
+    va_dir = tmp_path / ".autoskillit" / "temp" / "validate-audit"
+    va_dir.mkdir(parents=True)
+    # Leave directory empty
+    result = batch_create_issues(workspace=str(tmp_path))
+    assert result == {"issue_urls": "", "issue_count": "0"}
+
```
Trecek, line 464: [warning] tests: test_batch_create_issues_handles_no_tickets calls batch_create_issues with no mocking of subprocess.run. If the function makes any subprocess calls on the empty-directory early-return path, the test will fail with an unexpected real subprocess invocation instead of the intended assertion.

src/autoskillit/recipes/full-audit.yaml
```diff
@@ -152,39 +148,24 @@ steps:
...
+      ticket_body_{source}_{N}_{ts}.md files in the validate-audit temp directory,
+      parses titles and bodies, includes validation summaries, resolves repository
+      and label IDs, and executes a single batched createIssue mutation with aliases.
+      Chunks at 20 issues per request if needed.
+
+      No headless sessions are spawned. Rate-limit cost: 5 pts per chunk
+      (vs 90 pts for 18 sequential REST calls).
+    capture:
+      issue_urls: "${{ result.issue_urls }}"
+      issue_count: "${{ result.issue_count }}"
```
Trecek, line 167: [warning] cohesion: The captured output 'issue_count' and step name 'create_issues' are hard-coded as string literals in _analysis_detectors.py at L149. If the step or capture key is renamed, the detector silently breaks. These names should have a single source of truth rather than four mirrored copies across full-audit.yaml, _cmd_rpc.py, skill_contracts.yaml, and the detector.

src/autoskillit/recipe/_analysis_detectors.py
```diff
@@ -144,6 +144,11 @@ def _is_observability_capture(cap_key: str, step_name: str, step: RecipeStep) ->
     if cap_key == "local_bundle_path" and step_name == "export_local_bundle":
         return True
 
+    # batch_create_issues issue_count: captured output used for telemetry/logging,
+    # not threaded to downstream recipe steps (the done message uses issue_urls only).
+    if cap_key == "issue_count" and step_name == "create_issues" and step.tool == "run_python":
```
Trecek, line 153: [warning] cohesion: Observability exemption hard-codes capture key ('issue_count') and step name ('create_issues') as string literals. These names are already defined in full-audit.yaml, _cmd_rpc.py, and skill_contracts.yaml — this creates a fourth mirrored copy that will silently diverge on rename. Consider driving the exemption from a contract-level annotation.

Trecek (commented):

Trecek (commented):
AutoSkillit review: changes required. 2 blocking (critical) findings detected — see inline comments.

**Blocking findings:**
- `tests/recipe/test_cmd_rpc.py:235` [critical/tests]: Syntax error — missing closing parenthesis in `_make_side_effect` call; will cause SyntaxError at import time for the entire test file
- `tests/recipe/test_cmd_rpc.py:185` [critical/slop]: Duplicate dead code block at L185-187 (duplicate `if issue_data is None` + `alias_data = {}`); unreachable and left by AI edit

**Warning findings:**
- `tests/recipe/test_cmd_rpc.py:307` [warning/tests]: `test_batch_create_issues_constructs_graphql_mutation` writes ticket files without `validated: true` frontmatter — may silently pass vacuously
- `tests/recipe/test_cmd_rpc.py:464` [warning/tests]: `test_batch_create_issues_handles_no_tickets` has no subprocess.run mock — real subprocess calls on the early-return path would cause a test failure
- `src/autoskillit/recipe/_analysis_detectors.py:149` [warning/cohesion]: Observability exemption hard-codes step/capture key names — four mirrored copies of these names will diverge silently on rename
- `src/autoskillit/recipes/full-audit.yaml:168` [warning/cohesion]: Same naming fragility issue from the recipe side

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Added bounds check: `if len(parts) < 2: raise RuntimeError(...)` before unpacking `parts[0]`/`parts[1]`.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. owner and repo_name are now sanitized via `json.dumps(owner)[1:-1]` before interpolation into GraphQL f-strings.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. `json.loads()` in `_resolve_repo_identity` is now wrapped in `try/except JSONDecodeError` with a context-rich `RuntimeError`.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Added `data.get('errors')` check after the repo ID GraphQL response before accessing `data['data']['repository']['id']`.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. `json.loads()` in `_ensure_and_resolve_labels` is now wrapped in `try/except JSONDecodeError` with context-rich `RuntimeError`.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Added `if repo['impl'] is None or repo['enh'] is None: raise RuntimeError(...)` guard before accessing `['id']`.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Added `if not workspace or not Path(workspace).is_dir(): raise ValueError(...)` at `batch_create_issues` entry point.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. `int(chunk_size)` is now wrapped in `try/except ValueError` and followed by `if chunk_sz <= 0: raise ValueError(...)` to guard against the infinite-loop case.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Removed unused `repoId` and `labelIds` keys from the top-level `variables` dict. The mutation signature only declares `$i{k}: CreateIssueInput!` variables, not `$repoId`/`$labelIds`.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. `json.loads()` in the mutation loop is now guarded with `try/except JSONDecodeError`. `data.get('errors')` is checked before accessing `data['data'][alias]`.

src/autoskillit/recipe/_cmd_rpc.py
Trecek: Agreed — fixed in 46432ad2. Added null-guard: `issue_node = data['data'].get(alias, {})` with check for `None` before accessing `['issue']`, preventing KeyError/TypeError on partial-success or error responses.

src/autoskillit/recipe/_analysis_detectors.py
Trecek: Agreed — fixed in 46432ad2. Replaced slop comment describing what the code does with a rationale comment explaining why the `message` field scanner is separate from the recipe-args scanner.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. `_make_side_effect` default changed to 1-issue `alias_data` (matching the default `issue_data` list). Tests that need more issues now pass explicit `issue_data` arguments.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Removed narration comments (`# gh repo view`, `# gh api graphql (repo ID)`, etc.) from `_make_side_effect`.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Added `assert mutation_call, "no createIssue mutation call found in mock_run calls"` after the loop in `test_batch_create_issues_strips_body_content`.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Added `assert found, "no createIssue mutation call found in mock_run calls"` post-loop guard in `test_batch_create_issues_extracts_h1_title`.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Added `assert found, "no createIssue mutation call found in mock_run calls"` post-loop guard in `test_batch_create_issues_constructs_graphql_mutation`.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Removed narration comments (`# repo view`, `# repo ID query`, `# label create`, etc.) from `side_effect_factory`.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Chunking test now captures result and asserts `result['issue_count'] == '25'` as a total-correctness guard, anchoring chunk-relative alias indexing to the correct total.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. `test_batch_create_issues_appends_validation_summary` now passes explicit 1-issue `issue_data` to `_make_side_effect`, eliminating the mock/request alias mismatch.

tests/recipe/test_cmd_rpc.py
Trecek: Agreed — fixed in a0058d5e. Added `assert found, "no createIssue mutation call found in mock_run calls"` post-loop guard in `test_batch_create_issues_appends_validation_summary`.

tests/recipe/test_cmd_rpc.py
Trecek: Investigated — this is intentional. `py_compile.compile` confirms the file is syntactically valid. Lines 232-238 show `_make_side_effect(...)` is a properly closed multi-line call with the closing parenthesis on line 238, before `result = batch_create_issues(...)` on line 239. The reviewer likely saw an older commit state before a0058d5e.

tests/recipe/test_cmd_rpc.py
Trecek: Investigated — this is intentional. Lines 185-186 are the for-loop body (`for idx, issue in enumerate(issue_data): alias_data[f"issue{idx}"] = {"issue": issue}`), not duplicate dead code. `alias_data = {}` at line 184 initializes the dict; lines 185-186 populate it. Removing them would leave `alias_data` always empty, breaking `_make_side_effect` entirely.

tests/recipe/test_cmd_rpc.py
Trecek: Investigated — this is intentional. `batch_create_issues` does not filter files by `validated: true`. The glob at line 613 picks up all `ticket_body_*.md` files unconditionally. The only `validated: true` logic is in `_strip_ticket_body()` (line 497), which strips that line from the body content but does not gate file inclusion. The test correctly exercises the mutation path regardless of this frontmatter.

tests/recipe/test_cmd_rpc.py
Trecek: Investigated — this is intentional. The early-return path at lines 614-615 (`if not ticket_bodies: return {...}`) fires before any subprocess call. `_resolve_repo_identity()` and `_ensure_and_resolve_labels()` (which contain `subprocess.run` calls) are only invoked at lines 630-631, after the ticket count check. No mock is needed for this path.

src/autoskillit/recipes/full-audit.yaml
Trecek: Valid observation — flagged for design decision. The 'issue_count' and 'create_issues' string literals do appear in four locations: `full-audit.yaml`, `_cmd_rpc.py`, `skill_contracts.yaml`, and `_analysis_detectors.py:149`. The coupling is real, but the current pattern is consistent with existing per-step exemptions in `_is_observability_capture()` (e.g., `export_local_bundle`, `merge_worktree`). Driving exemptions from contract-level annotations requires schema changes, detector refactoring, and contract format updates — a non-trivial architectural trade-off. Flagging for human decision.

src/autoskillit/recipe/_analysis_detectors.py
Trecek: Valid observation — flagged for design decision. The exemption at L149 was introduced in d0180ea1 as a deliberate fix for a false-positive dead-output warning on the `create_issues` step. The pattern mirrors existing per-step exemptions in the same function. The reviewer's suggestion to drive exemptions from contract-level annotations is architecturally sound but requires non-trivial changes to the recipe schema and BFS analysis pipeline. Flagging for human decision.

Trecek (commented):

PR #1961, opened 2026-05-06
Token Summary — Per-Step Model Column and Per-Model Aggregate Breakdown

### LINKED ISSUE #1906: Token summary: add per-step model column and per-model aggregate breakdown
## Summary

With multi-provider support, the token summary table needs to show which model was used for each step, and provide a per-model aggregate breakdown so users can see cost distribution across models.

## Motivation

When steps route through different providers/models (e.g. Minimax M2.7-highspeed for implement, Sonnet for plan, Haiku for triage), the current token summary gives no visibility into which tokens came from which model. This makes cost attribution and provider performance comparison impossible.

### 1. Per-step model column in the existing token summary table

Add a `Model` column to the step-level token breakdown:

```
| Step      | Model                    | Input   | Output | Cache Read | Duration |
|-----------|--------------------------|---------|--------|------------|----------|
| plan      | claude-sonnet-4-6        | 40,275  | 9,767  | 722,007    | 3m 30s   |
| implement | MiniMax-M2.7-highspeed   | 85,000  | 22,000 | 1,200,000  | 8m 12s   |
| verify    | claude-sonnet-4-6        | 44,457  | 16,616 | 842,524    | 4m 56s   |
```

### 2. Per-model aggregate breakdown table

A separate summary table grouping all token usage by model across all steps:

```
| Model                    | Steps Used | Total Input | Total Output | Total Cache Read | Est. Cost |
|--------------------------|------------|-------------|--------------|------------------|-----------|
| claude-sonnet-4-6        | 3          | 128,000     | 42,000       | 2,800,000        | $0.XX     |
| MiniMax-M2.7-highspeed   | 1          | 85,000      | 22,000       | 1,200,000        | $0.XX     |
| claude-haiku-4-5         | 1          | 12,000      | 3,000        | 0                | $0.XX     |
```

## Implementation Notes

- The `model_identifier` is already captured in session diagnostics (`summary.json` → `versions.model_identifier`)
- The token summary hook lives at `src/autoskillit/hooks/token_summary_hook.py`
- Token accumulation is in `src/autoskillit/pipeline/tokens.py`
- The per-step token data likely needs a `model` field added to whatever struct accumulates step results
- `telemetry_fmt.py` handles canonical token display formatting
- Cost estimation per model would need a pricing lookup (could be config-driven or hardcoded for known models)

## Acceptance Criteria

- [ ] Token summary table includes a Model column showing the model used for each step
- [ ] A per-model aggregate table is appended below the step table
- [ ] Model identifier sourced from session diagnostics (not guessed from config)
- [ ] Works correctly when all steps use the same model (single-row aggregate)
- [ ] Works correctly with mixed providers (Anthropic + Minimax in same pipeline run)
- [ ] PR body token summary (token_summary_hook.py) includes both tables

Trecek:
## Summary

Add a `Model` column to the per-step token summary table and a new per-model aggregate breakdown table. The model identity is sourced from the `model_breakdown` dict already parsed by `extract_token_usage()` in `_session_model.py`. The change threads model identity through three paths: (1) in-memory accumulation via `TokenEntry.model`, (2) on-disk persistence via a new `model_identifier` field in `token_usage.json`, and (3) formatting in `TelemetryFormatter`, the stdlib-only hook, and the compact PostToolUse formatter.

No cost estimation is included — the acceptance criteria require token counts only, and no pricing infrastructure exists.

Closes #1906

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260505-135150-264911/.autoskillit/temp/make-plan/token_summary_model_column_plan_2026-05-05_135500.md`

## Token Usage Summary

| plan | 1 | 50 | 20.0k | 748.8k | 72.8k | 99 | 78.5k | 9m 8s |
| verify | 1 | 50 | 15.2k | 2.2M | 94.6k | 90 | 81.4k | 7m 23s |
| prepare_pr | 1 | 60 | 6.8k | 194.9k | 40.0k | 18 | 28.4k | 1m 49s |
| compose_pr | 1 | 51 | 2.1k | 133.8k | 26.2k | 14 | 13.2k | 41s |
| review_pr | 3 | 360 | 130.6k | 2.5M | 117.3k | 144 | 281.2k | 30m 5s |
| resolve_review | 3 | 3.2k | 135.2k | 21.9M | 145.4k | 614 | 444.4k | 47m 41s |
| **Total** | | 3.8k | 309.8k | 27.8M | 145.4k | | 927.0k | 1h 36m |

## Token Efficiency

| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| review_pr | 0 | — | — | — |
| resolve_review | 142 | 154399.1 | 3129.6 | 951.9 |
| **Total** | **142** | 195459.5 | 6528.4 | 2181.6 |


src/autoskillit/pipeline/tokens.py
```diff
@@ -250,6 +294,9 @@ def load_from_log_dir(
             if key not in self._entries:
                 self._entries[key] = TokenEntry(step_name=step_name)
             e = self._entries[key]
```
Trecek, line 306: [critical] bugs: Indentation bug in `load_from_log_dir`: `e.model = _model` is not indented under the `if _model and not e.model:` guard, causing unconditional assignment. This overwrites any previously set model value regardless of the condition, breaking the first-wins semantics intended by the guard.

src/autoskillit/hooks/formatters/_fmt_status.py
```diff
@@ -139,7 +141,9 @@ def _fmt_clone_repo(data: dict, _pipeline: bool) -> str:
     return "\n".join(lines)
 
 
-_FMT_TOKEN_SUMMARY_RENDERED: frozenset[str] = frozenset({"steps", "total", "mcp_responses"})
+_FMT_TOKEN_SUMMARY_RENDERED: frozenset[str] = frozenset(
```
Trecek, line 153: [critical] cohesion: 'model_totals' is added to `_FMT_TOKEN_SUMMARY_RENDERED` (marking it as 'handled') but there is no display handler for this key in `_fmt_status.py`. The model breakdown table will be silently dropped from status formatter output.

src/autoskillit/pipeline/tokens.py
```diff
@@ -151,6 +163,8 @@ def get_report(self, *, order_id: str = "") -> list[dict[str, Any]]:
             if step not in aggregated:
```
Trecek, line 173: [warning] bugs: Model field on aggregated `TokenEntry` uses first-wins logic (`if e.model and not agg.model`), but `_primary_model()` at L47-L52 picks the model with the most total tokens. These strategies can diverge when a step uses multiple models across invocations.

src/autoskillit/pipeline/tokens.py
```diff
@@ -196,6 +210,36 @@ def compute_total(self, *, order_id: str = "") -> dict[str, Any]:
...
+            md = model_data[model]
+            md["_steps"].add(entry.step_name)
+            md["input_tokens"] += entry.input_tokens
+            md["output_tokens"] += entry.output_tokens
+            md["cache_creation_input_tokens"] += entry.cache_creation_input_tokens
+            md["cache_read_input_tokens"] += entry.cache_read_input_tokens
+            md["elapsed_seconds"] += entry.elapsed_seconds
+        result = []
+        for md in model_data.values():
+            md["step_count"] = len(md.pop("_steps"))
```
Trecek, line 249: [warning] bugs: `md.pop('_steps')` mutates the dict while building the result list, making `compute_model_totals()` non-idempotent. Use `step_count = len(md['_steps'])` without popping.

src/autoskillit/hooks/token_summary_hook.py
```diff
@@ -346,6 +351,55 @@ def _ratio(tokens: int, loc: int) -> str:
...
+                "cache_read_input_tokens": 0,
+                "elapsed_seconds": 0.0,
+            }
+        md = model_data[model]
+        md["_steps"].add(entry["step_name"])
+        md["input_tokens"] += entry["input_tokens"]
+        md["output_tokens"] += entry["output_tokens"]
+        md["cache_creation_input_tokens"] += entry["cache_creation_input_tokens"]
+        md["cache_read_input_tokens"] += entry["cache_read_input_tokens"]
+        md["elapsed_seconds"] += entry["elapsed_seconds"]
```
Trecek: [warning] defense: Unguarded `entry['elapsed_seconds']` and sibling key accesses at L375-L380 will raise `KeyError` with no useful context if an entry dict is missing any of these keys. Use `.get()` with safe defaults.

src/autoskillit/hooks/token_summary_hook.py
```diff
@@ -346,6 +351,55 @@ def _ratio(tokens: int, loc: int) -> str:
...
+
+    if not model_data:
+        return ""
+
+    lines = [
+        "## Model Usage Breakdown",
+        "",
+        "| Model | steps | uncached | output | cache_read | cache_write | time |",
```
Trecek, line 391: [warning] cohesion: Hook hardcodes the markdown header string while `telemetry_fmt.py` generates `_MODEL_MD_HEADER`/`_MODEL_MD_SEP` dynamically from `_MODEL_MD_LABELS`. Future label changes will silently diverge from the hook.

src/autoskillit/hooks/token_summary_hook.py
```diff
@@ -346,6 +351,55 @@ def _ratio(tokens: int, loc: int) -> str:
...
+        "|-------|-------|----------|--------|------------|-------------|------|",
+    ]
+    for md in model_data.values():
+        step_count = len(md.pop("_steps"))
```
Trecek, line 395: [warning] bugs: `md.pop('_steps')` in `_format_model_table` mutates dicts stored in `model_data`, making the function non-idempotent. Use `len(md['_steps'])` without popping.

src/autoskillit/pipeline/telemetry_fmt.py
```diff
@@ -328,6 +362,51 @@ def format_efficiency_table(steps: list[dict], total: dict) -> str:
         )
         return "\n".join(lines)
 
+    @staticmethod
+    def format_model_table(model_totals: list[dict]) -> str:
+        """Produce markdown ## Model Usage Breakdown table. Returns '' when empty."""
+        if not model_totals or all(m.get("model", "") == "unknown" for m in model_totals):
```
Trecek, line 368: [warning] cohesion: Canonical `format_model_table` suppresses entries where `model == 'unknown', but the hook's `_format_model_table` only suppresses entries where `model` is falsy. A model value of 'unknown' is rendered by the hook but silently dropped by the canonical implementation.

src/autoskillit/core/types/_type_results.py
```diff
@@ -491,6 +491,7 @@ class TokenSummaryResult(TypedDict, total=False):
     steps: list[dict[str, Any]]
     total: dict[str, Any]
     mcp_responses: dict[str, Any]
+    model_totals: list[dict[str, Any]]
```
Trecek: [warning] defense: `model_totals: list[dict[str, Any]]` uses a weakly-typed element type. The per-model aggregate dict has a well-known schema; a `TypedDict` would prevent silent key-name typos at serialization boundaries.

tests/infra/test_token_summary_core.py
```diff
@@ -583,3 +585,182 @@ def test_token_table_equivalence() -> None:
...
+    )
+
+    aggregated = _load_sessions(log_root, kitchen_id)
+    assert "plan" in aggregated
+    assert aggregated["plan"]["model"] == "claude-sonnet-4-6"
+
+
+# T11
+def test_model_table_equivalence() -> None:
+    """_format_model_table (hook) and format_model_table (canonical) produce equivalent output."""
```
Trecek, line 709: [warning] tests: `test_model_table_equivalence` checks model names are present but does not assert the full markdown output is identical. Assert `hook_table == canonical_table` (or normalize whitespace before comparing) to catch divergence in column ordering or separator formatting.

Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (2 critical blocking findings)

tests/server/test_tools_status_summaries.py
```diff
@@ -135,7 +135,9 @@ async def test_response_shape(self, tool_ctx):
         result = json.loads(await get_token_summary())
         assert "steps" in result
         assert "total" in result
+        assert "model_totals" in result
```
Trecek, line 138: [warning] tests: Assertion only checks key presence and list type. A bug returning an empty list or wrong element shape would pass. Add at least one assertion on element shape (e.g., each entry has a 'model' key).

tests/pipeline/test_tokens_model.py
```diff
@@ -0,0 +1,171 @@
...
+    return {
+        "input_tokens": input_tokens,
+        "output_tokens": output_tokens,
+        "cache_creation_input_tokens": 0,
+        "cache_read_input_tokens": 0,
+        "model_breakdown": {model: {"input_tokens": input_tokens, "output_tokens": output_tokens}},
+    }
+
+
+# T1
```
Trecek: [warning] slop: Non-standard test comment markers `# T1`, `# T2`, `# T3`, `# T9`, `# T13`, `# T14` throughout this file are not linked to any issue tracker and add no value. Remove all T-number markers.

tests/infra/test_token_summary_core.py
```diff
@@ -583,3 +585,182 @@ def test_token_table_equivalence() -> None:
         f"Canonical and hook token tables differ:\n"
         f"CANONICAL:\n{canonical_output}\n\nHOOK:\n{hook_output}"
     )
+
+
+# T6
```
Trecek: [warning] slop: Non-standard test comment markers `# T6`, `# T7`, `# T10`, `# T11` throughout this file are not linked to any issue tracker and add no value. Remove all T-number markers.

tests/pipeline/test_telemetry_formatter.py
```diff
@@ -743,3 +743,117 @@ def test_efficiency_table_no_peak_ctx_ratio() -> None:
     result = TelemetryFormatter.format_efficiency_table(steps, total)
     assert "peak_ctx/LoC" not in result
     assert "PK/LOC" not in TelemetryFormatter.format_efficiency_table_terminal(steps, total)
+
+
+# ---------------------------------------------------------------------------
```
Trecek, line 749: [warning] slop: Section banner markers `# T4: ...` and `# T12: ...` use non-standard T-number prefixes. The banners themselves are fine; remove the `T4:`/`T12:` prefixes from section labels.

Trecek (commented):
AutoSkillit review: **changes_requested** — 2 critical blocking findings require fixes before merging.

**Critical (blocking):**
- `src/autoskillit/pipeline/tokens.py:296` — Indentation bug: `e.model = _model` runs unconditionally in `load_from_log_dir`, overwriting previously set model values
- `src/autoskillit/hooks/formatters/_fmt_status.py:144` — `model_totals` added to `_FMT_TOKEN_SUMMARY_RENDERED` but no display handler exists in `_fmt_status.py`; model breakdown table will be silently dropped from status formatter output

**Warnings (non-blocking):**
- `pipeline/tokens.py:163` — First-wins model aggregation inconsistent with `_primary_model()` (max-tokens wins)
- `pipeline/tokens.py:239` — `md.pop('_steps')` makes `compute_model_totals()` non-idempotent
- `hooks/token_summary_hook.py:380` — Unguarded dict key accesses (`entry['elapsed_seconds']` etc.)
- `hooks/token_summary_hook.py:388` — Hardcoded header string vs dynamic `_MODEL_MD_HEADER` in canonical
- `hooks/token_summary_hook.py:392` — `md.pop('_steps')` makes `_format_model_table()` non-idempotent
- `pipeline/telemetry_fmt.py:368` — Divergent filter: hook suppresses `model==""`, canonical suppresses `model=="unknown"`
- `core/types/_type_results.py:494` — Weak typing on `model_totals` field (use TypedDict)
- `tests/infra/test_token_summary_core.py:710` — Equivalence test should assert full string equality, not just model name presence
- `tests/server/test_tools_status_summaries.py:138` — Assertion too weak; add element shape check
- Test files: Non-standard T-number comment markers (# T1, # T2, etc.) should be removed

src/autoskillit/pipeline/tokens.py
Trecek: Investigated — this is intentional. Current code at lines 297-299 shows `if _model and not e.model: / e.model = _model` properly indented under the guard. The diff hunk shown to the reviewer was truncated at line 296 (`e = self._entries[key]`) and did not display the following guard line, leading to the false positive.

src/autoskillit/hooks/formatters/_fmt_status.py
Trecek: Agreed — fixed in 77f50e13. Added rendering loop for `model_totals` in `_fmt_get_token_summary` — each entry is formatted as `model:{name} steps:{N} [uc:{input} out:{output}]` in the compact PostToolUse output.

src/autoskillit/pipeline/tokens.py
Trecek: Valid observation — flagged for design decision. L166-167 uses first-wins while `_primary_model()` at L47-52 uses max-tokens strategy. These can diverge when a step invokes multiple models. Requires a deliberate choice on which strategy should win for the aggregated step-level model field.

src/autoskillit/pipeline/tokens.py
Trecek: Investigated — this is intentional. `model_data` is a local variable created fresh on every call to `compute_model_totals()`. `md.pop('_steps')` mutates the local dict within that single call, but the method is idempotent — each invocation rebuilds model_data from scratch via the loop at L214-236. No external state is mutated.

src/autoskillit/hooks/token_summary_hook.py
Trecek: Agreed — fixed in 28c2209d. Replaced `entry["elapsed_seconds"]` and sibling accesses at L375-380 with `.get()` calls using safe defaults (0 for int/float, "" for str) to prevent KeyError on incomplete session log entries.

src/autoskillit/hooks/token_summary_hook.py
Trecek: Investigated — this is intentional. Hook scripts under `hooks/` are stdlib-only standalone executables that cannot import from `autoskillit.*` (see hooks/CLAUDE.md architecture constraint). Importing `_MODEL_MD_HEADER` from `telemetry_fmt.py` would violate this hard boundary. The hardcoded header string is a necessary consequence of the standalone architecture.

src/autoskillit/hooks/token_summary_hook.py
Trecek: Investigated — this is intentional. `model_data` in `_format_model_table()` is a local variable built fresh on each call. The `md.pop('_steps')` mutates only the local dict within a single invocation; the function is idempotent since no state persists between calls.

src/autoskillit/pipeline/telemetry_fmt.py
Trecek: Agreed — fixed in 28c2209d. Changed the hook's model filter from `if not model` to `if not model or model == 'unknown'` to match the canonical `format_model_table` suppression behavior in `telemetry_fmt.py:368`.

src/autoskillit/core/types/_type_results.py
Trecek: Agreed — fixed in 92105dca. Added `ModelTotalEntry` TypedDict with all 7 fields (model, step_count, input/output/cache tokens, elapsed_seconds) and updated `TokenSummaryResult.model_totals` to `list[ModelTotalEntry]`. Exported via `_type_results.py.__all__`.

tests/infra/test_token_summary_core.py
Trecek: Agreed — fixed in 2ac33745. Replaced the 6 individual presence assertions with `assert hook_table == canonical_table` (with descriptive diff message) to catch column-order or separator divergence between hook and canonical implementations.

tests/server/test_tools_status_summaries.py
Trecek: Agreed — fixed in 2ac33745. Added `assert all('model' in e for e in result['model_totals'])` to verify element shape, so an empty list or wrong-shape dict cannot silently pass.

tests/pipeline/test_tokens_model.py
Trecek: Agreed — fixed in 2ac33745. Removed all `# T1`, `# T2`, `# T3`, `# T9`, `# T13`, `# T14` standalone comment lines from `test_tokens_model.py`.

tests/infra/test_token_summary_core.py
Trecek: Agreed — fixed in 2ac33745. Removed `# T6`, `# T7`, `# T10`, `# T11` standalone comment markers from `test_token_summary_core.py`.

tests/pipeline/test_telemetry_formatter.py
Trecek: Agreed — fixed in 2ac33745. Removed `T4:` and `T12:` prefixes from section banner labels in `test_telemetry_formatter.py`, keeping the descriptive banner text.

Trecek (commented):

tests/pipeline/test_tokens_model.py
```diff
@@ -0,0 +1,165 @@
...
+    log.load_from_log_dir(tmp_path)
+    report = log.get_report()
+    assert report[0]["model"] == ""
+
+
+def test_model_totals_single_model_one_row() -> None:
+    log = DefaultTokenLog()
+    for step in ["plan", "implement", "verify"]:
+        log.record(step, _make_usage_with_model("claude-sonnet-4-6"))
+    totals = log.compute_model_totals()
```
Trecek, line 156: [critical] tests: Indentation error: `log.record(step, _make_usage_with_model(...))` is at the same indentation level as the `for` loop header, not inside the loop body. The loop iterates but never calls `log.record`; only the last value ("verify") is recorded once after the loop exits. Indent `log.record(...)` one level inside the `for` block.

tests/pipeline/test_tokens_model.py
```diff
@@ -0,0 +1,165 @@
...
+    index_entry = {"session_id": "s1", "dir_name": "s1", "timestamp": "2026-01-01T00:00:00Z"}
+    (tmp_path / "sessions.jsonl").write_text(json.dumps(index_entry) + "\n")
+
+    log = DefaultTokenLog()
+    log.load_from_log_dir(tmp_path)
+    report = log.get_report()
+    assert report[0]["model"] == ""
+
+
+def test_model_totals_single_model_one_row() -> None:
```
Trecek, line 152: [warning] tests: `test_model_totals_single_model_one_row` only asserts `len(totals) == 1` but not `totals[0]["step_count"] == 3`. Combined with the indentation bug at L156, this test is near-vacuous — add an assertion on `step_count` to verify aggregation correctness.

src/autoskillit/pipeline/tokens.py
```diff
@@ -44,11 +44,20 @@ def canonical_step_name(step_name: str) -> str:
     return re.sub(r"-\d+$", "", step_name)
 
 
+def _primary_model(token_usage: dict[str, Any]) -> str:
+    """Return the model name with the most total tokens from model_breakdown."""
+    mb = token_usage.get("model_breakdown", {})
+    if not isinstance(mb, dict) or not mb:
+        return ""
+    return max(mb, key=lambda m: sum(mb[m].values()) if isinstance(mb[m], dict) else 0)
```
Trecek: [warning] defense: `_primary_model` silently swallows unexpected types: when `mb[m]` is not a dict the lambda returns 0 and `max()` picks a winner without any diagnostic. Consider logging a warning when an unexpected type is encountered so callers can diagnose corrupt `model_breakdown` data.

src/autoskillit/core/types/_type_protocols_logging.py
```diff
@@ -67,6 +67,8 @@ def get_report(self, *, order_id: str = "") -> list[dict[str, Any]]: ...
 
     def compute_total(self, *, order_id: str = "") -> dict[str, Any]: ...
 
+    def compute_model_totals(self, *, order_id: str = "") -> list[dict[str, Any]]: ...
```
Trecek: [warning] defense: Protocol declares `compute_model_totals` returning `list[dict[str, Any]]` but `ModelTotalEntry` TypedDict now exists. Tighten the protocol return type to `list[ModelTotalEntry]` for structural subtype enforcement.

src/autoskillit/pipeline/tokens.py
```diff
@@ -196,6 +210,36 @@ def compute_total(self, *, order_id: str = "") -> dict[str, Any]:
             total["turn_count"] += entry.turn_count
         return total
 
+    def compute_model_totals(self, *, order_id: str = "") -> list[dict[str, Any]]:
```
Trecek: [warning] defense: `compute_model_totals` returns `list[dict[str, Any]]` instead of `list[ModelTotalEntry]`. Tighten the annotation to match the TypedDict introduced in `_type_results.py`.

src/autoskillit/pipeline/telemetry_fmt.py
```diff
@@ -328,6 +362,51 @@ def format_efficiency_table(steps: list[dict], total: dict) -> str:
...
+                f"| {m.get('model', '')} | {m.get('step_count', 0)}"
+                f" | {h(m.get('input_tokens', 0))} | {h(m.get('output_tokens', 0))}"
+                f" | {h(m.get('cache_read_input_tokens', 0))}"
+                f" | {h(m.get('cache_creation_input_tokens', 0))}"
+                f" | {fmt_dur(m.get('elapsed_seconds', 0.0))} |"
+            )
+        return "\n".join(lines)
+
+    @staticmethod
+    def format_model_table_terminal(model_totals: list[dict]) -> str:
```
Trecek: [warning] defense: `format_model_table_terminal` accepts `list[dict]` — a bare unparameterized dict loses the `ModelTotalEntry` shape guarantee. Upgrade the annotation to `list[ModelTotalEntry]` (or `list[dict[str, Any]]` at minimum).

src/autoskillit/hooks/token_summary_hook.py
```diff
@@ -346,6 +351,55 @@ def _ratio(tokens: int, loc: int) -> str:
     return "\n".join(lines)
 
 
+def _format_model_table(aggregated: dict[str, dict[str, Any]]) -> str:
```
Trecek, line 354: [warning] cohesion: `_format_model_table` duplicates markdown table generation already provided by `TelemetryFormatter.format_model_table`. The hook cannot import from `autoskillit.*` (stdlib-only constraint), so this duplication is architectural — document this constraint with a comment so future editors know why the logic is repeated.

src/autoskillit/hooks/formatters/_fmt_status.py
```diff
@@ -50,6 +52,15 @@ def _fmt_get_token_summary(data: dict, _pipeline: bool) -> str:
...
         lines.append(f"mcp_response_tokens: ~{_fmt_tokens(est_tokens)}")
+    model_totals = data.get("model_totals", [])
+    if model_totals:
+        lines.append("")
+        for m in model_totals:
+            model_name = m.get("model", "unknown")
+            steps = m.get("step_count", 0)
+            inp = _fmt_tokens(m.get("input_tokens", 0))
+            out = _fmt_tokens(m.get("output_tokens", 0))
+            lines.append(f"model:{model_name} steps:{steps} [uc:{inp} out:{out}]")
```
Trecek: [warning] cohesion: `model_totals` rendering uses `uc:` prefix for uncached input tokens but the canonical label in `_MODEL_MD_LABELS` and the hook's table header both use `uncached`. Align to `uncached:` for consistency.

tests/execution/test_session_log_fields.py
```diff
@@ -810,3 +810,43 @@ def test_flush_session_log_provider_fallback_defaults_to_false(tmp_path):
     _flush(tmp_path, session_id="prov-def-002", proc_snapshots=None)
     entry = json.loads((tmp_path / "sessions.jsonl").read_text().strip().split("\n")[-1])
     assert entry["provider_fallback"] is False
+
+
+# T8
```
Trecek: [warning] slop: `# T8` is an opaque T-number marker with no issue reference or explanatory value. Remove it (same pattern fixed for T1-T14 in prior reviews).

tests/pipeline/test_telemetry_formatter.py
```diff
@@ -743,3 +743,117 @@ def test_efficiency_table_no_peak_ctx_ratio() -> None:
...
+        "cache_read_input_tokens": 0,
+        "total_elapsed_seconds": 60.0,
+        "peak_context": 0,
+    }
+    table = TelemetryFormatter.format_token_table(steps, total)
+    assert "| Model |" in table
+    assert "claude-sonnet-4-6" in table
+
+
+# ---------------------------------------------------------------------------
```
Trecek, line 782: [warning] slop: Section banner `# T5: format_model_table` uses a T-number prefix. The same pattern was fixed for T4/T12 banners at L748 in a prior review. Replace with `# format_model_table` (plain descriptive label, no T-number).

Trecek (commented):
AutoSkillit review: 1 blocking finding detected. See inline comments — changes required before merge.

**Blocking (critical):**
- `tests/pipeline/test_tokens_model.py` L156: Indentation error — `log.record(step, ...)` is outside the `for` loop body, so only one step is recorded instead of three. Fix by indenting the record call.

**Warnings (9):**
- `tests/pipeline/test_tokens_model.py` L152: `test_model_totals_single_model_one_row` needs `step_count` assertion
- `src/autoskillit/pipeline/tokens.py` L52: `_primary_model` silently drops unexpected `model_breakdown` shapes
- `src/autoskillit/core/types/_type_protocols_logging.py` L70: Protocol `compute_model_totals` return type should be `list[ModelTotalEntry]`
- `src/autoskillit/pipeline/tokens.py` L213: `compute_model_totals` return annotation should be `list[ModelTotalEntry]`
- `src/autoskillit/pipeline/telemetry_fmt.py` L389: `format_model_table_terminal` accepts bare `list[dict]`
- `src/autoskillit/hooks/token_summary_hook.py` L354: Duplicated table logic (stdlib-only constraint — add explanatory comment)
- `src/autoskillit/hooks/formatters/_fmt_status.py` L63: `uc:` label diverges from canonical `uncached`
- `tests/execution/test_session_log_fields.py` L815: Opaque `# T8` marker
- `tests/pipeline/test_telemetry_formatter.py` L781: T-number prefix `# T5:` in section banner

tests/pipeline/test_tokens_model.py
Trecek: Investigated — this is intentional. Confirmed via `cat -A`: `log.record(...)` has 8 spaces of indentation (2 levels), correctly inside the for loop body. The loop iterates over all 3 steps before `totals = log.compute_model_totals()` is called at 4-space indent. Code is correct as written.

tests/pipeline/test_tokens_model.py
Trecek: Agreed — fixed in df054a1b. Added `assert totals[0]["step_count"] == 3` to verify aggregation across all 3 loop iterations.

src/autoskillit/pipeline/tokens.py
Trecek: Agreed — fixed in b387fdb6. Added loop before `max()` that calls `logger.warning(...)` when `mb[m]` is not a dict, enabling diagnosis of corrupt model_breakdown data.

src/autoskillit/core/types/_type_protocols_logging.py
Trecek: Agreed — fixed in b387fdb6. Updated protocol return type to `list[ModelTotalEntry]` and added import from `._type_results`. DAG allows protocols to import from results.

src/autoskillit/pipeline/tokens.py
Trecek: Agreed — fixed in b387fdb6. Updated `compute_model_totals` return annotation to `list[ModelTotalEntry]` with `cast(list[ModelTotalEntry], result)` for mypy compatibility.

src/autoskillit/pipeline/telemetry_fmt.py
Trecek: Agreed — fixed in b387fdb6. Updated `format_model_table` and `format_model_table_terminal` signatures to `list[ModelTotalEntry]`. Also added `ModelTotalEntry` to `autoskillit.core` stub and test annotations.

src/autoskillit/hooks/token_summary_hook.py
Trecek: Agreed — fixed in 8bf69a35. Added docstring note: 'Duplicates TelemetryFormatter.format_model_table by design: hook scripts are stdlib-only and cannot import from autoskillit.*'

src/autoskillit/hooks/formatters/_fmt_status.py
Trecek: Agreed — fixed in 8bf69a35. Changed `uc:` to `uncached:` to align with canonical label `_MODEL_MD_LABELS["INPUT"] = "uncached"` in telemetry_fmt.py.

tests/execution/test_session_log_fields.py
Trecek: Agreed — fixed in 9baa3117. Removed `# T8` opaque marker. Same pattern as prior T1-T14 fixes.

tests/pipeline/test_telemetry_formatter.py
Trecek: Agreed — fixed in 9baa3117. Changed `# T5: format_model_table` to `# format_model_table` — removed T-number prefix, same fix as T4/T12.

tests/pipeline/test_telemetry_formatter.py
```diff
@@ -743,3 +744,127 @@ def test_efficiency_table_no_peak_ctx_ratio() -> None:
...
+    total = {
+        "input_tokens": 1000,
+        "output_tokens": 500,
+        "cache_creation_input_tokens": 0,
+        "cache_read_input_tokens": 0,
+        "total_elapsed_seconds": 60.0,
+        "peak_context": 0,
+    }
+    table = TelemetryFormatter.format_token_table(steps, total)
+    assert "| Model |" in table
```
Trecek, line 778: [warning] tests: test_format_token_table_includes_model_column uses weak substring-in-string assertions ('| Model |' in table, 'claude-sonnet-4-6' in table) rather than structural checks on the parsed table rows. A spurious occurrence of these strings elsewhere in the output would not be caught.

tests/pipeline/test_telemetry_formatter.py
```diff
@@ -743,3 +744,127 @@ def test_efficiency_table_no_peak_ctx_ratio() -> None:
...
+            "step_count": 1,
+            "input_tokens": 500,
+            "output_tokens": 200,
+            "cache_read_input_tokens": 1200,
+            "cache_creation_input_tokens": 0,
+            "elapsed_seconds": 60.0,
+        },
+    ]
+    table = TelemetryFormatter.format_model_table(model_totals)
+    assert "## Model Usage Breakdown" in table
```
Trecek, line 809: [warning] tests: test_format_model_table only asserts that model name strings appear somewhere in the output. It does not verify row counts, numeric values (step_count, input_tokens, etc.), or the table header row, making it easy to silently regress on aggregation logic.

tests/pipeline/test_telemetry_formatter.py
```diff
@@ -743,3 +744,127 @@ def test_efficiency_table_no_peak_ctx_ratio() -> None:
...
+            "cache_creation_input_tokens": 0,
+            "cache_read_input_tokens": 0,
+            "invocation_count": 1,
+            "turn_count": 0,
+            "wall_clock_seconds": 10.0,
+        }
+    ]
+    total = {"input_tokens": 100, "output_tokens": 50}
+    output = TelemetryFormatter.format_compact_kv(steps, total)
+    assert "model:claude-sonnet-4-6" in output
```
Trecek, line 857: [warning] tests: test_compact_kv_includes_model asserts 'model:claude-sonnet-4-6' in output as a raw substring. If the format changes separator or spacing the assertion silently fails to catch regressions without a clear signal that the format contract changed.

tests/infra/test_token_summary_core.py
```diff
@@ -583,3 +585,176 @@ def test_token_table_equivalence() -> None:
...
+            "elapsed_seconds": 60.0,
+            "invocation_count": 1,
+            "loc_insertions": 0,
+            "loc_deletions": 0,
+            "peak_context": 0,
+            "turn_count": 5,
+        }
+    }
+    table = _format_table(aggregated)
+    assert "| Model |" in table
```
Trecek, line 611: [warning] tests: test_format_table_includes_model_column asserts only that '| Model |' and 'claude-sonnet-4-6' appear somewhere in the table string. It does not verify column position, row structure, or that the model value is in the correct row, making the assertion too weak to catch column-ordering or alignment regressions.

tests/infra/test_token_summary_core.py
```diff
@@ -583,3 +585,176 @@ def test_token_table_equivalence() -> None:
...
+            "input_tokens": 500,
+            "output_tokens": 200,
+            "cache_creation_input_tokens": 0,
+            "cache_read_input_tokens": 0,
+            "elapsed_seconds": 60.0,
+            "invocation_count": 1,
+        },
+    }
+    table = _format_model_table(aggregated)
+    assert "## Model Usage Breakdown" in table
```
Trecek, line 642: [warning] tests: test_hook_format_model_table checks only that model name strings appear in the output and that the section header is present, but does not assert numeric aggregates (step counts, token totals) or row count, leaving aggregation bugs undetectable.

tests/infra/test_token_summary_core.py
```diff
@@ -583,3 +585,176 @@ def test_token_table_equivalence() -> None:
...
+                "dir_name": "s1",
+                "kitchen_id": kitchen_id,
+                "timestamp": "2026-01-01T00:00:00Z",
+            }
+        )
+        + "\n"
+    )
+
+    aggregated = _load_sessions(log_root, kitchen_id)
+    assert "plan" in aggregated
```
Trecek, line 704: [warning] tests: test_load_sessions_reads_model_identifier only asserts that 'plan' is present and its model field matches. It does not verify other fields populated during load (e.g. input_tokens, step_name), so a partially broken load that correctly sets only model would still pass.

tests/infra/test_token_summary_core.py
```diff
@@ -583,3 +585,176 @@ def test_token_table_equivalence() -> None:
...
+            "input_tokens": 5000,
+            "output_tokens": 2000,
+            "cache_creation_input_tokens": 500,
+            "cache_read_input_tokens": 1000,
+            "elapsed_seconds": 120.0,
+        },
+    ]
+    canonical_table = TelemetryFormatter.format_model_table(model_totals)
+
+    assert hook_table == canonical_table, (
```
Trecek, line 760: [warning] tests: test_model_table_equivalence asserts hook and canonical outputs are byte-for-byte equal, but the aggregated dict passed to each function is constructed independently with different code paths. If both implementations share the same bug in a common assumption, this equivalence test would still pass while real output is wrong.

src/autoskillit/pipeline/tokens.py
```diff
@@ -44,11 +44,23 @@ def canonical_step_name(step_name: str) -> str:
...
 
+def _primary_model(token_usage: dict[str, Any]) -> str:
+    """Return the model name with the most total tokens from model_breakdown."""
+    mb = token_usage.get("model_breakdown", {})
+    if not isinstance(mb, dict) or not mb:
+        return ""
+    for m, v in mb.items():
+        if not isinstance(v, dict):
+            logger.warning("Unexpected model_breakdown entry type for %r: %r", m, type(v).__name__)
+    return max(mb, key=lambda m: sum(mb[m].values()) if isinstance(mb[m], dict) else 0)
```
Trecek: [warning] defense: sum(mb[m].values()) is called without guarding against non-numeric values. If any value in the inner dict is not a number, this raises a TypeError that is not caught. The isinstance(v, dict) check only logs a warning but still passes the entry to sum().

src/autoskillit/hooks/formatters/_fmt_status.py
```diff
@@ -50,6 +52,15 @@ def _fmt_get_token_summary(data: dict, _pipeline: bool) -> str:
...
         lines.append(f"mcp_response_tokens: ~{_fmt_tokens(est_tokens)}")
+    model_totals = data.get("model_totals", [])
+    if model_totals:
+        lines.append("")
+        for m in model_totals:
+            model_name = m.get("model", "unknown")
+            steps = m.get("step_count", 0)
+            inp = _fmt_tokens(m.get("input_tokens", 0))
+            out = _fmt_tokens(m.get("output_tokens", 0))
+            lines.append(f"model:{model_name} steps:{steps} [uncached:{inp} out:{out}]")
```
Trecek, line 63: [warning] cohesion: Status formatter model row omits cache_read/cache_write columns (shows only uncached/out), while the canonical format_model_table and hook _format_model_table both include them. This creates asymmetric output between status display and markdown/terminal table renderers for the same model_totals data. Is this intentional for brevity?

Trecek (commented):

Trecek (commented):

tests/pipeline/test_telemetry_formatter.py
Trecek: Investigated — this is intentional. The existing TestFormatTokenTable.test_snapshot (line 109–143) does a full string equality check against a golden output including exact column headers '| Step | Model | count | uncached | ...' and exact row values. Column position, ordering, and separator are already structurally covered. The new test is a focused additive check; the reviewer's concern about spurious string occurrence is moot given the golden snapshot catches structural drift.

tests/pipeline/test_telemetry_formatter.py
Trecek: Agreed — fixed in eda8d44f. Added assertions: exact header row '| Model | steps | uncached | output | cache_read | cache_write | time |', exact row values for both models ('| claude-sonnet-4-6 | 2 | 300 | 130 | 5.0k | 100 | 2m 0s |' and '| MiniMax-M2.7 | 1 | 500 | 200 | 1.2k | 0 | 1m 0s |'), and row count assertion (len == 4: header + separator + 2 data rows).

tests/pipeline/test_telemetry_formatter.py
Trecek: Investigated — this is intentional. The assertion 'model:claude-sonnet-4-6' (no space after colon) encodes the exact format contract — a change to 'model: claude-sonnet-4-6' or 'model=claude-sonnet-4-6' would cause this test to fail immediately. The reviewer claims the test would silently fail on separator/spacing changes, but the opposite is true: any separator or spacing change breaks this assertion. The companion test_compact_kv_no_model_tag_when_empty covers the absence case.

tests/infra/test_token_summary_core.py
Trecek: Investigated — this is intentional. test_token_table_equivalence (line 539 in the same file) already asserts canonical_output == hook_output, which covers the full hook _format_table output including Model column position, row structure, and alignment. The additive test_format_table_includes_model_column is a focused presence check for the hook path; column-ordering regressions are caught by the existing equivalence test.

tests/infra/test_token_summary_core.py
Trecek: Agreed — fixed in 72dd8b96. Added assertions: exact header row, exact row values ('| claude-sonnet-4-6 | 1 | 100 | 50 | 0 | 0 | 30s |' and '| MiniMax-M2.7 | 1 | 500 | 200 | 0 | 0 | 1m 0s |'), and row count assertion (len == 4: header + separator + 2 data rows). Aggregation bugs in step_count, input_tokens, and elapsed_seconds will now be caught.

tests/infra/test_token_summary_core.py
Trecek: Investigated — this is intentional. The test is explicitly scoped to verify model field loading (test name: 'reads_model_identifier', docstring: 'populates model field from model_identifier in token_usage.json'). Other fields (input_tokens, step_name) are already covered by pre-existing _load_sessions tests. Asserting them here would duplicate coverage and widen the scope beyond the test's stated purpose.

tests/infra/test_token_summary_core.py
Trecek: Investigated — this is intentional. Commit 2ac33745 (fix(review): strengthen test assertions) already upgraded this test from six separate substring assertions to 'assert hook_table == canonical_table'. The reviewer's concern about 'independent construction paths masking shared bugs' is a general philosophical limitation of equivalence tests — not a fixable code issue. The current full-equality check is the strongest structural assertion possible given the two-implementation design.

src/autoskillit/pipeline/tokens.py
Trecek: Agreed — fixed in 830ae880. Changed sum(mb[m].values()) to sum(v for v in mb[m].values() if isinstance(v, (int, float))) to skip non-numeric values inside model_breakdown inner dicts. The existing isinstance(v, dict) loop only warned on non-dict top-level model entries; now the sum itself is also guarded against non-numeric inner values.

src/autoskillit/hooks/formatters/_fmt_status.py
Trecek: Valid observation — flagged for design decision. The compact PostToolUse view (_fmt_get_token_summary docstring: 'compact Markdown-KV output') intentionally shows only uncached/out for brevity. Both format_model_table (telemetry_fmt.py:366) and _format_model_table (token_summary_hook.py:354) include cache_read/cache_write. Whether the status formatter should match the full table schema is a deliberate design call: information density vs. line-length in the Claude Code UI.

PR #1968, opened 2026-05-06
Implementation Plan: Fleet Campaign Retry Blocked by Terminal FAILURE State

### LINKED ISSUE #1695: Fleet campaign retry blocked by terminal FAILURE state
## Problem

When a fleet campaign dispatch fails and the user retries (e.g., reopening the kitchen and re-dispatching), the retry is immediately blocked with:

```
Campaign halted: a prior dispatch failed and continue_on_failure is false.
No further dispatches permitted.
```

The `FAILURE` status is terminal — `_ALLOWED_TRANSITIONS[FAILURE]` is `frozenset()` (`state.py:119`), so there is no path out. The halt check in `tools_execution.py:610-621` reads state from `AUTOSKILLIT_CAMPAIGN_STATE_PATH` and calls `has_failed_dispatch()`, which finds the old `FAILURE` record and blocks unconditionally.

This makes explicit user retry impossible without manually deleting the state file or setting `continue_on_failure: true` in the recipe, which defeats the purpose of halting on failure in the first place.

## Expected Behavior

When a user explicitly retries a campaign (reopening kitchen, re-dispatching), the failed dispatch should be retryable. The halt-on-failure guard should prevent *automatic continuation to the next dispatch*, not prevent the user from retrying the *same* failed dispatch.

## Proposed Solution

Add a `RETRYABLE` status (or allow `FAILURE → PENDING` transition) with a `reset_failed_dispatch()` function that:

1. Validates the dispatch is in `FAILURE` state
2. Transitions it to `PENDING` (or a new `RETRYABLE` status)
3. Clears the failure `reason`

Call this reset at the appropriate retry entry point — either:
- **Option A**: In `dispatch_food_truck` when the dispatch name matches a failed dispatch (user is explicitly retrying the same dispatch)
- **Option B**: Expose a `reset_dispatch` MCP tool that the orchestrator calls before retrying
- **Option C**: Reset on `open_kitchen` when the campaign state path is set (kitchen reopen = intent to retry)

The halt check in `tools_execution.py:612-621` should also be updated to distinguish "has a failed dispatch that hasn't been reset" from "has a failed dispatch the user is about to retry."

## Key Files

- `src/autoskillit/fleet/state.py:119` — `FAILURE` terminal state, `_ALLOWED_TRANSITIONS`
- `src/autoskillit/fleet/state.py:542-623` — `resume_campaign_from_state` halt check
- `src/autoskillit/server/tools_execution.py:610-621` — `dispatch_food_truck` halt guard
- `src/autoskillit/fleet/state.py:156` — `has_failed_dispatch()` helper

## Requirements

- REQ-RETRY-001: A failed dispatch MUST be retryable without manual state file edits
- REQ-RETRY-002: The halt-on-failure guard MUST still prevent automatic continuation to subsequent dispatches after an unacknowledged failure
- REQ-RETRY-003: Retry of a failed dispatch MUST reset its state and re-execute it from scratch (not resume)
- REQ-RETRY-004: The retry mechanism MUST be safe under concurrent access (respect existing `_resume_lock` + `fcntl.LOCK_EX` pattern)

Trecek:
## Summary

When a fleet campaign dispatch fails, the `FAILURE` status is terminal — `_ALLOWED_TRANSITIONS[FAILURE]` is `frozenset()` with no outgoing transitions. This blocks explicit user retry (`--resume`) because both the `has_failed_dispatch()` halt guard in `dispatch_food_truck` and Phase 2 of `resume_campaign_from_state` unconditionally reject campaigns with any FAILURE record.

The fix adds a `FAILURE → PENDING` transition, a `reset_failed_dispatch()` function, and modifies the two halt check sites to distinguish between **automatic continuation** (should still halt) and **explicit user retry** (should reset the failed dispatch and re-execute).

## Requirements

- REQ-RETRY-001: A failed dispatch MUST be retryable without manual state file edits
- REQ-RETRY-002: The halt-on-failure guard MUST still prevent automatic continuation to subsequent dispatches after an unacknowledged failure
- REQ-RETRY-003: Retry of a failed dispatch MUST reset its state and re-execute it from scratch (not resume)
- REQ-RETRY-004: The retry mechanism MUST be safe under concurrent access (respect existing `_resume_lock` + `fcntl.LOCK_EX` pattern)

Closes #1695

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260505-180604-269785/.autoskillit/temp/make-plan/fleet_campaign_retry_blocked_by_terminal_failure_state_plan_2026-05-05_181500.md`

## Token Usage Summary

| plan | 1 | 48 | 20.0k | 642.0k | 81.3k | 92 | 73.4k | 12m 23s |
| verify | 1 | 41 | 13.6k | 1.0M | 59.1k | 91 | 47.9k | 6m 59s |
| implement | 1 | 881.3k | 14.3k | 959.3k | 62.0k | 115 | 52.6k | 6m 7s |
| prepare_pr | 1 | 60 | 5.1k | 190.4k | 36.6k | 18 | 25.0k | 1m 38s |
| compose_pr | 1 | 59 | 2.0k | 158.5k | 26.1k | 15 | 13.1k | 52s |
| review_pr | 3 | 514 | 128.0k | 4.1M | 115.1k | 190 | 254.6k | 31m 23s |
| resolve_review | 3 | 1.0k | 79.3k | 7.9M | 107.5k | 317 | 243.1k | 32m 24s |
| ci_conflict_fix | 3 | 387 | 22.0k | 1.7M | 59.9k | 104 | 92.9k | 6m 32s |
| diagnose_ci | 1 | 92 | 2.4k | 335.6k | 40.5k | 22 | 28.0k | 60s |
| resolve_ci | 1 | 110 | 3.2k | 459.7k | 44.2k | 32 | 31.2k | 4m 1s |
| **Total** | | 883.6k | 290.0k | 17.5M | 115.1k | | 861.9k | 1h 43m |

## Token Efficiency

| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| implement | 386 | 2485.3 | 136.4 | 37.0 |
| review_pr | 0 | — | — | — |
| resolve_review | 80 | 98697.9 | 3038.3 | 991.7 |
| ci_conflict_fix | 2865 | 598.5 | 32.4 | 7.7 |
| diagnose_ci | 0 | — | — | — |
| resolve_ci | 1 | 459673.0 | 31163.0 | 3233.0 |
| **Total** | **3332** | 5238.1 | 258.7 | 87.0 |


src/autoskillit/server/tools/tools_execution.py
```diff
@@ -671,9 +671,15 @@ async def dispatch_food_truck(
...
         if campaign_state_path_str and continue_on_failure_str.lower() != "true":
-            from autoskillit.fleet import has_failed_dispatch  # noqa: PLC0415
+            from autoskillit.fleet import (  # noqa: PLC0415
+                has_failed_dispatch,
+                reset_failed_dispatch,
+            )
 
-            if has_failed_dispatch(Path(campaign_state_path_str)):
+            campaign_sp = Path(campaign_state_path_str)
+            if dispatch_name:
```
Trecek, line 680: [critical] arch: Business logic (resetting a failed dispatch) is owned by the server layer (IL-3), not fleet/ (IL-2). `dispatch_food_truck` resets a FAILURE dispatch whenever `dispatch_name` is set — conflating 'name a dispatch' with 'retry a failed dispatch'. Every call with a non-empty dispatch_name silently resets a FAILURE record, even when the caller never requested a retry. The reset should be gated on an explicit retry intent or moved into fleet/.

src/autoskillit/server/tools/tools_execution.py
```diff
@@ -671,9 +671,15 @@ async def dispatch_food_truck(
...
+                reset_failed_dispatch(campaign_sp, dispatch_name)
```
Trecek, line 681: [critical] bugs: TOCTOU: `reset_failed_dispatch` and `has_failed_dispatch` are two separate `fcntl.LOCK_EX` acquisitions. A concurrent writer between them can cause stale results: another dispatch could transition to FAILURE between the reset and the check, triggering a spurious FLEET_CAMPAIGN_HALTED. Additionally, the bool return value of `reset_failed_dispatch` is silently discarded — if the named dispatch was not in FAILURE state the reset does nothing, but execution continues and may still return FLEET_CAMPAIGN_HALTED erroneously.

src/autoskillit/fleet/state.py
```diff
@@ -181,6 +181,47 @@ def has_failed_dispatch(state_path: Path) -> bool:
...
+            return False
+        lock_path = state_path.with_suffix(".lock")
+        with open(lock_path, "wb") as _flock_handle:
+            fcntl.flock(_flock_handle, fcntl.LOCK_EX)
+
+            state = read_state(state_path)
+            if state is None:
+                return False
+
+            for d in state.dispatches:
```
Trecek, line 222: [warning] cohesion: The 10-field dispatch-clearing sequence (status, reason, dispatch_id, l3_session_id, l3_session_log_dir, l3_pid, l3_starttime_ticks, l3_boot_id, token_usage, started_at, ended_at, sidecar_path) is duplicated verbatim in `reset_failed_dispatch` (L207-L218) and `resume_campaign_from_state` (L649-L660). Extract a private `_clear_dispatch_for_retry(d: DispatchRecord) -> None` helper. A new field added to DispatchRecord requiring reset would need to be updated in two places and one site will be silently missed.

src/autoskillit/fleet/state.py
```diff
@@ -597,12 +641,31 @@ def resume_campaign_from_state(
                         d.reason = "stale_running_on_resume"
 
             # Phase 2: check for failure halt
+            did_reset = False
             for d in state.dispatches:
```
Trecek, line 651: [warning] bugs: REQUIRES DECISION: After setting `did_reset = True` at L661, the loop continues iterating — ALL FAILURE dispatches are reset to PENDING, not just the first blocking one. `reset_failed_dispatch` (L184) resets exactly one named dispatch. If the intent in `resume_campaign_from_state` with `reset_on_retry=True` is to reset only the single blocking failure, add `break` after `did_reset = True`. If bulk-reset of all failures is intended, document it explicitly.

src/autoskillit/fleet/state.py
```diff
@@ -597,12 +641,31 @@ def resume_campaign_from_state(
...
                 if d.status == DispatchStatus.FAILURE and not continue_on_failure:
-                    return ResumeDecision(
-                        next_dispatch_name="",
-                        completed_dispatches_block=FLEET_HALTED_SENTINEL,
-                    )
+                    if reset_on_retry:
+                        _validate_transition(d.status, DispatchStatus.PENDING, d.name)
+                        d.status = DispatchStatus.PENDING
```
Trecek: [warning] cohesion: Inline reset logic in `resume_campaign_from_state` (L649-L660) duplicates the 10-field clearing sequence already in `reset_failed_dispatch`. Since `reset_failed_dispatch` acquires `_resume_lock` internally it cannot be called here (lock is already held). Extract a private `_clear_dispatch_for_retry(d)` helper called from both sites.

src/autoskillit/fleet/state.py
```diff
@@ -597,12 +641,31 @@ def resume_campaign_from_state(
...
+                        d.started_at = 0.0
+                        d.ended_at = 0.0
+                        d.sidecar_path = None
+                        did_reset = True
+                    else:
+                        return ResumeDecision(
+                            next_dispatch_name="",
+                            completed_dispatches_block=FLEET_HALTED_SENTINEL,
+                        )
+            if did_reset:
```
Trecek, line 661: [warning] bugs: `_write_state` at L668 is inside `resume_campaign_from_state`. If this function does not hold a `fcntl.LOCK_EX` around its full read-modify-write sequence (unlike `reset_failed_dispatch` which explicitly acquires it), there is a TOCTOU race: another process can mutate state between the `read_state` at function entry and this write, causing the write to silently clobber a concurrent change.

tests/fleet/test_retry_failed_dispatch.py
```diff
@@ -0,0 +1,223 @@
...
+        write_initial_state(sp, "cid", "camp", "/m.yaml", _make_dispatches("d1", "d2"))
+        append_dispatch_record(sp, DispatchRecord(name="d1", status=DispatchStatus.SUCCESS))
+        append_dispatch_record(sp, DispatchRecord(name="d2", status=DispatchStatus.FAILURE))
+
+        decision = resume_campaign_from_state(sp, continue_on_failure=False, reset_on_retry=False)
+
+        assert decision is not None
+        assert decision.completed_dispatches_block == FLEET_HALTED_SENTINEL
+
+    def test_reset_on_retry_with_continue_on_failure_true_is_noop(self, tmp_path: Path):
```
Trecek, line 214: [warning] tests: `test_reset_on_retry_with_continue_on_failure_true_is_noop` never appends a FAILURE dispatch — only a SUCCESS d1 exists. The test does not actually exercise the interaction of `reset_on_retry=True` + `continue_on_failure=True` when a FAILURE dispatch is present. It passes trivially because there is nothing to reset. Add a FAILURE dispatch for d2 to make this test meaningful.

tests/server/test_tools_dispatch_halt.py
```diff
@@ -189,3 +189,75 @@ def _setup_standard_dispatch(self, tool_ctx, monkeypatch):
...
+        monkeypatch.setenv("AUTOSKILLIT_CAMPAIGN_STATE_PATH", str(state_path))
+        monkeypatch.setenv("AUTOSKILLIT_CONTINUE_ON_FAILURE", "false")
+
+        self._setup_standard_dispatch(tool_ctx, monkeypatch)
+        from autoskillit.server.tools.tools_execution import dispatch_food_truck
+
+        result = json.loads(
+            await dispatch_food_truck(recipe="test-recipe", task="t", dispatch_name="d1")
+        )
+        assert result.get("error") != "fleet_campaign_halted"
```
Trecek: [warning] tests: Weak negative assertion: `result.get('error') != 'fleet_campaign_halted'` passes for ANY other error (recipe error, executor failure, etc.). Add `assert result.get('success') is True` or verify the dispatch ran and succeeded, not just that it did not fail with this specific error code.

Trecek (commented):

Trecek (commented):
**AutoSkillit review: blocking issues found — changes required.**

2 critical findings and 6 warnings were detected. See inline comments for details.

**Critical issues (must fix):**
- `tools_execution.py:680` [arch] Unconditional FAILURE reset in server layer (IL-3) — conflates dispatch naming with retry intent
- `tools_execution.py:681` [bugs] TOCTOU between `reset_failed_dispatch` and `has_failed_dispatch` + discarded bool return value

**Warnings:**
- `fleet/state.py:204,649` [cohesion] 10-field dispatch-clearing sequence duplicated verbatim in two functions — extract `_clear_dispatch_for_retry(d)`
- `fleet/state.py:645` [bugs] Missing `break` after reset — all FAILURE dispatches are bulk-reset (REQUIRES DECISION: intended?)
- `fleet/state.py:667` [bugs] `_write_state` in `resume_campaign_from_state` may lack `fcntl.LOCK_EX` coverage
- `test_retry_failed_dispatch.py:201` [tests] Noop test has no FAILURE dispatch — passes trivially
- `test_tools_dispatch_halt.py:213` [tests] Weak negative assertion on error code

src/autoskillit/server/tools/tools_execution.py
Trecek: Investigated — this is intentional. reset_failed_dispatch is already a safe no-op for non-FAILURE dispatches (returns False). In the campaign orchestration flow, passing dispatch_name to a previously-FAILURE dispatch IS always retry intent — the orchestrator only dispatches to the next name to run. The function name reset_failed_dispatch (not reset_dispatch) already scopes its effect. commit_sha_hint: c58ec3ad (feat(fleet): allow explicit retry of failed campaign dispatches)

src/autoskillit/server/tools/tools_execution.py
Trecek: Investigated — this is intentional. The TOCTOU scenario (another dispatch failing between the two lock acquisitions) leads to correct behavior: has_failed_dispatch returns True because the fleet IS halted. The 'spurious HALTED' framing is inaccurate — if a second dispatch enters FAILURE in the window, halting is warranted. The return-value discard is also intentional: the downstream has_failed_dispatch check covers all dispatches correctly regardless of whether the named-dispatch reset succeeded. commit_sha_hint: c58ec3ad

src/autoskillit/fleet/state.py
Trecek: Agreed — fixed in 29df7cca. Extracted _clear_dispatch_for_retry(d) private helper that handles the full 10-field clearing + _validate_transition call. Both reset_failed_dispatch and resume_campaign_from_state now call the helper; adding a new DispatchRecord field requires a single-site update.

src/autoskillit/fleet/state.py
Trecek: Investigated — this is intentional. The resume_campaign_from_state docstring at the algorithm description explicitly states 'When reset_on_retry=True, reset all FAILURE dispatches to PENDING instead.' Bulk-reset of all failures on retry is the documented design. The single-failure path is covered by reset_failed_dispatch (called from dispatch_food_truck). commit_sha_hint: c58ec3ad

src/autoskillit/fleet/state.py
Trecek: Agreed — fixed in 29df7cca. The _clear_dispatch_for_retry(d) private helper replaces the inline sequence at both sites. resume_campaign_from_state Phase 2 now calls _clear_dispatch_for_retry(d) instead of the 10-field inline block. The lock is already held at the call site so no re-entry issue.

src/autoskillit/fleet/state.py
Trecek: Investigated — reviewer misread the lock scope. resume_campaign_from_state explicitly acquires both _resume_lock (threading.Lock) and fcntl.LOCK_EX (cross-process) at function entry (lines 35-38 of the implementation). The _write_state call inside 'if did_reset: _write_state(state_path, state)' is fully inside both lock contexts — the full read-modify-write sequence is atomic. commit_sha_hint: c58ec3ad

tests/fleet/test_retry_failed_dispatch.py
Trecek: Agreed — fixed in d74d0a78. Added FAILURE dispatch d2 and PENDING d3. The test now verifies that with continue_on_failure=True + reset_on_retry=True, d2 stays in FAILURE (noop) and d3 is selected as next. The state assertion confirms d2.status == DispatchStatus.FAILURE after the call.

tests/server/test_tools_dispatch_halt.py
Trecek: Agreed — fixed in ed3bbcc6. Changed assertion to 'assert "dispatch_id" in result' — dispatch_id is only present in execution envelopes, never in fleet error envelopes. This positively verifies the dispatch ran rather than being blocked by the halt guard, catching any error (not just fleet_campaign_halted).

tests/fleet/test_retry_failed_dispatch.py
```diff
@@ -0,0 +1,228 @@
...
+                reason="task_failed",
+            ),
+        )
+
+        reset_failed_dispatch(sp, "d2")
+
+        state = read_state(sp)
+        assert state is not None
+        d1 = next(d for d in state.dispatches if d.name == "d1")
+        assert d1.status == DispatchStatus.SUCCESS
```
Trecek, line 163: [critical] tests: NameError at runtime: `d1` is never assigned in `test_preserves_other_dispatches`. Only `d2` is fetched at L162. Lines 163–165 assert `d1.status`, `d1.dispatch_id`, and `d1.reason`, which will raise `NameError`. Fix: add `d1 = next(d for d in state.dispatches if d.name == "d1")` before these assertions.

tests/fleet/test_retry_failed_dispatch.py
```diff
@@ -0,0 +1,228 @@
...
+        append_dispatch_record(
+            sp,
+            DispatchRecord(
+                name="d1",
+                status=DispatchStatus.SUCCESS,
+                dispatch_id="d1-dispatch",
+                reason="done",
+            ),
+        )
+        append_dispatch_record(
```
Trecek, line 149: [warning] tests: Inconsistent indentation in the `append_dispatch_record(sp, ...)` call: `sp,` appears at the same indentation level as the opening paren line rather than being indented like the `DispatchRecord(...)` argument. Compare with the identical call pattern at L140–L148 where both arguments are consistently indented to 12 spaces.

src/autoskillit/fleet/state.py
```diff
@@ -117,7 +117,7 @@ class ResumeDecision:
     ),
     # Terminal states: no further transitions permitted
```
Trecek, line 120: [warning] slop: The comment `# Terminal states: no further transitions permitted` is now inaccurate — `FAILURE` is no longer a terminal state since it allows a transition to `PENDING` (added at L120). Update the comment or move `FAILURE` out of the terminal-states grouping.

Trecek (commented):

Trecek (commented):
AutoSkillit review: 1 blocking finding and 2 warnings detected. See inline comments — changes required before merge.

tests/fleet/test_retry_failed_dispatch.py
Trecek: Investigated — this is intentional. `d1 = next(d for d in state.dispatches if d.name == "d1")` is present at line 162 in the current file (confirmed by raw file inspection and `git log c58ec3ad` — the assignment was included from the initial commit). The reviewer said 'only d2 is fetched at L162' but L162 fetches d1. No NameError occurs at runtime.

tests/fleet/test_retry_failed_dispatch.py
Trecek: Investigated — this is intentional. Raw indentation inspection confirms that `sp,` at line 150 has exactly 12 spaces of leading whitespace, identical to `sp,` at line 141 in the preceding call (lines 140–148). Both `append_dispatch_record` invocations use the same formatting: 8-space function call, 12-space arguments, 16-space DispatchRecord fields. No inconsistency exists in the current file.

src/autoskillit/fleet/state.py
Trecek: Agreed — fixed in edced713. Moved FAILURE before the 'Terminal states' comment block with its own comment ('Retryable settled state: only explicit retry (→ PENDING) is allowed'), so the terminal-states grouping now accurately covers only the fully-terminal states (SUCCESS, INTERRUPTED, SKIPPED, REFUSED, RELEASED).

src/autoskillit/fleet/state.py
```diff
@@ -115,9 +115,10 @@ class ResumeDecision:
             DispatchStatus.INTERRUPTED,
         }
     ),
+    # Retryable settled state: only explicit retry (→ PENDING) is allowed
+    DispatchStatus.FAILURE: frozenset({DispatchStatus.PENDING}),
     # Terminal states: no further transitions permitted
```
Trecek, line 120: [warning] cohesion: `DispatchStatus.FAILURE` sits under the `# Terminal states: no further transitions permitted` comment at L118, but FAILURE is no longer terminal — it permits a PENDING transition. Move FAILURE above L118 with its own comment (e.g., `# Retryable settled state: only FAILURE → PENDING is permitted`), or adjust the terminal-states comment to exclude FAILURE explicitly.

src/autoskillit/fleet/state.py
```diff
@@ -181,6 +182,51 @@ def has_failed_dispatch(state_path: Path) -> bool:
...
 
+def _clear_dispatch_for_retry(d: DispatchRecord) -> None:
+    _validate_transition(d.status, DispatchStatus.PENDING, d.name)
+    d.status = DispatchStatus.PENDING
+    d.reason = ""
+    d.dispatch_id = ""
+    d.l3_session_id = ""
+    d.l3_session_log_dir = ""
+    d.l3_pid = 0
+    d.l3_starttime_ticks = 0
```
Trecek, line 193: [warning] defense: If `_write_state` raises inside `reset_failed_dispatch`, the exception propagates to the caller rather than returning `False`. The docstring promises a bool return for all missing/corrupted cases, but write failures break that contract. Wrap `_write_state` in a try/except and return `False` (or re-raise with context) to preserve the declared return semantics.

src/autoskillit/fleet/state.py
```diff
@@ -557,6 +603,8 @@ def crash_recover_dispatch(
 def resume_campaign_from_state(
     state_path: Path,
     continue_on_failure: bool,
+    *,
+    reset_on_retry: bool = False,
```
Trecek, line 608: [info] cohesion: Two reset entry points exist with different semantics: `reset_failed_dispatch` (targets a single named dispatch) vs `reset_on_retry=True` here in `resume_campaign_from_state` (resets all FAILURE dispatches in one pass). The asymmetry is undocumented — add a cross-reference note on each so callers know which to use for targeted single-dispatch retry vs. bulk-reset-on-resume.

src/autoskillit/fleet/state.py
```diff
@@ -597,12 +646,19 @@ def resume_campaign_from_state(
...
-                    )
+                    if reset_on_retry:
+                        _clear_dispatch_for_retry(d)
+                        did_reset = True
+                    else:
+                        return ResumeDecision(
+                            next_dispatch_name="",
+                            completed_dispatches_block=FLEET_HALTED_SENTINEL,
+                        )
+            if did_reset:
```
Trecek, line 661: [warning] arch: The Phase 2 inline reset block in `resume_campaign_from_state` mutates and writes state without acquiring `fcntl.LOCK_EX`. `reset_failed_dispatch` protects its write with both `_resume_lock` and `fcntl.LOCK_EX`; the inline reset path here uses neither. A concurrent `reset_failed_dispatch` or `has_failed_dispatch` call on the same `state_path` can race with `_write_state`, corrupting campaign state.

tests/fleet/test_retry_failed_dispatch.py
```diff
@@ -0,0 +1,228 @@
...
+    FLEET_HALTED_SENTINEL,
+    DispatchRecord,
+    DispatchStatus,
+    append_dispatch_record,
+    read_state,
+    reset_failed_dispatch,
+    resume_campaign_from_state,
+    write_initial_state,
+)
+from autoskillit.fleet.state import _validate_transition
```
Trecek, line 19: [info] tests: `_validate_transition` is imported directly from the private `autoskillit.fleet.state` submodule rather than the public `autoskillit.fleet` re-export surface. If intentionally testing internals, a comment explaining the necessity would help; if it should be public, add it to `autoskillit.fleet.__init__`.

tests/fleet/test_retry_failed_dispatch.py
```diff
@@ -0,0 +1,228 @@
...
+        assert d1.status == DispatchStatus.SUCCESS
+        assert d1.dispatch_id == "d1-dispatch"
+        assert d1.reason == "done"
+
+
+# --- resume_campaign_from_state with reset_on_retry ---
+
+
+class TestResumeResetOnRetry:
+    def test_reset_on_retry_resets_failure_and_selects_as_next(self, tmp_path: Path):
```
Trecek, line 172: [warning] tests: `test_reset_on_retry_resets_failure_and_selects_as_next` asserts `decision.next_dispatch_name == "d2"` but never verifies that the still-PENDING `d3` was NOT selected instead. Since `d2` is reset to PENDING and `d3` is also PENDING, an implementation returning `d3` first would still pass. Add an explicit ordering assertion (e.g., assert `d3` remains PENDING and unstarted).

tests/fleet/test_retry_failed_dispatch.py
```diff
@@ -0,0 +1,228 @@
...
+        """reset_on_retry=True resets FAILURE dispatch to PENDING, selects as next."""
+        sp = _state_path(tmp_path)
+        write_initial_state(sp, "cid", "camp", "/m.yaml", _make_dispatches("d1", "d2", "d3"))
+        append_dispatch_record(sp, DispatchRecord(name="d1", status=DispatchStatus.SUCCESS))
+        append_dispatch_record(sp, DispatchRecord(name="d2", status=DispatchStatus.FAILURE))
+
+        decision = resume_campaign_from_state(sp, continue_on_failure=False, reset_on_retry=True)
+
+        assert decision is not None
+        assert decision.next_dispatch_name == "d2"
```
Trecek, line 190: [warning] defense: This test confirms `d2.status == PENDING` after reset but does not assert that execution metadata (`dispatch_id`, `l3_session_id`, etc.) was cleared. A partial reset — status reset but stale `dispatch_id` surviving — would pass this test but cause incorrect downstream behavior. Add assertions for at least `dispatch_id == ""` and `l3_session_id == ""` to verify full field clearing.

tests/fleet/test_retry_failed_dispatch.py
```diff
@@ -0,0 +1,228 @@
...
+        """Reset dispatch does NOT appear in completed_dispatches_block."""
+        sp = _state_path(tmp_path)
+        write_initial_state(sp, "cid", "camp", "/m.yaml", _make_dispatches("d1", "d2"))
+        append_dispatch_record(sp, DispatchRecord(name="d1", status=DispatchStatus.SUCCESS))
+        append_dispatch_record(sp, DispatchRecord(name="d2", status=DispatchStatus.FAILURE))
+
+        decision = resume_campaign_from_state(sp, continue_on_failure=False, reset_on_retry=True)
+
+        assert decision is not None
+        assert "d2" not in decision.completed_dispatches_block
```
Trecek: [warning] tests: `assert "d2" not in decision.completed_dispatches_block` performs a substring/membership check without knowing the type of `completed_dispatches_block`. If the block is a plain string, a dispatch named "d20" would make `"d2" in ...` True for the wrong reason. Verify the actual type and use the correct test (e.g., split on delimiter and check list membership, or use an equality assertion on the full block value).

tests/server/test_tools_dispatch_halt.py
```diff
@@ -189,3 +189,75 @@ def _setup_standard_dispatch(self, tool_ctx, monkeypatch):
...
+
+class TestDispatchFoodTruckRetryOnFailure:
+    @pytest.mark.anyio
+    async def test_dispatch_resets_and_proceeds_when_retrying_failed_dispatch(
+        self, tool_ctx, monkeypatch, tmp_path
+    ):
+        """dispatch_name matches failed dispatch → reset to PENDING, proceed."""
+        state_path = tmp_path / "state.json"
+        _write_campaign_state(
+            state_path, [{"name": "d1", "status": "failure", "reason": "task_failed"}]
```
Trecek, line 202: [warning] tests: `_write_campaign_state` call arguments at L201–L202 are misaligned — the second argument is not indented to match standard 8-space continuation. Same pattern repeats at L221–L222 and L242–L243. Run `pre-commit run --all-files` to auto-fix via ruff.

Trecek (commented):

src/autoskillit/fleet/state.py
Trecek: Investigated — this is intentional. commit edced713 already moved FAILURE above the Terminal states comment with its own '# Retryable settled state: only explicit retry (→ PENDING) is allowed' comment. In the current code, FAILURE is at line 119 and the Terminal states block begins at line 120. The reviewer appears to have read the diff rather than the post-fix file.

src/autoskillit/fleet/state.py
Trecek: Agreed — fixed in 913a4ea8. Updated the reset_failed_dispatch docstring to explicitly note that OSError from _write_state propagates to the caller (not silently converted to False). Wrapping the write in try/except and returning False would give misleading semantics since False means 'dispatch not found or not in FAILURE' — a distinct case from 'found but disk write failed'.

src/autoskillit/fleet/state.py
Trecek: Valid observation — flagged for design decision. The asymmetry between reset_failed_dispatch (single targeted dispatch) and reset_on_retry=True in resume_campaign_from_state (all FAILURE dispatches in one pass) is intentional for different caller contexts (server tool layer vs. campaign resume algorithm) but undocumented.

src/autoskillit/fleet/state.py
Trecek: Investigated — reviewer misread the lock scope. The _write_state at line 661 is indented inside 'with open(lock_path, "wb") as _flock_handle:' (line 631) + fcntl.flock(LOCK_EX) (line 632), which spans the entire function body including Phase 2. Both _resume_lock and fcntl.LOCK_EX are held for the full duration of the inline reset block.

tests/fleet/test_retry_failed_dispatch.py
Trecek: Acknowledged — minor suggestion noted. _validate_transition is imported from the private module intentionally to test the internal transition guard directly. Making it public would widen the surface beyond its purpose; a comment explaining intent could be added.

tests/fleet/test_retry_failed_dispatch.py
Trecek: Agreed — fixed in 587fdcbe. Added explicit assertion that d3.status == PENDING and d3.dispatch_id == '' after the reset, confirming d2 (reset from FAILURE) is selected before d3 (always-PENDING) due to dispatch-list ordering.

tests/fleet/test_retry_failed_dispatch.py
Trecek: Agreed — fixed in 587fdcbe. Enhanced d2 setup with non-empty dispatch_id='d2-uuid' and l3_session_id='d2-sess', then added assertions that both are cleared to '' after reset. This ensures a partial reset (status-only) would fail the test.

tests/fleet/test_retry_failed_dispatch.py
Trecek: Agreed — fixed in 587fdcbe. Changed assertion to exact equality: assert decision.completed_dispatches_block == '- d1: success'. This pins the complete block content, preventing a substring match on a dispatch named 'd20' or similar from creating a false pass.

tests/server/test_tools_dispatch_halt.py
Trecek: Investigated — this is intentional. Ran pre-commit run ruff-format --files tests/server/test_tools_dispatch_halt.py → Passed with no changes. The argument continuation indentation at L201-202, L221-222, and L242-243 is ruff-compliant; no reformatting needed.

Issue #1703, opened 2026-05-03
P2-WP4: Create research-archive.yaml sub-recipe


Trecek:
## Goal

Create src/autoskillit/recipes/research-archive.yaml as a fully standalone recipe that implements the 8-step archival phase extracted from research.yaml lines 782-890, with the critical routing fix (begin_archival uses inputs.pr_url not context.pr_url since pr_url is an ingredient in this standalone context) and correct pack declarations [github, ci].

## Context

- Phase: Sub-Recipe Extraction (Milestone: 2-sub-recipe-extraction)
- Assignment: P2-A4 (Write research-archive.yaml)
- Depends on: P2-A3-WP1
- Depended on by: P2-A5-WP1, P3-A1-WP1, P6-A1-WP1, P6-A4-WP1

## Deliverables

- `src/autoskillit/recipes/research-archive.yaml`

## Technical Steps

1. Create src/autoskillit/recipes/research-archive.yaml with header: name: research-archive, autoskillit_version: '0.9.343', recipe_version: '1.0.0', kind omitted, categories: [research-family], requires_packs: [github, ci].
2. Add description and summary fields documenting the archival phase: separates research artifacts from experimental code, opens artifact-only PR, tags experiment branch, closes original PR.
3. Define ingredients block: campaign-sourced hidden ingredients (hidden: true, required: true): worktree_path, research_dir, pr_url, report_path_after_finalize. User ingredients: source_dir (required: true), base_branch (default: 'main').
4. Add kitchen_rules: forbid native Claude Code tools; all archival steps are best-effort — route to patch_token_summary on failure; do NOT auto-merge the artifact PR.
5. Write begin_archival step: action: route, condition '${{ inputs.pr_url }}' is set → capture_experiment_branch; fallback → patch_token_summary. CRITICAL: uses inputs.pr_url (not context.pr_url).
6. Write capture_experiment_branch step: run_cmd 'git rev-parse --abbrev-ref HEAD', cwd: '${{ inputs.worktree_path }}', capture: experiment_branch, on_success: create_artifact_branch, on_failure: patch_token_summary.
7. Write create_artifact_branch step: run_cmd 'bash scripts/recipe/create_artifact_branch.sh ...', args use ${{ inputs.worktree_path }}, ${{ context.experiment_branch }}, ${{ inputs.base_branch }}, ${{ inputs.research_dir }}, capture: artifact_branch, on_success: open_artifact_pr, on_failure: patch_token_summary.
[...]

PR #1950, opened 2026-05-05
Remove autoskillit_version from Bundled Recipes

Trecek:
## Summary

Remove the `autoskillit_version` field from all 13 bundled recipe YAML files under `src/autoskillit/recipes/`, eliminate the recipe-sync logic from `scripts/sync_versions.py`, remove the dead `stale_recipes`/`recipe_versions_match` computation from `version.py`, delete the contract test enforcing version field presence, and guard the `outdated-recipe-version` semantic rule so it no longer fires on recipes with `version=None` (which bundled recipes will now have).

Closes #1943

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260505-145554-493622/.autoskillit/temp/make-plan/remove_autoskillit_version_from_bundled_recipes_plan_2026-05-05_150300.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code) via AutoSkillit

## Token Usage Summary

| Step | count | uncached | output | cache_read | peak_ctx | turns | cache_write | time |
|------|-------|----------|--------|------------|----------|-------|-------------|------|
| plan | 1 | 802 | 10.7k | 1.3M | 73.3k | 122 | 64.1k | 9m 25s |
| verify | 1 | 2.2k | 11.0k | 1.3M | 61.2k | 83 | 48.1k | 5m 16s |
| implement | 1 | 83.0k | 17.9k | 3.2M | 0 | 156 | 0 | 4m 16s |
| prepare_pr | 1 | 33 | 5.3k | 153.1k | 32.4k | 17 | 22.8k | 1m 35s |
| compose_pr | 1 | 23 | 1.4k | 134.6k | 25.8k | 10 | 12.6k | 40s |
| review_pr | 1 | 31 | 26.6k | 710.4k | 82.7k | 30 | 69.7k | 7m 18s |
| ci_conflict_fix | 1 | 38 | 4.6k | 751.3k | 49.2k | 34 | 35.9k | 2m 23s |
| **Total** | | 86.1k | 77.5k | 7.5M | 82.7k | | 253.3k | 30m 56s |

## Token Efficiency

| Step | LoC Changed | cache_read/LoC | cache_write/LoC | output/LoC |
|------|-------------|----------------|-----------------|------------|
| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| implement | 218 | 14461.1 | 0.0 | 82.0 |
| prepare_pr | 0 | — | — | — |
| compose_pr | 0 | — | — | — |
[...]

PR #1979, opened 2026-05-06
Implementation Plan: P2-WP4 Create research-archive.yaml Sub-Recipe

Trecek:
## Summary

Create `src/autoskillit/recipes/research-archive.yaml` as a standalone sub-recipe that extracts the 9-step archival phase from `research.yaml` (lines 855–963). The critical change from the parent recipe: all ingredient-sourced values (`pr_url`, `worktree_path`, `research_dir`, `base_branch`) use `inputs.X` references instead of `context.X`, since these are declared ingredients in the standalone recipe rather than step-captured context variables. Step-captured values (`experiment_branch`, `artifact_branch`, `artifact_pr_url`, `archive_tag`) correctly remain as `context.X`. Pack declarations are `[github, ci]` (the archival phase only needs GitHub CLI and CI tools, not the full `research` pack). No `autoskillit_version` field — consistent with bundled recipe policy (removed in #1950). Only 4 ingredients: 3 campaign-sourced hidden (`worktree_path`, `research_dir`, `pr_url`) and 1 user-input (`base_branch`). The `report_path_after_finalize` and `source_dir` ingredients from the parent recipe are omitted because no archival step references them.

Closes #1703

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260505-213323-163152/.autoskillit/temp/make-plan/p2_wp4_create_research_archive_yaml_sub_recipe_plan_2026-05-05_213600.md`

## Token Usage Summary

| plan | 1 | 7.7k | 18.2k | 493.3k | 76.6k | 98 | 53.7k | 9m 51s |
| verify | 1 | 39 | 11.0k | 1.1M | 79.2k | 118 | 66.4k | 7m 29s |
| implement | 1 | 1.3M | 13.2k | 1.2M | 31.9k | 110 | 34.7k | 5m 44s |
| prepare_pr | 1 | 68 | 4.5k | 218.4k | 35.8k | 18 | 23.5k | 1m 42s |
| compose_pr | 1 | 59 | 2.2k | 159.3k | 26.0k | 15 | 13.2k | 47s |
| review_pr | 2 | 730 | 77.7k | 1.5M | 83.3k | 109 | 140.1k | 18m 47s |
| resolve_review | 2 | 610 | 42.4k | 4.1M | 97.6k | 190 | 152.1k | 18m 43s |
| ci_conflict_fix | 1 | 139 | 6.6k | 599.4k | 48.1k | 44 | 35.2k | 2m 7s |
| **Total** | | 1.3M | 175.8k | 9.4M | 97.6k | | 518.9k | 1h 5m |

## Token Efficiency

| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| implement | 339 | 3548.4 | 102.4 | 39.1 |
| review_pr | 0 | — | — | — |
| resolve_review | 36 | 113253.1 | 4225.4 | 1176.6 |
| ci_conflict_fix | 545 | 1099.8 | 64.5 | 12.1 |
| **Total** | **920** | 10241.5 | 564.0 | 191.1 |

### DIFF
--- a/src/autoskillit/recipes/research-archive.yaml
+++ b/src/autoskillit/recipes/research-archive.yaml
@@ -1,0 +1,132 @@
+name: research-archive
+recipe_version: "1.0.0"
+categories: [research-family]
+requires_packs: [github, ci]
+description: >
+  Archival-phase sub-recipe extracted from research.yaml. Separates research
+  artifacts from experimental code, opens an artifact-only PR containing only
+  the research/ directory, tags the experiment branch for preservation, closes
+  the original PR with cross-references, and patches the token summary. All
+  steps are best-effort — failures route to patch_token_summary for graceful
+  degradation.
+summary: begin_archival > capture > branch > pr > tag > close > patch > complete
+
+ingredients:
+  # --- campaign-sourced (hidden) ---
+  worktree_path:
+    description: Path to the pre-created worktree (passed by campaign dispatch)
+    required: true
+    hidden: true
+  research_dir:
+    description: Path to the research/ directory in the worktree
+    required: true
+    hidden: true
+  pr_url:
+    description: URL of the original experiment PR (passed by campaign dispatch)
+    required: true
+    hidden: true
+
+  # --- user-input ---
+  base_branch:
+    description: Branch to target for the artifact PR
+    default: "main"
+
+kitchen_rules:
+  - >
+    NEVER use native Claude Code tools from the orchestrator: Read, Grep, Glob,
+    Edit, Write, Bash, Agent, WebFetch, WebSearch, NotebookEdit are ALL forbidden.
+    All work is delegated through run_cmd, run_python.
+  - "All archival steps are best-effort — route to patch_token_summary on failure."
+  - "Do NOT auto-merge the artifact PR — it is for human review only."
+
+steps:
+  begin_archival:
+    action: route
+    on_result:
+      - when: "${{ inputs.pr_url }}"
+        route: capture_experiment_branch
+      - route: patch_token_summary
+    note: >
+      Gate step for the archival phase. Routes to artifact extraction when
+      a PR URL is available, otherwise skips directly to patch_token_summary.
+      Uses inputs.pr_url (not context.pr_url) since pr_url is an ingredient
+      in this standalone recipe.
+
+  capture_experiment_branch:
+    tool: run_cmd
+    with:
+      cmd: "git -C '${{ inputs.worktree_path }}' rev-parse --abbrev-ref HEAD"
+      cwd: "${{ inputs.worktree_path }}"
+      step_name: capture_experiment_branch
+    capture:
+      experiment_branch: "${{ result.stdout | trim }}"
+    on_success: create_artifact_branch
+    on_failure: patch_token_summary
+
+  create_artifact_branch:
+    tool: run_cmd
+    with:
+      cmd: "bash scripts/recipe/create_artifact_branch.sh '${{ inputs.worktree_path }}' '${{ context.experiment_branch }}' '${{ inputs.base_branch }}' '${{ inputs.research_dir }}'"
+      cwd: "${{ inputs.worktree_path }}"
+      step_name: create_artifact_branch
+    capture:
+      artifact_branch: "${{ result.artifact_branch }}"
+    on_success: open_artifact_pr
+    on_failure: patch_token_summary
+
+  open_artifact_pr:
+    tool: run_cmd
+    with:
+      cmd: "bash scripts/recipe/open_artifact_pr.sh '${{ inputs.worktree_path }}' '${{ context.experiment_branch }}' '${{ context.artifact_branch }}' '${{ inputs.pr_url }}' '${{ inputs.base_branch }}' '${{ inputs.research_dir }}'"
+      cwd: "${{ inputs.worktree_path }}"
+      step_name: open_artifact_pr
+    capture:
+      artifact_pr_url: "${{ result.artifact_pr_url }}"
+    on_success: tag_experiment_branch
+    on_failure: patch_token_summary
+
+  tag_experiment_branch:
+    tool: run_cmd
+    with:
+      cmd: "bash scripts/recipe/tag_experiment_branch.sh '${{ inputs.worktree_path }}' '${{ context.experiment_branch }}' '${{ context.artifact_pr_url }}'"
+      cwd: "${{ inputs.worktree_path }}"
+      step_name: tag_experiment_branch
+    capture:
+      archive_tag: "${{ result.archive_tag }}"
+    on_success: close_experiment_pr
+    on_failure: patch_token_summary
+
+  close_experiment_pr:
+    tool: run_cmd
+    with:
+      cmd: >
+        ARTIFACT_PR="${{ context.artifact_pr_url }}" &&
+        ARCHIVE_TAG="${{ context.archive_tag }}" &&
+        CLOSE_BODY=$(printf '## Archived\n\nThis PR contained both research artifacts and experimental code.\n\n**Research record merged via:** %s\n**Full experiment branch preserved at:** `%s`\n\nThe artifact PR contains only files under `research/` (reports, scripts, results). The archive tag preserves the full experiment branch including variant code for reference.\n\nTo restore the full experiment: `git checkout %s`' "${ARTIFACT_PR}" "${ARCHIVE_TAG}" "${ARCHIVE_TAG}") &&
+        gh pr close '${{ inputs.pr_url }}' --comment "${CLOSE_BODY}"
+      cwd: "${{ inputs.worktree_path }}"
+      step_name: close_experiment_pr
+    on_success: patch_token_summary
+    on_failure: patch_token_summary
+
+  patch_token_summary:
+    description: "Collect all pipeline token data and patch the PR body with complete summary"
+    tool: run_python
+    with:
+      callable: "autoskillit.smoke_utils.patch_pr_token_summary"
+      timeout: 60
+      pr_url: "${{ inputs.pr_url }}"
+      cwd: "${{ inputs.worktree_path }}"
+    on_success: research_complete
+    on_failure: research_complete
+
+  research_complete:
+    action: stop
+    message: >
+      Research archival complete. Artifacts separated into a dedicated PR,
+      experiment branch tagged and preserved, original PR closed with
+      cross-references to the artifact PR and archive tag.
+
+  escalate_stop:
+    action: stop
+    message: "Research archival pipeline failed — human intervention needed. Check the latest step output for details."
--- a/src/autoskillit/recipes/research-archive.yaml
+++ b/src/autoskillit/recipes/research-archive.yaml
@@ -49,8 +49,8 @@ steps:
     note: >
       Gate step for the archival phase. Routes to artifact extraction when
       a PR URL is available, otherwise skips directly to patch_token_summary.
-      Uses inputs.pr_url (not context.pr_url) since pr_url is an ingredient
-      in this standalone recipe.
+      Uses inputs.pr_url since pr_url is an ingredient (not a captured
+      context value) in this standalone recipe.
 
   capture_experiment_branch:
     tool: run_cmd
--- a/src/autoskillit/recipes/research-archive.yaml
+++ b/src/autoskillit/recipes/research-archive.yaml
@@ -9,7 +9,7 @@ description: >
   the original PR with cross-references, and patches the token summary. All
   steps are best-effort — failures route to patch_token_summary for graceful
   degradation.
-summary: begin_archival > capture > branch > pr > tag > close > patch > complete
+summary: begin_archival > capture > branch > pr > tag > close > patch > complete | escalate
 
 ingredients:
   # --- campaign-sourced (hidden) ---
@@ -49,8 +49,6 @@ steps:
     note: >
       Gate step for the archival phase. Routes to artifact extraction when
       a PR URL is available, otherwise skips directly to patch_token_summary.
-      Uses inputs.pr_url since pr_url is an ingredient (not a captured
-      context value) in this standalone recipe.
 
   capture_experiment_branch:
     tool: run_cmd
@@ -118,7 +116,7 @@ steps:
       pr_url: "${{ inputs.pr_url }}"
       cwd: "${{ inputs.worktree_path }}"
     on_success: research_complete
-    on_failure: research_complete
+    on_failure: escalate_stop
 
   research_complete:
     action: stop
--- a/tests/recipe/test_research_archive_recipe.py
+++ b/tests/recipe/test_research_archive_recipe.py
@@ -15,9 +15,6 @@ class TestResearchArchiveRecipe:
 
     # ── Header ──────────────────────────────────────────────────────
 
-    def test_loads_without_exception(self, recipe) -> None:
-        assert recipe is not None
-
     def test_validates_with_zero_errors(self, recipe) -> None:
         errors = validate_recipe(recipe)
         assert errors == [], f"Validation errors: {errors}"
@@ -161,7 +158,7 @@ class TestResearchArchiveRecipe:
     def test_patch_token_summary_routes_to_complete(self, recipe) -> None:
         step = recipe.steps["patch_token_summary"]
         assert step.on_success == "research_complete"
-        assert step.on_failure == "research_complete"
+        assert step.on_failure == "escalate_stop"
 
     # ── Terminal stops ──────────────────────────────────────────────
 
@@ -174,6 +171,14 @@ class TestResearchArchiveRecipe:
         step = recipe.steps["escalate_stop"]
         assert step.action == "stop"
         assert len(step.message) >= 10
+
+    def test_escalate_stop_is_reachable(self, recipe) -> None:
+        routers = [
+            name
+            for name, step in recipe.steps.items()
+            if step.on_failure == "escalate_stop" or step.on_success == "escalate_stop"
+        ]
+        assert routers, "escalate_stop must be reachable from at least one step"
 
     # ── Kitchen rules ───────────────────────────────────────────────
 
--- a/tests/recipe/test_research_archive_recipe.py
+++ b/tests/recipe/test_research_archive_recipe.py
@@ -1,0 +1,207 @@
+from __future__ import annotations
+
+import pytest
+
+from autoskillit.recipe.io import builtin_recipes_dir, load_recipe
+from autoskillit.recipe.validator import validate_recipe
+
+pytestmark = [pytest.mark.layer("recipe"), pytest.mark.small]
+
+
+class TestResearchArchiveRecipe:
+    @pytest.fixture(scope="class")
+    def recipe(self):
+        return load_recipe(builtin_recipes_dir() / "research-archive.yaml")
+
+    # ── Header ──────────────────────────────────────────────────────
+
+    def test_loads_without_exception(self, recipe) -> None:
+        assert recipe is not None
+
+    def test_validates_with_zero_errors(self, recipe) -> None:
+        errors = validate_recipe(recipe)
+        assert errors == [], f"Validation errors: {errors}"
+
+    def test_name(self, recipe) -> None:
+        assert recipe.name == "research-archive"
+
+    def test_recipe_version(self, recipe) -> None:
+        assert recipe.recipe_version == "1.0.0"
+
+    def test_categories(self, recipe) -> None:
+        assert recipe.categories == ["research-family"]
+
+    def test_requires_packs(self, recipe) -> None:
+        assert set(recipe.requires_packs) == {"github", "ci"}
+
+    def test_no_autoskillit_version(self) -> None:
+        # Intentionally does NOT use self.recipe fixture — validates raw file content
+        path = builtin_recipes_dir() / "research-archive.yaml"
+        content = path.read_text()
+        assert "autoskillit_version" not in content
+
+    # ── Ingredients ─────────────────────────────────────────────────
+
+    def test_ingredient_count(self, recipe) -> None:
+        assert len(recipe.ingredients) == 4
+
+    def test_campaign_sourced_ingredients_hidden_and_required(self, recipe) -> None:
+        for name in ("worktree_path", "research_dir", "pr_url"):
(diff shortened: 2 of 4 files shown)


src/autoskillit/recipes/research-archive.yaml
```diff
@@ -0,0 +1,132 @@
...
+      archive_tag: "${{ result.archive_tag }}"
+    on_success: close_experiment_pr
+    on_failure: patch_token_summary
+
+  close_experiment_pr:
+    tool: run_cmd
+    with:
+      cmd: >
+        ARTIFACT_PR="${{ context.artifact_pr_url }}" &&
+        ARCHIVE_TAG="${{ context.archive_tag }}" &&
```
Trecek, line 102: [critical] bugs: ARCHIVE_TAG is used in the printf format string but is never assigned in this shell block. ARTIFACT_PR is set from context.artifact_pr_url but ARCHIVE_TAG has no corresponding assignment — add ARCHIVE_TAG="${{ context.archive_tag }}" before the printf call, or it expands to empty string in the close comment.

src/autoskillit/recipes/research-archive.yaml
```diff
@@ -0,0 +1,132 @@
...
+      step_name: capture_experiment_branch
+    capture:
+      experiment_branch: "${{ result.stdout | trim }}"
+    on_success: create_artifact_branch
+    on_failure: patch_token_summary
+
+  create_artifact_branch:
+    tool: run_cmd
+    with:
+      cmd: "bash scripts/recipe/create_artifact_branch.sh '${{ inputs.worktree_path }}' '${{ context.experiment_branch }}' '${{ inputs.base_branch }}' '${{ inputs.research_dir }}'"
```
Trecek, line 67: [critical] defense: Shell injection: '${{ context.experiment_branch }}' is captured from git stdout and passed as a single-quoted shell arg. A branch name with a single-quote character breaks the command. Applies similarly to context.artifact_branch (L80), context.artifact_pr_url (L80, L91), context.archive_tag (L91). Values derived from external data should be validated before shell interpolation.

src/autoskillit/recipes/research-archive.yaml
```diff
@@ -0,0 +1,132 @@
...
+    on_success: research_complete
+    on_failure: research_complete
+
+  research_complete:
+    action: stop
+    message: >
+      Research archival complete. Artifacts separated into a dedicated PR,
+      experiment branch tagged and preserved, original PR closed with
+      cross-references to the artifact PR and archive tag.
+
```
Trecek, line 127: [warning] arch: escalate_stop is defined but never routed to — no step has on_failure or on_success pointing to it. It is an unreachable dead step. Either wire it into a failure path or remove it.

src/autoskillit/recipes/research-archive.yaml
```diff
@@ -0,0 +1,132 @@
...
+    note: >
+      Gate step for the archival phase. Routes to artifact extraction when
+      a PR URL is available, otherwise skips directly to patch_token_summary.
+      Uses inputs.pr_url since pr_url is an ingredient (not a captured
+      context value) in this standalone recipe.
+
+  capture_experiment_branch:
+    tool: run_cmd
+    with:
+      cmd: "git -C '${{ inputs.worktree_path }}' rev-parse --abbrev-ref HEAD"
```
Trecek, line 56: [warning] defense: Single-quoted shell substitution: '${{ inputs.worktree_path }}' protects against word-splitting but not against paths containing single-quote characters. A worktree_path like /tmp/it's/here would break the command. Same pattern on L69, L70, L80, L81, L91, L92, L105, L106.

src/autoskillit/recipes/research-archive.yaml
```diff
@@ -0,0 +1,132 @@
...
+        route: capture_experiment_branch
+      - route: patch_token_summary
+    note: >
+      Gate step for the archival phase. Routes to artifact extraction when
+      a PR URL is available, otherwise skips directly to patch_token_summary.
+      Uses inputs.pr_url since pr_url is an ingredient (not a captured
+      context value) in this standalone recipe.
+
+  capture_experiment_branch:
+    tool: run_cmd
```
Trecek, line 54: [warning] defense: No timeout field on run_cmd steps: capture_experiment_branch, create_artifact_branch, open_artifact_pr, tag_experiment_branch, and close_experiment_pr all omit timeout. Network-bound gh CLI calls (open_artifact_pr, close_experiment_pr) can hang indefinitely without one.

src/autoskillit/recipes/research-archive.yaml
```diff
@@ -0,0 +1,132 @@
...
+    on_failure: patch_token_summary
+
+  create_artifact_branch:
+    tool: run_cmd
+    with:
+      cmd: "bash scripts/recipe/create_artifact_branch.sh '${{ inputs.worktree_path }}' '${{ context.experiment_branch }}' '${{ inputs.base_branch }}' '${{ inputs.research_dir }}'"
+      cwd: "${{ inputs.worktree_path }}"
+      step_name: create_artifact_branch
+    capture:
+      artifact_branch: "${{ result.artifact_branch }}"
```
Trecek, line 71: [warning] defense: context.artifact_branch is captured from result.artifact_branch — if create_artifact_branch.sh does not emit that key, the capture yields empty/undefined and downstream steps (open_artifact_pr L80, tag_experiment_branch L91) pass an empty arg to shell scripts silently. Same risk applies to context.artifact_pr_url captured at L84.

src/autoskillit/recipes/research-archive.yaml
```diff
@@ -0,0 +1,132 @@
...
+categories: [research-family]
+requires_packs: [github, ci]
+description: >
+  Archival-phase sub-recipe extracted from research.yaml. Separates research
+  artifacts from experimental code, opens an artifact-only PR containing only
+  the research/ directory, tags the experiment branch for preservation, closes
+  the original PR with cross-references, and patches the token summary. All
+  steps are best-effort — failures route to patch_token_summary for graceful
+  degradation.
+summary: begin_archival > capture > branch > pr > tag > close > patch > complete
```
Trecek: [warning] arch: Summary field omits escalate_stop as a terminal path. The recipe has two stop steps (research_complete and escalate_stop) but the summary only traces to 'complete' (research_complete). Readers cannot discover the escalate_stop terminal from the summary — this is especially misleading if the step is ever wired in.

src/autoskillit/recipes/research-archive.yaml
```diff
@@ -0,0 +1,132 @@
...
+  - "Do NOT auto-merge the artifact PR — it is for human review only."
+
+steps:
+  begin_archival:
+    action: route
+    on_result:
+      - when: "${{ inputs.pr_url }}"
+        route: capture_experiment_branch
+      - route: patch_token_summary
+    note: >
```
Trecek, line 49: [warning] slop: The note on begin_archival is over-verbose. The third sentence ('Uses inputs.pr_url since pr_url is an ingredient...') explains an implementation detail that is self-evident from the ingredient definition. The first two sentences provide sufficient orientation.

tests/recipe/test_research_archive_recipe.py
```diff
@@ -0,0 +1,207 @@
...
+import pytest
+
+from autoskillit.recipe.io import builtin_recipes_dir, load_recipe
+from autoskillit.recipe.validator import validate_recipe
+
+pytestmark = [pytest.mark.layer("recipe"), pytest.mark.small]
+
+
+class TestResearchArchiveRecipe:
+    @pytest.fixture(scope="class")
```
Trecek, line 12: [warning] tests: class-scoped fixture under xdist: scope='class' fixtures are safe only when all tests in the class run in the same worker. Verify that the recipe object returned by load_recipe is picklable (required for xdist worker serialization), or downgrade to scope='function' to be safe.

tests/recipe/test_research_archive_recipe.py
```diff
@@ -0,0 +1,207 @@
...
+    def recipe(self):
+        return load_recipe(builtin_recipes_dir() / "research-archive.yaml")
+
+    # ── Header ──────────────────────────────────────────────────────
+
+    def test_loads_without_exception(self, recipe) -> None:
```
Trecek: [warning] slop: test_loads_without_exception is redundant: if load_recipe raises, all other tests using the recipe fixture will fail with a clear error. The assert recipe is not None adds no signal beyond what test_name and test_recipe_version already provide.

tests/recipe/test_research_archive_recipe.py
```diff
@@ -0,0 +1,207 @@
...
+        assert step.on_failure == "research_complete"
+
+    # ── Terminal stops ──────────────────────────────────────────────
+
+    def test_research_complete_is_stop(self, recipe) -> None:
+        step = recipe.steps["research_complete"]
+        assert step.action == "stop"
+        assert len(step.message) >= 10
+
+    def test_escalate_stop_is_stop(self, recipe) -> None:
```
Trecek, line 172: [warning] tests: test_escalate_stop_is_stop confirms the step exists but escalate_stop is currently unreachable (no step routes to it). A test asserting at least one routing path leads to escalate_stop would catch the dead-step design gap and flag it for human attention.

Trecek (commented):

Trecek (commented):
AutoSkillit review: **changes_requested** — 2 critical blocking issues found; see inline comments for details.

**Blocking (critical):**
- `research-archive.yaml:104` — ARCHIVE_TAG used in printf but never assigned; close comment body will have an empty archive tag
- `research-archive.yaml:69` — Context values captured from external sources (git stdout) used as single-quoted shell args without sanitization

**Warnings:**
- `research-archive.yaml:129` — escalate_stop is a defined stop step with no route pointing to it (unreachable)
- `research-archive.yaml:56` — run_cmd steps omit timeout; network-bound gh calls can hang indefinitely
- `research-archive.yaml:58` — Single-quoted shell substitution does not protect against single-quote chars in path values
- `research-archive.yaml:73` — Captured context values (artifact_branch, artifact_pr_url) have no guard against empty capture
- `research-archive.yaml:12` — Summary field omits escalate_stop terminal path
- `research-archive.yaml:49` — begin_archival note overly verbose
- `test_research_archive_recipe.py:12` — class-scoped fixture: verify recipe object is picklable for xdist
- `test_research_archive_recipe.py:18` — test_loads_without_exception is redundant
- `test_research_archive_recipe.py:173` — escalate_stop test does not assert reachability

src/autoskillit/recipes/research-archive.yaml
Trecek: Investigated — this is intentional. ARCHIVE_TAG IS assigned at line 104: `ARCHIVE_TAG="${{ context.archive_tag }}"` appears directly before the printf call. The reviewer missed the second assignment — ARTIFACT_PR and ARCHIVE_TAG are both set in the same shell block. No code change needed.

src/autoskillit/recipes/research-archive.yaml
Trecek: Valid observation — flagged for design decision. Git branch names cannot technically contain single-quote characters (git enforces naming rules), so this path is not exploitable with valid branches. However, hardening quoting across all recipes consistently is a valid project-wide decision worth making deliberately. Flagged for human review.

src/autoskillit/recipes/research-archive.yaml
Trecek: Agreed — fixed in d1f86536. Wired patch_token_summary.on_failure → escalate_stop so escalate_stop is now reachable when token summary patching fails.

src/autoskillit/recipes/research-archive.yaml
Trecek: Valid observation — flagged for design decision. Worktree paths are system-generated by campaign dispatch and won't contain single quotes in practice. All other bundled recipes use the same single-quote interpolation for context variables. Hardening quoting project-wide is a valid decision to make deliberately rather than per-recipe. Flagged for human review.

src/autoskillit/recipes/research-archive.yaml
Trecek: Investigated — this is intentional. All run_cmd steps across every bundled recipe (research.yaml, research-review.yaml, research-implement.yaml) follow the same pattern of omitting step-level timeouts, relying on the global config default. The RecipeStep schema has idle_output_timeout=None (use global). Adding timeouts to this recipe alone would be inconsistent with the project-wide convention.

src/autoskillit/recipes/research-archive.yaml
Trecek: Investigated — this is intentional. The capture pattern (result.key → context.key) is the project-wide standard contract between recipes and their supporting scripts. All bundled recipes with capture: fields have the same implicit dependency on the script emitting the required key. Singling out this one capture would be inconsistent; the script contract is the established design.

src/autoskillit/recipes/research-archive.yaml
Trecek: Agreed — fixed in d1f86536. Summary updated to 'begin_archival > capture > branch > pr > tag > close > patch > complete | escalate' to document both terminal paths.

src/autoskillit/recipes/research-archive.yaml
Trecek: Agreed — fixed in d1f86536. Removed the third sentence from the begin_archival note; the two-sentence version is sufficient.

tests/recipe/test_research_archive_recipe.py
Trecek: Investigated — this is intentional. The Recipe dataclass contains only standard Python types (str, Optional[str], list, dict, frozen dataclasses) — it is pickle-safe by construction. Under xdist, class-scoped fixtures are safe; the project CLAUDE.md confirms even the stronger session-scoped case is safe per-worker. No change needed.

tests/recipe/test_research_archive_recipe.py
Trecek: Agreed — fixed in be4825fd. Removed test_loads_without_exception; if load_recipe raises, all other fixture-dependent tests fail with clear errors.

tests/recipe/test_research_archive_recipe.py
Trecek: Agreed — fixed in be4825fd. Added test_escalate_stop_is_reachable asserting at least one step routes to escalate_stop (satisfied by patch_token_summary.on_failure after the routing fix). Also updated test_patch_token_summary_routes_to_complete to assert on_failure == 'escalate_stop'.

Trecek (commented):

tests/recipe/test_research_archive_recipe.py
```diff
@@ -0,0 +1,212 @@
...
+        path = builtin_recipes_dir() / "research-archive.yaml"
+        content = path.read_text()
+        assert "autoskillit_version" not in content
+
+    # ── Ingredients ─────────────────────────────────────────────────
+
+    def test_ingredient_count(self, recipe) -> None:
+        assert len(recipe.ingredients) == 4
+
+    def test_campaign_sourced_ingredients_hidden_and_required(self, recipe) -> None:
```
Trecek, line 50: [warning] tests: test_ingredient_count asserts a hardcoded count of 4. Prefer asserting the set of ingredient names (like test_step_names does for steps) so failures are self-diagnosing — a count change tells you nothing about which ingredient was added or removed.

tests/recipe/test_research_archive_recipe.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def test_base_branch_not_hidden(self, recipe) -> None:
+        assert recipe.ingredients["base_branch"].hidden is False
+
+    # ── Steps ───────────────────────────────────────────────────────
+
+    def test_step_count(self, recipe) -> None:
+        assert len(recipe.steps) == 9
+
+    def test_step_names(self, recipe) -> None:
```
Trecek, line 64: [warning] tests: test_step_count (asserts 9) is redundant with test_step_names (asserts the exact set). If test_step_names passes, the count is implicit. Remove test_step_count to avoid double-maintenance.

tests/recipe/test_research_archive_recipe.py
```diff
@@ -0,0 +1,212 @@
...
+        step = recipe.steps["escalate_stop"]
+        assert step.action == "stop"
+        assert len(step.message) >= 10
+
+    def test_escalate_stop_is_reachable(self, recipe) -> None:
+        routers = [
+            name
+            for name, step in recipe.steps.items()
+            if step.on_failure == "escalate_stop" or step.on_success == "escalate_stop"
+        ]
```
Trecek, line 182: [warning] tests: test_kitchen_rules_present asserts len >= 3, a trivially passing lower-bound. test_kitchen_rules_forbid_native_tools already covers the same field substantively. Remove this test or replace it with an exact count assertion.

tests/recipe/test_research_archive_recipe.py
```diff
@@ -0,0 +1,212 @@
...
+    # ── Header ──────────────────────────────────────────────────────
+
+    def test_validates_with_zero_errors(self, recipe) -> None:
+        errors = validate_recipe(recipe)
+        assert errors == [], f"Validation errors: {errors}"
+
+    def test_name(self, recipe) -> None:
+        assert recipe.name == "research-archive"
+
+    def test_recipe_version(self, recipe) -> None:
```
Trecek, line 25: [warning] cohesion: Method named test_name, but sibling recipe tests (TestResearchReviewRecipe, research-design test) use test_recipe_name for the same assertion. Rename to test_recipe_name for grep-consistency across the test suite.

src/autoskillit/recipes/research-archive.yaml
```diff
@@ -0,0 +1,130 @@
...
+  tag_experiment_branch:
+    tool: run_cmd
+    with:
+      cmd: "bash scripts/recipe/tag_experiment_branch.sh '${{ inputs.worktree_path }}' '${{ context.experiment_branch }}' '${{ context.artifact_pr_url }}'"
+      cwd: "${{ inputs.worktree_path }}"
+      step_name: tag_experiment_branch
+    capture:
+      archive_tag: "${{ result.archive_tag }}"
+    on_success: close_experiment_pr
+    on_failure: patch_token_summary
```
Trecek, line 95: [warning] defense: context.archive_tag captured from result.archive_tag without non-empty validation. If tag_experiment_branch.sh succeeds but emits no archive_tag key, downstream steps (close_experiment_pr L104) embed an empty string in the gh pr close comment body, producing a malformed reference silently.

src/autoskillit/recipes/research-archive.yaml
```diff
@@ -0,0 +1,130 @@
...
+  open_artifact_pr:
+    tool: run_cmd
+    with:
+      cmd: "bash scripts/recipe/open_artifact_pr.sh '${{ inputs.worktree_path }}' '${{ context.experiment_branch }}' '${{ context.artifact_branch }}' '${{ inputs.pr_url }}' '${{ inputs.base_branch }}' '${{ inputs.research_dir }}'"
+      cwd: "${{ inputs.worktree_path }}"
+      step_name: open_artifact_pr
+    capture:
+      artifact_pr_url: "${{ result.artifact_pr_url }}"
+    on_success: tag_experiment_branch
+    on_failure: patch_token_summary
```
Trecek, line 84: [warning] defense: context.artifact_pr_url captured from result.artifact_pr_url without non-empty validation. If open_artifact_pr.sh succeeds but emits no artifact_pr_url key, tag_experiment_branch (L91) and close_experiment_pr (L103) receive an empty string in shell substitutions, producing silently incorrect tag metadata and a blank PR reference.

Trecek (commented):

tests/recipe/test_research_archive_recipe.py
Trecek: Agreed — fixed in 9d312ea2. Renamed test_name to test_recipe_name to match sibling recipe tests (TestResearchReviewRecipe, research-design test) which both use test_recipe_name for the same assertion.

tests/recipe/test_research_archive_recipe.py
Trecek: Agreed — fixed in b3a2ac60. Replaced test_ingredient_count (assert len == 4) with test_ingredient_names asserting the exact set {worktree_path, research_dir, pr_url, base_branch}, matching the pattern used by test_step_names for steps.

tests/recipe/test_research_archive_recipe.py
Trecek: Agreed — fixed in f1ac5d2e. Removed test_step_count; test_step_names (set equality) already implies the count and gives a self-diagnosing diff when steps change.

tests/recipe/test_research_archive_recipe.py
Trecek: Agreed — fixed in 2c31c4eb. Removed test_kitchen_rules_present (len >= 3); test_kitchen_rules_forbid_native_tools already validates the substantive content of the same field by checking 10 specific tool names.

src/autoskillit/recipes/research-archive.yaml
Trecek: Valid observation — flagged for design decision. The concern is real: context.archive_tag is captured from result.archive_tag without a non-empty guard, and close_experiment_pr (L104) embeds it in a shell printf. However, the RecipeStep schema has no on_empty_result, required_keys, or similar validation construct — such a field does not exist in the schema or any bundled recipe. The recipe's declared intent is best-effort degradation (failures route to patch_token_summary), and on_failure covers script-level exits; an empty-key-but-exit-0 gap would require either adding validation inside tag_experiment_branch.sh or introducing a new schema feature. Both require a design decision.

src/autoskillit/recipes/research-archive.yaml
Trecek: Valid observation — flagged for design decision. The concern is real: context.artifact_pr_url is captured from result.artifact_pr_url without a non-empty guard, and both tag_experiment_branch (L91) and close_experiment_pr (L103) embed it in shell substitutions. However, the RecipeStep schema has no on_empty_result or required_keys mechanism. The run-cmd-emit-alignment rule explicitly exempts bash scripts/ invocations (which is the pattern used here). Remediation requires adding validation inside open_artifact_pr.sh or introducing a new schema feature — both are design-level decisions.

Issue #2135, opened 2026-05-07
Perf: Verify api-simulator wheel caching in CI


Trecek:
## Problem

`api-simulator` is a private Rust/PyO3 crate compiled from source. It's pinned to git rev `8f711f9...` and only used in 3 test files (guarded by `pytest.importorskip`). Cold Rust compilation takes 3-8 minutes.

## Status: CONFIRMED — Wheel is NOT Cached

Investigation completed 2026-05-07 via diagnostic PR #2161 (run 25503753214). The api-simulator wheel recompiles from Rust source on **every CI run** (45–64s). Evidence collected across 6 runs — zero showed cache reuse.

### Root Cause: Three Compounding Failures

**1. Cache poisoning by `preflight` job**
The `preflight` job runs `setup-uv@v7` and only executes `uv lock --check` (no install, no compilation). It saves the uv cache in its `Post Install uv` step — this cache contains metadata (82MB) but no compiled wheels. The `test` job restores this same cache (same key), compiles the wheel during `uv sync` (cache grows to 493MB with the 3.6MB wheel), but GitHub Actions skips saves on exact key matches: `Cache hit occurred on key ..., not saving cache.` The wheel is built and then discarded every run.

**2. No `restore-keys` fallback**
`setup-uv` uses a single exact cache key hashing `uv.lock`. Any `uv.lock` change → total miss, no partial restore. `uv.lock` changes 36–51 times per busy day from version bumps (733 of 2,444 total commits).

**3. `uv-version` parameter silently ignored**
`setup-uv@v7` renamed `uv-version` to `version`. The workflow uses `uv-version: "0.9.21"`, which produces `##[warning]Unexpected input(s) 'uv-version'` and is silently ignored. CI runs uv 0.11.11 instead of the pinned 0.9.21.

### Diagnostic Evidence (Run 25503753214)

**Pre-install** — cache hit is empty of wheels:
```
Cache hit: true
=== Looking for api-simulator in uv cache ===
(empty — no files found)
=== uv cache size === 82M
```

**During install** — uv explicitly reports uncached, compiles from Rust:
```
DEBUG Identified uncached distribution: api-simulator @ git+...
Building api-simulator @ git+... (60 seconds)
```

**Post-install** — wheel exists but won't be saved:
```
api_simulator-0.1.0-cp39-abi3-linux_x86_64.whl (3.6M)
uv cache size AFTER install: 493M
[...]

PR #2161, opened 2026-05-07
ci: api-simulator cache diagnostics (issue #2135)

Trecek:
## Summary
- Adds diagnostic CI steps to verify whether the api-simulator Rust/PyO3 wheel is being cached or recompiled on every run
- Captures pre-install cache state, verbose `uv sync` output with compilation/cache pattern matching, and post-install cache contents
- This is a diagnostic-only PR for investigation — not intended to be merged as-is

## Context
Investigation for #2135. The hypothesis (supported by log analysis of recent runs) is that the wheel is recompiled every time because `uv.lock` changes frequently (version bumps), invalidating the entire `setup-uv` cache key.

## What the diagnostic steps check
1. **Pre-install**: Whether api-simulator wheel exists in the uv cache before `uv sync`
2. **During install**: Verbose `uv sync` output filtered for Rust compilation vs cache-hit patterns
3. **Post-install**: What's in the uv cache after install completes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -1,0 +1,17 @@
+# GitHub Actions CI — Agent Instructions
+
+## `setup-uv@v7` Cache Behavior
+
+- **Parameter name**: Use `version:` not `uv-version:` — the old name is silently ignored.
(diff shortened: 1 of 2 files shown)


Trecek:
Diagnostic-only PR for #2135 investigation. Evidence captured — closing.

Issue #2231, opened 2026-05-08
Campaign dispatch validation: capture-vs-sentinel and required-ingredient cross-checks


Trecek:
## Summary

Two classes of campaign recipe wiring errors are invisible to static validation: (1) capture blocks referencing fields that the target recipe's sentinel never emits, and (2) target recipes declaring `required: true` ingredients that the campaign dispatch never provides. These caused the research campaign chain to be fundamentally broken without any validation warning. Two new semantic rules are needed.

## Architectural Gap

The existing `dispatch-capture-value-references-result` rule (`rules_campaign.py:512-535`) validates only `${{ result.<field> }}` syntax — it does not load the target recipe to check whether the stop step's sentinel message actually instructs emission of that field.

The existing `campaign-dangling-ingredient` rule (`rules_campaign.py:291-331`) checks campaign-level ingredients outward (campaign→target) but not target-required ingredients inward (target→campaign). A target recipe's `required: true` ingredients that aren't in the campaign's own ingredient schema are invisible to validation.

### `dispatch-required-ingredient-provided` (ERROR)

For each non-gate dispatch, iterate the target recipe's `ingredients` where `required == True AND default is None`. Check if the key is present in the dispatch's `ingredients:` block OR is auto-injected (`_AUTO_INJECTED_CAMPAIGN_INGREDIENTS = {"task"}`). If absent, emit ERROR.

Must mirror the exact condition from `_run_dispatch` at `_api.py:334-345`:
```python
missing_required = [
    key for key, ing in full_recipe.ingredients.items()
    if getattr(ing, "required", False)
    and getattr(ing, "default", None) is None
    and key not in effective_ingredients
]
```

### `dispatch-capture-field-in-sentinel` (WARNING)

For each dispatch with a `capture:` block, load the target recipe and find all stop steps whose message contains "sentinel" (matching the existing `food-truck-has-sentinel-stop` heuristic). Parse field names from the JSON example block in the sentinel message (more robust than prose parsing). Check that each `result.<field>` referenced in the capture block appears in at least one sentinel's field list.

Design considerations:
- For multi-sentinel recipes (e.g., `research-review.yaml` with both PR-mode and local-mode sentinels), checking if a field appears in ANY sentinel path is the correct strategy — a field valid in PR mode but absent in local mode is a conditional gap, not an unconditional one. Consider a stronger variant that checks ALL paths, but use WARNING severity.
[...]

PR #2252, opened 2026-05-08
Rectify: Campaign Dispatch Validation — Capture-vs-Sentinel and Required-Ingredient Cross-Checks

Trecek:
## Summary

The campaign semantic validation system validates dispatch wiring *syntactically* (template patterns, identifier format, ref resolution) but lacks *semantic* cross-recipe validation: it never loads a target recipe to verify that (a) captured `${{ result.<field> }}` fields actually appear in the target's sentinel message, or (b) the target's `required: true` (no default) ingredients are provided by the dispatch. These two gaps allowed the entire research campaign chain to carry 6 field-level wiring errors without any validation warning. Two new semantic rules and a shared sentinel field extraction helper close both gaps, following the established `_load_dispatch_target` pattern already used by 4 existing rules.

## Requirements

Two classes of campaign recipe wiring errors are invisible to static validation: (1) capture blocks referencing fields that the target recipe's sentinel never emits, and (2) target recipes declaring `required: true` ingredients that the campaign dispatch never provides. These caused the research campaign chain to be fundamentally broken without any validation warning. Two new semantic rules are needed.

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260508-153155-490823/.autoskillit/temp/rectify/rectify_campaign_dispatch_validation_2026-05-08_153155.md`

Closes #2231

## Token Usage Summary

| rectify | claude-sonnet-4-6 | 1 | 1.1k | 17.4k | 1.4M | 98.4k | 174 | 85.0k | 10m 31s |
| dry_walkthrough | claude-opus-4-6 | 1 | 47 | 14.5k | 1.1M | 76.4k | 78 | 64.9k | 7m 12s |
| implement* | MiniMax-M2.7 | 1 | 80.6k | 11.0k | 1.8M | 21.0k | 70 | 0 | 7m 2s |
| prepare_pr* | MiniMax-M2.7 | 1 | 45.7k | 2.6k | 247.8k | 0 | 19 | 0 | 34s |
| compose_pr* | MiniMax-M2.7 | 1 | 37.5k | 1.5k | 338.5k | 0 | 21 | 0 | 23s |
| review_pr | claude-opus-4-6 | 3 | 1.1k | 70.8k | 2.2M | 90.6k | 160 | 204.3k | 21m 54s |
| resolve_review | claude-opus-4-6 | 2 | 112 | 24.7k | 2.3M | 75.5k | 106 | 122.3k | 12m 15s |
| **Total** | | | 166.1k | 142.5k | 9.4M | 98.4k | | 476.4k | 59m 52s |

## Token Efficiency

| rectify | 0 | — | — | — |
| implement | 642 | 2751.7 | 0.0 | 17.1 |
| review_pr | 0 | — | — | — |
| resolve_review | 12 | 195594.5 | 10187.9 | 2056.2 |
| **Total** | **654** | 14304.0 | 728.4 | 217.9 |

### DIFF
--- a/src/autoskillit/recipe/rules/rules_campaign.py
+++ b/src/autoskillit/recipe/rules/rules_campaign.py
@@ -376,11 +376,7 @@ def _check_dispatch_required_ingredient_provided(
         for auto in _AUTO_INJECTED_CAMPAIGN_INGREDIENTS:
             effective_ingredients.add(auto)
         for key, ing in target.ingredients.items():
-            if (
-                getattr(ing, "required", False)
-                and getattr(ing, "default", None) is None
-                and key not in effective_ingredients
-            ):
+            if ing.required and ing.default is None and key not in effective_ingredients:
                 findings.append(
                     RuleFinding(
                         rule="dispatch-required-ingredient-provided",
--- a/src/autoskillit/recipe/rules/rules_campaign.py
+++ b/src/autoskillit/recipe/rules/rules_campaign.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import json
 from collections import Counter
 from pathlib import Path
 from typing import TYPE_CHECKING
@@ -24,6 +25,27 @@ _KEBAB_RE = re.compile(r"^[a-z0-9]+(-[a-z0-9]+)*$")
 # dispatch task: field), so campaigns that rely on this injection pattern should not
 # be flagged for not explicitly forwarding them in the ingredients block.
 _AUTO_INJECTED_CAMPAIGN_INGREDIENTS: frozenset[str] = frozenset({"task"})
+
+
+_SENTINEL_JSON_RE = re.compile(r"[Ee]xample\s+sentinel:\s*(\{[^}]+\})", re.DOTALL)
+
+
+def _extract_sentinel_fields(recipe: Recipe) -> frozenset[str]:
+    """Extract declared field names from all sentinel stop step JSON examples."""
+    fields: set[str] = set()
+    for step in recipe.steps.values():
+        if step.action != "stop" or not step.message:
+            continue
+        if "sentinel" not in step.message.lower():
+            continue
+        for match in _SENTINEL_JSON_RE.finditer(step.message):
+            try:
+                parsed = json.loads(match.group(1))
+                if isinstance(parsed, dict):
+                    fields.update(parsed.keys())
+            except (json.JSONDecodeError, ValueError):
+                continue
+    return frozenset(fields)
 
 
 def _load_dispatch_target(dispatch: CampaignDispatch, project_dir: Path | None) -> Recipe | None:
@@ -332,6 +354,50 @@ def _check_campaign_dangling_ingredient(ctx: ValidationContext) -> list[RuleFind
 
 
 @semantic_rule(
+    name="dispatch-required-ingredient-provided",
+    description=(
+        "Target recipe required ingredients (no default) must be provided by the dispatch"
+    ),
+    severity=Severity.ERROR,
+)
+def _check_dispatch_required_ingredient_provided(
+    ctx: ValidationContext,
+) -> list[RuleFinding]:
+    if ctx.recipe.kind != RecipeKind.CAMPAIGN:
+        return []
+    findings: list[RuleFinding] = []
+    for d in ctx.recipe.dispatches:
+        if d.gate:
+            continue
+        target = _load_dispatch_target(d, ctx.project_dir)
+        if target is None:
+            continue
+        effective_ingredients = set(d.ingredients.keys())
+        for auto in _AUTO_INJECTED_CAMPAIGN_INGREDIENTS:
+            effective_ingredients.add(auto)
+        for key, ing in target.ingredients.items():
+            if (
+                getattr(ing, "required", False)
+                and getattr(ing, "default", None) is None
+                and key not in effective_ingredients
+            ):
+                findings.append(
+                    RuleFinding(
+                        rule="dispatch-required-ingredient-provided",
+                        severity=Severity.ERROR,
+                        step_name="(top-level)",
+                        message=(
+                            f"Dispatch {d.name!r} targets recipe {d.recipe!r} which "
+                            f"declares ingredient {key!r} as required (no default), "
+                            f"but the dispatch does not provide it. "
+                            f"Provided: {sorted(d.ingredients)}."
+                        ),
+                    )
+                )
+    return findings
+
+
+@semantic_rule(
     name="dispatch-ingredient-values-are-strings",
     description="All dispatch ingredient values must be strings",
     severity=Severity.ERROR,
@@ -480,6 +546,7 @@ def _check_campaign_task_non_empty(ctx: ValidationContext) -> list[RuleFinding]:
 
 _IDENT_RE = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
 _RESULT_TEMPLATE_RE = re.compile(r"^\$\{\{\s*result\.[\w-]+\s*\}\}$")
+_RESULT_FIELD_RE = re.compile(r"^\$\{\{\s*result\.([\w-]+)\s*\}\}$")
 
 
 @semantic_rule(
@@ -529,6 +596,47 @@ def _check_dispatch_capture_value_references_result(ctx: ValidationContext) -> l
                         message=(
                             f"Dispatch {d.name!r} capture[{key!r}] value {val!r} must use "
                             "${{ result.<field_name> }} syntax."
+                        ),
+                    )
+                )
+    return findings
+
+
+@semantic_rule(
+    name="dispatch-capture-field-in-sentinel",
+    description="Captured result fields should appear in target recipe's sentinel message",
+    severity=Severity.WARNING,
+)
+def _check_dispatch_capture_field_in_sentinel(ctx: ValidationContext) -> list[RuleFinding]:
+    if ctx.recipe.kind != RecipeKind.CAMPAIGN:
+        return []
+    findings: list[RuleFinding] = []
+    for d in ctx.recipe.dispatches:
+        if d.gate or not d.capture:
+            continue
+        target = _load_dispatch_target(d, ctx.project_dir)
+        if target is None:
+            continue
+        sentinel_fields = _extract_sentinel_fields(target)
+        if not sentinel_fields:
+            continue
+        for cap_key, cap_val in d.capture.items():
+            match = _RESULT_FIELD_RE.match(cap_val.strip())
+            if not match:
+                continue
+            field_name = match.group(1)
+            if field_name not in sentinel_fields:
+                findings.append(
+                    RuleFinding(
+                        rule="dispatch-capture-field-in-sentinel",
+                        severity=Severity.WARNING,
+                        step_name="(top-level)",
+                        message=(
+                            f"Dispatch {d.name!r} captures {cap_key!r} as "
+                            f"${{{{ result.{field_name} }}}} but target recipe "
+                            f"{d.recipe!r} has no sentinel stop step listing "
+                            f"field {field_name!r}. "
+                            f"Known sentinel fields: {sorted(sentinel_fields)}."
                         ),
                     )
                 )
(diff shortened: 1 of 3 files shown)


src/autoskillit/recipe/rules/rules_campaign.py
```diff
@@ -26,6 +27,27 @@
 _AUTO_INJECTED_CAMPAIGN_INGREDIENTS: frozenset[str] = frozenset({"task"})
 
 
+_SENTINEL_JSON_RE = re.compile(r"[Ee]xample\s+sentinel:\s*(\{[^}]+\})", re.DOTALL)
```
Trecek, line 30: [warning] defense: The regex `_SENTINEL_JSON_RE` uses `[^}]+` which stops at the first `}`. This means it cannot match sentinel JSON with nested objects (e.g., `{"meta": {"count": 1}}`). The multiline test only exercises flat objects. A sentinel with a nested object value will only partially match and fail `json.loads`. Consider using a balanced-brace approach or a non-greedy `.*?}` with validation.

src/autoskillit/recipe/rules/rules_campaign.py
```diff
@@ -331,6 +353,46 @@ def _check_campaign_dangling_ingredient(ctx: ValidationContext) -> list[RuleFind
...
+    for d in ctx.recipe.dispatches:
+        if d.gate:
+            continue
+        target = _load_dispatch_target(d, ctx.project_dir)
+        if target is None:
+            continue
+        effective_ingredients = set(d.ingredients.keys())
+        for auto in _AUTO_INJECTED_CAMPAIGN_INGREDIENTS:
+            effective_ingredients.add(auto)
+        for key, ing in target.ingredients.items():
```
Trecek, line 378: [warning] defense: The condition `ing.required and ing.default is None` conflates two concerns. The test at L1058 exercises `required=True, default="main"` which passes (correctly). But if `required` is always True when no default exists, the `ing.required` check is redundant. If `required` can be True with a non-None default (as the test shows), then the conjunction is the intended semantics — but this means `required=True` alone does NOT trigger the rule, which may surprise users expecting "required means must be provided".

src/autoskillit/recipe/rules/rules_campaign.py
```diff
@@ -331,6 +353,46 @@ def _check_campaign_dangling_ingredient(ctx: ValidationContext) -> list[RuleFind
...
+            if ing.required and ing.default is None and key not in effective_ingredients:
```
Trecek, line 379: [warning] bugs: Edge case: if a `RecipeIngredient` has `required=True` but `default=""` (empty string), this rule will NOT fire because `ing.default is None` is False. An empty-string default effectively suppresses the validation. Verify this is the intended behavior — an empty string default may not be a meaningful value for all ingredient types.

src/autoskillit/recipe/rules/rules_campaign.py
```diff
@@ -480,6 +542,7 @@ def _check_campaign_task_non_empty(ctx: ValidationContext) -> list[RuleFinding]:
 
 _IDENT_RE = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
 _RESULT_TEMPLATE_RE = re.compile(r"^\$\{\{\s*result\.[\w-]+\s*\}\}$")
+_RESULT_FIELD_RE = re.compile(r"^\$\{\{\s*result\.([\w-]+)\s*\}\}$")
```
Trecek, line 545: [info] cohesion: `_RESULT_TEMPLATE_RE` (L544) and `_RESULT_FIELD_RE` (L545) share nearly identical patterns — one is a validator, the other a capturing extractor. Both are needed by different rules, but the structural similarity makes the distinction non-obvious on first read.

src/autoskillit/recipe/rules/rules_campaign.py
```diff
@@ -535,6 +598,47 @@ def _check_dispatch_capture_value_references_result(ctx: ValidationContext) -> l
...
+        return []
+    findings: list[RuleFinding] = []
+    for d in ctx.recipe.dispatches:
+        if d.gate or not d.capture:
+            continue
+        target = _load_dispatch_target(d, ctx.project_dir)
+        if target is None:
+            continue
+        sentinel_fields = _extract_sentinel_fields(target)
+        if not sentinel_fields:
```
Trecek, line 617: [warning] defense: When `sentinel_fields` is empty (no parseable sentinel JSON found), the rule silently skips via `continue`. A dispatch declaring `capture` against a target with no parseable sentinel is a configuration smell worth surfacing. The regex does match case-insensitively (`[Ee]xample`), but a completely unrecognizable sentinel pattern produces silent absorption rather than a diagnostic.

src/autoskillit/recipe/rules/rules_campaign.py
```diff
@@ -535,6 +598,47 @@ def _check_dispatch_capture_value_references_result(ctx: ValidationContext) -> l
...
+            continue
+        for cap_key, cap_val in d.capture.items():
```
Trecek, line 619: [warning] bugs: `cap_val.strip()` assumes `cap_val` is a string. While the `dispatch-ingredient-values-are-strings` rule validates this upstream, there is no local guard. If schema enforcement is bypassed (e.g., programmatic recipe construction), this will raise `AttributeError`. Low risk given the existing validation chain, but a defensive gap.

tests/recipe/test_research_campaign_rules.py
```diff
@@ -78,15 +78,18 @@ def test_depends_on_acyclic_passes(validation_ctx) -> None:
 
 
 def test_dispatch_capture_fields_in_sentinel_contract_passes(validation_ctx) -> None:
-    rule_name = "dispatch-capture-fields-in-sentinel-contract"
-    registered_names = {r.name for r in _RULE_REGISTRY}
-    if rule_name not in registered_names:
-        pytest.skip(f"Rule {rule_name!r} not yet registered")
+    rule_name = "dispatch-capture-field-in-sentinel"
```
Trecek, line 81: [info] tests: The old test guarded itself with `pytest.skip` if the rule wasn't registered. Now that the guard is removed and the rule name changed from plural `dispatch-capture-fields-in-sentinel-contract` to singular `dispatch-capture-field-in-sentinel`, confirm the registry name is final — the test will hard-fail (not skip) on any future rename.

tests/recipe/test_rules_campaign.py
```diff
@@ -967,3 +967,524 @@ def test_campaign_dangling_ingredient_exempts_task(tmp_path: Path):
...
+            CampaignDispatch(
+                name="phase-two",
+                recipe="target-recipe",
+                task="do it",
+                ingredients={"task": "do it"},
+                # source_dir missing
+            ),
+        ],
+    )
+    found = _findings(recipe, "dispatch-required-ingredient-provided", project_dir=tmp_path)
```
Trecek, line 1170: [info] tests: The negative assertion `"phase-one" not in found[0].message` is fragile — the error template includes `Provided: {sorted(d.ingredients)}` which could contain the substring 'phase-one' if an ingredient key happened to match. Asserting on the dispatch name field directly would be more robust.

Trecek (commented):

tests/recipe/test_rules_campaign.py
```diff
@@ -967,3 +967,524 @@ def test_campaign_dangling_ingredient_exempts_task(tmp_path: Path):
...
+
+
+def test_dispatch_required_ingredient_provided_skips_non_campaign(tmp_path: Path):
+    """Rule skips non-campaign recipes."""
+    recipe = _standard_recipe()
+    found = _findings(recipe, "dispatch-required-ingredient-provided", project_dir=tmp_path)
+    assert found == []
+
+
+# ---------------------------------------------------------------------------
```
Trecek, line 1183: [warning] cohesion: Section label is 'T-S2' but the required-ingredient section above is 'T-R1'. If 'R' = required and 'S' = sentinel, the convention is intentional but undocumented. The prior commit 'renumber colliding T-S1 test section labels' confirms T-S was already in use — consider documenting the prefix convention in the test file header.

tests/recipe/test_rules_campaign.py
```diff
@@ -967,3 +967,524 @@ def test_campaign_dangling_ingredient_exempts_task(tmp_path: Path):
...
+                task="do it",
+                capture={"alpha": "${{ result.alpha }}"},
+            ),
+        ]
+    )
+    found = _findings(recipe, "dispatch-capture-field-in-sentinel", project_dir=tmp_path)
+    assert found == []
+
+
+def test_dispatch_capture_field_in_sentinel_skips_no_sentinel(tmp_path: Path):
```
Trecek, line 1302: [warning] tests: `test_dispatch_capture_field_in_sentinel_skips_no_sentinel` writes a full target recipe YAML to disk to test a code path that returns `[]` when `sentinel_fields` is empty. The filesystem setup is heavier than needed — the rule's early `continue` at L617 fires before any file content matters beyond the absence of a sentinel marker. Consider whether an unloadable-target variant (like L1376) would suffice.

tests/recipe/test_rules_campaign.py
```diff
@@ -967,3 +967,524 @@ def test_campaign_dangling_ingredient_exempts_task(tmp_path: Path):
...
+                task="do it",
+                capture={"output_path": "${{ result.output_path }}"},
+            ),
+        ]
+    )
+    found = _findings(recipe, "dispatch-capture-field-in-sentinel", project_dir=tmp_path)
+    assert found == []
+
+
+def test_dispatch_capture_field_in_sentinel_skips_no_capture(tmp_path: Path):
```
Trecek, line 1331: [warning] tests: `test_dispatch_capture_field_in_sentinel_skips_no_capture` passes `capture={}` (empty dict) rather than `capture=None`. The production guard `not d.capture` at L611 is falsy for both `None` and `{}`, so both paths are covered. But the docstring says "dispatch has no capture block" — if the schema distinguishes `None` (absent) from `{}` (empty), a second variant with `capture=None` would give full branch coverage.

tests/recipe/test_rules_campaign.py
```diff
@@ -967,3 +967,524 @@ def test_campaign_dangling_ingredient_exempts_task(tmp_path: Path):
...
+            },
+        },
+    )
+    recipe = _campaign(
+        dispatches=[
+            CampaignDispatch(
+                name="phase-one",
+                recipe="target-recipe",
+                task="do it",
+                capture={},
```
Trecek, line 1357: [warning] bugs: Same `capture={}` vs `None` concern as the no-capture test above (L1331). The docstring says "dispatch has no capture block" but passes an empty dict. This test passes because `not d.capture` handles both, but the intent documentation is imprecise.

tests/recipe/test_rules_campaign.py
```diff
@@ -967,3 +967,524 @@ def test_campaign_dangling_ingredient_exempts_task(tmp_path: Path):
...
+
+def test_dispatch_capture_field_in_sentinel_skips_non_campaign(tmp_path: Path):
+    """Rule skips non-campaign recipes."""
+    recipe = _standard_recipe()
+    found = _findings(recipe, "dispatch-capture-field-in-sentinel", project_dir=tmp_path)
+    assert found == []
+
+
+# ---------------------------------------------------------------------------
+# T-S3: _extract_sentinel_fields
```
Trecek, line 1400: [info] cohesion: Section 'T-S3: _extract_sentinel_fields' tests a private helper directly via inline import — the only helper in this file tested that way. Other rules are exercised through the public `_findings`/`run_semantic_rules` path. Consider grouping T-S3 as a sub-section of T-S2 to signal the structural relationship.

tests/recipe/test_rules_campaign.py
```diff
@@ -967,3 +967,524 @@ def test_campaign_dangling_ingredient_exempts_task(tmp_path: Path):
...
+# ---------------------------------------------------------------------------
+
+
+def test_extract_sentinel_fields_parses_json_example():
```
Trecek, line 1404: [warning] tests: The four `_extract_sentinel_fields` tests (L1404, L1427, L1442, L1469) each repeat `from autoskillit.recipe.rules.rules_campaign import _extract_sentinel_fields` and `from autoskillit.recipe.schema import Recipe, RecipeStep` inline. These imports should be hoisted to module level or a shared fixture to reduce repetition and simplify maintenance if the internal path changes.

Trecek (commented):

PR #2324, opened 2026-05-10
Split CI Test Suite into Two Parallel Shards

Trecek:
## Summary

Split the single `test` job in `.github/workflows/tests.yml` into two parallel shards using a matrix strategy. Shard A ("execution") positive-selects the heavy test directories (`tests/execution/`, `tests/contracts/`, `tests/core/`, `tests/planner/`, `tests/pipeline/`, `tests/migration/`, `tests/integration/`, and root-level `tests/test_*.py` files). Shard B ("general") is a catch-all that runs `tests/` with `--ignore` flags for every Shard A path, ensuring new test directories automatically land in Shard B without workflow changes.

The Taskfile is modified to accept `PYTEST_TEST_PATHS` and `PYTEST_IGNORE_PATHS` environment variables (defaulting to `tests/` and empty respectively), keeping local usage unchanged while enabling CI sharding. `lint-imports` is extracted to a separate CI step so it runs exactly once, not per-shard.

## Conflict Resolution Decisions

The following files had merge conflicts that were automatically resolved.

### New (★):
tests/infra/test_ci_shard_config.py

### Modified (●):
.github/workflows/tests.yml
Taskfile.yml
tests/infra/CLAUDE.md

## Implementation Plan

Plan file: /home/talon/projects/autoskillit-runs/impl-20260509-174835-412085/.autoskillit/temp/make-plan/split_ci_test_shards_plan_2026-05-09_175800.md

## Token Usage Summary

| plan | claude-opus-4-6 | 1 | 1.6k | 13.8k | 697.4k | 55.6k | 45 | 46.7k | 6m 4s |
| verify | claude-opus-4-6 | 1 | 47 | 17.9k | 467.1k | 62.4k | 58 | 49.2k | 7m 51s |
| implement* | MiniMax-M2.7-highspeed | 1 | 440.5k | 5.1k | 661.4k | 28.8k | 53 | 66.1k | 2m 24s |
| prepare_pr* | MiniMax-M2.7 | 1 | 43.2k | 2.9k | 239.1k | 0 | 17 | 0 | 1m 3s |
| compose_pr* | MiniMax-M2.7 | 1 | 41.3k | 1.9k | 196.2k | 0 | 17 | 0 | 1m 8s |
| review_pr | claude-opus-4-6 | 3 | 122 | 67.0k | 1.2M | 65.7k | 81 | 198.8k | 17m 12s |
| resolve_review | claude-opus-4-6 | 3 | 159 | 40.1k | 3.6M | 73.1k | 156 | 158.1k | 17m 14s |
| **Total** | | | 526.9k | 148.8k | 7.0M | 73.1k | | 518.9k | 52m 58s |

## Token Efficiency

| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| implement | 74 | 8937.4 | 893.4 | 68.4 |
| review_pr | 0 | — | — | — |
| resolve_review | 83 | 42879.3 | 1904.2 | 483.2 |
| **Total** | **157** | 44854.7 | 3305.4 | 947.5 |

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -92,15 +92,19 @@ jobs:
         id: test-paths
         run: |
           SHARD_A_DIRS="tests/execution tests/contracts tests/core tests/planner tests/pipeline tests/migration tests/integration"
+          if find tests/ -maxdepth 1 -name 'test_*.py' -print0 | grep -qz ' '; then
+            echo "::error::Test filenames with spaces are not supported by the shard path splitter"
+            exit 1
+          fi
           if [[ "${{ matrix.shard }}" == "execution" ]]; then
-            ROOT_FILES=$(find tests/ -maxdepth 1 -name "test_*.py" | sort | tr '\n' ' ')
+            ROOT_FILES=$(find tests/ -maxdepth 1 -name 'test_*.py' | sort | tr '\n' ' ')
             echo "PYTEST_TEST_PATHS=${SHARD_A_DIRS} ${ROOT_FILES}" >> "$GITHUB_ENV"
           else
             IGNORES=""
             for d in $SHARD_A_DIRS; do
               IGNORES="$IGNORES --ignore=$d"
             done
-            ROOT_IGNORES=$(find tests/ -maxdepth 1 -name "test_*.py" -exec echo "--ignore={}" \; | sort | tr '\n' ' ')
+            ROOT_IGNORES=$(find tests/ -maxdepth 1 -name 'test_*.py' | sort | sed 's|^|--ignore=|' | tr '\n' ' ')
             echo "PYTEST_TEST_PATHS=tests/" >> "$GITHUB_ENV"
             echo "PYTEST_IGNORE_PATHS=${IGNORES} ${ROOT_IGNORES}" >> "$GITHUB_ENV"
           fi
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,13 +44,14 @@ jobs:
           uv run python --version
 
   test:
-    name: Test on ${{ matrix.os }}
+    name: Test (${{ matrix.shard }}) on ${{ matrix.os }}
     needs: preflight
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: ${{ fromJSON(needs.preflight.outputs.os-matrix) }}
+        shard: [execution, general]
 
     steps:
       - uses: actions/checkout@v4
@@ -87,6 +88,27 @@ jobs:
       - name: Generate recipe diagrams
         run: uv run autoskillit recipes render
 
+      - name: Compute test paths
+        id: test-paths
+        run: |
+          SHARD_A_DIRS="tests/execution tests/contracts tests/core tests/planner tests/pipeline tests/migration tests/integration"
+          if [[ "${{ matrix.shard }}" == "execution" ]]; then
+            ROOT_FILES=$(find tests/ -maxdepth 1 -name "test_*.py" | sort | tr '\n' ' ')
+            echo "PYTEST_TEST_PATHS=${SHARD_A_DIRS} ${ROOT_FILES}" >> "$GITHUB_ENV"
+          else
+            IGNORES=""
+            for d in $SHARD_A_DIRS; do
+              IGNORES="$IGNORES --ignore=$d"
+            done
+            ROOT_IGNORES=$(find tests/ -maxdepth 1 -name "test_*.py" -exec echo "--ignore={}" \; | sort | tr '\n' ' ')
+            echo "PYTEST_TEST_PATHS=tests/" >> "$GITHUB_ENV"
+            echo "PYTEST_IGNORE_PATHS=${IGNORES} ${ROOT_IGNORES}" >> "$GITHUB_ENV"
+          fi
+
+      - name: Lint imports
+        if: matrix.shard == 'execution'
+        run: uv run lint-imports
+
       - name: Run tests
         env:
           AUTOSKILLIT_FILTER_STATS_FILE: ${{ github.workspace }}/.autoskillit/temp/filter-stats.json
@@ -100,12 +122,12 @@ jobs:
           if [[ "${{ github.event_name }}" == "pull_request" && "${{ github.base_ref }}" == "develop" ]]; then
             export AUTOSKILLIT_TEST_FILTER=conservative
           fi
-          task test-all
+          task test-check
 
       - name: Upload filter stats
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: filter-stats-${{ matrix.os }}
+          name: filter-stats-${{ matrix.os }}-${{ matrix.shard }}
           path: .autoskillit/temp/filter-stats.json
           if-no-files-found: ignore
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -38,7 +38,7 @@ tasks:
 
         set -o pipefail
         set +e
-        $PYTEST_CMD tests/ -q --tb=short -n 4 -m "not smoke and not canary" --disable-warnings -o "addopts=" --basetemp={{.PYTEST_TMPDIR}} -o "cache_dir={{.PYTEST_CACHEDIR}}" 2>&1 | tee "$TEST_OUTPUT"
+        $PYTEST_CMD ${PYTEST_TEST_PATHS:-tests/} -q --tb=short -n 4 -m "not smoke and not canary" ${PYTEST_IGNORE_PATHS:-} --disable-warnings -o "addopts=" --basetemp={{.PYTEST_TMPDIR}} -o "cache_dir={{.PYTEST_CACHEDIR}}" 2>&1 | tee "$TEST_OUTPUT"
         PYTEST_EXIT=$?
         set +o pipefail
         set -e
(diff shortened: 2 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -87,6 +88,27 @@ jobs:
       - name: Generate recipe diagrams
         run: uv run autoskillit recipes render
 
+      - name: Compute test paths
+        id: test-paths
+        run: |
+          SHARD_A_DIRS="tests/execution tests/contracts tests/core tests/planner tests/pipeline tests/migration tests/integration"
+          if [[ "${{ matrix.shard }}" == "execution" ]]; then
+            ROOT_FILES=$(find tests/ -maxdepth 1 -name "test_*.py" | sort | tr '\n' ' ')
```
Trecek: [warning] arch: ROOT_FILES uses `find` output injected unquoted into PYTEST_TEST_PATHS. If any test filename contains spaces, the expansion will split incorrectly. Unlikely in practice but a latent fragility in the shell word-splitting.

Trecek (commented):

.github/workflows/tests.yml
```diff
@@ -87,6 +88,27 @@ jobs:
...
+            echo "PYTEST_TEST_PATHS=${SHARD_A_DIRS} ${ROOT_FILES}" >> "$GITHUB_ENV"
+          else
+            IGNORES=""
+            for d in $SHARD_A_DIRS; do
+              IGNORES="$IGNORES --ignore=$d"
+            done
+            ROOT_IGNORES=$(find tests/ -maxdepth 1 -name "test_*.py" -exec echo "--ignore={}" \; | sort | tr '\n' ' ')
```
Trecek: [warning] arch: ROOT_IGNORES is computed inline and appended to PYTEST_IGNORE_PATHS in a single echo. While correct, adding a second assignment would silently overwrite. Consider building the full ignore list in a single variable before the final echo to GITHUB_ENV.

.github/workflows/tests.yml
```diff
@@ -100,12 +122,12 @@ jobs:
           if [[ "${{ github.event_name }}" == "pull_request" && "${{ github.base_ref }}" == "develop" ]]; then
             export AUTOSKILLIT_TEST_FILTER=conservative
           fi
-          task test-all
+          task test-check
```
Trecek, line 129: [warning] arch: `task test-check` replaces `task test-all` but lint-imports is now only run on the execution shard (L109). Previously lint ran for all matrix entries via test-all. Confirm that import linting only on one shard provides acceptable coverage.

tests/infra/test_ci_shard_config.py
```diff
@@ -0,0 +1,78 @@
...
+from pathlib import Path
+
+import pytest
+
+pytestmark = [pytest.mark.layer("infra"), pytest.mark.small]
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+WORKFLOW_PATH = REPO_ROOT / ".github" / "workflows" / "tests.yml"
+
+KNOWN_GENERAL_SHARD_DIRS = {
```
Trecek, line 14: [warning] cohesion: KNOWN_GENERAL_SHARD_DIRS stores bare directory names while SHARD_A_DIRS stores tests/-prefixed paths. The union in test_all_test_dirs_assigned_to_shard uses Path(d).name to normalize. If SHARD_A_DIRS ever contains a nested path deeper than one level (e.g. tests/foo/bar), Path(d).name would silently resolve to just bar. Consider standardizing both sets to bare directory names.

tests/infra/test_ci_shard_config.py
```diff
@@ -0,0 +1,78 @@
...
+    if not WORKFLOW_PATH.exists():
+        pytest.skip(f"Workflow file not found: {WORKFLOW_PATH}")
+    text = WORKFLOW_PATH.read_text()
+    match = re.search(r'SHARD_A_DIRS="([^"]+)"', text)
+    if not match:
+        pytest.skip("SHARD_A_DIRS not found in tests.yml — skipping shard config tests")
+    return match.group(1).split()
+
+
+SHARD_A_DIRS = _parse_shard_a_dirs_from_workflow()
```
Trecek, line 47: [warning] tests: Module-level pytest.skip() via _parse_shard_a_dirs_from_workflow() is unreliable under pytest-xdist (-n 4). Collection in subprocesses can surface skip exceptions as collection errors rather than clean skips. Defer the parse into a session-scoped fixture or use allow_module_level=True explicitly.

.github/workflows/tests.yml
Trecek: **Observation from local review round 1:**

[warning] cohesion: The execution shard includes root-level test_*.py files alongside the named SHARD_A_DIRS directories. This groups infrastructure-style root test files with execution-domain tests, weakening the shard's conceptual identity. The root files are filtered out of general via ROOT_IGNORES, but their assignment to execution is implicit and undocumented.

**Evidence:** The reviewer correctly observes that root-level test_*.py files are assigned to the execution shard implicitly with no inline comment explaining the rationale. The logic is functionally correct (no tests are dropped), but the design choice is undocumented.

Trecek (commented):
Observations accumulated from 5 local review rounds:

.github/workflows/tests.yml
```diff
@@ -87,6 +88,27 @@ jobs:
...
+            for d in $SHARD_A_DIRS; do
+              IGNORES="$IGNORES --ignore=$d"
+            done
+            ROOT_IGNORES=$(find tests/ -maxdepth 1 -name "test_*.py" -exec echo "--ignore={}" \; | sort | tr '\n' ' ')
+            echo "PYTEST_TEST_PATHS=tests/" >> "$GITHUB_ENV"
+            echo "PYTEST_IGNORE_PATHS=${IGNORES} ${ROOT_IGNORES}" >> "$GITHUB_ENV"
+          fi
+
+      - name: Lint imports
+        if: matrix.shard == 'execution'
```
Trecek, line 113: **Observation from local review round 1:**

[warning] cohesion: lint-imports is gated only on the execution shard. Import layer violations can originate in any package (server, fleet, recipe, etc.) running under the general shard. Pinning lint-imports to execution means a shard rename or rebalancing would silently drop import linting from CI.

**Evidence:** lint-imports analyzes source packages not test directories, so running it once in either shard is sufficient. The fragility concern is legitimate: the lint step is coupled to the 'execution' shard name string literal.

tests/infra/test_ci_shard_config.py
```diff
@@ -0,0 +1,78 @@
...
+
```
Trecek, line 48: **Observation from local review round 1:**

[info] tests: Error message 'Shard A directory {d} does not exist' implies d is a bare name, but d is a relative path like tests/execution. Minor clarity issue only.

**Evidence:**

tests/infra/test_ci_shard_config.py
```diff
@@ -0,0 +1,78 @@
...
+
+class TestCIShardConfig:
```
Trecek, line 50: **Observation from local review round 1:**

[info] arch: Path(d).name extraction at L57 works correctly for current SHARD_A entries (tests/execution -> execution), but would break if a SHARD_A entry were added outside tests/ whose basename collides with an existing tests/ subdirectory.

**Evidence:**

Taskfile.yml
```diff
@@ -38,7 +38,7 @@ tasks:
 
         set -o pipefail
         set +e
-        $PYTEST_CMD tests/ -q --tb=short -n 4 -m "not smoke and not canary" --disable-warnings -o "addopts=" --basetemp={{.PYTEST_TMPDIR}} -o "cache_dir={{.PYTEST_CACHEDIR}}" 2>&1 | tee "$TEST_OUTPUT"
+        $PYTEST_CMD ${PYTEST_TEST_PATHS:-tests/} -q --tb=short -n 4 -m "not smoke and not canary" ${PYTEST_IGNORE_PATHS:-} --disable-warnings -o "addopts=" --basetemp={{.PYTEST_TMPDIR}} -o "cache_dir={{.PYTEST_CACHEDIR}}" 2>&1 | tee "$TEST_OUTPUT"
```
Trecek, line 41: **Observation from local review round 1:**

[info] arch: PYTEST_TEST_PATHS and PYTEST_IGNORE_PATHS env vars are now consumed by task test-check and test-all. A developer who happens to have these set in their shell environment will silently run a subset of tests locally. Consider documenting the variables or scoping them to CI only.

**Evidence:**

.github/workflows/tests.yml
Trecek: Agreed — fixed in cde92ca1. Added an early guard that fails CI if any test filename contains spaces, and replaced the `-exec echo` pattern with `sed` for ROOT_IGNORES.

.github/workflows/tests.yml
Trecek: Investigated — this is intentional. PYTEST_TEST_PATHS and PYTEST_IGNORE_PATHS are separate env vars, each written via a single echo. IGNORES and ROOT_IGNORES are intermediate shell variables combined before the final echo — there is no overwrite risk. The code at L99-105 builds the full ignore list in two variables (IGNORES via loop, ROOT_IGNORES via find) and concatenates them in a single GITHUB_ENV write.

.github/workflows/tests.yml
Trecek: Valid observation — flagged for design decision. lint-imports checks source packages (not test directories), so running it once in any shard is sufficient for correctness. However, the coupling to the 'execution' shard name is a maintenance concern if shards are renamed or rebalanced.

tests/infra/test_ci_shard_config.py
Trecek: Agreed — fixed in 33154f54. Replaced `Path(d).name` with `d.removeprefix("tests/")` to avoid silent misresolution on deeper nested paths.

tests/infra/test_ci_shard_config.py
Trecek: Agreed — fixed in 33154f54. Added `allow_module_level=True` to both `pytest.skip()` calls in `_parse_shard_a_dirs_from_workflow()` to prevent xdist collection errors.

Issue #2585, opened 2026-05-18
Recipe-scoped provider step overrides (specificity-based resolution)


Trecek:
## Problem

`step_overrides` in the provider configuration is a flat `step_name → profile_name` map. When multiple recipes share the same step name (e.g., both `implementation.yaml` and `remediation.yaml` have a step named `implement`), there is no way to route them to different providers. The override applies universally to all recipes.

This is a problem for sensitive recipes like `remediation`, where routing the `implement` step to a cheaper/faster provider is undesirable even though the same override is cost-effective for the standard `implementation` recipe.

## Current Behavior

The resolution in `_resolve_provider_profile()` (`server/_guards.py:157-201`) uses a four-tier waterfall:

1. **Tier 1** — `step_overrides[step_name]` (requires `recipe_name` to be non-empty, but does not use the recipe name as a routing key)
2. **Tier 2** — `step_overrides["*"]` wildcard (requires `recipe_name` non-empty)
3. **Tier 3** — step YAML `model:` field resolved as a profile name
4. **Tier 4** — `providers.default_provider` or `"anthropic"`

`ProvidersConfig` (`config/_config_dataclasses.py:402-423`) defines `step_overrides` as `dict[str, str]` — no recipe dimension.

Both recipes declare `step_name: "implement"` (`remediation.yaml:244`, `implementation.yaml:235`), so both hit the same Tier 1 match.

## Desired Behavior

Add a recipe-scoped override layer that takes precedence over the global `step_overrides`. The principle is **specificity**: the more specific the match, the higher priority it gets.

### Resolution priority (first match wins)

```
 Highest   recipe_overrides.{recipe}.{step}   — per-recipe, per-step
    ↓      step_overrides.{step}              — global, per-step (existing)
    ↓      step_overrides.*                   — global wildcard (existing)
    ↓      Step YAML model: field             — inline in recipe YAML (existing)
 Lowest    providers.default_provider          — fallback (existing)
```

### Example config

```yaml
providers:
  step_overrides:
    implement: minimax        # global default for all implement steps

  recipe_overrides:
    remediation:
      implement: anthropic    # overrides the global for remediation only
```

### Result

| Recipe | Step | Resolution | Provider |
|--------|------|-----------|----------|
| implementation | implement | global `step_overrides` | minimax |
| remediation | implement | `recipe_overrides.remediation.implement` | anthropic |
| any other | implement | global `step_overrides` | minimax |

### Affected components

[...]

PR #2590, opened 2026-05-18
Recipe-Scoped Provider Step Overrides with Specificity-Based Resolution

Trecek:
## Summary

Add a `recipe_overrides` field to `ProvidersConfig` that enables per-recipe, per-step provider routing. The existing four-tier waterfall in `_resolve_provider_profile()` becomes a six-tier waterfall with two new tiers at the top: (1) per-recipe per-step override and (2) per-recipe wildcard override. This follows the specificity principle — the more specific the match, the higher priority it gets. The existing `step_overrides` behavior is unchanged; the new tiers only fire when `recipe_overrides` contains a matching entry.

Including the per-recipe wildcard (`recipe_overrides.{recipe}.*`) is the correct design. It completes the specificity matrix — without it, there's no way to say "route all steps of recipe X to provider Y" without enumerating every step name. This is a natural need (e.g., "all remediation steps should use anthropic") and deferring it would require a future breaking change to the resolution order.

Closes #2585

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-20260518-101157-129835/.autoskillit/temp/make-plan/recipe_scoped_provider_overrides_plan_2026-05-18_101500.md`

## Token Usage Summary

| plan | claude-sonnet-4-6 | 1 | 76 | 8.8k | 555.6k | 80.8k | 129 | 51.1k | 7m 6s |
| verify | claude-opus-4-6 | 1 | 1.6k | 12.0k | 1.6M | 70.0k | 71 | 55.4k | 4m 57s |
| implement | claude-sonnet-4-6 | 1 | 885 | 14.5k | 1.4M | 63.1k | 89 | 96.2k | 4m 53s |
| prepare_pr* | <synthetic> | 3 | 52 | 4.4k | 149.0k | 35.4k | 18 | 36.5k | 8m 5s |
| compose_pr | claude-sonnet-4-6 | 1 | 59 | 1.8k | 172.7k | 27.9k | 15 | 13.4k | 40s |
| review_pr | claude-sonnet-4-6 | 3 | 318 | 118.0k | 1.8M | 91.8k | 127 | 205.9k | 23m 45s |
| resolve_review | claude-sonnet-4-6 | 3 | 655 | 45.8k | 3.9M | 75.3k | 188 | 158.7k | 25m 15s |
| **Total** | | | 3.6k | 205.3k | 9.5M | 91.8k | | 617.1k | 1h 14m |

## Token Efficiency

| plan | 0 | — | — | — |
| verify | 0 | — | — | — |
| implement | 190 | 7455.4 | 506.3 | 76.4 |
| review_pr | 0 | — | — | — |
| resolve_review | 74 | 52231.9 | 2143.9 | 619.5 |
| **Total** | **264** | 36133.4 | 2337.6 | 777.5 |

### DIFF
--- a/src/autoskillit/server/_guards.py
+++ b/src/autoskillit/server/_guards.py
@@ -160,6 +160,39 @@ def _resolve_provider_profile(
     config_providers: ProvidersConfig,
 ) -> tuple[str, dict[str, str]]:
 
+    # Tier 0: per-recipe per-step override (highest specificity)
+    if recipe_name and step_name:
+        recipe_map = config_providers.recipe_overrides.get(recipe_name)
+        if recipe_map:
+            recipe_step_override = recipe_map.get(step_name)
+            if recipe_step_override:
+                logger.debug(
+                    "provider_profile_resolved",
+                    tier="recipe_step_override",
+                    profile=recipe_step_override,
+                )
+                if recipe_step_override == "anthropic":
+                    return ("anthropic", {})
+                return (
+                    recipe_step_override,
+                    config_providers.profiles.get(recipe_step_override, {}),
+                )
+
+    # Tier 0W: per-recipe wildcard override
+    if recipe_name:
+        recipe_map = config_providers.recipe_overrides.get(recipe_name)
+        if recipe_map:
+            recipe_wildcard = recipe_map.get("*")
+            if recipe_wildcard:
+                logger.debug(
+                    "provider_profile_resolved",
+                    tier="recipe_wildcard_override",
+                    profile=recipe_wildcard,
+                )
+                if recipe_wildcard == "anthropic":
+                    return ("anthropic", {})
+                return (recipe_wildcard, config_providers.profiles.get(recipe_wildcard, {}))
+
     # Tier 1: per-step config override (requires recipe context)
     if recipe_name and step_name:
         step_override = config_providers.step_overrides.get(step_name)
--- a/src/autoskillit/server/_guards.py
+++ b/src/autoskillit/server/_guards.py
@@ -154,6 +154,15 @@ def _check_dry_walkthrough(skill_command: str, cwd: str) -> str | None:
     return None
 
 
+def _provider_result(
+    provider: str,
+    profiles: dict[str, dict[str, str]],
+) -> tuple[str, dict[str, str]]:
+    if provider == "anthropic":
+        return ("anthropic", {})
+    return (provider, profiles.get(provider, {}))
+
+
 def _resolve_provider_profile(
     step_name: str,
     recipe_name: str,
@@ -173,12 +182,7 @@ def _resolve_provider_profile(
                         tier="recipe_step_override",
                         profile=recipe_step_override,
                     )
-                    if recipe_step_override == "anthropic":
-                        return ("anthropic", {})
-                    return (
-                        recipe_step_override,
-                        config_providers.profiles.get(recipe_step_override, {}),
-                    )
+                    return _provider_result(recipe_step_override, config_providers.profiles)
 
             # Tier 0W: per-recipe wildcard override
             recipe_wildcard = recipe_map.get("*")
@@ -188,9 +192,7 @@ def _resolve_provider_profile(
                     tier="recipe_wildcard_override",
                     profile=recipe_wildcard,
                 )
-                if recipe_wildcard == "anthropic":
-                    return ("anthropic", {})
-                return (recipe_wildcard, config_providers.profiles.get(recipe_wildcard, {}))
+                return _provider_result(recipe_wildcard, config_providers.profiles)
 
     # Tier 1: per-step config override (requires recipe context)
     if recipe_name and step_name:
--- a/src/autoskillit/server/_guards.py
+++ b/src/autoskillit/server/_guards.py
@@ -160,28 +160,27 @@ def _resolve_provider_profile(
     config_providers: ProvidersConfig,
 ) -> tuple[str, dict[str, str]]:
 
-    # Tier 0: per-recipe per-step override (highest specificity)
-    if recipe_name and step_name:
-        recipe_map = config_providers.recipe_overrides.get(recipe_name)
-        if recipe_map:
-            recipe_step_override = recipe_map.get(step_name)
-            if recipe_step_override:
-                logger.debug(
-                    "provider_profile_resolved",
-                    tier="recipe_step_override",
-                    profile=recipe_step_override,
-                )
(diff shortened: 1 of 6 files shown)


src/autoskillit/server/_guards.py
```diff
@@ -160,6 +160,38 @@ def _resolve_provider_profile(
...
+            # Tier 0: per-recipe per-step override (highest specificity)
+            if step_name:
+                recipe_step_override = recipe_map.get(step_name)
+                if recipe_step_override:
+                    logger.debug(
+                        "provider_profile_resolved",
+                        tier="recipe_step_override",
+                        profile=recipe_step_override,
+                    )
+                    if recipe_step_override == "anthropic":
```
Trecek: [warning] cohesion: The 'anthropic' short-circuit (returning ('anthropic', {}) instead of looking up profiles) is duplicated at Tier 0 and Tier 0W (L176-L177 and L191-L192). If this special-case ever changes — e.g., to pass a non-empty env dict — both branches must be updated in sync. Centralize into a shared helper or post-lookup normalisation step.

src/autoskillit/server/_guards.py
```diff
@@ -160,6 +160,38 @@ def _resolve_provider_profile(
...
+                        return ("anthropic", {})
+                    return (
+                        recipe_step_override,
+                        config_providers.profiles.get(recipe_step_override, {}),
+                    )
+
+            # Tier 0W: per-recipe wildcard override
+            recipe_wildcard = recipe_map.get("*")
```
Trecek, line 188: [warning] arch: The Tier 0W wildcard lookup (recipe_map.get('*')) executes even when step_name is empty, because only the Tier 0 per-step branch guards on 'if step_name'. This means a recipe-scoped wildcard override fires for calls with no step context, which may be unintended. Either document that wildcards apply regardless of step_name, or add a matching 'if step_name' guard before the wildcard branch for consistency.

tests/config/test_providers_config.py
```diff
@@ -97,6 +100,18 @@ def test_providers_config_profiles_non_string_value_raises(self) -> None:
         with pytest.raises(ValueError, match=r"profiles\[.+\]\[.+\] must be a string"):
             ProvidersConfig(profiles={"my_profile": {"model": 42}})  # type: ignore[arg-type]
 
+    def test_recipe_overrides_non_string_value_raises(self) -> None:
+        from autoskillit.config.settings import ProvidersConfig
+
+        with pytest.raises(ValueError, match="recipe_overrides"):
```
Trecek: [warning] tests: pytest.raises(ValueError, match="recipe_overrides") is too weak — it matches any error containing the substring. The existing profiles test uses the specific pattern r"profiles\[.+\]\[.+\] must be a string". Use r"recipe_overrides\[.+\]\[.+\] must be a string" to ensure the correct error branch fires.

tests/config/test_providers_config.py
```diff
@@ -97,6 +100,18 @@ def test_providers_config_profiles_non_string_value_raises(self) -> None:
...
+            ProvidersConfig(recipe_overrides={"remediation": {"implement": 42}})  # type: ignore[arg-type]
+
+    def test_recipe_overrides_non_dict_inner_value_raises(self) -> None:
+        from autoskillit.config.settings import ProvidersConfig
+
+        with pytest.raises(ValueError, match="recipe_overrides"):
```
Trecek: [warning] tests: pytest.raises(ValueError, match="recipe_overrides") is too weak for the non-dict inner value test. Use r"recipe_overrides\[.+\] must be a dict" to match the exact error message produced by the __post_init__ guard, consistent with the specificity of the profiles test.

tests/server/test_resolve_provider_profile.py
```diff
@@ -100,3 +100,103 @@ def test_no_recipe_name_skips_step_override():
...
+
+    cfg = _make_config(
+        recipe_overrides={"remediation": {"implement": "vertex"}},
+        profiles={"vertex": {"K": "V"}},
+    )
+    result = _resolve_provider_profile("implement", "", cfg)
+    assert result == ("implement", {})
+
+
+def test_recipe_override_requires_step_name():
```
Trecek, line 182: [warning] tests: test_recipe_override_requires_step_name does not isolate the step_name guard: the config has no wildcard ('*') entry, so the wildcard branch is absent rather than bypassed. Adding a '*' key to recipe_overrides would reveal whether step_name='' correctly skips the wildcard check or falls through it — that distinction is what the test claims to verify per the comment at L191.

Trecek (commented):

Trecek (commented):

src/autoskillit/server/_guards.py
Trecek: Agreed — fixed in a5ab3817. Extracted `_provider_result(provider, profiles)` helper used at both Tier 0 and Tier 0W, eliminating the duplicate `if provider == 'anthropic': return ('anthropic', {})` branches.

src/autoskillit/server/_guards.py
Trecek: Valid observation — flagged for design decision. Tier 0W wildcard (`recipe_map.get('*')`) fires regardless of step_name because only Tier 0 has the `if step_name:` guard. Whether a recipe-scoped wildcard should apply to all calls within a recipe (including those without step context) or only to calls with an explicit step_name is a design question requiring human input.

tests/config/test_providers_config.py
Trecek: Agreed — fixed in 6ae45055. Updated to `match=r"recipe_overrides\[.+\]\[.+\] must be a string"` matching the exact error from `ProvidersConfig.__post_init__`: `recipe_overrides[{recipe!r}][{step!r}] must be a string`.

tests/config/test_providers_config.py
Trecek: Agreed — fixed in 6ae45055. Updated to `match=r"recipe_overrides\[.+\] must be a dict"` matching the exact error from `ProvidersConfig.__post_init__`: `recipe_overrides[{recipe!r}] must be a dict`.

tests/server/test_resolve_provider_profile.py
Trecek: Valid observation — flagged for design decision. Adding `'*': 'vertex'` to the test config would change the assertion because the current Tier 0W wildcard fires for empty step_name. The fix depends on the design decision for the wildcard behavior (see thread on `_guards.py:184`).

PR #2597, opened 2026-05-18
Rectify: Telemetry PR Body Section Contract

Trecek:
## Summary

The token telemetry subsystem has no shared contract defining which sections belong in a PR body. Four independent writers (`token_summary_hook.py`, `patch_pr_token_summary()`, `get_token_summary()`, `write_telemetry_files()`) each independently decide which `TelemetryFormatter.format_*` methods to call and how to assemble them. When `format_model_table` was added in PR #1961, only the hook and `get_token_summary` were updated — `patch_pr_token_summary()` was missed, leaving stale `## Model Usage Breakdown` data on every pipeline PR. The fix introduces a **section registry constant**, a **canonical assembly function**, and a **full-assembly parity test** so that adding a new telemetry section to the formatter automatically propagates to all writers or fails CI.

### REQ-FIX-1: Extend `patch_pr_token_summary()` to generate and replace `## Model Usage Breakdown`

After `total = token_log.compute_total(**scope_kwargs)`, add:
```python
model_totals = token_log.compute_model_totals(**scope_kwargs)
```

After `efficiency = TelemetryFormatter.format_efficiency_table(steps, total)`, add:
```python
model_table = TelemetryFormatter.format_model_table(model_totals)
```

Change the `combined =` line to:
```python
combined = table + ("\n\n" + efficiency if efficiency else "") + ("\n\n" + model_table if model_table else "")
```

Both `DefaultTokenLog` and `TelemetryFormatter` are already imported. `compute_model_totals(**scope_kwargs)` is compatible with the existing scoping pattern. `format_model_table` returns `""` for empty/unknown-only data, so it is safe to call unconditionally.

### REQ-FIX-2: Expand `section_re` to consume `## Model Usage Breakdown`

Add `(?:\n## Model Usage Breakdown\n.*?)?` as a third optional arm before the lookahead:
```python
section_re = re.compile(
    r"\n*## Token Usage Summary\n.*?(?:\n## Token Efficiency\n.*?)?(?:\n## Model Usage Breakdown\n.*?)?(?=\n## |\Z)",
    re.DOTALL,
)
```

This assumes canonical ordering (Token Efficiency before Model Breakdown), enforced by the hook's write order at `token_summary_hook.py:470-474`.

### REQ-FIX-3: Add test coverage

- Add `model_identifier` to test session data in `_write_test_sessions` so `compute_model_totals` returns real model entries
- Add a test for three-section replacement: existing body contains Token Usage Summary + Token Efficiency + Model Usage Breakdown, verify all three are replaced with complete data
- Add a test for two-section replacement without Token Efficiency: Token Usage Summary + Model Usage Breakdown (no LoC data)

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260518-133634-508324/.autoskillit/temp/rectify/rectify_telemetry_section_contract_2026-05-18_134500.md`

## Token Usage Summary

| rectify | claude-sonnet-4-6 | 1 | 3.0k | 10.6k | 722.3k | 88.3k | 112 | 56.8k | 6m 26s |
| dry_walkthrough | claude-sonnet-4-6 | 1 | 1.2k | 6.7k | 859.7k | 76.1k | 117 | 52.7k | 4m 24s |
| implement | claude-opus-4-6 | 1 | 79 | 18.3k | 3.0M | 90.3k | 155 | 76.9k | 7m 3s |
| prepare_pr* | MiniMax-M2.7-highspeed | 1 | 99.3k | 3.9k | 209.7k | 30.0k | 23 | 42.7k | 1m 9s |
| compose_pr* | MiniMax-M2.7-highspeed | 1 | 49.7k | 1.7k | 206.8k | 30.0k | 14 | 15.4k | 40s |
| review_pr | claude-sonnet-4-6 | 3 | 296 | 108.5k | 1.7M | 96.4k | 132 | 211.8k | 25m 5s |
| resolve_review | claude-sonnet-4-6 | 3 | 1.1k | 95.6k | 9.4M | 129.0k | 340 | 276.2k | 43m 42s |
| **Total** | | | 154.6k | 245.4k | 16.1M | 129.0k | | 732.5k | 1h 28m |

## Token Efficiency

| rectify | 0 | — | — | — |
| implement | 353 | 8440.1 | 217.9 | 51.8 |
| review_pr | 0 | — | — | — |
| resolve_review | 142 | 66140.2 | 1944.9 | 673.2 |
| **Total** | **495** | 32429.4 | 1479.9 | 495.7 |


tests/pipeline/test_telemetry_formatter.py
```diff
@@ -1217,3 +1217,98 @@ def test_non_anthropic_step_marked_compact_kv() -> None:
...
+    # Only those methods are PR-bound; other formatters (e.g. format_timing_table) are not.
+    pr_block_source = textwrap.dedent(
+        inspect.getsource(TelemetryFormatter.format_pr_telemetry_block)
+    )
+    pr_block_tree = ast.parse(pr_block_source)
+    called_methods: set[str] = set()
+    for node in ast.walk(pr_block_tree):
+        if isinstance(node, ast.Attribute) and isinstance(node.ctx, ast.Load):
+            if node.attr.startswith("format_") and node.attr != "format_pr_telemetry_block":
+                called_methods.add(node.attr)
```
Trecek, line 1299: [critical] arch: Indentation error: `called_methods.add(node.attr)` is at the same indentation level as the `if` guard on L1298, not inside it. The inner `if node.attr != 'format_pr_telemetry_block':` guard is effectively an empty block (SyntaxError) — or if tolerated, `called_methods.add` runs for every `format_*` attribute unconditionally, defeating the exclusion guard.

src/autoskillit/smoke_utils.py
```diff
@@ -8,10 +8,21 @@
...
 from pathlib import Path
 
-from autoskillit.core import DISPATCH_ID_ENV_VAR, get_logger
+import regex as _regex
+
+from autoskillit.core import DISPATCH_ID_ENV_VAR, PR_TELEMETRY_SECTIONS, get_logger
 
 logger = get_logger(__name__)
 
+_PR_SECTION_RE = _regex.compile(
```
Trecek, line 20: [warning] arch: _PR_SECTION_RE is built at import time treating PR_TELEMETRY_SECTIONS[0] as a mandatory anchor while PR_TELEMETRY_SECTIONS[1:] become optional. If the constant is reordered or extended, the regex silently changes behavior. Build uniformly across all entries or add an assert on tuple length.

tests/pipeline/test_telemetry_formatter.py
```diff
@@ -1217,3 +1217,98 @@ def test_non_anthropic_step_marked_compact_kv() -> None:
...
+    ]
+    total_with_loc = {**_TOTAL, "loc_insertions": 250, "loc_deletions": 40}
+    result = TelemetryFormatter.format_pr_telemetry_block(
+        steps_with_loc, total_with_loc, model_totals
+    )
+    positions = [result.index(section) for section in PR_TELEMETRY_SECTIONS]
+    assert positions == sorted(positions), "Sections are not in PR_TELEMETRY_SECTIONS order"
+
+
+def test_pr_telemetry_sections_exhaustive() -> None:
```
Trecek, line 1281: [warning] tests: test_pr_telemetry_sections_exhaustive uses inspect.getsource + AST walk to discover sub-formatters and find '## ' headers. Brittle: silently misses transitively-called helpers and false-positives on non-header '## ' string constants. The L1299 indentation bug also breaks the called_methods filtering, making the test audit incorrect.

tests/test_smoke_utils.py
```diff
@@ -741,6 +743,90 @@ def test_pts_reads_order_id_from_dispatch_env(
...
+            },
+        ],
+    )
+    mock_run.side_effect = _make_gh_mock(get_body=existing_body)
+    result = patch_pr_token_summary(PR_URL, cwd="/clone/test", log_dir=str(tmp_path))
+    assert result["success"] == "true"
+    patch_call = mock_run.call_args_list[-1]
+    body_arg = next((a for a in patch_call[0][0] if a.startswith("body=")), None)
+    assert body_arg is not None, "body= argument not found in gh patch call"
+    assert "## Token Usage Summary" in body_arg
```
Trecek, line 783: [warning] tests: T_PTS12 asserts all three headers appear once in body_arg but does not assert that the old section content ('old', 'old eff', 'old model') was removed. The test would pass if the new sections were appended rather than replaced, as long as duplicates happen to produce count==1 failures elsewhere.

src/autoskillit/pipeline/telemetry_fmt.py
```diff
@@ -422,6 +422,22 @@ def format_efficiency_table(steps: list[dict], total: dict) -> str:
         )
         return "\n".join(lines)
 
+    @staticmethod
+    def format_pr_telemetry_block(
+        steps: list[dict],
```
Trecek, line 427: [warning] defense: Public static method format_pr_telemetry_block uses unparameterized `list[dict]` and `dict` for steps and total. Use `list[dict[str, Any]]` / `dict[str, Any]` or TypedDicts for precise contracts at this public boundary.

src/autoskillit/smoke_utils.py
```diff
@@ -323,14 +331,8 @@ def patch_pr_token_summary(
...
-    # Match from "## Token Usage Summary" through an optional "## Token Efficiency"
-    # block, stopping at the next "## " heading or end-of-string.
-    section_re = re.compile(
-        r"\n*## Token Usage Summary\n.*?(?:\n## Token Efficiency\n.*?)?(?=\n## |\Z)",
-        re.DOTALL,
-    )
-    if section_re.search(current_body):
-        new_body = section_re.sub("\n\n" + combined, current_body)
+    if _PR_SECTION_RE.search(current_body):
+        new_body = _PR_SECTION_RE.sub("\n\n" + combined, current_body)
```
Trecek: [warning] defense: _PR_SECTION_RE.sub replaces only the first match. If the PR body somehow contains duplicate telemetry blocks, stale sections accumulate. Pass count=1 explicitly to document intent and prevent silent double-append on repeated calls.

Trecek (commented):

Trecek (commented):
AutoSkillit review: 1 blocking issue found (indentation error in test). See inline comments. Verdict: changes_requested.

Trecek (commented):
**Accumulated observation from local review round 0 (src/autoskillit/smoke_utils.py ~line 325):**

Section header strings ('## Token Usage Summary', '## Token Efficiency', '## Model Usage Breakdown') are hardcoded inline in the regex rather than referencing PR_TELEMETRY_SECTIONS. If headers change in PR_TELEMETRY_SECTIONS, this regex will silently diverge and miss or double-insert sections.

**Evidence:** PR_TELEMETRY_SECTIONS in _type_constants.py contains exactly the three strings hardcoded in the regex. The commit (8ad25b4f) states PR_TELEMETRY_SECTIONS is the SOT but section_re still hardcodes them. Refactoring non-trivial: regex uses \n* prefix, optional groups, lookahead — can't be mechanically built from a flat tuple of headers. Design decision needed on whether to build a regex-from-constants generator or accept the duplication.

src/autoskillit/smoke_utils.py
Trecek: Agreed — fixed in 5688a76f43487a59932145e776d5014b9bb4cf6a. Added `assert len(PR_TELEMETRY_SECTIONS) == 3` before the `_PR_SECTION_RE` compile call so any future reordering or extension fails immediately at import time rather than silently changing regex behavior.

src/autoskillit/smoke_utils.py
Trecek: Agreed — fixed in 5688a76f43487a59932145e776d5014b9bb4cf6a. Added `count=1` to `_PR_SECTION_RE.sub` to explicitly document the replace-first intent and prevent stale-section accumulation if duplicate telemetry blocks ever appear in a PR body.

tests/pipeline/test_telemetry_formatter.py
Trecek: Investigated — this is intentional. The indentation is correct: `called_methods.add(node.attr)` (line 1299) is at 16 spaces (4 indentation levels: function body → for body → first-if body → second-if body), while the `if node.attr != 'format_pr_telemetry_block':` guard (line 1298) is at 12 spaces. The exclusion guard is working as intended. The diff hunk's leading '+' prefix can make indentation appear shallower than it is.

tests/test_smoke_utils.py
Trecek: Investigated — this is intentional. The test already asserts `body_arg.count("## Token Usage Summary") == 1`, `.count("## Token Efficiency") == 1`, and `.count("## Model Usage Breakdown") == 1` (lines 786–788). Since `existing_body` contains each header exactly once, an append-instead-of-replace bug would produce a count of 2, failing all three assertions. The test does guard against the append scenario.

tests/pipeline/test_telemetry_formatter.py
Trecek: Valid observation — flagged for design decision. The brittleness concern about transitively-called helpers and false-positive '## ' string constants is real. Note: the secondary premise (L1299 indentation bug breaks filtering) does not hold — the indentation is correct. Fixing the exhaustive check would require running the actual formatters and scanning their output, rather than static AST analysis. This is a design choice on whether to invest in a runtime-based check.

src/autoskillit/pipeline/telemetry_fmt.py
Trecek: Valid observation — flagged for design decision. Using `list[dict[str, Any]]` / `dict[str, Any]` at this public boundary is correct. However, other methods in `TelemetryFormatter` (e.g., `format_efficiency_table`) also use bare `list[dict]` / `dict`. Changing only `format_pr_telemetry_block` in isolation would create annotation inconsistency within the class. A decision on whether to upgrade all method signatures holistically is needed.

PR #2630, opened 2026-05-20
Rectify: Recipe Validation Version Skew Immunity

Trecek:
## Summary

The recipe validation pipeline has a structural asymmetry: YAML data hot-reloads via mtime-based `_LOAD_CACHE` invalidation, but Python rule logic (`_RULE_REGISTRY`, `_BLOCK_RULE_REGISTRY`) and `lru_cache`-pinned helpers are frozen at import time for the process lifetime. When on-disk code changes while the MCP server is running, freshly-read YAML is evaluated against stale rules, producing false validation errors that block fleet dispatch.

The fix applies the hook registry drift detection pattern — already proven in `hook_registry.py` + `_lifespan.py` — to the rule registry: a content hash of `_RULE_REGISTRY` is embedded in `_LoadCacheEntry` and validated on every cache hit. A complementary process-level staleness guard compares `pkg_root()` directory mtime at call time against the mtime captured at module load time, catching all forms of code drift. Four new tests and an arch test enforce the guarantees.

## Requirements

- **Problem Summary:** Recipe YAML files hot-reload from disk via mtime-based cache invalidation, but Python validation rules are loaded once at import time and never reloaded. When a commit changes both recipe YAML and validation rules while the MCP server is running, only the YAML change is picked up — the rule change requires a server restart. This causes false validation errors that block dispatches.
- **Observed Failure:** Commit `d4002070c` added `route_ci_applicable` to `implementation.yaml` and extended `rules_ci.py` to recognize `action: route` steps. The MCP server running v0.10.152 re-read the YAML (now v0.10.158 content) but evaluated it against v0.10.152 `rules_ci.py` (no `_CI_APPLICABLE_RE`), resulting in 3 `enqueue-missing-ci-gate` errors blocking all further dispatches.
- **Test Suite Blind Spot:** All tests run at HEAD — no test simulates cross-version evaluation where cached YAML is validated by stale rules.
- **Related Risks:** `_block_budgets()`, `load_bundled_manifest()`, and `load_ml_sub_area_folding()` are `@lru_cache`-pinned helpers reading YAML that could change on disk.

## Implementation Plan

Plan file: `.autoskillit/temp/rectify/rectify_recipe_validation_version_skew_immunity_2026-05-20_083554.md`

## Changed Files

**Modified:**
- `src/autoskillit/recipe/__init__.py` — call `_finalize_registry()` after all rule imports
- `src/autoskillit/recipe/_api.py` — add `rule_registry_hash` to `_LoadCacheEntry`, cache hit validation, process staleness guard, lru_cache cleanup
- `src/autoskillit/recipe/registry.py` — add `compute_rule_registry_hash()`, `RULE_REGISTRY_HASH`, `_finalize_registry()`, finalization guard in decorators
- `tests/arch/test_recipe_rule_registration.py` — arch tests for rule registry hash presence
- `tests/recipe/test_api.py` — tests for cache invalidation on rule change, staleness detection, lru_cache clearing
- `tests/recipe/test_rules_registry.py` — tests for hash stability and change detection

## Token Usage Summary

| investigate | claude-sonnet-4-6 | 1 | 49 | 10.6k | 1.2M | 100.6k | 248 | 84.7k | 10m 6s |
| rectify | claude-sonnet-4-6 | 1 | 53 | 12.0k | 908.0k | 81.1k | 112 | 72.4k | 7m 1s |
| dry_walkthrough | claude-opus-4-6 | 1 | 1.7k | 21.8k | 2.8M | 87.4k | 145 | 72.6k | 9m 24s |
| implement | claude-opus-4-6 | 1 | 144 | 17.4k | 4.8M | 113.3k | 122 | 98.5k | 7m 37s |
| prepare_pr* | MiniMax-M2.7 | 1 | 66.9k | 3.4k | 245.7k | 0 | 20 | 29.2k | 2m 45s |
| compose_pr* | MiniMax-M2.7 | 1 | 104.3k | 3.0k | 313.9k | 26.3k | 27 | 2.9k | 2m 25s |
| review_pr | claude-opus-4-6 | 3 | 117 | 88.5k | 1.9M | 85.6k | 100 | 188.4k | 20m 56s |
| resolve_review | claude-opus-4-6 | 3 | 183 | 47.1k | 6.6M | 105.1k | 197 | 209.8k | 21m 19s |
| **Total** | | | 173.4k | 203.8k | 18.7M | 113.3k | | 758.5k | 1h 21m |

## Token Efficiency

| rectify | 0 | — | — | — |
| implement | 321 | 14887.4 | 306.9 | 54.3 |
| review_pr | 0 | — | — | — |
| resolve_review | 98 | 67643.8 | 2140.6 | 480.1 |
| **Total** | **419** | 44703.4 | 1810.2 | 486.4 |

## Model Usage Breakdown

| claude-sonnet-4-6 | 2 | 102 | 22.6k | 2.1M | 157.1k | 17m 8s |
| claude-opus-4-6 | 4 | 2.1k | 174.8k | 16.1M | 569.3k | 59m 18s |
| MiniMax-M2.7 | 2 | 171.2k | 6.4k | 559.5k | 32.1k | 5m 10s |

### DIFF
--- a/src/autoskillit/recipe/__init__.py
+++ b/src/autoskillit/recipe/__init__.py
@@ -7,6 +7,7 @@ from autoskillit.core import get_logger
 logger = get_logger(__name__)
 
 # Rule registration — import triggers @semantic_rule registration.
+from autoskillit.recipe import registry as _reg  # noqa: E402, PLC0415
 from autoskillit.recipe._api import (  # noqa: E402
     format_recipe_list_response,
     list_all,
@@ -158,6 +159,9 @@ from autoskillit.recipe.validator import (  # noqa: E402
     run_semantic_rules,
     validate_recipe_structure,
 )
+
+_reg._finalize_registry()  # pyright: ignore[reportAttributeAccessIssue]
+del _reg
 
 __all__ = [
     "GROUP_LABELS",
--- a/src/autoskillit/recipe/_api.py
+++ b/src/autoskillit/recipe/_api.py
@@ -180,7 +180,8 @@ def _clear_stale_caches() -> None:
     _block_budgets.cache_clear()
     load_bundled_manifest.cache_clear()
     load_ml_sub_area_folding.cache_clear()
-    _LOAD_CACHE.clear()
+    with _LOAD_CACHE_LOCK:
+        _LOAD_CACHE.clear()
 
 
 def format_recipe_list_response(result: LoadResult[RecipeInfo]) -> dict[str, object]:
--- a/src/autoskillit/recipe/_api.py
+++ b/src/autoskillit/recipe/_api.py
@@ -142,8 +142,8 @@ _LOAD_CACHE: dict[tuple, _LoadCacheEntry] = {}
 _LOAD_CACHE_LOCK = threading.Lock()
 
 _PROCESS_START_PKG_MTIME: int | None = None
-_staleness_last_check: float = 0.0
-_staleness_is_stale: bool = False
+_STALENESS_LAST_CHECK: float = 0.0
+_STALENESS_IS_STALE: bool = False
 _STALENESS_TTL: float = 30.0
 
 
@@ -156,17 +156,17 @@ def _get_process_start_mtime() -> int:
 
 def _check_process_staleness() -> bool:
     """Return True if the package directory was modified after process start."""
-    global _staleness_last_check, _staleness_is_stale  # noqa: PLW0603
+    global _STALENESS_LAST_CHECK, _STALENESS_IS_STALE  # noqa: PLW0603
     now = time.monotonic()
-    if now - _staleness_last_check < _STALENESS_TTL:
-        return _staleness_is_stale
-    _staleness_last_check = now
+    if now - _STALENESS_LAST_CHECK < _STALENESS_TTL:
+        return _STALENESS_IS_STALE
+    _STALENESS_LAST_CHECK = now
     try:
-        _staleness_is_stale = _path_mtime_ns(pkg_root()) != _get_process_start_mtime()
+        _STALENESS_IS_STALE = _path_mtime_ns(pkg_root()) != _get_process_start_mtime()
     except (OSError, RuntimeError):
         logger.warning("pkg_root() unavailable during staleness check; assuming non-stale")
-        _staleness_is_stale = False
-    return _staleness_is_stale
+        _STALENESS_IS_STALE = False
+    return _STALENESS_IS_STALE
 
 
 def _clear_stale_caches() -> None:
--- a/src/autoskillit/recipe/_api.py
+++ b/src/autoskillit/recipe/_api.py
@@ -134,11 +134,46 @@ class _LoadCacheEntry:
     project_dir_mtime: int
     builtin_dir_mtime: int
     pkg_version: str
+    rule_registry_hash: str
     result: LoadRecipeResult
 
 
 _LOAD_CACHE: dict[tuple, _LoadCacheEntry] = {}
 _LOAD_CACHE_LOCK = threading.Lock()
+
+# ---------------------------------------------------------------------------
+# Process staleness guard
+# ---------------------------------------------------------------------------
+
+_PROCESS_START_PKG_MTIME: int = _path_mtime_ns(pkg_root())
+_staleness_last_check: float = 0.0
+_staleness_is_stale: bool = False
+_STALENESS_TTL: float = 30.0
+
+
+def _check_process_staleness() -> bool:
+    """Return True if the package directory was modified after process start."""
+    global _staleness_last_check, _staleness_is_stale  # noqa: PLW0603
+    now = time.monotonic()
+    if now - _staleness_last_check < _STALENESS_TTL:
+        return _staleness_is_stale
+    _staleness_last_check = now
+    _staleness_is_stale = _path_mtime_ns(pkg_root()) != _PROCESS_START_PKG_MTIME
+    return _staleness_is_stale
+
+
+def _clear_stale_caches() -> None:
+    """Clear all lru_cache helpers and the load cache when staleness is detected."""
+    from autoskillit.recipe.contracts import load_bundled_manifest  # noqa: PLC0415
+    from autoskillit.recipe.methodology_venue_appendix import (  # noqa: PLC0415
+        load_ml_sub_area_folding,
+    )
+    from autoskillit.recipe.rules.rules_blocks import _block_budgets  # noqa: PLC0415
+
+    _block_budgets.cache_clear()
+    load_bundled_manifest.cache_clear()
+    load_ml_sub_area_folding.cache_clear()
+    _LOAD_CACHE.clear()
 
 
 def format_recipe_list_response(result: LoadResult[RecipeInfo]) -> dict[str, object]:
@@ -357,6 +392,16 @@ def load_and_validate(
         {"content": str, "suggestions": list, "valid": bool}
         On not-found: {"error": str}
     """
+    if _check_process_staleness():
+        _clear_stale_caches()
+        return {
+            "error": (
+                "Process is running stale code — package directory was modified on disk "
+                "since server startup. Restart the MCP server via reload_session."
+            ),
+            "valid": False,
+        }
+
     _pdir = project_dir if project_dir is not None else Path.cwd()
     pkg_version = _get_pkg_version()
     project_recipes_dir = _pdir / ".autoskillit" / "recipes"
@@ -390,7 +435,14 @@ def load_and_validate(
     with _LOAD_CACHE_LOCK:
         cached = _LOAD_CACHE.get(cache_key)
 
-    if cached is not None and cached.pkg_version == pkg_version:
+    from autoskillit.recipe import registry as _registry  # noqa: PLC0415
+
+    _rule_hash: str = _registry.RULE_REGISTRY_HASH  # pyright: ignore[reportAttributeAccessIssue]
+    if (
+        cached is not None
+        and cached.pkg_version == pkg_version
+        and cached.rule_registry_hash == _rule_hash
+    ):
         pm = _path_mtime_ns(project_recipes_dir)
         bm = _path_mtime_ns(_builtin_dir)
         rm = _path_mtime_ns(cached.recipe_path)
@@ -619,6 +671,7 @@ def load_and_validate(
             project_dir_mtime=_path_mtime_ns(project_recipes_dir),
             builtin_dir_mtime=_path_mtime_ns(_builtin_dir),
             pkg_version=pkg_version,
+            rule_registry_hash=_rule_hash,
             result=result,
         )
         with _LOAD_CACHE_LOCK:
--- a/src/autoskillit/recipe/_api.py
+++ b/src/autoskillit/recipe/_api.py
@@ -450,6 +450,8 @@ def load_and_validate(
     from autoskillit.recipe import registry as _registry  # noqa: PLC0415
 
     _rule_hash: str = _registry.RULE_REGISTRY_HASH  # pyright: ignore[reportAttributeAccessIssue]
+    if not _rule_hash:
+        logger.warning("RULE_REGISTRY_HASH is empty — _finalize_registry() was never called")
     if (
         cached is not None
         and cached.pkg_version == pkg_version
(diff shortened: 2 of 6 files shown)


Trecek (commented):

src/autoskillit/recipe/registry.py
```diff
@@ -86,6 +88,10 @@ def semantic_rule(
     def decorator(
         fn: Callable[[ValidationContext], list[RuleFinding]],
     ) -> Callable[[ValidationContext], list[RuleFinding]]:
+        if _REGISTRY_FINALIZED:
```
Trecek, line 91: [warning] defense: The idempotency guard `if any(r.name == name for r in _RULE_REGISTRY): return fn` silently swallows re-registration of a rule with the *same name but different behavior* (e.g., reloaded module). This can mask bugs where a stale rule function is retained. Consider logging a warning when a duplicate is silently accepted post-finalization.

src/autoskillit/recipe/registry.py
```diff
@@ -109,6 +115,10 @@ def block_rule(
     def decorator(
         fn: Callable[[BlockContext], list[RuleFinding]],
     ) -> Callable[[BlockContext], list[RuleFinding]]:
+        if _REGISTRY_FINALIZED:
```
Trecek, line 118: [warning] defense: Same silent idempotency issue for `block_rule`: re-registering a block rule with the same name after finalization returns the original function with no warning, hiding potential staleness or reload bugs.

src/autoskillit/recipe/_api.py
```diff
@@ -134,12 +134,55 @@ class _LoadCacheEntry:
...
 
+_PROCESS_START_PKG_MTIME: int | None = None
+_STALENESS_LAST_CHECK: float = 0.0
+_STALENESS_IS_STALE: bool = False
+_STALENESS_TTL: float = 30.0
+
+
+def _get_process_start_mtime() -> int:
+    global _PROCESS_START_PKG_MTIME  # noqa: PLW0603
+    if _PROCESS_START_PKG_MTIME is None:
```
Trecek, line 153: [warning] defense: `_get_process_start_mtime()` is not thread-safe: two threads can both see `_PROCESS_START_PKG_MTIME is None` and both call `_path_mtime_ns(pkg_root())`. If the filesystem mtime changes between the two reads, one thread captures a different baseline. Add a module-level lock or initialize at module import time.

src/autoskillit/recipe/_api.py
```diff
@@ -357,6 +400,16 @@ def load_and_validate(
         {"content": str, "suggestions": list, "valid": bool}
         On not-found: {"error": str}
     """
+    if _check_process_staleness():
```
Trecek, line 406: [warning] defense: `_clear_stale_caches()` is called unconditionally every time `_check_process_staleness()` returns True, which happens on every `load_and_validate` call for the stale process lifetime (after 30s TTL). Repeated cache clears with no benefit once stale is detected. The stale state should short-circuit without re-clearing on every invocation.

src/autoskillit/recipe/__init__.py
```diff
@@ -159,6 +160,9 @@
     validate_recipe_structure,
 )
 
+_reg._finalize_registry()  # pyright: ignore[reportAttributeAccessIssue]
```
Trecek, line 163: [warning] cohesion: `_reg._finalize_registry()` calls a private function across a module boundary with a `pyright: ignore` suppression. This is asymmetric with the public API surface — `_finalize_registry` should either be made public (removing the suppression) or the call site should be moved inside `registry.py` itself.

src/autoskillit/recipe/registry.py
```diff
@@ -117,6 +127,28 @@ def decorator(
     return decorator
 
 
+_REGISTRY_FINALIZED: bool = False
```
Trecek, line 130: [warning] defense: `_REGISTRY_FINALIZED` and `RULE_REGISTRY_HASH` are initialized to sentinel values (`False`/`""`). If `_finalize_registry()` is never called (e.g., import order changes), `RULE_REGISTRY_HASH` will be empty and cache entries built against it will never match post-finalization entries, causing every lookup to miss. No guard in `load_and_validate` asserts non-empty hash.

src/autoskillit/recipe/_api.py
```diff
@@ -390,7 +443,14 @@ def load_and_validate(
     with _LOAD_CACHE_LOCK:
         cached = _LOAD_CACHE.get(cache_key)
 
-    if cached is not None and cached.pkg_version == pkg_version:
+    from autoskillit.recipe import registry as _registry  # noqa: PLC0415
```
Trecek, line 450: [warning] cohesion: Rule registry hash lookup via deferred local import inside `load_and_validate` is asymmetric with how staleness state is managed (module-level globals with dedicated accessors like `_get_process_start_mtime()`). Extract hash read to a small helper for consistency.

Trecek (commented):

src/autoskillit/recipe/_api.py
```diff
@@ -134,12 +134,55 @@ class _LoadCacheEntry:
...
+    if _PROCESS_START_PKG_MTIME is None:
+        _PROCESS_START_PKG_MTIME = _path_mtime_ns(pkg_root())
+    return _PROCESS_START_PKG_MTIME
+
+
+def _check_process_staleness() -> bool:
+    """Return True if the package directory was modified after process start."""
+    global _STALENESS_LAST_CHECK, _STALENESS_IS_STALE  # noqa: PLW0603
+    now = time.monotonic()
+    if now - _STALENESS_LAST_CHECK < _STALENESS_TTL:
```
Trecek, line 162: **Observation from local review round 0:**

_check_process_staleness() reads and writes _staleness_last_check, _staleness_is_stale, and _PROCESS_START_PKG_MTIME without any lock. Two threads can both observe the TTL expired and both compute staleness, causing redundant cache clears. Tolerable under CPython GIL but architecturally fragile.

**Evidence:** Globals at L148-151 are simple scalars; under CPython GIL worst case is idempotent redundant staleness checks. No data corruption. Design observation about thread safety patterns.

src/autoskillit/recipe/registry.py
Trecek, line 91: **Observation from local review round 0:**

The semantic_rule and block_rule decorators check _REGISTRY_FINALIZED and conditionally append without a lock. TOCTOU pattern at L91-L94. Low practical risk since rule imports are module-level.

**Evidence:** Module-level imports execute under CPython import lock. list.append is GIL-atomic. Re-registration tolerance added deliberately in c14bf3b3. No practical concurrency issue.

src/autoskillit/recipe/_api.py
```diff
@@ -357,6 +400,16 @@ def load_and_validate(
         {"content": str, "suggestions": list, "valid": bool}
         On not-found: {"error": str}
     """
```
Trecek, line 405: **Observation from local review round 0:**

_clear_stale_caches() does not reset _staleness_is_stale back to False. Staleness guard fires repeatedly for the full TTL period. Consider resetting or documenting the intentional always-error-until-restart design.

**Evidence:** _get_process_start_mtime() captures mtime at first call; once diverged, process IS stale and cannot become unstale without restart. Repeated _LOAD_CACHE.clear() is idempotent. Always-error-until-restart is correct design.

src/autoskillit/recipe/_api.py
```diff
@@ -134,12 +134,55 @@ class _LoadCacheEntry:
...
+        _STALENESS_IS_STALE = _path_mtime_ns(pkg_root()) != _get_process_start_mtime()
+    except (OSError, RuntimeError):
+        logger.warning("pkg_root() unavailable during staleness check; assuming non-stale")
+        _STALENESS_IS_STALE = False
+    return _STALENESS_IS_STALE
+
+
+def _clear_stale_caches() -> None:
+    """Clear all lru_cache helpers and the load cache when staleness is detected."""
+    from autoskillit.recipe.contracts import load_bundled_manifest  # noqa: PLC0415
```
Trecek, line 176: **Observation from local review round 1:**

Deferred intra-package imports inside _clear_stale_caches() import private symbol _block_budgets from rules.rules_blocks.

**Evidence:** _block_budgets is @lru_cache decorated; cache_clear() requires direct reference to the function object. A public clear_caches() API in rules_blocks would be cleaner but is a design decision.

src/autoskillit/recipe/registry.py
Trecek, line 130: **Observation from local review round 1:**

_REGISTRY_FINALIZED and RULE_REGISTRY_HASH are module-level mutable globals with no concurrency protection.

**Evidence:** Written once at import time by _finalize_registry(). After init, RULE_REGISTRY_HASH is read-only. CPython GIL makes scalar reads atomic. Low practical risk.

src/autoskillit/recipe/_api.py
```diff
@@ -390,7 +443,14 @@ def load_and_validate(
     with _LOAD_CACHE_LOCK:
         cached = _LOAD_CACHE.get(cache_key)
 
-    if cached is not None and cached.pkg_version == pkg_version:
+    from autoskillit.recipe import registry as _registry  # noqa: PLC0415
+
+    _rule_hash: str = _registry.RULE_REGISTRY_HASH  # pyright: ignore[reportAttributeAccessIssue]
```
Trecek, line 452: **Observation from local review round 1:**

RULE_REGISTRY_HASH is accessed with a pyright: ignore[reportAttributeAccessIssue] suppression.

**Evidence:** Suppression is a consequence of the deferred import pattern inside load_and_validate(). Removing requires resolving the import architecture (circular dependency between _api.py and registry.py).

src/autoskillit/recipe/registry.py
```diff
@@ -117,6 +127,28 @@ def decorator(
...
+RULE_REGISTRY_HASH: str = ""
```
Trecek, line 131: **Observation from local review round 1:**

RULE_REGISTRY_HASH breaks naming symmetry — public while managed by private _finalize_registry().

**Evidence:** All other module-level mutable state uses _prefix (_RULE_REGISTRY, _BLOCK_RULE_REGISTRY, _REGISTRY_FINALIZED). Adding a getter function or renaming to _RULE_REGISTRY_HASH is a design decision.

src/autoskillit/recipe/registry.py
Trecek, line 91: **Observation from local review round 1:**

Idempotency shortcut silently accepts re-registration of a rule whose callable may differ from the original.

**Evidence:** Deliberately added in c14bf3b3 to tolerate rule re-registration (module re-import). requires_decision flag set by reviewer. A duplicate-name-different-fn check would add safety but is a design decision.

Trecek (commented):
Observations accumulated from 2 local review rounds:

src/autoskillit/recipe/_api.py
Trecek: Agreed — fixed in 46d9aa1e. Added _STALENESS_CACHES_CLEARED flag to short-circuit repeated cache clearing after the first stale detection.

src/autoskillit/recipe/registry.py
Trecek: Agreed — fixed in 7f5a46f2. Added a warning log when RULE_REGISTRY_HASH is empty at cache-comparison time, guarding against silent cache mismatches if _finalize_registry() was never called.

src/autoskillit/recipe/registry.py
Trecek: Investigated — this is intentional. Commit c14bf3b3 explicitly added re-registration tolerance for the module-reimport scenario in tests. The guard silently returns fn because the reimported module produces a functionally identical callable.

src/autoskillit/recipe/registry.py
Trecek: Investigated — this is intentional. Same rationale as semantic_rule: commit c14bf3b3 added symmetric idempotency guards to both decorator types for module-reimport tolerance.

src/autoskillit/recipe/_api.py
Trecek: Investigated — this is intentional. CPython GIL makes int assignment atomic. Two threads both seeing None and racing to assign are idempotent — both capture the same pkg_root() mtime (directory mtime doesn't change between nanosecond-level reads). No corruption risk.

src/autoskillit/recipe/__init__.py
Trecek: Valid observation — flagged for design decision. Making _finalize_registry public would eliminate the pyright suppression and be consistent with __init__.py as the canonical lifecycle caller. Alternatively the pattern can be documented as intentional. Requires human decision.

src/autoskillit/recipe/_api.py
Trecek: Valid observation — flagged for design decision. A _get_rule_hash() helper would be symmetric with _get_process_start_mtime(). The deferred import was needed to avoid circular import at module level. Whether to extract a helper is a design choice.

PR #2764, opened 2026-05-23
fix: introduce SidecarReadResult to distinguish missing vs empty sidecar and add ORCHESTRATOR label sweep

Trecek:
## Summary

Introduce `SidecarReadResult` to distinguish "file missing" (unverifiable state) from "file genuinely empty" (verified no-op), so `cleanup_orphaned_labels` correctly returns `False` when a sidecar path is recorded but the file no longer exists on disk. Fix the boot-path asymmetry where `sweep_stale_dispatch_labels` ran only on FLEET session startup but not ORCHESTRATOR, leaving corrupted label records unswept in food-truck-only environments. Also fix `check_review_loop` to yield `had_blocking="false"` for `approved_with_comments` regardless of `local_review_rounds` position, and align L3 prompt timeout derivation with `fleet.default_timeout_sec`.

## Requirements

- REQ-REM-001: `cleanup_orphaned_labels` must return `False` when `sidecar_path` is set but the file does not exist on disk
- REQ-REM-002: `check_review_loop` must return `had_blocking="false"` when `previous_verdict` is `approved_with_comments`, regardless of `local_review_rounds` position. `approved` verdicts continue to force re-review until local rounds are exhausted (existing behavior, per docstring and test).
- REQ-REM-003: The L3 prompt timeout (`l3_timeout_sec`), `AUTOSKILLIT_SESSION_DEADLINE` env var, and subprocess kill timeout must all derive from the same source (`fleet.default_timeout_sec` or explicit `timeout_sec` override)
- REQ-REM-004: `sweep_stale_dispatch_labels` must run on ORCHESTRATOR session startup, not only FLEET sessions
- REQ-REM-005: Tests must cover: (a) missing-sidecar cleanup returns `False`, (b) `check_review_loop` with `approved_with_comments` + configured `local_review_rounds` + non-exhausted iteration, (c) L3 prompt timeout consistency with config default
- REQ-REM-006: `check_review_loop` docstring (`smoke_utils.py:183-185`) and recipe note (`implementation.yaml:699-701`) must be updated to reflect that `approved_with_comments` yields `had_blocking=false` unconditionally, not only when `local_review_rounds` is absent or exhausted

## Implementation Plan

Plan file: `.autoskillit/temp/rectify/rectify_sidecar_result_integrity_label_recovery_2026-05-22_220537.md`

## Token Usage Summary

| rectify | claude-sonnet-4-6 | 1 | 4.5k | 23.7k | 1.3M | 123.3k | 163 | 84.7k | 12m 52s |
| dry_walkthrough | claude-opus-4-6 | 2 | 113 | 21.0k | 3.1M | 98.1k | 179 | 162.8k | 10m 48s |
| implement | claude-sonnet-4-6 | 2 | 806 | 44.6k | 6.0M | 109.5k | 305 | 247.2k | 17m 10s |
| prepare_pr* | MiniMax-M2.7 | 1 | 75.9k | 3.6k | 379.8k | 29.8k | 25 | 42.8k | 1m 21s |
| compose_pr* | MiniMax-M2.7 | 1 | 55.2k | 1.9k | 264.6k | 0 | 20 | 0 | 51s |
| review_pr | claude-opus-4-6 | 3 | 129 | 84.6k | 2.2M | 95.2k | 107 | 219.6k | 23m 38s |
| resolve_review | claude-sonnet-4-6 | 2 | 798 | 65.9k | 6.5M | 121.6k | 248 | 160.0k | 24m 49s |
| **Total** | | | 137.4k | 245.3k | 19.7M | 123.3k | | 917.1k | 1h 31m |

## Token Efficiency

| rectify | 0 | — | — | — |
| implement | 443 | 13533.5 | 558.0 | 100.6 |
| review_pr | 0 | — | — | — |
| resolve_review | 60 | 108383.1 | 2666.0 | 1098.9 |
| **Total** | **503** | 39210.7 | 1823.2 | 487.7 |

## Model Usage Breakdown

| claude-sonnet-4-6 | 3 | 6.1k | 134.2k | 13.8M | 491.9k | 54m 52s |
| claude-opus-4-6 | 2 | 242 | 105.5k | 5.2M | 382.4k | 34m 27s |
| MiniMax-M2.7 | 2 | 131.1k | 5.6k | 644.4k | 42.8k | 2m 12s |

### DIFF
--- a/src/autoskillit/fleet/_api.py
+++ b/src/autoskillit/fleet/_api.py
@@ -48,6 +48,20 @@ logger = get_logger(__name__)
 ENVELOPE_STDERR_MAX = 2000
 
 
+def resolve_dispatch_timeout(
+    timeout_sec: int | None,
+    default_timeout_sec: int,
+) -> float:
+    """Resolve dispatch timeout to a concrete float.
+
+    Single source of truth for all three timeout surfaces (prompt, process kill,
+    session deadline). Uses ``is not None`` to correctly handle ``timeout_sec=0``.
+    """
+    if timeout_sec is not None:
+        return float(timeout_sec)
+    return float(default_timeout_sec)
+
+
 class CaptureCompletenessError(RuntimeError):
     """Raised when a capture spec extracts zero fields from the payload."""
 
@@ -757,13 +771,16 @@ async def _run_dispatch(
 
     dispatch_sidecar_path = str(compute_sidecar_path(dispatch_id, tool_ctx.project_dir))
 
+    resolved_timeout = resolve_dispatch_timeout(
+        timeout_sec, tool_ctx.config.fleet.default_timeout_sec
+    )
     prompt = prompt_builder(
         recipe=recipe,
         task=task,
         ingredients=effective_ingredients,
         dispatch_id=dispatch_id,
         campaign_id=campaign_id,
-        l3_timeout_sec=timeout_sec or 1800,
+        l3_timeout_sec=int(resolved_timeout),
         capture=capture,
     )
 
@@ -856,7 +873,7 @@ async def _run_dispatch(
                     project_dir=str(tool_ctx.project_dir),
                     marker_dir=marker_dir,
                     session_id=caller_session_id,
-                    timeout=float(timeout_sec) if timeout_sec else None,
+                    timeout=resolved_timeout,
                     idle_output_timeout=float(idle_output_timeout)
                     if idle_output_timeout is not None
                     else None,
@@ -864,14 +881,7 @@ async def _run_dispatch(
                         "AUTOSKILLIT_PROJECT_DIR": str(tool_ctx.project_dir),
                         "AUTOSKILLIT_CAMPAIGN_ID": campaign_id,
                         "AUTOSKILLIT_DISPATCH_ID": dispatch_id,
-                        "AUTOSKILLIT_SESSION_DEADLINE": str(
-                            started_at
-                            + (
-                                float(timeout_sec)
-                                if timeout_sec is not None
-                                else float(tool_ctx.config.fleet.default_timeout_sec)
-                            )
-                        ),
+                        "AUTOSKILLIT_SESSION_DEADLINE": str(started_at + resolved_timeout),
                     },
                     requires_packs=list(full_recipe.requires_packs) or ["kitchen-core"],
                     on_spawn=_on_spawn,
--- a/src/autoskillit/fleet/_api.py
+++ b/src/autoskillit/fleet/_api.py
@@ -903,7 +903,7 @@ async def _run_dispatch(
         from autoskillit.fleet._checkpoint_bridge import checkpoint_from_sidecar  # noqa: PLC0415
         from autoskillit.fleet.sidecar import read_sidecar_from_path  # noqa: PLC0415
 
-        sidecar_entries = read_sidecar_from_path(sidecar_file)
+        sidecar_entries = read_sidecar_from_path(sidecar_file).entries
         if sidecar_entries:
             dispatch_checkpoint = checkpoint_from_sidecar(sidecar_entries)
 
--- a/src/autoskillit/fleet/_label_cleanup.py
+++ b/src/autoskillit/fleet/_label_cleanup.py
@@ -16,7 +16,7 @@ from autoskillit.core import (
     get_logger,
 )
 from autoskillit.fleet._liveness import is_dispatch_session_alive
-from autoskillit.fleet.sidecar import read_sidecar_from_path
+from autoskillit.fleet.sidecar import SidecarReadStatus, read_sidecar_from_path
 from autoskillit.fleet.state import (
     TERMINAL_UNCLEANED_STATUSES,
     CampaignStateMutator,
@@ -58,7 +58,7 @@ async def cleanup_orphaned_labels(
         return True
 
     try:
-        entries = read_sidecar_from_path(Path(sidecar_path))
+        sidecar_result = read_sidecar_from_path(Path(sidecar_path))
     except Exception:
         logger.warning(
             "infra_label_cleanup_sidecar_read_failed",
@@ -67,11 +67,19 @@ async def cleanup_orphaned_labels(
         )
         return False
 
-    if not entries:
+    if sidecar_result.source != SidecarReadStatus.FOUND:
+        logger.warning(
+            "infra_label_cleanup_sidecar_unavailable",
+            sidecar_path=sidecar_path,
+            source=sidecar_result.source,
+        )
+        return False
+
+    if not sidecar_result.entries:
         return True
 
     all_succeeded = True
-    for entry in entries:
+    for entry in sidecar_result.entries:
         try:
             owner, repo, number = _parse_issue_ref(entry.issue_url)
         except ValueError:
--- a/src/autoskillit/fleet/sidecar.py
+++ b/src/autoskillit/fleet/sidecar.py
@@ -3,12 +3,19 @@ from __future__ import annotations
 import json
 from collections.abc import Sequence
 from dataclasses import asdict, dataclass
+from enum import StrEnum
 from pathlib import Path
-from typing import Literal
+from typing import Literal, NamedTuple
 
 from autoskillit.core import ensure_project_temp, get_logger
 
 logger = get_logger()
+
+
+class SidecarReadStatus(StrEnum):
+    FOUND = "found"
+    MISSING = "missing"
+    ERROR = "error"
 
 
 @dataclass(frozen=True, slots=True)
@@ -18,6 +25,11 @@ class IssueSidecarEntry:
     ts: str
     pr_url: str | None = None
     reason: str | None = None
+
+
+class SidecarReadResult(NamedTuple):
+    entries: list[IssueSidecarEntry]
+    source: SidecarReadStatus
 
 
 def sidecar_path(dispatch_id: str, project_dir: Path) -> Path:
@@ -58,13 +70,18 @@ def read_sidecar(dispatch_id: str, project_dir: Path) -> list[IssueSidecarEntry]
     return entries
 
 
-def read_sidecar_from_path(path: Path) -> list[IssueSidecarEntry]:
-    """Read and parse a sidecar JSONL at path; skips corrupt lines; returns [] on OSError."""
+def read_sidecar_from_path(path: Path) -> SidecarReadResult:
+    """Read and parse a sidecar JSONL at path.
+
+    Returns a SidecarReadResult with source indicating whether the file was
+    found, missing, or unreadable. Callers must inspect source to distinguish
+    'file missing' from 'file empty'.
+    """
     entries: list[IssueSidecarEntry] = []
     try:
         lines = path.read_text().splitlines()
     except OSError:
-        return []
+        return SidecarReadResult(entries=[], source=SidecarReadStatus.MISSING)
     for line in lines:
         line = line.strip()
         if not line:
@@ -83,7 +100,7 @@ def read_sidecar_from_path(path: Path) -> list[IssueSidecarEntry]:
         except (json.JSONDecodeError, KeyError) as exc:
             logger.debug("sidecar: skipping corrupt JSONL line", path=str(path), error=str(exc))
             continue
-    return entries
+    return SidecarReadResult(entries=entries, source=SidecarReadStatus.FOUND)
 
 
 def compute_remaining_issues(
--- a/src/autoskillit/fleet/sidecar.py
+++ b/src/autoskillit/fleet/sidecar.py
@@ -80,8 +80,11 @@ def read_sidecar_from_path(path: Path) -> SidecarReadResult:
     entries: list[IssueSidecarEntry] = []
     try:
         lines = path.read_text().splitlines()
-    except OSError:
+    except FileNotFoundError:
         return SidecarReadResult(entries=[], source=SidecarReadStatus.MISSING)
+    except OSError as exc:
+        logger.warning("sidecar: failed to read file", path=str(path), error=str(exc))
+        return SidecarReadResult(entries=[], source=SidecarReadStatus.ERROR)
     for line in lines:
         line = line.strip()
         if not line:
--- a/src/autoskillit/fleet/state_recovery.py
+++ b/src/autoskillit/fleet/state_recovery.py
@@ -130,7 +130,10 @@ def classify_stale_dispatch(
 
     Returns (new_status, sidecar_path_or_empty).
     """
-    from autoskillit.fleet.sidecar import read_sidecar_from_path  # noqa: PLC0415
+    from autoskillit.fleet.sidecar import (  # noqa: PLC0415
+        SidecarReadStatus,
+        read_sidecar_from_path,
+    )
 
     sidecar = Path(dispatch.sidecar_path) if dispatch.sidecar_path else None
     sidecar_path_str = ""
@@ -142,7 +145,10 @@ def classify_stale_dispatch(
         else:
             sidecar_path_str = str(sidecar)
             try:
-                has_entries = bool(read_sidecar_from_path(sidecar).entries)
+                sidecar_result = read_sidecar_from_path(sidecar)
+                has_entries = sidecar_result.source == SidecarReadStatus.FOUND and bool(
+                    sidecar_result.entries
+                )
             except Exception:
                 logger.warning(
                     "classify_stale_dispatch: read_sidecar_from_path failed for %s",
--- a/src/autoskillit/fleet/state_recovery.py
+++ b/src/autoskillit/fleet/state_recovery.py
@@ -142,7 +142,7 @@ def classify_stale_dispatch(
         else:
             sidecar_path_str = str(sidecar)
             try:
-                has_entries = bool(read_sidecar_from_path(sidecar))
+                has_entries = bool(read_sidecar_from_path(sidecar).entries)
             except Exception:
                 logger.warning(
                     "classify_stale_dispatch: read_sidecar_from_path failed for %s",
@@ -305,7 +305,7 @@ def find_dispatch_for_issue(
             if d.sidecar_path is None:
                 continue
             if d.status == DispatchStatus.RUNNING:
-                entries = read_sidecar_from_path(Path(d.sidecar_path))
+                entries = read_sidecar_from_path(Path(d.sidecar_path)).entries
                 if any(e.issue_url == issue_url for e in entries):
                     return d
             elif (
@@ -313,7 +313,7 @@ def find_dispatch_for_issue(
                 and d.status in TERMINAL_UNCLEANED_STATUSES
                 and not d.labels_cleaned
             ):
-                entries = read_sidecar_from_path(Path(d.sidecar_path))
+                entries = read_sidecar_from_path(Path(d.sidecar_path)).entries
                 if any(e.issue_url == issue_url for e in entries):
                     terminal_match = d
     return terminal_match
--- a/src/autoskillit/server/_lifespan.py
+++ b/src/autoskillit/server/_lifespan.py
@@ -304,6 +304,21 @@ async def _food_truck_auto_gate_boot(ctx: Any) -> None:
     except Exception:
         logger.warning("food_truck_auto_gate_boot_registry_failed", exc_info=True)
 
+    try:
+        from autoskillit.fleet import (  # noqa: PLC0415
+            discover_campaign_state_files,
+            sweep_stale_dispatch_labels,
+        )
+
+        _campaign_state_paths = discover_campaign_state_files(ctx.project_dir)
+        if _campaign_state_paths and ctx.github_client is not None:
+            create_background_task(
+                sweep_stale_dispatch_labels(_campaign_state_paths, ctx.github_client),
+                label="startup_label_recovery_sweep",
+            )
+    except Exception:
+        logger.warning("food_truck_auto_gate_boot_label_recovery_failed", exc_info=True)
+
 
 _LIFESPAN_BOOT_REGISTRY: dict[SessionType, Callable[[Any], Awaitable[None]] | None] = {
     SessionType.FLEET: _fleet_auto_gate_boot,
--- a/src/autoskillit/server/_lifespan.py
+++ b/src/autoskillit/server/_lifespan.py
@@ -311,13 +311,17 @@ async def _food_truck_auto_gate_boot(ctx: Any) -> None:
         )
 
         _campaign_state_paths = discover_campaign_state_files(ctx.project_dir)
+    except Exception:
+        logger.warning("food_truck_auto_gate_boot_state_discovery_failed", exc_info=True)
+        return
+    try:
         if _campaign_state_paths and ctx.github_client is not None:
             create_background_task(
                 sweep_stale_dispatch_labels(_campaign_state_paths, ctx.github_client),
                 label="startup_label_recovery_sweep",
             )
     except Exception:
-        logger.warning("food_truck_auto_gate_boot_label_recovery_failed", exc_info=True)
+        logger.warning("food_truck_auto_gate_boot_label_sweep_failed", exc_info=True)
 
 
 _LIFESPAN_BOOT_REGISTRY: dict[SessionType, Callable[[Any], Awaitable[None]] | None] = {
--- a/src/autoskillit/smoke_utils.py
+++ b/src/autoskillit/smoke_utils.py
@@ -163,6 +163,19 @@ def annotate_pr_diff(
     }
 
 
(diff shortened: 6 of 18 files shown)


Trecek (commented):

src/autoskillit/fleet/sidecar.py
```diff
@@ -58,13 +70,18 @@ def read_sidecar(dispatch_id: str, project_dir: Path) -> list[IssueSidecarEntry]
...
+    Returns a SidecarReadResult with source indicating whether the file was
+    found, missing, or unreadable. Callers must inspect source to distinguish
+    'file missing' from 'file empty'.
+    """
     entries: list[IssueSidecarEntry] = []
     try:
         lines = path.read_text().splitlines()
     except OSError:
-        return []
+        return SidecarReadResult(entries=[], source=SidecarReadStatus.MISSING)
```
Trecek, line 84: [warning] arch: OSError is mapped to SidecarReadStatus.MISSING regardless of the actual error type. A permission denial (EACCES), I/O error, or other non-ENOENT OSError will be silently treated as 'file missing' rather than 'error'. The MISSING status should be reserved for FileNotFoundError; other OSError subtypes should map to SidecarReadStatus.ERROR.

src/autoskillit/fleet/_label_cleanup.py
```diff
@@ -67,11 +67,19 @@ async def cleanup_orphaned_labels(
         )
         return False
 
-    if not entries:
+    if sidecar_result.source != SidecarReadStatus.FOUND:
```
Trecek, line 70: [warning] arch: The new MISSING/FOUND branch returns False (failure) when the sidecar is missing, but the old code returned True (success/no-op) for an empty or absent sidecar. For a dispatch whose sidecar file was never written (e.g. aborted before first write), this now logs a warning and returns False, which may block label cleanup from being reported as successful.

src/autoskillit/server/_lifespan.py
```diff
@@ -304,6 +304,21 @@ async def _food_truck_auto_gate_boot(ctx: Any) -> None:
     except Exception:
         logger.warning("food_truck_auto_gate_boot_registry_failed", exc_info=True)
 
+    try:
+        from autoskillit.fleet import (  # noqa: PLC0415
+            discover_campaign_state_files,
+            sweep_stale_dispatch_labels,
+        )
+
+        _campaign_state_paths = discover_campaign_state_files(ctx.project_dir)
```
Trecek, line 313: [warning] arch: No guard ensures this boot step is idempotent. If _food_truck_auto_gate_boot is called more than once (e.g. during reconnect), multiple sweep tasks could be spawned concurrently against the same state files.

src/autoskillit/fleet/sidecar.py
```diff
@@ -58,13 +70,18 @@ def read_sidecar(dispatch_id: str, project_dir: Path) -> list[IssueSidecarEntry]
...
+    """Read and parse a sidecar JSONL at path.
+
+    Returns a SidecarReadResult with source indicating whether the file was
+    found, missing, or unreadable. Callers must inspect source to distinguish
+    'file missing' from 'file empty'.
+    """
     entries: list[IssueSidecarEntry] = []
     try:
         lines = path.read_text().splitlines()
     except OSError:
```
Trecek: [warning] defense: OSError is caught but the error is silently discarded with no logging. A file that exists but fails to read (e.g. permission error) is indistinguishable from a genuinely missing file at the call site. The ERROR status variant exists in SidecarReadStatus but is never produced — consider distinguishing FileNotFoundError from other OSErrors.

src/autoskillit/server/_lifespan.py
```diff
@@ -304,6 +304,21 @@ async def _food_truck_auto_gate_boot(ctx: Any) -> None:
...
+        if _campaign_state_paths and ctx.github_client is not None:
+            create_background_task(
+                sweep_stale_dispatch_labels(_campaign_state_paths, ctx.github_client),
+                label="startup_label_recovery_sweep",
+            )
+    except Exception:
```
Trecek, line 323: [warning] defense: The bare except Exception block swallows all exceptions from both discover_campaign_state_files and sweep_stale_dispatch_labels construction. The warning message does not distinguish between a discovery failure and a sweep scheduling failure, making post-incident triage harder.

tests/fleet/test_sidecar.py
```diff
@@ -160,7 +162,35 @@ def test_crash_scenario_2_of_4_done(self, tmp_path: Path) -> None:
 class TestReadSidecarFromPath:
     def test_nonexistent_parent_returns_empty(self) -> None:
         result = read_sidecar_from_path(Path("/nonexistent/dir/issues.jsonl"))
-        assert result == []
+        assert result.source == SidecarReadStatus.MISSING
+        assert result.entries == []
+
+    def test_read_sidecar_from_path_returns_missing_when_file_absent(self) -> None:
```
Trecek: [warning] tests: test_read_sidecar_from_path_returns_missing_when_file_absent is a near-duplicate of test_nonexistent_parent_returns_empty (lines 163-166). Both call read_sidecar_from_path with the same nonexistent path and assert the same two conditions. The original test was already updated — this second test adds no additional coverage.

tests/fleet/test_sidecar.py
```diff
@@ -160,7 +162,35 @@ def test_crash_scenario_2_of_4_done(self, tmp_path: Path) -> None:
...
+    def test_read_sidecar_from_path_returns_found_with_entries(self, tmp_path: Path) -> None:
+        sidecar = tmp_path / "issues.jsonl"
+        sidecar.write_text(
+            '{"issue_url":"https://github.com/o/r/issues/1","status":"completed","ts":"t"}\n'
+        )
+        result = read_sidecar_from_path(sidecar)
+        assert result.source == SidecarReadStatus.FOUND
+        assert len(result.entries) == 1
+
+    def test_result_is_sidecar_read_result_type(self, tmp_path: Path) -> None:
```
Trecek: [warning] tests: test_result_is_sidecar_read_result_type only asserts isinstance(result, SidecarReadResult). Every other test in the class already implicitly validates this by accessing result.source and result.entries without error. This test is redundant and can be removed.

tests/fleet/test_label_cleanup.py
```diff
@@ -452,3 +452,19 @@ async def test_cleanup_returns_false_when_swap_returns_failure(self, tmp_path: P
...
+    @pytest.mark.anyio
+    async def test_cleanup_returns_false_when_sidecar_file_missing_on_disk(
+        self, tmp_path: Path
+    ) -> None:
+        """cleanup_orphaned_labels returns False when sidecar_path points to nonexistent file."""
+        from autoskillit.fleet._label_cleanup import cleanup_orphaned_labels
+
+        missing_path = str(tmp_path / "deleted_issues.jsonl")
+        github_client = AsyncMock()
+        github_client.swap_labels = AsyncMock(return_value={"success": True})
```
Trecek: [warning] tests: github_client.swap_labels is set up with a success return value even though the test expects it to never be called. The mock result is irrelevant — remove the swap_labels setup to clarify intent.

tests/server/test_server_init_session_visibility.py
```diff
@@ -775,6 +775,52 @@ async def test_run_skill_succeeds_after_food_truck_auto_gate_boot(
...
+        monkeypatch.setenv(FOOD_TRUCK_TOOL_TAGS_ENV_VAR, "kitchen-core")
+
+        mock_create_bg_task = MagicMock(return_value=MagicMock())
+
+        with patch("autoskillit.server.tools.tools_kitchen._write_hook_config"):
+            with patch("autoskillit.server._misc._prime_quota_cache", new=AsyncMock()):
+                with patch("autoskillit.pipeline.create_background_task", mock_create_bg_task):
+                    with patch("autoskillit.core.register_active_kitchen"):
+                        with patch(
+                            "autoskillit.fleet.sweep_stale_dispatch_labels",
```
Trecek, line 807: [warning] tests: sweep_stale_dispatch_labels is patched with new_callable=AsyncMock but the mock is never captured (no 'as sweep_mock'). The test only verifies create_background_task was called with label 'startup_label_recovery_sweep' — capturing the patched mock and asserting it was called would provide stronger coverage.

src/autoskillit/fleet/sidecar.py
```diff
@@ -83,7 +100,7 @@ def read_sidecar_from_path(path: Path) -> list[IssueSidecarEntry]:
         except (json.JSONDecodeError, KeyError) as exc:
             logger.debug("sidecar: skipping corrupt JSONL line", path=str(path), error=str(exc))
             continue
-    return entries
+    return SidecarReadResult(entries=entries, source=SidecarReadStatus.FOUND)
```
Trecek, line 106: [warning] bugs: read_sidecar_from_path returns SidecarReadStatus.FOUND even when some lines were skipped due to json.JSONDecodeError or KeyError. A file with only corrupt lines returns FOUND with empty entries, which callers treating FOUND as 'file exists and is valid' may misinterpret. Consider returning ERROR or PARTIAL when all lines fail to parse.

src/autoskillit/fleet/_api.py
```diff
@@ -856,22 +873,15 @@ def _on_spawn(pid: int, ticks: int) -> None:
                     project_dir=str(tool_ctx.project_dir),
                     marker_dir=marker_dir,
                     session_id=caller_session_id,
-                    timeout=float(timeout_sec) if timeout_sec else None,
+                    timeout=resolved_timeout,
```
Trecek, line 876: [warning] bugs: Previously timeout=None was passed when timeout_sec was falsy (0 or None), which may have signalled 'no timeout' to the caller. Now resolved_timeout always returns a concrete float (never None). If any downstream consumer treats None as 'wait forever' and a float as a bounded timeout, behavior has changed silently for the None case. Verify the subprocess/session runner interprets this correctly.

src/autoskillit/fleet/state_recovery.py
```diff
@@ -142,7 +142,7 @@ def classify_stale_dispatch(
         else:
             sidecar_path_str = str(sidecar)
             try:
-                has_entries = bool(read_sidecar_from_path(sidecar))
+                has_entries = bool(read_sidecar_from_path(sidecar).entries)
```
Trecek: [warning] cohesion: Inline .entries access discards the SidecarReadResult.source discriminator, making it impossible to distinguish a missing sidecar from an empty one. The call-sites at L308 and L316 share the same pattern. _label_cleanup.py correctly checks .source first — state_recovery.py should follow the same pattern for symmetry.

src/autoskillit/smoke_utils.py
```diff
@@ -163,6 +163,19 @@ def annotate_pr_diff(
     }
 
 
+LOCAL_ROUND_EXEMPT_VERDICTS: frozenset[str] = frozenset(
```
Trecek, line 170: [warning] cohesion: LOCAL_ROUND_EXEMPT_VERDICTS is defined in smoke_utils.py. While its only consumer (check_review_loop) is in the same file, verdict constants and the broader verdict taxonomy live in the pipeline/fleet layer. Placing a verdict classification set in smoke_utils scatters feature code across layers.

src/autoskillit/smoke_utils.py
```diff
@@ -163,6 +163,19 @@ def annotate_pr_diff(
...
+    {
+        "approved_with_comments",
+    }
+)
+"""Verdicts exempt from local_review_rounds re-review.
```
Trecek: [warning] slop: Module-level string literal placed after the constant assignment is not a docstring — it is a dead expression statement that will not appear in help() or IDE tooltips. Move the text into an inline comment block preceding the constant, or remove it if the constant name is self-documenting.

Trecek (commented):

src/autoskillit/fleet/sidecar.py
Trecek: Agreed — fixed in c193e1a4518e78542595c4468115e87366b18307. Split `except OSError` into `except FileNotFoundError → MISSING` and `except OSError as exc → log warning + ERROR`. `SidecarReadStatus.ERROR` now activates for permission-denied and I/O-error cases.

src/autoskillit/fleet/_label_cleanup.py
Trecek: Investigated — this is intentional. The commit message for 963fb6b7 explicitly states: 'Update cleanup_orphaned_labels to return False for MISSING sidecar (was incorrectly True)'. The old True behavior was a documented bug. The new False behavior is the correct fix: an unavailable sidecar means label cleanup cannot proceed safely.

src/autoskillit/server/_lifespan.py
Trecek: Investigated — this is intentional. `_food_truck_auto_gate_boot` is registered in `_LIFESPAN_BOOT_REGISTRY` and called exactly once from `_autoskillit_lifespan` context manager. FastMCP's lifespan is server-scoped, not connection-scoped — it does not re-enter on client reconnect. An idempotency guard here would be defensive coding for a scenario that cannot occur in this framework's lifecycle model.

src/autoskillit/fleet/sidecar.py
Trecek: Agreed — fixed in c193e1a4518e78542595c4468115e87366b18307. Other OSErrors (non-FileNotFoundError) now log a `logger.warning` with the path and error details before returning `SidecarReadStatus.ERROR`. Same fix as arch:3292223384 — single code change covering both dimensions.

src/autoskillit/server/_lifespan.py
Trecek: Agreed — fixed in c193e1a4518e78542595c4468115e87366b18307. Split the single try/except into two separate blocks: `food_truck_auto_gate_boot_state_discovery_failed` for `discover_campaign_state_files` and `food_truck_auto_gate_boot_label_sweep_failed` for the background task scheduling. State discovery failure now also early-returns to avoid accessing `_campaign_state_paths` in the second block.

tests/fleet/test_sidecar.py
Trecek: Agreed — fixed in c193e1a4518e78542595c4468115e87366b18307. Removed `test_read_sidecar_from_path_returns_missing_when_file_absent`: it was byte-for-byte identical to `test_nonexistent_parent_returns_empty` (same path `/nonexistent/dir/issues.jsonl`, same assertions).

tests/fleet/test_sidecar.py
Trecek: Agreed — fixed in c193e1a4518e78542595c4468115e87366b18307. Removed `test_result_is_sidecar_read_result_type`: the isinstance check is already proven by every other test in the class that accesses `.source` and `.entries` attributes. Also removed the now-unused `SidecarReadResult` import.

tests/fleet/test_label_cleanup.py
Trecek: Agreed — fixed in c193e1a4518e78542595c4468115e87366b18307. Removed `github_client.swap_labels = AsyncMock(return_value={"success": True})`. The setup was misleading since the mock is immediately asserted as `assert_not_called()`. An `AsyncMock()` with no configured return value suffices.

tests/server/test_server_init_session_visibility.py
Trecek: Investigated — this is intentional. The test's assertion target is `mock_create_bg_task` with `label='startup_label_recovery_sweep'` — this validates the correct behavioral contract (the background task scheduler was invoked with the right label). The `sweep_stale_dispatch_labels` patch prevents real network calls; there is no need to assert on it directly. Capturing it would over-assert implementation details rather than the observable scheduling behavior.

src/autoskillit/fleet/sidecar.py
Trecek: Investigated — this is intentional. `SidecarReadStatus.FOUND` means 'the file was accessible' — which is true even if all lines are corrupt. All callers already check `.entries` separately from `.source`: `cleanup_orphaned_labels` checks `if not sidecar_result.entries: return True` and `state_recovery.py` checks `bool(sidecar_result.entries)`. No caller assumes FOUND implies non-empty entries. The design is intentional and correctly handled throughout.

src/autoskillit/fleet/_api.py
Trecek: Investigated — this is intentional. Commit 262ac8d8 documents: 'Replace float(timeout_sec) if timeout_sec else None (process kill) with resolved_timeout'. The `is not None` check in `resolve_dispatch_timeout` was specifically designed to fix `timeout_sec=0` being treated as None (falsy) under the old `float(timeout_sec) if timeout_sec else None` pattern. This is a deliberate consolidation and correctness fix, not a regression.

src/autoskillit/fleet/state_recovery.py
Trecek: Agreed — fixed in c193e1a4518e78542595c4468115e87366b18307. Changed `bool(read_sidecar_from_path(sidecar).entries)` to capture `sidecar_result = read_sidecar_from_path(sidecar)` and check `sidecar_result.source == SidecarReadStatus.FOUND and bool(sidecar_result.entries)`. Also updated the local import to include `SidecarReadStatus` alongside `read_sidecar_from_path`.

src/autoskillit/smoke_utils.py
Trecek: Investigated — this is intentional. `LOCAL_ROUND_EXEMPT_VERDICTS` is consumed exclusively by `check_review_loop` in the same file. The package CLAUDE.md documents `smoke_utils.py` as 'Callables for smoke-test pipeline run_python steps'; `check_review_loop` is such a callable and the frozenset is its behavioral parameter. Co-locating a constant with its only consumer is the correct pattern; moving it to a types module would be premature abstraction.

src/autoskillit/smoke_utils.py
Trecek: Agreed — fixed in c193e1a4518e78542595c4468115e87366b18307. Converted the dead module-level string literal (placed after the assignment, making it a discarded expression) to a comment block (`# ...`) placed before the constant. Python only recognizes docstrings at the start of a module/class/function body as the first statement.

tests/arch/test_boot_step_symmetry.py
```diff
@@ -0,0 +1,41 @@
...
+
+pytestmark = [pytest.mark.layer("arch"), pytest.mark.small]
+
+LIFESPAN_PATH = Path(__file__).parents[2] / "src" / "autoskillit" / "server" / "_lifespan.py"
+
+_SWEEP_SYMBOL = "sweep_stale_dispatch_labels"
+_BOOT_FUNCTIONS = ("_fleet_auto_gate_boot", "_food_truck_auto_gate_boot")
+
+
+def _function_body_contains_symbol(tree: ast.Module, func_name: str, symbol: str) -> bool:
```
Trecek, line 18: [warning] tests: _function_body_contains_symbol uses ast.walk which finds the symbol name as any ast.Name node, not specifically as an ast.Call target. While ast.Name correctly excludes string literals, it would match a bare reference (e.g. assignment without invocation). Consider checking for ast.Call nodes whose func attribute references the symbol to ensure the guard validates an actual call, not just a name binding.

tests/arch/test_dispatch_timeout_guard.py
```diff
@@ -0,0 +1,51 @@
...
+    func_source = _collect_function_source(tree, _RUN_DISPATCH)
+    assert func_source, f"Function '{_RUN_DISPATCH}' not found in {_API_PATH}"
+
+    assert _RESOLVE_SYMBOL in func_source, (
+        f"'{_RUN_DISPATCH}' must call '{_RESOLVE_SYMBOL}' but no call was found. "
+        "All three timeout surfaces (prompt, process kill, session deadline) must use "
+        "a single resolved value from resolve_dispatch_timeout."
+    )
+
+    assert "or 1800" not in func_source, (
```
Trecek, line 43: [warning] tests: The guard checks for the resolve_dispatch_timeout symbol via string containment in ast.unparse output. This would be satisfied by a dead import or a variable assignment referencing the name without an actual function call. Consider asserting that an ast.Call node with func referencing resolve_dispatch_timeout exists inside _run_dispatch for a stronger guarantee.

tests/fleet/test_startup_label_recovery.py
```diff
@@ -333,3 +333,48 @@ async def test_startup_sweep_skips_terminal_dispatch_with_labels_already_cleaned
...
+    @pytest.mark.anyio
+    async def test_sweep_does_not_mark_cleaned_when_sidecar_file_missing_on_disk(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """Dispatch with deleted sidecar file must NOT get labels_cleaned=True."""
+        from autoskillit.fleet import DispatchRecord, read_state, write_initial_state
+        from autoskillit.fleet._label_cleanup import sweep_stale_dispatch_labels
+        from autoskillit.fleet.state import upsert_dispatch_record_by_name
+
+        monkeypatch.setattr(
```
Trecek, line 346: [warning] tests: monkeypatch.setattr patches is_dispatch_session_alive with a synchronous lambda. If is_dispatch_session_alive is async, the lambda will return a coroutine object (truthy) rather than False, causing the test to pass vacuously. Verify the production function is synchronous; if async, use AsyncMock(return_value=False) instead.

Trecek (commented):

Trecek (commented):

src/autoskillit/fleet/sidecar.py
Trecek: Investigated — this is intentional. Already fixed in commit c193e1a4. The current code at lines 83-87 catches FileNotFoundError separately (→ MISSING) and OSError as a distinct branch (→ ERROR with a warning log).

src/autoskillit/fleet/_label_cleanup.py
Trecek: Valid observation — flagged for design decision. The behavior change from True to False for a missing sidecar was intentional per commit 963fb6b7. The reviewer raises a valid design question: for dispatches aborted before first write, returning False may cause the caller to treat it as cleanup failure. Requires human confirmation of intended semantics.

src/autoskillit/fleet/sidecar.py
Trecek: Investigated — this is intentional. All callers handle FOUND with empty entries correctly. _label_cleanup.py returns True for empty entries (nothing to clean up). state_recovery.py gates on non-empty entries explicitly. The reviewer's concern is theoretical but doesn't apply to existing callers.

src/autoskillit/fleet/_api.py
Trecek: Investigated — this is intentional. The downstream consumer (dispatch_food_truck) handled None by falling back to fleet_cfg.default_timeout_sec — not 'wait forever'. resolve_dispatch_timeout uses default_timeout_sec identically. Behavior is preserved for None and actually improved for 0 (which old code mishandled as None).

tests/server/test_server_init_session_visibility.py
Trecek: Investigated — this is intentional. The test's design is intentional: it verifies the scheduling contract (create_background_task called with right label), not the sweep internals. Capturing the sweep mock and asserting it was called would test the wrong layer.

tests/arch/test_boot_step_symmetry.py
Trecek: Valid observation — flagged for design decision. _function_body_contains_symbol uses ast.Name which would match bare references. The reviewer's ast.Call suggestion would produce a stronger guard. This is a valid improvement suggestion that requires human judgment on acceptable guard rigor.

tests/arch/test_dispatch_timeout_guard.py
Trecek: Valid observation — flagged for design decision. String containment on ast.unparse output would be satisfied by in-function assignments. The reviewer's ast.Call suggestion would be more precise. This is a valid improvement suggestion that requires human judgment.

src/autoskillit/server/_lifespan.py
Trecek: Acknowledged — minor suggestion noted.

src/autoskillit/server/_lifespan.py
Trecek: Acknowledged — minor suggestion noted.

src/autoskillit/smoke_utils.py
Trecek: Acknowledged — minor suggestion noted.

tests/fleet/test_startup_label_recovery.py
Trecek: Acknowledged — minor suggestion noted.

PR #2782, opened 2026-05-23
Worktree Path Mismatch — Type-Safe Path Propagation

Trecek:
## Summary

The worktree path propagation pipeline had no structural guarantee that a `worktree_path` value refers to an actual directory on disk. The value flowed as a raw `str | None` from extraction (`_extract_worktree_path`) through `SkillResult.worktree_path` to MCP tool consumers (`run_skill`, `run_cmd`, `test_check`) without any filesystem validation gate. When `parse_session_result` failed to extract text from a non-Anthropic backend, `_extract_worktree_path` returned `None`, and the orchestrating model constructed a wrong path from context — propagating a hallucinated cwd to `test_check` and causing test failures.

The fix introduces `ValidatedWorktreePath` type following the `ValidatedAddDir` precedent, centralizes cwd validation at the subprocess boundary (`_run_subprocess`), and adds `isdir()` guards to `test_check` and `run_skill`. Four layers of defense: (1) `_extract_worktree_path` rejects non-existent paths via `isdir()` check, (2) `_build_skill_result` wraps extracted paths through `validate_worktree_path()` factory, (3) `_run_subprocess` centralized guard catches invalid cwd before subprocess spawn, (4) direct `isdir` guards in `run_skill` and `test_check` provide tool-level validation.

## Implementation Plan

Plan file: `.autoskillit/temp/rectify/rectify_worktree_path_mismatch_2026-05-22_234254.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code) via AutoSkillit

## Token Usage Summary

| Step | Model | count | uncached | output | cache_read | peak_ctx | turns | cache_write | time |
|------|-------|-------|----------|--------|------------|----------|-------|-------------|------|
| rectify | claude-sonnet-4-6 | 1 | 5.4k | 15.9k | 1.3M | 110.7k | 163 | 66.2k | 7m 52s |
| dry_walkthrough | claude-opus-4-6 | 1 | 74 | 20.3k | 2.6M | 92.1k | 172 | 76.8k | 10m 0s |
| implement | claude-opus-4-6 | 1 | 124 | 21.8k | 7.1M | 118.2k | 166 | 200.4k | 9m 45s |
| prepare_pr* | MiniMax-M2.7 | 1 | 52.4k | 4.9k | 430.1k | 29.8k | 29 | 42.7k | 1m 17s |
[...]

PR #3258, opened 2026-05-29
[FEATURE] Rectify: Server-Side Input Contract Enforcement for `skill_command` Path Arguments

### LINKED ISSUE #3245: Orchestrator plan_path argument strips .md extension and appends dispatch timestamp — resolve-failures receives nonexistent path
## Summary

The orchestrator constructs `plan_path` arguments for resolve-failures with the `.md` extension stripped and a worktree dispatch timestamp appended as a suffix. The resulting path does not exist on disk. The agent recovered by improvising a directory scan, but this is outside the skill spec and fragile.

Partially related to #2771 (worktree path mismatch: wrong cwd propagated to test_check and resolve-failures), which is staged/merged to develop. That fix addressed the cwd component but not the plan_path construction defect.

## Observed Behavior

**Pipeline:** remediation run `9389c358-d927-49c1-b9b1-84074ebbf22e`
**Session:** `7e010672` (assess step)

1. Orchestrator passed `plan_path` as `...rectify_merge_ref_coherence_2026-05-28_194500-20260528-201752` — missing the `.md` extension and with the worktree dispatch timestamp (`-20260528-201752`) appended as a suffix.
2. No file exists at that path.
3. The agent received a `ToolUseError: directory does not exist` when attempting to read it.
4. The agent recovered by running `ls` on the parent directory and selecting the single `.md` file matching the stem.
5. This worked because exactly one candidate file was present.

## Risk

The skill's documented behavior assumes a valid `plan_path`. The improvised directory scan:
- Could mismatch with multiple candidates if more than one plan file shares the stem prefix
- Costs extra turns and tokens on every invocation
- Is not in the skill spec — a different model or temperature could fail to improvise the same recovery

## Root Cause

The plan_path construction logic (likely in the orchestrator's step argument interpolation) appears to:
1. Strip the `.md` extension from the plan filename
2. Append the worktree dispatch timestamp as a suffix

Both transformations produce an invalid path.

## Evidence

- JSONL session `7e010672` line 16: `ToolUseError: directory does not exist`
- JSONL session `7e010672` line 24: `ls` recovery selecting the correct `.md` file
- `summary.json` `write_path_warnings` for session `7e010672`

## Related

- #2771 (staged) — Worktree path mismatch: wrong cwd propagated to test_check and resolve-failures (addresses cwd, not plan_path construction)

Trecek:
## Summary

The orchestrator LLM mangles `context.plan_path` when constructing `skill_command` strings — stripping the `.md` extension and appending worktree timestamp fragments. This happens because `${{ context.* }}` substitution is entirely LLM-managed with no code-level enforcement of the substituted values. The `skill_contracts.yaml` declares `plan_path` as `type: file_path, required: true` for 35+ skills, and `worktree_path` as `type: directory_path` for 15+ skills, but these input type declarations are purely advisory — no runtime validator enforces them.

The architectural weakness is: **the server enforces output contracts (patterns, write behavior, read-only) but has zero enforcement of input contracts**. The `_check_dry_walkthrough` guard is a one-off hand-coded validation for implement skills only. It cannot scale to cover all 320+ `file_path`/`directory_path` input declarations across 80+ skills.

The immunity solution is a **contract-driven input validator** in the `run_skill` guard chain that reads input type declarations from `skill_contracts.yaml` and enforces them against the actual `skill_command` arguments at dispatch time. This catches `plan_path` mangling, `worktree_path` corruption, and any other `file_path`/`directory_path` violation — for every skill — without per-skill hand-coding.

## Related Issues Found

1. **Every `file_path` input across all skills is unvalidated.** There are 263 `type: file_path` declarations and 57 `type: directory_path` declarations in `skill_contracts.yaml`. None are enforced at dispatch time. Any LLM substitution error that produces a path-like token (starts with `/`) will pass through the hook chain and server guards uncaught.

2. **`skill_cmd_guard.py` and `_check_dry_walkthrough` duplicate path detection logic.** Both implement `_looks_like_path()` with the same `_PATH_PREFIXES = ("/", "./", ".autoskillit/")`. The hook checks argument ordering; the guard checks file existence + marker. Neither checks argument type conformance.

3. **`extract_path_arg()` returns only the first path-like token.** Skills like `resolve-failures` take multiple path arguments (`worktree_path`, `plan_path`, `diagnosis_path`). The server can only extract the first one — there's no `extract_all_path_args()` function to access positional arguments by index.

4. **`ValidatedWorktreePath` exists but `ValidatedPlanPath` does not.** PR #2782 introduced a type-safe wrapper for `worktree_path` with `is_absolute()` + `is_dir()` validation. No equivalent exists for `plan_path` (`is_absolute()` + `is_file()` + `.md` extension check). The same class of fix was applied to one path type but not the other.

5. **The `campaign-path-coherence` rule watches `worktree_path` staleness but not `plan_path` staleness.** The semantic rule in `rules_campaign_flow.py` catches cross-dispatch `worktree_path` decay. No analogous rule exists for `plan_path`.

### New (★):
tests/core/test_type_helpers.py

### Modified (●):
- src/autoskillit/core/__init__.pyi
- src/autoskillit/core/types/_type_helpers.py
- src/autoskillit/core/types/_type_protocols_execution.py
- src/autoskillit/core/types/_type_results.py
- src/autoskillit/hooks/guards/skill_cmd_guard.py
- src/autoskillit/pipeline/context.py
- src/autoskillit/server/_factory.py
- src/autoskillit/server/_guards.py
- src/autoskillit/server/tools/tools_execution.py
- tests/_test_filter.py
- tests/core/CLAUDE.md
- tests/core/test_type_protocol_shards.py
- tests/pipeline/test_context.py
- tests/recipe/test_bundled_recipes_pipeline_structure.py
- tests/server/test_guards_module.py
- tests/server/test_tools_execution_input_gates.py

Closes #3245

## Implementation Plan

Plan file: `.autoskillit/temp/rectify/rectify_plan_path_contract_enforcement_2026-05-28_234500.md`

## Token Usage Summary

| investigate* | opus | 1 | 75 | 11.4k | 1.8M | 116.4k | 325 | 70.8k | 8m 47s |
| rectify* | opus[1m] | 1 | 72 | 24.3k | 3.4M | 151.2k | 270 | 198.6k | 20m 59s |
| review_approach* | sonnet | 1 | 52 | 5.8k | 183.0k | 40.8k | 33 | 28.2k | 3m 24s |
| dry_walkthrough* | opus | 1 | 75 | 16.0k | 2.0M | 87.0k | 180 | 70.8k | 7m 29s |
| implement* | sonnet | 1 | 2.6k | 14.9k | 1.8M | 93.9k | 102 | 78.2k | 5m 18s |
| audit_impl* | sonnet | 1 | 558 | 15.5k | 381.6k | 60.6k | 69 | 45.7k | 6m 59s |
| prepare_pr* | sonnet | 1 | 161.0k | 4.0k | 367.6k | 30.9k | 24 | 15.7k | 1m 30s |
| compose_pr* | sonnet | 1 | 95.1k | 3.4k | 277.9k | 30.9k | 23 | 43.5k | 1m 26s |
| review_pr* | sonnet | 3 | 2.3k | 136.2k | 4.9M | 144.6k | 362 | 332.6k | 42m 31s |
| resolve_review* | opus | 4 | 420 | 101.9k | 21.3M | 126.1k | 557 | 401.4k | 57m 54s |
| post_run_diagnostics* | sonnet | 1 | 18 | 3.8k | 146.9k | 93.4k | 441 | 6.8k | 10m 33s |
| **Total** | | | 262.3k | 337.3k | 36.7M | 151.2k | | 1.3M | 2h 46m |

## Token Efficiency

| rectify | 0 | — | — | — |
| review_approach | 0 | — | — | — |
| implement | 239 | 7597.4 | 327.1 | 62.5 |
| review_pr | 0 | — | — | — |
| resolve_review | 269 | 79329.1 | 1492.4 | 378.7 |
| post_run_diagnostics | 0 | — | — | — |
| **Total** | **508** | 72155.9 | 2544.0 | 664.0 |

## Model Usage Breakdown

| opus | 3 | 570 | 129.3k | 25.2M | 543.1k | 1h 14m |
| opus[1m] | 1 | 72 | 24.3k | 3.4M | 198.6k | 20m 59s |
| sonnet | 7 | 261.7k | 183.7k | 8.1M | 550.7k | 1h 11m |

### DIFF
--- a/src/autoskillit/core/types/_type_protocols_execution.py
+++ b/src/autoskillit/core/types/_type_protocols_execution.py
@@ -7,9 +7,10 @@ from pathlib import Path
 from typing import Protocol, runtime_checkable
 
 from ._type_checkpoint import SessionCheckpoint  # noqa: F401, TC001
-from ._type_results import SkillResult, TestResult, ValidatedAddDir, WriteBehaviorSpec
+from ._type_results import InputSpec, SkillResult, TestResult, ValidatedAddDir, WriteBehaviorSpec
 
 __all__ = [
+    "InputContractResolver",
     "TestRunner",
     "HeadlessExecutor",
     "OutputPatternResolver",
@@ -116,3 +117,10 @@ class WriteExpectedResolver(Protocol):
     """Protocol for resolving write-expectation metadata from skill contracts."""
 
     def __call__(self, skill_command: str) -> WriteBehaviorSpec: ...
+
+
+@runtime_checkable
+class InputContractResolver(Protocol):
+    """Protocol for resolving input contract specs from skill contracts."""
+
+    def __call__(self, skill_command: str) -> Sequence[InputSpec]: ...
--- a/src/autoskillit/core/types/_type_results.py
+++ b/src/autoskillit/core/types/_type_results.py
@@ -20,6 +20,7 @@ T = TypeVar("T")
 
 __all__ = [
     "ContaminationOutcome",
+    "InputSpec",
     "LoadReport",
     "LoadResult",
     "TestResult",
@@ -144,6 +145,16 @@ class WriteBehaviorSpec:
 
     mode: str | None = None
     expected_when: tuple[str, ...] = ()
+
+
+@dataclass(frozen=True, slots=True)
+class InputSpec:
+    """Input contract specification for a single file_path or directory_path argument."""
+
+    name: str
+    type: Literal["file_path", "directory_path", "string"]
+    required: bool
+    position: int
 
 
 @dataclass
--- a/src/autoskillit/core/types/_type_results.py
+++ b/src/autoskillit/core/types/_type_results.py
@@ -152,7 +152,7 @@ class InputSpec:
     """Input contract specification for a single file_path or directory_path argument."""
 
     name: str
-    type: Literal["file_path", "directory_path", "string"]
+    type: Literal["file_path", "directory_path"]
     required: bool
     position: int
 
--- a/src/autoskillit/core/types/_type_results.py
+++ b/src/autoskillit/core/types/_type_results.py
@@ -156,6 +156,10 @@ class InputSpec:
     required: bool
     position: int
 
+    def __post_init__(self) -> None:
+        if self.position < 0:
+            raise ValueError(f"InputSpec.position must be >= 0, got {self.position}")
+
 
 @dataclass
 class FailureRecord:
--- a/src/autoskillit/hooks/guards/skill_cmd_guard.py
+++ b/src/autoskillit/hooks/guards/skill_cmd_guard.py
@@ -51,6 +51,14 @@ PATH_ARG_SKILLS: frozenset[str] = frozenset(
 # Token prefixes that unambiguously identify a filesystem path argument.
 _PATH_PREFIXES: tuple[str, ...] = ("/", "./", ".autoskillit/")
 
+# Position of the plan file among path-like tokens for each skill.
+_PLAN_PATH_POSITION: dict[str, int] = {
+    "implement-worktree-no-merge": 0,
+    "implement-worktree": 0,
+    "retry-worktree": 0,
+    "resolve-failures": 1,
+}
+
 # Captures the skill short-name from a skill_command string such as:
 #   /implement-worktree-no-merge ...
 #   implement-worktree-no-merge ...
@@ -103,8 +111,22 @@ def main() -> None:
     tokens = args_str.split()
     first = tokens[0]
 
-    # Correct format: first token is a path → allow.
+    # Correct format: first token is a path.
     if _looks_like_path(first):
+        plan_pos = _PLAN_PATH_POSITION.get(skill_name)
+        if plan_pos is not None:
+            path_tokens = [t for t in tokens if _looks_like_path(t)]
+            if plan_pos < len(path_tokens):
+                plan_token = path_tokens[plan_pos]
+                basename = plan_token.rsplit("/", 1)[-1] if "/" in plan_token else plan_token
+                if basename and "." not in basename:
+                    _deny(
+                        f"{SKILL_CMD_DENY_TRIGGER} '{skill_name}': "
+                        f"plan path argument '{plan_token}' has no file extension. "
+                        f"This may indicate a mangled path (e.g. stripped .md extension). "
+                        f"Verify the path is correct and includes the file extension."
+                    )
+                    sys.exit(0)
         sys.exit(0)
 
     # First token is not a path. Check whether a path token appears later.
--- a/src/autoskillit/hooks/guards/skill_cmd_guard.py
+++ b/src/autoskillit/hooks/guards/skill_cmd_guard.py
@@ -65,7 +65,7 @@ _PLAN_PATH_POSITION: dict[str, int] = {
 _SKILL_RE = re.compile(r"^/?(?:autoskillit:)?(\S+)")
 
 
-def _looks_like_path(token: str) -> bool:
+def is_path_like_token(token: str) -> bool:
     """Return True if token begins with a recognised filesystem path prefix."""
     return any(token.startswith(p) for p in _PATH_PREFIXES)
 
@@ -112,10 +112,10 @@ def main() -> None:
     first = tokens[0]
 
     # Correct format: first token is a path.
-    if _looks_like_path(first):
+    if is_path_like_token(first):
         plan_pos = _PLAN_PATH_POSITION.get(skill_name)
         if plan_pos is not None:
-            path_tokens = [t for t in tokens if _looks_like_path(t)]
+            path_tokens = [t for t in tokens if is_path_like_token(t)]
             if plan_pos < len(path_tokens):
                 plan_token = path_tokens[plan_pos]
                 basename = plan_token.rsplit("/", 1)[-1] if "/" in plan_token else plan_token
@@ -130,7 +130,7 @@ def main() -> None:
         sys.exit(0)
 
     # First token is not a path. Check whether a path token appears later.
-    path_token = next((t for t in tokens[1:] if _looks_like_path(t)), None)
+    path_token = next((t for t in tokens[1:] if is_path_like_token(t)), None)
     if path_token is None:
         # No path-like token found at all — could be pasted plan content.
         # Allow; the skill's Step 0 will handle it.
@@ -140,8 +140,8 @@ def main() -> None:
     # Reconstruct corrected command: all path tokens first, then non-path tokens.
     # When args_str is multiline, separate prose from paths with a blank line.
     has_newlines = "\n" in args_str
-    path_tokens = [t for t in tokens if _looks_like_path(t)]
-    non_path_tokens = [t for t in tokens if not _looks_like_path(t)]
+    path_tokens = [t for t in tokens if is_path_like_token(t)]
+    non_path_tokens = [t for t in tokens if not is_path_like_token(t)]
     path_part = " ".join(path_tokens)
     if has_newlines and non_path_tokens:
         prose_block = " ".join(non_path_tokens)
--- a/src/autoskillit/server/_guards.py
+++ b/src/autoskillit/server/_guards.py
@@ -6,7 +6,16 @@ import os
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from autoskillit.core import SessionType, extract_path_arg, get_logger, session_type
+from autoskillit.core import (
+    InputContractResolver,
+    SessionType,
+    _looks_like_path,
+    extract_path_arg,
+    extract_positional_args,
+    extract_skill_name,
+    get_logger,
+    session_type,
+)
 from autoskillit.pipeline import gate_error_result, headless_error_result
 
 if TYPE_CHECKING:
@@ -160,6 +169,56 @@ def _check_dry_walkthrough(skill_command: str, cwd: str) -> str | None:
             f"The implement gate requires the plan at its make-plan/ or rectify/ origin — "
             f"copies in other directories are not accepted."
         )
+
+    return None
+
+
+def _check_input_contracts(
+    skill_command: str,
+    cwd: str,
+    resolver: InputContractResolver | None,
+) -> str | None:
+    """Validate skill_command arguments against input contracts.
+
+    Returns gate_error_result JSON string on validation failure, None if OK.
+    Fails open if resolver is None or skill has no contracts.
+    """
+    if resolver is None:
+        return None
+    specs = resolver(skill_command)
+    if not specs:
+        return None
+
+    args = extract_positional_args(skill_command)
+    path_args = [a for a in args if _looks_like_path(a)]
+
+    if not path_args:
+        return None
+
+    for spec in specs:
+        if spec.position >= len(path_args):
+            if spec.required:
+                return gate_error_result(
+                    f"Missing required {spec.type} argument '{spec.name}' "
+                    f"for {extract_skill_name(skill_command)}"
+                )
+            continue
+
+        value = path_args[spec.position]
+        resolved = Path(cwd) / value if not Path(value).is_absolute() else Path(value)
+
+        if spec.type == "file_path":
+            if not resolved.is_file():
+                return gate_error_result(
+                    f"Input '{spec.name}' for {extract_skill_name(skill_command)}: "
+                    f"expected a file, path does not exist or is a directory: {resolved}"
+                )
+        elif spec.type == "directory_path":
+            if not resolved.is_dir():
+                return gate_error_result(
+                    f"Input '{spec.name}' for {extract_skill_name(skill_command)}: "
+                    f"expected a directory, path does not exist or is a file: {resolved}"
+                )
 
     return None
 
--- a/src/autoskillit/server/_guards.py
+++ b/src/autoskillit/server/_guards.py
@@ -9,11 +9,11 @@ from typing import TYPE_CHECKING
 from autoskillit.core import (
     InputContractResolver,
     SessionType,
-    _looks_like_path,
     extract_path_arg,
     extract_positional_args,
     extract_skill_name,
     get_logger,
+    is_path_like_token,
     session_type,
 )
 from autoskillit.pipeline import gate_error_result, headless_error_result
@@ -194,7 +194,7 @@ def _check_input_contracts(
         return None
 
     args = extract_positional_args(skill_command)
-    path_args = [a for a in args if _looks_like_path(a)]
+    path_args = [a for a in args if is_path_like_token(a)]
 
     for spec in specs:
         if spec.position >= len(path_args):
--- a/src/autoskillit/server/_guards.py
+++ b/src/autoskillit/server/_guards.py
@@ -192,9 +192,6 @@ def _check_input_contracts(
     args = extract_positional_args(skill_command)
     path_args = [a for a in args if _looks_like_path(a)]
 
-    if not path_args:
-        return None
-
     for spec in specs:
         if spec.position >= len(path_args):
             if spec.required:
--- a/src/autoskillit/server/_guards.py
+++ b/src/autoskillit/server/_guards.py
@@ -185,7 +185,11 @@ def _check_input_contracts(
     """
     if resolver is None:
         return None
-    specs = resolver(skill_command)
+    try:
+        specs = resolver(skill_command)
+    except Exception:
+        logger.warning("input_contract_resolver_failed", skill_command=skill_command)
+        return None
     if not specs:
         return None
 
--- a/src/autoskillit/server/_guards.py
+++ b/src/autoskillit/server/_guards.py
@@ -188,7 +188,9 @@ def _check_input_contracts(
     try:
         specs = resolver(skill_command)
     except Exception:
-        logger.warning("input_contract_resolver_failed", skill_command=skill_command)
+        logger.warning(
+            "input_contract_resolver_failed", skill_command=skill_command, exc_info=True
+        )
         return None
     if not specs:
         return None
@@ -201,7 +203,7 @@ def _check_input_contracts(
             if spec.required:
                 return gate_error_result(
                     f"Missing required {spec.type} argument '{spec.name}' "
-                    f"for {extract_skill_name(skill_command)}"
+                    f"for {extract_skill_name(skill_command) or skill_command}"
                 )
             continue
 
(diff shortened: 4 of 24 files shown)


tests/server/test_tools_execution_input_gates.py
```diff
@@ -316,6 +316,189 @@ async def test_gate_still_fires_for_implement_skill(self, tool_ctx_kitchen_open,
         assert "dry-walked" in result["result"].lower()
 
 
+class TestInputContractValidation:
```
Trecek, line 319: [critical] tests: TestInputContractValidation tests all call _check_input_contracts directly, bypassing the run_skill integration path. The production wiring in tools_execution.py calls _check_input_contracts only when resume_session_id is falsy, pulling tool_ctx.input_contract_resolver from ToolContext. There is no async test in the diff that exercises _check_input_contracts firing via run_skill with an input_contract_resolver wired into tool_ctx_kitchen_open and returning a gate_error. The integration wiring at tools_execution.py:338-344 is untested end-to-end.

src/autoskillit/server/_guards.py
```diff
@@ -164,6 +173,57 @@ def _check_dry_walkthrough(skill_command: str, cwd: str) -> str | None:
...
+    try:
+        specs = resolver(skill_command)
+    except Exception:
+        logger.warning("input_contract_resolver_failed", skill_command=skill_command)
+        return None
+    if not specs:
+        return None
+
+    args = extract_positional_args(skill_command)
+    path_args = [a for a in args if is_path_like_token(a)]
```
Trecek, line 199: [critical] defense: path_args is built by filtering args through is_path_like_token, which requires tokens to start with /, ./, or .autoskillit/. A path argument passed as a bare relative filename (e.g. plan.md without a prefix) will be excluded from path_args. Because InputSpec.position counts positions among path-type tokens only, a required path input whose value is a bare filename will trigger Missing required file_path argument — a false positive that incorrectly rejects a valid command. The guard should either document that only path-prefixed tokens are validated, or apply validation to all positional args using the full args list and spec position.

src/autoskillit/server/_guards.py
```diff
@@ -164,6 +173,57 @@ def _check_dry_walkthrough(skill_command: str, cwd: str) -> str | None:
...
+    """Validate skill_command arguments against input contracts.
+
+    Returns gate_error_result JSON string on validation failure, None if OK.
+    Fails open if resolver is None or skill has no contracts.
+    """
+    if resolver is None:
+        return None
+    try:
+        specs = resolver(skill_command)
+    except Exception:
```
Trecek, line 190: [warning] defense: The bare except Exception block discards exception context before logging. The logger.warning call does not pass exc_info=True so the actual exception type and traceback are silently dropped. Diagnosis of resolver failures is impossible after the fact. Use logger.warning(..., exc_info=True) or pass the exception as a structured field to preserve error context.

src/autoskillit/server/_guards.py
```diff
@@ -164,6 +173,57 @@ def _check_dry_walkthrough(skill_command: str, cwd: str) -> str | None:
...
+    if not specs:
+        return None
+
+    args = extract_positional_args(skill_command)
+    path_args = [a for a in args if is_path_like_token(a)]
+
+    for spec in specs:
+        if spec.position >= len(path_args):
+            if spec.required:
+                return gate_error_result(
```
Trecek, line 204: [warning] defense: extract_skill_name(skill_command) returns str | None, and its return value is interpolated directly into the gate_error_result f-string. If skill_command is malformed the error message will contain the literal string None. Guard with or skill_command fallback before interpolation.

src/autoskillit/server/_guards.py
```diff
@@ -164,6 +173,57 @@ def _check_dry_walkthrough(skill_command: str, cwd: str) -> str | None:
     return None
 
 
+def _check_input_contracts(
```
Trecek, line 176: [warning] cohesion: _check_input_contracts takes a third injected parameter (resolver: InputContractResolver | None) while _check_dry_walkthrough fetches its dependencies internally via _get_config(). The two guard functions follow opposite dependency patterns, making the guard layer inconsistent. Either the resolver should be fetched from ctx or _check_dry_walkthrough should also receive its dependency as an argument.

src/autoskillit/core/types/_type_results.py
```diff
@@ -146,6 +147,16 @@ class WriteBehaviorSpec:
...
 
 
+@dataclass(frozen=True, slots=True)
+class InputSpec:
+    """Input contract specification for a single file_path or directory_path argument."""
+
+    name: str
+    type: Literal["file_path", "directory_path"]
+    required: bool
+    position: int
```
Trecek, line 157: [warning] defense: InputSpec.position has no construction-time validation. A negative value would silently wrap around in path_args indexing. Since this is a frozen=True dataclass, a __post_init__ guard asserting position >= 0 would enforce the contract at construction.

src/autoskillit/core/types/_type_protocols_execution.py
```diff
@@ -116,3 +117,10 @@ class WriteExpectedResolver(Protocol):
     """Protocol for resolving write-expectation metadata from skill contracts."""
 
     def __call__(self, skill_command: str) -> WriteBehaviorSpec: ...
+
+
+@runtime_checkable
```
Trecek, line 122: [warning] cohesion: InputContractResolver.__call__ declares return type Sequence[InputSpec], but the sole concrete implementation resolve_input_specs returns tuple[InputSpec, ...]. WriteExpectedResolver.__call__ returns WriteBehaviorSpec exactly, matching its implementation. For symmetry the protocol should declare the narrower tuple[InputSpec, ...] return type.

Trecek (commented):

src/autoskillit/hooks/guards/skill_cmd_guard.py
```diff
@@ -51,13 +51,21 @@
 # Token prefixes that unambiguously identify a filesystem path argument.
 _PATH_PREFIXES: tuple[str, ...] = ("/", "./", ".autoskillit/")
 
+# Position of the plan file among path-like tokens for each skill.
+_PLAN_PATH_POSITION: dict[str, int] = {
```
Trecek, line 55: [warning] cohesion: _PLAN_PATH_POSITION encodes per-skill argument position knowledge that is now formally captured by InputSpec.position in the contract layer. Every future skill addition or argument reordering requires updating both this dict and skill_contracts.yaml. The position semantics that InputSpec.position was introduced to own canonically are duplicated here.

tests/server/test_tools_execution_input_gates.py
```diff
@@ -316,6 +316,189 @@ async def test_gate_still_fires_for_implement_skill(self, tool_ctx_kitchen_open,
...
+        result = _check_input_contracts(
+            f"/resolve-failures {worktree} {mangled} main",
+            str(tmp_path),
+            resolver,
+        )
+        assert result is not None
+        parsed = json.loads(result)
+        assert parsed["success"] is False
+
+    def test_guard_fires_when_called_directly(self, tmp_path):
```
Trecek: [warning] tests: test_guard_fires_when_called_directly is a near-duplicate of test_file_path_input_rejects_nonexistent_path: both call _check_input_contracts with the same nonexistent paths and assert result is not None. No additional assertion, no new coverage. Remove it.

tests/server/test_tools_execution_input_gates.py
```diff
@@ -316,6 +316,189 @@ async def test_gate_still_fires_for_implement_skill(self, tool_ctx_kitchen_open,
...
+        assert specs[2] == InputSpec(
+            name="diagnosis_path", type="file_path", required=False, position=2
+        )
+
+    def test_input_contract_resolver_returns_empty_for_unknown_skill(self):
+        resolver = _make_input_contract_resolver()
+        specs = resolver("/unknown-skill-not-in-contracts /some/path")
+        assert list(specs) == []
+
+    def test_input_contract_resolver_skips_string_type_inputs(self):
```
Trecek: [warning] tests: test_input_contract_resolver_skips_string_type_inputs calls the resolver with the same command as test_input_contract_resolver_returns_specs_for_resolve_failures and only checks base_branch is absent from names. This is already implied by the exhaustive equality check in the first test. Remove the redundant test.

tests/server/test_guards_module.py
```diff
@@ -30,3 +30,10 @@ def test_require_fleet_doc_mentions_l3():
     assert "L3" in doc
     assert "L1" in doc
     assert "L2" in doc
+
+
+def test_check_input_contracts_has_docstring():
```
Trecek: [warning] slop: test_check_input_contracts_has_docstring only verifies the docstring contains the word contract. The other tests here check semantically load-bearing content (L1/L2/L3 tier keywords). This test adds no behavioral protection. Remove it.

Trecek (commented):
AutoSkillit review: 2 critical blocking findings detected. See inline comments.

### ⚠️ Outside Diff Range

These findings target lines not in the pre-computed diff range and could not be posted as inline comments:

**src/autoskillit/recipe/_contracts_manifest.py**
- **L10** [warning/slop]: Dead `if TYPE_CHECKING: pass` block. The body is a bare `pass` with no imports, making the entire block a no-op. Remove it.

**tests/server/test_tools_run_skill_retry.py**
- **L140** [warning/tests]: test_context_limit_result_is_actionable sets `tool_ctx_kitchen_open.input_contract_resolver = None` via direct attribute mutation. Other tests in the same class that do not explicitly set this field will inherit whatever `input_contract_resolver` the fixture wires by default. If the real resolver later rejects the command used in those tests, they will fail for the wrong reason. Set the field in the fixture or use monkeypatch consistently.

Trecek (commented):
(Empty review - clearing pending state)

src/autoskillit/core/types/_type_results.py
```diff
@@ -146,6 +147,16 @@ class WriteBehaviorSpec:
     expected_when: tuple[str, ...] = ()
 
 
+@dataclass(frozen=True, slots=True)
+class InputSpec:
+    """Input contract specification for a single file_path or directory_path argument."""
+
+    name: str
+    type: Literal["file_path", "directory_path"]
```
Trecek, line 155: **Observation from local review round 0:**

`InputSpec.type` includes `"string"` in its `Literal` union, but the class docstring says it is for file_path or directory_path arguments and `_factory.py` only constructs InputSpec for those types. A spec with `type="string"` silently falls through validation. Either remove `"string"` from the Literal or add an explicit validation branch.

**Evidence:** L155: `type: Literal["file_path", "directory_path", "string"]` — factory at L416 guards `if inp.type in ("file_path", "directory_path")` so string specs never reach InputSpec. _check_input_contracts has no elif for string type so it passes silently.

src/autoskillit/hooks/guards/skill_cmd_guard.py
Trecek, line 55: **Observation from local review round 0:**

`_PLAN_PATH_POSITION` hardcodes per-skill path-argument positions in the hook guard, while `_resolve_input_contracts` in `server/_factory.py` derives the same positional data from SKILL.md contracts via `InputSpec.position`. Two independent sources of truth that can silently diverge.

**Evidence:** Hook is stdlib-only (can't import autoskillit). Positions at L55-60 mirror contract data in skill_contracts.yaml. No automated sync mechanism exists.

tests/server/test_tools_execution_input_gates.py
```diff
@@ -316,6 +316,189 @@ async def test_gate_still_fires_for_implement_skill(self, tool_ctx_kitchen_open,
...
+        specs = resolver("/resolve-failures /worktrees/foo /plans/bar.md main")
+        names = [s.name for s in specs]
+        assert "base_branch" not in names
+
+
+def _make_input_contract_resolver():
+    """Create a concrete InputContractResolver using the bundled manifest."""
+    from autoskillit.recipe._contracts_manifest import resolve_input_specs
+
+    return resolve_input_specs
```
Trecek, line 481: **Observation from local review round 0:**

`_make_input_contract_resolver()` in the test file is a verbatim copy of the `_resolve_input_contracts` closure in `_factory.py`. The duplicated resolution logic can silently diverge.

**Evidence:** Test L499-526 reproduces factory L406-426 line-for-line. Extracting a shared helper would cross the IL-2 recipe -> test boundary. Alternative: structural parity assertion.

src/autoskillit/hooks/guards/skill_cmd_guard.py
Trecek, line 55: **Observation from local review round 1:**

`_PLAN_PATH_POSITION` hardcodes per-skill path-argument positions in the hook guard, while `_resolve_input_contracts` in `server/_factory.py` derives the same positional data from SKILL.md contracts via `InputSpec.position`. These are now two independent sources of truth. If a skill's contract changes, `_PLAN_PATH_POSITION` will silently diverge. The hook is stdlib-only and cannot import from autoskillit, but a maintenance strategy for keeping them in sync should be established.

**Evidence:** The reviewer correctly identifies a real divergence risk: `_PLAN_PATH_POSITION` is a static dict introduced in commit 7d7504f9 that encodes the positional index of the plan file among path-like tokens, while `_resolve_input_contracts` in `server/_factory.py` derives equivalent positional data dynamically from SKILL.md contracts. The values currently agree (e.g., `resolve-failures` plan at position 1 because a directory_path argument occupies position 0), but no test or generator enforces this alignment. The stdlib-only constraint is real and acknowledged — the hook cannot import from autoskillit to read contracts at runtime. However, the fix strategy is a design decision: options include (a) a test in tests/infra/ or tests/contracts/ that imports both the hook dict and the server contract resolver to assert agreement, (b) a build-time generated constant, or (c) an inline comment referencing the SKILL.md contracts as the authoritative source.

Trecek (commented):
Observations accumulated from 2 local review rounds:

tests/server/test_tools_execution_input_gates.py
Trecek: Agreed — fixed in b8fa7055. Added TestInputContractIntegration.test_run_skill_rejects_nonexistent_path_via_input_contract exercising the production wiring at tools_execution.py:338-344.

src/autoskillit/server/_guards.py
Trecek: Agreed — fixed in 2aff6201. Added exc_info=True to the logger.warning call.

src/autoskillit/server/_guards.py
Trecek: Agreed — fixed in 2aff6201. Guarded with `or skill_command` fallback.

src/autoskillit/core/types/_type_results.py
Trecek: Agreed — fixed in 61800e77. Added __post_init__ guard asserting position >= 0.

tests/server/test_tools_execution_input_gates.py
Trecek: Agreed — fixed in 72478f63. Removed test_guard_fires_when_called_directly.

tests/server/test_tools_execution_input_gates.py
Trecek: Agreed — fixed in 72478f63. Removed test_input_contract_resolver_skips_string_type_inputs — coverage already implied by exhaustive equality check.

tests/server/test_guards_module.py
Trecek: Agreed — fixed in 72478f63. Removed test_check_input_contracts_has_docstring — no behavioral protection.

src/autoskillit/server/_guards.py
Trecek: Investigated — this is intentional. is_path_like_token filtering distinguishes path arguments from non-path arguments (e.g. branch names like "main"). InputSpec.position indexes among path-typed tokens specifically to avoid conflating branch names with file paths. All production callers (orchestrator, hook guard) supply absolute paths. The _PATH_PREFIXES contract is consistent between server/_guards.py and hooks/guards/skill_cmd_guard.py.

src/autoskillit/core/types/_type_protocols_execution.py
Trecek: Investigated — this is intentional. Sequence is the correct Protocol return type per variance principles — a Protocol should declare the widest practical return type; implementations can return narrower types (tuple). Narrowing to tuple would restrict future implementations unnecessarily. The WriteExpectedResolver comparison is a false equivalence (returns a single concrete type, not a collection).

src/autoskillit/core/types/_type_results.py
Trecek: Investigated — this is intentional. Already fixed in commit a0b72250 which removed "string" from InputSpec.type Literal union. The current code has `type: Literal["file_path", "directory_path"]`.

tests/server/test_tools_execution_input_gates.py
Trecek: Investigated — this is intentional. Already addressed in commit 03f4f8e9 which extracted resolve_input_specs as a shared importable helper. The test helper _make_input_contract_resolver now simply imports and returns resolve_input_specs.

src/autoskillit/server/_guards.py
Trecek: Valid observation — flagged for design decision. _check_input_contracts uses DI (resolver as parameter) while _check_dry_walkthrough fetches via _get_config(). The DI pattern is arguably better practice but changing _check_dry_walkthrough to match is a larger refactor requiring design review.

src/autoskillit/hooks/guards/skill_cmd_guard.py
Trecek: Valid observation — flagged for design decision. The duplication is architecturally necessary (stdlib-only hook constraint). Options: (a) sync test asserting agreement between hook dict and contract resolver, (b) build-time generated constant, (c) inline comment referencing SKILL.md as authoritative source.

Trecek: Valid observation — flagged for design decision. Same issue as comment on _PLAN_PATH_POSITION: stdlib-only constraint makes duplication necessary. A sync test would address the divergence risk.

Trecek: Valid observation — flagged for design decision. Design decision needed for sync strategy between _PLAN_PATH_POSITION and contract-layer positions. See thread on comment 3325776460 for options.

Issue #3086, opened 2026-05-27
Task 1: [P2-A3-WP1] Add recipe/rules/rules_phoropter_adjacency.py


Trecek:
---
plan_id: 1a30575f-9919-4d24-a9f1-b2c249665250
source_commit: 4841abd935780a2bc93e3a616ada821076b41218
---

## Goal

Create rules_phoropter_adjacency.py and register it in recipe/rules/__init__.py

## Context

- Phase: Decompose plan-visualization into Phoropter Steps and Add Adjacency Validation Rule (Milestone: 2-decompose-plan-visualization-into-phoropter-steps-and-add-adjacency-validation-rule)
- Assignment: P2-A3 (Add recipe/rules/rules_phoropter_adjacency.py)
- Depends on: P1-A1-WP1
- Depended on by: P2-A4-WP1

## Deliverables

- `src/autoskillit/recipe/rules/rules_phoropter_adjacency.py`
- `src/autoskillit/recipe/__init__.py — one new import line registering rules_phoropter_adjacency`

## Technical Steps

1. Read rules_actions.py to confirm the exact import signature: `from autoskillit.core import Severity, get_logger`, `from autoskillit.recipe._analysis import ValidationContext`, `from autoskillit.recipe.registry import RuleFinding, semantic_rule`. Use the same header for the new module.
2. Define the module docstring: 'Semantic validation rules enforcing phoropter step adjacency (dial→apply→synthesize).'
3. Define `_PHOROPTER_PHASES: tuple[str, ...] = ('dial', 'apply', 'synthesize')` as the ordered phase progression constant.
4. Implement `@semantic_rule(name='phoropter-phase-order', description='...', severity=Severity.ERROR)` decorated function `_check_phoropter_phase_order(ctx: ValidationContext) -> list[RuleFinding]`. Logic: iterate `ctx.recipe.steps.items()` in declaration order. Maintain `expected_next: dict[str, str]` mapping each phoropter_family to its expected next phase name (initially 'dial' for unseen families). For each step with `step.phoropter_family` set: look up expected phase; if `step.name` does not equal expected phase, append a `RuleFinding` with rule='phoropter-phase-order', the step_name, and a message naming both the actual and expected phase; otherwise advance expected_next[family] to the next phase in `_PHOROPTER_PHASES`.
[...]

PR #3269, opened 2026-05-29
[FIX] Rectify Pre-Staged Index Contamination: Execution-Layer Immunity (Part A)

Trecek:
## Summary

Pipeline sessions inherit a contaminated git index from prior sessions because no pre-session validation exists at the execution layer. When a prior skill session stages files without committing, the next session inherits those staged files. Skills with blind `git add -A` instructions sweep contamination into commits.

Part A implements the primary execution-layer barriers:
1. **Pre-session index validation** — new `validate_pre_session_index()` in `clone_guard.py`, decoupled from `should_snapshot` so it covers `CLONE_COMMIT_SKILLS`
2. **Selective revert fix** — add `git reset <snapshot_sha>` to `revert_contamination(selective=True)` to clear staged-only entries
3. **`main_repo_guard` fallback fix** — add `git reset HEAD` to the stash-failure fallback path

## Problem

Commit `eac9d652` in PR #3086's worktree was supposed to fix only the `test_no_subpackage_exceeds_10_files` threshold (49→50) and a missing import. Instead it committed 67 files with 1,606 lines of deletions — sweeping in pre-staged index contamination from a prior pipeline step.

## Root Cause

The `resolve-merge-conflicts` session ran a diagnostic sequence that left the git index contaminated:
```bash
git stash && git checkout upstream/develop -- . 2>/dev/null; \
pre-commit run check-contract-freshness --all-files 2>&1; \
git checkout - -- . 2>/dev/null; git stash pop 2>/dev/null
```
- `git checkout upstream/develop -- .` staged ALL files from upstream/develop into the index
- `git checkout - -- .` was supposed to restore the index, but with stderr suppressed any failure was silently swallowed
- This left 50+ files staged with pre-Rectify versions (deletions relative to HEAD)

When the subsequent `resolve-failures` session started, its `git status` showed 50+ files already staged. Per SKILL.md Step 0.5, the agent committed them alongside its legitimate fixes.

## Impact

The contaminated commit deleted code from multiple recent Rectify PRs:
- `rules_merge.py` (86 lines), `rules_skill_content.py` (64 lines) — entire rule files removed
- `_cmd_rpc_guards.py` (102 lines), `test_cmd_rpc.py` (209 lines) — entire modules deleted
- `test_backend_protocol_completeness.py` (59 lines), `test_backend_protocol.py` (101 lines)
- `test_liveness.py` (120 lines reduced), `_liveness.py` (47 lines reduced)
- 40+ additional files with partial deletions

## Two Fixes Needed

1. **Immediate**: Clean up the contaminated branch
2. **Structural**: Prevent index contamination in pipeline handoffs — resolve-merge-conflicts should use `git stash push --keep-index` or explicit ref-based checkout instead of `-`, and resolve-failures SKILL.md Step 0.5 should warn/abort when the pre-staged file count is suspiciously high.

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260528-232352-212742/.autoskillit/temp/rectify/rectify_pre_staged_index_contamination_2026-05-28_235500_part_a.md`

## Token Usage Summary

| investigate* | opus | 1 | 63 | 10.3k | 875.8k | 126.9k | 340 | 57.2k | 7m 58s |
| rectify* | opus[1m] | 1 | 10.3k | 36.8k | 2.1M | 139.6k | 302 | 170.4k | 17m 50s |
| review_approach* | sonnet | 1 | 3.2k | 5.7k | 175.4k | 39.6k | 37 | 26.9k | 3m 14s |
| dry_walkthrough* | opus | 2 | 1.0k | 23.5k | 1.8M | 88.8k | 272 | 124.6k | 14m 35s |
| implement* | sonnet | 2 | 346 | 30.4k | 2.8M | 106.4k | 148 | 143.9k | 10m 10s |
| audit_impl* | sonnet | 3 | 236 | 30.1k | 937.4k | 70.9k | 124 | 127.4k | 11m 31s |
| post_run_diagnostics* | sonnet | 1 | 78 | 17.6k | 270.4k | 82.1k | 276 | 58.1k | 10m 31s |
| prepare_pr* | sonnet | 1 | 146.5k | 6.2k | 308.7k | 30.9k | 24 | 43.6k | 1m 47s |
| compose_pr* | sonnet | 1 | 80.3k | 2.2k | 275.0k | 30.9k | 21 | 15.5k | 45s |
| review_pr* | sonnet | 3 | 2.1k | 137.8k | 5.4M | 156.0k | 224 | 349.9k | 39m 20s |
| resolve_review* | opus | 3 | 233 | 82.7k | 9.3M | 130.4k | 281 | 409.3k | 46m 3s |
| resolve_pre_review_conflicts* | opus | 1 | 44 | 11.1k | 1.5M | 85.0k | 51 | 85.5k | 3m 8s |
| **Total** | | | 244.5k | 394.3k | 25.7M | 156.0k | | 1.6M | 2h 47m |

## Token Efficiency

| rectify | 0 | — | — | — |
| review_approach | 0 | — | — | — |
| implement | 228 | 12253.3 | 631.3 | 133.1 |
| post_run_diagnostics | 0 | — | — | — |
| review_pr | 0 | — | — | — |
| resolve_review | 137 | 68016.0 | 2987.4 | 603.4 |
| resolve_pre_review_conflicts | 6082 | 251.8 | 14.1 | 1.8 |
| **Total** | **6447** | 3987.6 | 250.1 | 61.2 |

## Model Usage Breakdown

| opus | 4 | 1.4k | 127.6k | 13.5M | 676.6k | 1h 11m |
| opus[1m] | 1 | 10.3k | 36.8k | 2.1M | 170.4k | 17m 50s |
| sonnet | 7 | 232.8k | 229.9k | 10.1M | 765.2k | 1h 17m |

### DIFF
--- a/src/autoskillit/execution/clone_guard.py
+++ b/src/autoskillit/execution/clone_guard.py
@@ -428,11 +428,13 @@ async def revert_contamination(
             if reset_result.returncode != 0:
                 return dataclasses.replace(report, reverted=False)
         else:
-            await runner(
+            reset_result = await runner(
                 ["git", "reset", snapshot.head_sha],
                 cwd=Path(cwd),
                 timeout=_GIT_TIMEOUT,
             )
+            if reset_result.returncode != 0:
+                return dataclasses.replace(report, reverted=False)
         checkout_result = await runner(
             ["git", "checkout", "--", "."],
             cwd=Path(cwd),
--- a/src/autoskillit/execution/clone_guard.py
+++ b/src/autoskillit/execution/clone_guard.py
@@ -7,6 +7,7 @@ prevent contamination from propagating to retry sessions.
 Public API:
     is_worktree_skill(skill_command) -> bool
     snapshot_clone_state(cwd, runner) -> CloneSnapshot | None
+    validate_pre_session_index(cwd, runner) -> bool
     check_and_revert_clone_contamination(
         snapshot, skill_result, cwd, runner, audit
     ) -> tuple[SkillResult, bool]
@@ -226,6 +227,83 @@ async def snapshot_clone_state(cwd: str, runner: SubprocessRunner) -> CloneSnaps
     return CloneSnapshot(head_sha=head_sha, worktree_set=wt_set)
 
 
+async def validate_pre_session_index(
+    cwd: str,
+    runner: SubprocessRunner,
+    *,
+    exclude_prefix: str = ".autoskillit/",
+    pre_session_sha: str = "",
+) -> bool:
+    """Reset dirty index state from prior sessions before a new session starts.
+
+    Returns True if dirty state was found and reset, False if index was clean.
+    Uses pre_session_sha for mixed reset when available; falls back to
+    ``git rm --cached -r .`` on repos with no HEAD.
+    """
+    status_result = await runner(
+        ["git", "status", "--porcelain"],
+        cwd=Path(cwd),
+        timeout=_GIT_TIMEOUT,
+    )
+    if status_result.returncode != 0:
+        return False
+    status_lines = [line for line in status_result.stdout.splitlines() if line.strip()]
+    if exclude_prefix:
+        status_lines = [
+            line for line in status_lines if not _status_path_under_prefix(line, exclude_prefix)
+        ]
+    if not status_lines:
+        return False
+
+    file_count = len(status_lines)
+    logger.warning(
+        "pre_session_dirty_index_detected",
+        file_count=file_count,
+        sample_files=status_lines[:5],
+    )
+
+    reset_sha = pre_session_sha
+    if not reset_sha:
+        head_result = await runner(
+            ["git", "rev-parse", "HEAD"],
+            cwd=Path(cwd),
+            timeout=_GIT_TIMEOUT,
+        )
+        if head_result.returncode == 0 and head_result.stdout.strip():
+            reset_sha = head_result.stdout.strip()
+
+    if reset_sha:
+        await runner(
+            ["git", "reset", reset_sha],
+            cwd=Path(cwd),
+            timeout=_GIT_TIMEOUT,
+        )
+    else:
+        await runner(
+            ["git", "rm", "--cached", "-r", "."],
+            cwd=Path(cwd),
+            timeout=_GIT_TIMEOUT,
+        )
+
+    await runner(
+        ["git", "checkout", "--", "."],
+        cwd=Path(cwd),
+        timeout=_GIT_TIMEOUT,
+    )
+    await runner(
+        ["git", "clean", "-fd"],
+        cwd=Path(cwd),
+        timeout=_GIT_TIMEOUT,
+    )
+
+    logger.info(
+        "index_reset_pre_session",
+        file_count=file_count,
+        reset_strategy="sha" if reset_sha else "rm_cached",
+    )
+    return True
+
+
 def _status_path_under_prefix(status_line: str, prefix: str) -> bool:
     """Return True if the file path in a git status --porcelain line is under prefix."""
     path_part = status_line[3:]  # Skip "XY " status chars
@@ -325,6 +403,12 @@ async def revert_contamination(
             )
             if reset_result.returncode != 0:
                 return dataclasses.replace(report, reverted=False)
+        else:
+            await runner(
+                ["git", "reset", snapshot.head_sha],
+                cwd=Path(cwd),
+                timeout=_GIT_TIMEOUT,
+            )
         checkout_result = await runner(
             ["git", "checkout", "--", "."],
             cwd=Path(cwd),
--- a/src/autoskillit/execution/clone_guard.py
+++ b/src/autoskillit/execution/clone_guard.py
@@ -237,6 +237,7 @@ async def validate_pre_session_index(
     """Reset dirty index state from prior sessions before a new session starts.
 
     Returns True if dirty state was found and reset, False if index was clean.
+    Raises RuntimeError if git infrastructure commands fail.
     Uses pre_session_sha for mixed reset when available; falls back to
     ``git rm --cached -r .`` on repos with no HEAD.
     """
@@ -246,7 +247,9 @@ async def validate_pre_session_index(
         timeout=_GIT_TIMEOUT,
     )
     if status_result.returncode != 0:
-        return False
+        raise RuntimeError(
+            f"git status failed (rc={status_result.returncode}): {status_result.stderr.strip()}"
+        )
     status_lines = [line for line in status_result.stdout.splitlines() if line.strip()]
     if exclude_prefix:
         status_lines = [
@@ -273,17 +276,26 @@ async def validate_pre_session_index(
             reset_sha = head_result.stdout.strip()
 
     if reset_sha:
-        await runner(
+        reset_result = await runner(
             ["git", "reset", reset_sha],
             cwd=Path(cwd),
             timeout=_GIT_TIMEOUT,
         )
     else:
-        await runner(
+        reset_result = await runner(
             ["git", "rm", "--cached", "-r", "."],
             cwd=Path(cwd),
             timeout=_GIT_TIMEOUT,
         )
+
+    if reset_result.returncode != 0:
+        logger.warning(
+            "pre_session_reset_failed",
+            returncode=reset_result.returncode,
+            stderr=reset_result.stderr.strip(),
+            strategy="sha" if reset_sha else "rm_cached",
+        )
+        return False
 
     await runner(
         ["git", "checkout", "--", "."],
--- a/src/autoskillit/execution/clone_guard.py
+++ b/src/autoskillit/execution/clone_guard.py
@@ -297,16 +297,28 @@ async def validate_pre_session_index(
         )
         return False
 
-    await runner(
+    co_result = await runner(
         ["git", "checkout", "--", "."],
         cwd=Path(cwd),
         timeout=_GIT_TIMEOUT,
     )
-    await runner(
+    if co_result.returncode != 0:
+        logger.warning(
+            "pre_session_checkout_failed",
+            returncode=co_result.returncode,
+            stderr=co_result.stderr.strip(),
+        )
+    clean_result = await runner(
         ["git", "clean", "-fd"],
         cwd=Path(cwd),
         timeout=_GIT_TIMEOUT,
     )
+    if clean_result.returncode != 0:
+        logger.warning(
+            "pre_session_clean_failed",
+            returncode=clean_result.returncode,
+            stderr=clean_result.stderr.strip(),
+        )
 
     logger.info(
         "index_reset_pre_session",
--- a/src/autoskillit/execution/headless/_headless_execute.py
+++ b/src/autoskillit/execution/headless/_headless_execute.py
@@ -28,6 +28,7 @@ from autoskillit.core import (
     collect_version_snapshot,
     get_logger,
     is_feature_enabled,
+    is_git_main_checkout,
     is_git_worktree,
     is_in_git_repo,
 )
@@ -41,6 +42,7 @@ from autoskillit.execution.clone_guard import (
     is_path_under_exclude,
     is_worktree_skill,
     snapshot_clone_state,
+    validate_pre_session_index,
 )
 from autoskillit.execution.headless._headless_evidence import (
     _build_error_path_telemetry,
@@ -192,6 +194,18 @@ async def _execute_claude_headless(
         and _clone_guard_policy.should_snapshot
     ):
         _clone_snapshot = await snapshot_clone_state(cwd, runner)
+
+    if not skip_clone_guard and is_git_main_checkout(Path(cwd)):
+        _pre_sha = _clone_snapshot.head_sha if _clone_snapshot else ""
+        try:
+            await validate_pre_session_index(
+                cwd,
+                runner,
+                pre_session_sha=_pre_sha,
+                exclude_prefix=_derived_prefix or GUARD_EXCLUDE_PREFIX,
+            )
+        except Exception:
+            logger.debug("validate_pre_session_index_failed", exc_info=True)
 
     _watch_dirs: list[Path] = list(write_watch_dirs) if write_watch_dirs else []
     if not _watch_dirs:
--- a/src/autoskillit/execution/headless/_headless_execute.py
+++ b/src/autoskillit/execution/headless/_headless_execute.py
@@ -205,7 +205,7 @@ async def _execute_claude_headless(
                 exclude_prefix=_derived_prefix or GUARD_EXCLUDE_PREFIX,
             )
         except Exception:
-            logger.debug("validate_pre_session_index_failed", exc_info=True)
+            logger.warning("validate_pre_session_index_failed", exc_info=True)
 
     _watch_dirs: list[Path] = list(write_watch_dirs) if write_watch_dirs else []
     if not _watch_dirs:
--- a/src/autoskillit/recipe/_cmd_rpc_guards.py
+++ b/src/autoskillit/recipe/_cmd_rpc_guards.py
@@ -152,6 +152,20 @@ def _count_numstat_net(numstat_output: str) -> int:
     return total
 
 
+def _parse_numstat_per_file(numstat_output: str) -> dict[str, int]:
+    """Parse per-file net line changes from git diff --numstat output."""
+    result: dict[str, int] = {}
+    for line in numstat_output.splitlines():
+        parts = line.split("\t", 2)
+        if len(parts) < 3:
+            continue
+        try:
+            result[parts[2]] = int(parts[0]) - int(parts[1])
+        except ValueError:
+            continue
+    return result
+
+
 def _check_regression(
     worktree_path: str,
     files_to_add: list[str],
@@ -189,14 +203,13 @@ def _check_regression(
     if regression_lines <= 10:
         return None  # within tolerance
 
+    committed_per_file = _parse_numstat_per_file(committed_diff.stdout)
+    wt_per_file = _parse_numstat_per_file(wt_diff.stdout)
+
     reverted_files: list[str] = []
     for f in files_to_add:
-        c_diff = run_git(["diff", "--numstat", merge_base, "HEAD", "--", f], cwd=worktree_path)
-        w_diff = run_git(["diff", "--numstat", merge_base, "--", f], cwd=worktree_path)
-        if c_diff.returncode != 0 or w_diff.returncode != 0:
-            continue
-        c_net = _count_numstat_net(c_diff.stdout)
-        w_net = _count_numstat_net(w_diff.stdout)
+        c_net = committed_per_file.get(f, 0)
+        w_net = wt_per_file.get(f, 0)
         if c_net - w_net > 5:
             reverted_files.append(f)
 
--- a/src/autoskillit/recipe/_cmd_rpc_guards.py
+++ b/src/autoskillit/recipe/_cmd_rpc_guards.py
@@ -132,7 +132,78 @@ def main_repo_guard(clone_path: str) -> dict[str, str]:
     return {"cleaned": "true"}
 
 
-def commit_guard(worktree_path: str) -> dict[str, str]:
+def _count_numstat_net(numstat_output: str) -> int:
+    """Sum net line changes (insertions - deletions) from git diff --numstat output."""
+    total = 0
+    for line in numstat_output.splitlines():
+        parts = line.split("\t", 2)
+        if len(parts) < 2:
+            continue
+        try:
+            total += int(parts[0]) - int(parts[1])
+        except ValueError:
+            continue  # binary files show "-"
+    return total
+
+
+def _check_regression(
+    worktree_path: str,
+    files_to_add: list[str],
+    base_branch: str,
+) -> dict[str, str] | None:
+    """Detect if pending changes would revert implementation commits.
+
+    Returns a regression dict if the pending dirty files would net-revert
+    more than 10 implementation lines. Returns None to proceed normally.
+    """
+    mb = run_git(["merge-base", "HEAD", base_branch], cwd=worktree_path)
+    if mb.returncode != 0:
+        return None  # fresh repo, no merge-base
+    merge_base = mb.stdout.strip()
+
+    committed_diff = run_git(
+        ["diff", "--numstat", merge_base, "HEAD", "--"] + files_to_add,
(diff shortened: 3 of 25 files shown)


src/autoskillit/execution/clone_guard.py
```diff
@@ -226,6 +227,107 @@ async def snapshot_clone_state(cwd: str, runner: SubprocessRunner) -> CloneSnaps
...
+        reset_result = await runner(
+            ["git", "reset", reset_sha],
+            cwd=Path(cwd),
+            timeout=_GIT_TIMEOUT,
+        )
+    else:
+        reset_result = await runner(
+            ["git", "rm", "--cached", "-r", "."],
+            cwd=Path(cwd),
+            timeout=_GIT_TIMEOUT,
```
Trecek, line 288: [critical] bugs: `git checkout -- .` only restores tracked file modifications; it does NOT unstage or remove staged-new files added by a prior session (`A ` status in porcelain). In the `reset_sha` branch, `git reset reset_sha` (mixed mode) correctly unstages these first, making checkout safe. However this ordering dependency is fragile and undocumented — future callers relying on checkout to clear staged-new entries will be silently wrong.

src/autoskillit/execution/clone_guard.py
Trecek, line 1: [critical] bugs (L408 — outside diff hunk): New else branch calls git reset snapshot.head_sha without checking the returncode. Unlike the direct_commits branch above, this branch silently ignores reset failure. If snapshot.head_sha is empty string (default in CloneSnapshot), git reset fails and contamination is unreported.

src/autoskillit/recipe/_cmd_rpc_guards.py
```diff
@@ -131,7 +138,91 @@ def main_repo_guard(clone_path: str) -> dict[str, str]:
...
+    return result
+
+
+def _check_regression(
+    worktree_path: str,
+    files_to_add: list[str],
+    base_branch: str,
+) -> dict[str, str] | None:
+    """Detect if pending changes would revert implementation commits.
+
```
Trecek, line 175: [critical] bugs: `git diff --numstat merge_base -- files_to_add` compares merge-base to the working tree, not the staged index. In the contamination scenario where a prior session staged reverted content (index differs from working tree), the working-tree diff may not reflect the staged regression, causing `_check_regression` to return None and allow contamination through. Should use `git diff --cached --numstat merge_base -- files_to_add` or check both.

src/autoskillit/execution/clone_guard.py
```diff
@@ -226,6 +227,107 @@ async def snapshot_clone_state(cwd: str, runner: SubprocessRunner) -> CloneSnaps
...
+
+    reset_sha = pre_session_sha
+    if not reset_sha:
+        head_result = await runner(
+            ["git", "rev-parse", "HEAD"],
+            cwd=Path(cwd),
+            timeout=_GIT_TIMEOUT,
+        )
+        if head_result.returncode == 0 and head_result.stdout.strip():
+            reset_sha = head_result.stdout.strip()
```
Trecek, line 276: [warning] bugs: Return value of `git reset reset_sha` is not checked. If reset fails (e.g., invalid SHA), the function continues to `git checkout -- .` and `git clean -fd` against an unreset index, and returns True claiming the index was cleaned.

src/autoskillit/execution/clone_guard.py
```diff
@@ -226,6 +227,107 @@ async def snapshot_clone_state(cwd: str, runner: SubprocessRunner) -> CloneSnaps
...
+
+    if reset_sha:
+        reset_result = await runner(
+            ["git", "reset", reset_sha],
+            cwd=Path(cwd),
+            timeout=_GIT_TIMEOUT,
+        )
```
Trecek, line 283: [warning] bugs: Return value of `git rm --cached -r .` (no-HEAD fallback) is not checked. On failure, the subsequent `git checkout -- .` will also fail (no HEAD), leaving the index dirty while the function returns True.

src/autoskillit/execution/clone_guard.py
```diff
@@ -226,6 +227,107 @@ async def snapshot_clone_state(cwd: str, runner: SubprocessRunner) -> CloneSnaps
...
+    else:
+        reset_result = await runner(
+            ["git", "rm", "--cached", "-r", "."],
+            cwd=Path(cwd),
+            timeout=_GIT_TIMEOUT,
+        )
+
+    if reset_result.returncode != 0:
+        logger.warning(
+            "pre_session_reset_failed",
```
Trecek, line 293: [warning] defense: Return value of `git clean -fd` is not checked. Untracked contamination files that fail to be removed are silently ignored and True is returned to the caller.

src/autoskillit/execution/headless/_headless_execute.py
```diff
@@ -193,6 +195,18 @@ async def _execute_claude_headless(
...
 
+    if not skip_clone_guard and is_git_main_checkout(Path(cwd)):
+        _pre_sha = _clone_snapshot.head_sha if _clone_snapshot else ""
+        try:
+            await validate_pre_session_index(
+                cwd,
+                runner,
+                pre_session_sha=_pre_sha,
+                exclude_prefix=_derived_prefix or GUARD_EXCLUDE_PREFIX,
+            )
```
Trecek, line 206: [warning] defense: Bare `except Exception` swallows all errors from `validate_pre_session_index` and logs only at DEBUG level. Infrastructure failures (runner timeout, subprocess spawn failure) are indistinguishable from no-op paths, breaking observability for cross-session contamination failures.

src/autoskillit/recipe/rules/rules_skill_content.py
```diff
@@ -200,6 +200,56 @@ def _check_hardcoded_origin_remote(ctx: ValidationContext) -> list[RuleFinding]:
     return findings
 
 
+_BLIND_GIT_ADD_RE = re.compile(
```
Trecek, line 203: [warning] bugs: `_BLIND_GIT_ADD_RE` uses `\n` in the alternation group, but the regex is applied via `.search(line)` on strings from `block.splitlines()` — a literal newline never appears in such strings. The `\n` alternative is dead code. `git add .` at line-end is caught by `$` (end-of-string), not by `\n`.

tests/execution/test_clone_guard_pre_session.py
```diff
@@ -0,0 +1,295 @@
...
+    async def __call__(
+        self,
+        cmd: list[str],
+        *,
+        cwd: Path,
+        timeout: float,
+        env: Mapping[str, str] | None = None,
+        **kwargs: Any,
+    ) -> SubprocessResult:
+        run_env = dict(env) if env is not None else _GIT_ENV
```
Trecek, line 84: [warning] tests: `_RealAsyncRunner.__call__` has inverted env logic: `run_env = dict(env) if env is not None else _GIT_ENV` uses the passed `env` when not None, but since `validate_pre_session_index` never passes `env`, this branch is never reached in current tests. The `else _GIT_ENV` fallback fires for all actual calls.

tests/execution/test_clone_guard_pre_session.py
```diff
@@ -0,0 +1,295 @@
...
+            await proc.communicate()
+            return SubprocessResult(
+                returncode=-1,
+                stdout="",
+                stderr="timeout",
+                termination=TerminationReason.TIMED_OUT,
+                pid=proc.pid,
+            )
+        return SubprocessResult(
+            returncode=proc.returncode if proc.returncode is not None else -1,
```
Trecek, line 107: [warning] tests: When `proc.returncode` is None after `communicate()`, the runner returns `0` (success). `asyncio.create_subprocess_exec` guarantees returncode is set after `communicate()`, so this branch is unreachable, but silently maps a process error to success if ever reached.

tests/execution/test_clone_guard_pre_session.py
```diff
@@ -0,0 +1,295 @@
...
+    _git(["git", "commit", "-m", "add files"], repo)
+    sha = _head_sha(repo)
+    (repo / "a.py").write_text("modified a staged")
+    _git(["git", "add", "a.py"], repo)
+    (repo / "b.py").write_text("modified b unstaged")
+
+    snapshot = CloneSnapshot(head_sha=sha)
+    report = ContaminationReport(
+        pre_sha=sha,
+        post_sha=sha,
```
Trecek, line 269: [warning] tests: Uses `_git(["git", "add", "."], repo)` — a blind `git add .` in test setup — the exact anti-pattern this PR is eliminating from skill SKILL.md files.

tests/skills/test_generate_report_guards.py
```diff
@@ -0,0 +1,13 @@
...
+
+from autoskillit.core.paths import pkg_root
+
+SKILL_MD = pkg_root() / "skills_extended" / "generate-report" / "SKILL.md"
+
+
+def test_generate_report_no_blind_add() -> None:
+    """generate-report/SKILL.md must not contain blind 'git add -A'."""
+    text = SKILL_MD.read_text()
+    assert "add -A" not in text, "generate-report/SKILL.md still contains 'add -A'"
```
Trecek, line 13: [warning] tests: Guard checks only that `'add -A'` is absent, missing protection against `git add --all` and `git add .` — the other two blind-add forms caught by the `blind-git-add-in-skill` rule. Same gap in `test_resolve_review_guards.py` and `test_retry_worktree_guards.py`.

tests/skills/test_resolve_failures_guards.py
```diff
@@ -108,3 +108,18 @@ def test_resolve_failures_fix_loop_instructs_stdout_discard() -> None:
         "resolve-failures Step 3 must explicitly instruct discarding full pytest stdout "
         "after extracting failure info. This prevents context bloat across fix iterations."
     )
+
+
+# --- Blind git add guard ---
+
+
+def test_resolve_failures_no_blind_add() -> None:
```
Trecek, line 116: [warning] tests: `assert 'add --' in text` is a weak positive check — it passes if `add --` appears anywhere in the file including prose commentary, even if blind-add forms are still present in bash blocks.

tests/execution/test_clone_guard_pre_session.py
```diff
@@ -0,0 +1,295 @@
...
+        uncommitted_files=["M  somefile.py"],
+        direct_commits=False,
+        reverted=False,
+    )
+    runner = _RealAsyncRunner()
+    result = await revert_contamination(snapshot, report, str(repo), runner, selective=True)
+
+    assert result.reverted
+    assert _git_status(repo) == ""
+
```
Trecek, line 250: [warning] tests: `ContaminationReport(uncommitted_files=["M  somefile.py"])` uses the two-space porcelain staged format. If `revert_contamination` parses this field differently than raw `git status --porcelain` output, the test could pass with semantically wrong data.

tests/skills/test_generate_report_guards.py
```diff
@@ -0,0 +1,13 @@
+"""Guards for generate-report SKILL.md: blind git add prevention."""
+
+from __future__ import annotations
+
+from autoskillit.core.paths import pkg_root
+
+SKILL_MD = pkg_root() / "skills_extended" / "generate-report" / "SKILL.md"
+
+
+def test_generate_report_no_blind_add() -> None:
```
Trecek, line 10: [warning] cohesion: Three new skills (generate-report, resolve-review, retry-worktree) get standalone 13-line guard files, while resolve-failures gets its blind-add guard as an inline extension. Asymmetric pattern: either all four should be standalone guard files or all four should follow the inline-extension pattern.

src/autoskillit/recipe/_cmd_rpc_guards.py
Trecek, line 1: [warning] cohesion (L135 — outside diff hunk): `_count_numstat_net` mixes abstraction levels: `numstat` (a git subcommand) vs `net` (a computed metric). The sole caller uses `committed_net`/`wt_net`. A name like `_sum_numstat_net` or `_net_line_count` would align better.

tests/execution/test_clone_guard_pre_session.py
Trecek, line 1: [warning] tests (L310 — outside diff hunk): Disjunctive assertion `not evil.py.exists() OR ls-files returncode != 0` passes even if `evil.py` exists on disk but is merely unstaged. A stricter assertion should verify both that the file is absent from the index AND not present as untracked.

Trecek (commented):
AutoSkillit review: blocking issues found. See inline comments — changes required before merge.

3 critical findings and 14 warning findings were posted as inline comments. Critical issues:
- `clone_guard.py:288` — `git checkout -- .` ordering dependency undocumented for staged-new files
- `clone_guard.py:408` — new `else` branch lacks returncode check, allows silent contamination on empty head_sha  
- `_cmd_rpc_guards.py:175` — `_check_regression` diffs working tree instead of staged index, misses staged-only regressions

src/autoskillit/execution/clone_guard.py
Trecek, line 288: **Observation from local review round 1:**

validate_pre_session_index is named as an index-scoped operation but unconditionally runs git checkout -- . and git clean -fd, wiping working-tree changes and removing untracked files. The scope mismatch between name/docstring and implementation is an architectural contract violation.

**Evidence:** The function's purpose is to clean all contamination from prior sessions, which includes working-tree state. The name could be improved but the implementation is intentionally broader. Renaming is a separate refactor.

src/autoskillit/skills_extended/resolve-review/SKILL.md
```diff
@@ -67,7 +67,7 @@ normal commit protocol.
 
 **Before every test run and before emitting structured output tokens:**
 1. Run `git -C {work_dir} status --porcelain`
-2. If any files are dirty: `git -C {work_dir} add -A && git -C {work_dir} commit -m "fix: commit pending review changes"`
+2. If any files are dirty: `git -C {work_dir} add -- <files you modified> && git -C {work_dir} commit -m "fix: commit pending review changes"`
```
Trecek, line 70: **Observation from local review round 1:**

Uses {work_dir} as the worktree path placeholder while generate-report and resolve-failures use {worktree_path} and retry-worktree uses {WORKTREE_PATH}. Three distinct names for the same concept across the skill suite.

**Evidence:** SKILL.md placeholders are human-readable documentation markers, not code-interpolated. Normalizing requires coordinated changes across multiple skill files.

src/autoskillit/skills_extended/retry-worktree/SKILL.md
```diff
@@ -62,7 +62,7 @@ not yet committed. The recipe routes to `on_context_limit`, preserving the workt
 
 **Before emitting structured output tokens:**
 1. Run `git -C {WORKTREE_PATH} status --porcelain`
-2. If any files are dirty: `git -C {WORKTREE_PATH} add -A && git -C {WORKTREE_PATH} commit -m "chore: commit pending changes before context limit"`
+2. If any files are dirty: `git -C {WORKTREE_PATH} add -- <files you modified> && git -C {WORKTREE_PATH} commit -m "chore: commit pending changes before context limit"`
```
Trecek, line 65: **Observation from local review round 1:**

Uses {WORKTREE_PATH} (uppercase) as the worktree path placeholder while resolve-failures and generate-report use {worktree_path} (lowercase). Normalizing the placeholder convention was the natural opportunity.

**Evidence:** Same as above — SKILL.md placeholders are documentation-only. Requires coordinated normalization across the skill suite.

Trecek (commented):
Observations accumulated from 1 local review rounds:

tests/skills/test_resolve_failures_guards.py
```diff
@@ -108,3 +108,18 @@ def test_resolve_failures_fix_loop_instructs_stdout_discard() -> None:
...
+    """resolve-failures/SKILL.md must not use blind 'git add -A' as a command."""
+    text = _skill_text()
+    for line in text.splitlines():
+        stripped = line.strip()
+        if "add -A" in stripped and "never use" not in stripped.lower():
+            raise AssertionError(
+                f"resolve-failures/SKILL.md still uses 'add -A' as a command: {stripped!r}"
+            )
```
Trecek, line 124: **Observation from local review round 1:**

assert 'add --' in text verifies only that the literal string 'add --' appears anywhere in the SKILL.md — satisfied by prose documentation rather than an executable command. A more targeted check matching a scoped-add command pattern would be meaningful.

**Evidence:** The assertion is a contract guard ensuring the SKILL.md mentions scoped adds. While it could be more targeted, the current check serves its purpose. A regex-based match in a bash block would be stronger.

src/autoskillit/execution/clone_guard.py
Trecek: Agreed — fixed in 005bae13b1b64665716bba0c8b0fc66ed16ee10d. Stored the result of `git reset snapshot.head_sha` and added returncode check with early `return dataclasses.replace(report, reverted=False)` on failure, matching the direct_commits branch above.

src/autoskillit/recipe/rules/rules_skill_content.py
Trecek: Agreed — fixed in 2fc970a177d5607dcab9ec2eba20307c44f48096. Removed the dead `\n` alternative from the alternation group since the regex is applied on splitlines() output.

tests/skills/test_generate_report_guards.py
Trecek: Agreed — fixed in . Added assertions for `add --all` and `git add .` across all three guard files (generate-report, resolve-review, retry-worktree).

src/autoskillit/execution/clone_guard.py
Trecek: Investigated — this is intentional. Commit a74c797e stores the result in `reset_result` (lines 279-289) and checks `reset_result.returncode != 0` at lines 291-298, returning False on failure. Both the `git reset reset_sha` and `git rm --cached` branches flow into the same returncode check.

src/autoskillit/execution/clone_guard.py
Trecek: Investigated — this is intentional. The `git rm --cached -r .` result is stored in `reset_result` (lines 285-289, same variable as the SHA branch) and checked at lines 291-298. Commit a74c797e added both storage and the shared returncode check covering both branches.

src/autoskillit/execution/headless/_headless_execute.py
Trecek: Investigated — this is intentional. The exception handler at line 207-208 uses `logger.warning("validate_pre_session_index_failed", exc_info=True)`, not DEBUG. Commit a74c797e explicitly raised the log level from DEBUG to WARNING for this handler. The `exc_info=True` parameter ensures the full traceback is captured.

tests/execution/test_clone_guard_pre_session.py
Trecek: Investigated — this is intentional. `run_env = dict(env) if env is not None else _GIT_ENV` is standard optional-override fallback: when `env` is explicitly passed, use it; when absent (None), fall back to the module-level `_GIT_ENV` constant. `validate_pre_session_index` intentionally does not pass `env`, so the controlled `_GIT_ENV` is always used — correct test behavior.

tests/execution/test_clone_guard_pre_session.py
Trecek: Investigated — this is intentional. The code at line 107 returns `-1` for None returncode (`proc.returncode if proc.returncode is not None else -1`), not `0`. Commit a74c797e changed `else 0` to `else -1` in a prior review iteration.

tests/execution/test_clone_guard_pre_session.py
Trecek: Acknowledged — minor suggestion noted.

tests/execution/test_clone_guard_pre_session.py
Trecek: Investigated — this is intentional. `revert_contamination` only uses `len(report.uncommitted_files)` for logging at line 418 and never parses or iterates the string contents. The revert logic issues direct git commands (reset, checkout, clean) regardless of file list contents, so the porcelain format is irrelevant to test correctness.

tests/skills/test_generate_report_guards.py
Trecek: Investigated — this is intentional. The asymmetry is by design: generate-report, resolve-review, and retry-worktree each got new standalone guard files because no prior guard file existed. `test_resolve_failures_guards.py` already existed (6 polling-cascade and output-bloat tests from a prior PR), so the blind-add guard was appended inline. New skills get new files; existing guard files get extended.

tests/execution/test_clone_guard_pre_session.py
Trecek: Investigated — this is intentional. The current code at lines 293-295 has three independent `assert` statements (conjunctive AND), not a disjunctive OR. Commit e41da12c explicitly replaced the old disjunctive `or` form with two separate assert statements: `assert not (repo / "evil.py").exists()` and `assert _git(["git", "ls-files", ...]).returncode != 0`.

src/autoskillit/execution/clone_guard.py
Trecek: Valid observation — flagged for design decision. The ordering (reset → checkout → clean) is functionally correct: mixed `git reset` unstages new entries before `checkout -- .` restores tracked modifications. Whether to add an inline comment documenting this dependency is a design question for the author.

src/autoskillit/recipe/_cmd_rpc_guards.py
Trecek: Valid observation — flagged for design decision. `git diff --numstat merge_base -- files` compares to working tree, not staged index. However, the practical gap is limited: `commit_guard` calls `git add -- files_to_add` immediately after, overwriting staged content with working-tree state. A staged-only revert that differs from the working tree would be clobbered by the add. Whether to add `--cached` for semantic completeness is a design question.

src/autoskillit/execution/clone_guard.py
Trecek: Acknowledged — minor suggestion noted.

tests/skills/test_resolve_failures_guards.py
Trecek: Acknowledged — minor suggestion noted.

src/autoskillit/recipe/_cmd_rpc_guards.py
Trecek: Acknowledged — minor suggestion noted.

Issue #3462, opened 2026-05-31
Ingredient naming asymmetry causes orchestrator confusion (audit vs audit_impl)


Trecek:
## Problem

During a `process-issues` pipeline dispatching Group 1 issues (6 parallel dispatches), the orchestrator AI used `audit_impl` as the ingredient key when the correct name is `audit`. The dispatches failed because `audit_impl` is not a valid ingredient — it is the **step name**, not the ingredient name. The AI had to retry all 6 dispatches after discovering the error.

### Root Cause

The `audit` ingredient breaks the naming convention used by most 1:1 gating ingredients in the recipe system:

| Ingredient | Step It Gates | Skill It Invokes | Convention Match? |
|---|---|---|---|
| `investigate` | `investigate` | `investigate` | Yes |
| `review_approach` | `review_approach` | `review-approach` | Yes |
| `audit_claims` | `audit_claims` | `audit-claims` | Yes |
| `review_design` | `review_design` | `review-design` | Yes |
| **`audit`** | **`audit_impl`** | **`audit-impl`** | **No** |
| **`planner_assess_review_approach`** | **`assess_review_approach_step`** | **`planner-assess-review-approach`** | **No** |

Note: structural mode flags like `open_pr` (gates 41 steps), `post_run_diagnostics`, and `test_check_enabled` are intentionally named for their semantic meaning rather than a single step, so they are not outliers in the same way.

The AI reasonably inferred that ingredient names mirror step names (which they do in most 1:1 cases) and used `audit_impl`. This was compounded by:

1. **Silent acceptance of unknown ingredient keys.** `open_kitchen` and `lock_ingredients` accept arbitrary ingredient keys without validation. When the AI passed `audit_impl`, it received no error — the key was silently stored with no effect. Only the fleet path (`dispatch_food_truck` in `fleet/_api.py:488-495`) validates ingredient keys and returns `FLEET_UNKNOWN_INGREDIENT` with valid key suggestions.

2. **Ambiguous `audit` name.** The ingredient `audit` is a prefix of 10+ skill names (`audit-arch`, `audit-bugs`, `audit-claims`, `audit-cohesion`, `audit-docs`, `audit-friction`, `audit-impl`, `audit-review-decisions`, `audit-tests`, `audit-defense-standards`). Its broad name suggests it might control the entire audit family when it actually gates only `audit-impl`.

[...]

PR #3500, opened 2026-06-01
Rectify: Ingredient Naming Asymmetry Causes Orchestrator Confusion

Trecek:
## Summary

The recipe system allows ingredient names to diverge from the step names they gate via `skip_when_false`. When an ingredient named `audit` gates a step named `audit_impl`, AI orchestrators infer the ingredient key is `audit_impl` (matching the step name), call `lock_ingredients(locked={"audit_impl": "false"})`, and the call succeeds silently — the unknown key is stored with no effect. The architectural weakness is twofold: (1) no static rule enforces naming symmetry between 1:1 gating ingredients and their steps, and (2) `lock_ingredients` and `open_kitchen` accept arbitrary ingredient keys without schema validation, while only the fleet dispatch path (`dispatch_food_truck`) validates keys.

Closes #3462

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260531-182018-614340/.autoskillit/temp/rectify/rectify_ingredient_naming_asymmetry_2026-05-31_182018.md`


src/autoskillit/server/tools/tools_kitchen.py
```diff
@@ -806,6 +855,20 @@ def _apply_unlock_keys(current_pipeline_li: dict[str, str], unlock_keys: list[st
         current_pipeline_li.pop(key, None)
 
 
+def _build_ingredient_key_suggestions(
+    unknown: set[str], declared: frozenset[str]
+) -> dict[str, list[str]]:
+    from difflib import get_close_matches  # noqa: PLC0415
```
Trecek: [warning] arch: Deferred import of `difflib.get_close_matches` inside `_build_ingredient_key_suggestions` with `# noqa: PLC0415` suppression. `difflib` is stdlib with no import-layer concerns; this noqa suppresses a legitimate top-level import convention. Import at module level and remove the suppression.

src/autoskillit/server/tools/tools_kitchen.py
```diff
@@ -867,6 +930,30 @@ async def lock_ingredients(
             )
 
         active_steps = getattr(ctx, "active_recipe_steps", None) or {}
+        _raw_decl = getattr(ctx, "active_recipe_ingredients", None)
```
Trecek: [warning] arch: Accessing `active_recipe_ingredients` via `getattr(ctx, "active_recipe_ingredients", None)` followed by an `isinstance(_, frozenset)` guard, even though the field was explicitly typed on `ToolContext` in this same PR. Access `ctx.active_recipe_ingredients` directly; the isinstance guard is redundant and silently degrades validation to a no-op if the field is absent or mistyped.

Trecek (commented):

Trecek (commented):

src/autoskillit/server/tools/tools_kitchen.py
Trecek: Agreed — fixed in 61515afec541776f96874999c5ecb381fcf76471. Moved `import difflib` to module level and replaced the inline `from difflib import get_close_matches  # noqa: PLC0415` with `difflib.get_close_matches()` at call site. The noqa suppression is removed.

src/autoskillit/server/tools/tools_kitchen.py
Trecek: Agreed — fixed in 95a09dfc72940d93b3bffff9e20d617438f04876. Replaced `getattr(ctx, "active_recipe_ingredients", None)` + isinstance guard with direct `ctx.active_recipe_ingredients` access. The field is typed as `frozenset[str] | None` in `context.py:157`, so direct access is correct and type-safe.

Issue #4208, opened 2026-07-08
load_recipe re-serve corrupts routing in live kitchen: clone.on_success rewired to register_clone_failure on 2nd+ serving (0.10.853)


Trecek:
## Problem

Within a live kitchen session on 0.10.853, `load_recipe` re-serves of the remediation recipe carry **corrupted routing**: `clone.on_success` is rewired from `claim_and_resolve` to `register_clone_failure`. The first serving (via `open_kitchen`) is correct; the second and third servings (via `load_recipe`) are corrupted. An orchestrator that re-loads the recipe then dutifully routes a *successful* clone to the failure path and the dispatch dies.

## Evidence (dispatch e6a45f7f, session d8d50e44, 2026-07-07 ~20:52)

- Extraction of every recipe serving in the orchestrator session log (`~/.claude/projects/-home-talon-projects-generic-automation-mcp/d8d50e44-4da4-4de3-8cc0-6aeb2be5bae5.jsonl`), matching `clone: tool: bootstrap_clone … on_success: <target>`: **`['claim_and_resolve', 'register_clone_failure', 'register_clone_failure']`** — first serve correct, both re-serves corrupted.
- The orchestrator behaved correctly per anti-fabrication discipline: "The routing authority is the recipe YAML as loaded… I must follow it." — clone succeeded (work_dir, base_sha `cefeaff57`, 3052ms), then it followed the served `register_clone_failure`. Not a hallucination.
- On-disk recipes are correct (`remediation.yaml:175` and the project copy both `on_success: claim_and_resolve`).
- Offline reproduction attempt: `load_recipe` twice AND `_build_active_recipe` twice with the exact dispatch ingredients (`issue_url`, `source_dir`, `post_run_diagnostics=true`) — **does NOT reproduce** (both composes correct, original object unmutated). The corruption therefore lives in the server-side serve/recall path that only exists in a live kitchen session — prime suspect: the pre-reveal/deferred-recall cache (#4092) being mutated in place by a post-serve pass (pruning, capability overrides, or the new check_review_posted wiring pass from #4205 — this was the first remediation dispatch on 0.10.853).
- Same session's `kitchen_status` also warned: `package_version: 0.10.853, plugin_json_version: None, versions_match: False` — possibly unrelated, noted for completeness.
- Related lineage: #3992 (dangling route after pruning), #4089/#4092 (double open_kitchen recall).

## Impact

[...]

PR #4210, opened 2026-07-08
Implementation Plan: Serve Surface Symmetry — SERVE_SURFACES __all__ Alphabetical Fix

Trecek:
## Summary

A single entry in `__all__` of `src/autoskillit/core/types/_type_constants_registries.py` is out of alphabetical order. `"SERVE_SURFACES"` is currently appended as the very last entry in `__all__`, after `"SkillCapabilityDef"` and `"SKILL_CAPABILITY_REGISTRY"`. Because SERVE precedes SKILL alphabetically, `"SERVE_SURFACES"` must appear before `"SkillCapabilityDef"`. The fix is to remove `"SERVE_SURFACES"` from position 38 and insert it at position 36 (between `"ALL_VISIBILITY_TAGS"` and `"SkillCapabilityDef"`).

No other files require changes. The remediation document explicitly states that the implementation is complete and correct in all other respects.

Closes #4208

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260708-002748-757853/.autoskillit/temp/make-plan/serve_surface_symmetry_alphabetical_fix_plan_2026-07-08_030000.md`

## Token Usage Summary

| investigate* | fable | 1 | 8.8k | 23.3k | 547.1k | 91.2k | 26 | 96.7k | 14m 15s |
| rectify* | sonnet | 1 | 7.8k | 28.9k | 748.9k | 102.4k | 37 | 83.9k | 17m 19s |
| dry_walkthrough* | sonnet | 2 | 6.8k | 21.4k | 1.2M | 65.5k | 62 | 88.5k | 11m 40s |
| implement* | sonnet | 2 | 638 | 71.8k | 8.5M | 176.5k | 201 | 191.5k | 23m 42s |
| retry_worktree* | sonnet | 1 | 796 | 62.2k | 10.7M | 172.7k | 221 | 153.9k | 37m 55s |
| audit_impl* | sonnet | 3 | 196 | 24.0k | 1.3M | 107.2k | 100 | 103.6k | 18m 27s |
| make_plan* | sonnet | 1 | 79 | 7.1k | 392.5k | 56.5k | 21 | 42.2k | 2m 14s |
| merge_gate_assess* | sonnet | 1 | 342 | 46.9k | 4.3M | 143.7k | 103 | 125.1k | 21m 9s |
| prepare_pr* | MiniMax-M3 | 1 | 42.1k | 2.8k | 110.1k | 0 | 13 | 0 | 36s |
| compose_pr* | MiniMax-M3 | 1 | 40.4k | 3.3k | 261.8k | 0 | 16 | 0 | 46s |
| **Total** | | | 107.9k | 291.6k | 28.0M | 176.5k | | 885.4k | 2h 28m |

## Token Efficiency

| rectify | 0 | — | — | — |
| implement | 664 | 12832.6 | 288.4 | 108.1 |
| retry_worktree | 658 | 16263.6 | 233.9 | 94.5 |
| make_plan | 0 | — | — | — |
| merge_gate_assess | 2 | 2128658.0 | 62551.0 | 23430.5 |
| **Total** | **1324** | 21115.2 | 668.7 | 220.3 |

## Model Usage Breakdown

| fable | 1 | 8.8k | 23.3k | 547.1k | 96.7k | 14m 15s |
| sonnet | 7 | 16.6k | 262.3k | 27.0M | 788.7k | 2h 12m |
| MiniMax-M3 | 2 | 82.5k | 6.0k | 372.0k | 0 | 1m 21s |


src/autoskillit/server/tools/_serve_helpers.py
```diff
@@ -0,0 +1,103 @@
...
+    }
+    if resolved_defaults is not None:
+        kwargs["resolved_defaults"] = resolved_defaults
+    if effective_backend_map is not None:
+        kwargs["effective_backend_map"] = effective_backend_map
+    if suppressed is not None:
+        kwargs["suppressed"] = suppressed
+    if backend_name is not None:
+        kwargs["backend_name"] = backend_name
+    if ingredients_only:
```
Trecek, line 106: [critical] defense: assert ctx.recipes is not None is stripped by Python's -O flag, turning this into a bare AttributeError with no diagnostic context when ctx.recipes is None in optimized builds. Replace with: if ctx.recipes is None: raise RuntimeError('serve_recipe() called with ctx.recipes=None')

src/autoskillit/core/types/_type_protocols_recipe.py
```diff
@@ -13,7 +13,28 @@
 
 from ._type_results import LoadResult
 
-__all__ = ["RecipeRepository", "MigrationService", "DatabaseReader", "ReadOnlyResolver"]
+__all__ = [
+    "RecipeRepository",
+    "MigrationService",
+    "DatabaseReader",
+    "ReadOnlyResolver",
+    "ServeOverridesSnapshot",
```
Trecek, line 21: [warning] defense: ServeOverridesSnapshot Protocol is missing __contains__, get, items, and values. Any call site that does `key in snapshot`, `snapshot.get(k)`, `snapshot.items()`, or `snapshot.values()` will fail at runtime despite passing isinstance check. Add the missing methods or document that these operations are intentionally unsupported.

src/autoskillit/fleet/_api.py
```diff
@@ -777,7 +777,7 @@ def _on_session_id(session_id: str) -> None:
                     else None,
                 )
 
-        ended_at = time.time()
+        ended_at = max(time.time(), started_at + 1e-6)
```
Trecek, line 780: [warning] bugs: ended_at = max(time.time(), started_at + 1e-6) can silently corrupt elapsed duration under clock skew: if started_at was already ahead of wall-clock time, ended_at becomes started_at + 1e-6 and duration = ended_at - started_at collapses to ~0. Safer: compute elapsed = max(time.time() - started_at, 1e-6) at the use site instead of clamping ended_at.

tests/server/test_serve_idempotence.py
```diff
@@ -0,0 +1,464 @@
...
+    )
+    assert deferred_result.get("success") is True, (
+        f"deferred-recall open_kitchen failed: {deferred_result}"
+    )
+    deferred_content = deferred_result["content"]
+
+    assert first_content == deferred_content, (
+        "Deferred-recall open_kitchen content diverges from first serving — "
+        "session_serve_overrides not injected into deferred-recall _merged_overrides"
+    )
```
Trecek, line 170: [warning] tests: assume(False) silently discards examples where open_kitchen fails. If _open_kitchen_patched consistently fails due to a regression, Hypothesis exhausts its examples with all discarded and reports a passing test. Replace with a hard assertion to make failures visible.

tests/server/test_serve_idempotence.py
```diff
@@ -0,0 +1,464 @@
...
+
+
+async def test_session_serve_overrides_cleared_on_close_kitchen(
```
Trecek, line 173: [warning] tests: assume(False) on load_recipe failure silently discards examples where the system under test is broken — if load_recipe always fails (regression), the test passes vacuously. Use a hard assertion instead.

tests/server/test_serve_idempotence.py
```diff
@@ -0,0 +1,464 @@
...
+    )
+    assert ok_result.get("success") is True, f"open_kitchen failed: {ok_result}"
+    ok_content = ok_result["content"]
+
+    lr_result = json.loads(await load_recipe(name=_RECIPE))
+    assert "content" in lr_result, f"load_recipe missing content: {lr_result}"
+    lr_content = lr_result["content"]
+
+    assert ok_content == lr_content, (
+        "load_recipe content diverges from open_kitchen content — "
```
Trecek, line 93: [warning] tests: Hardcoded step path parsed_ok['steps']['clone']['on_success'] couples the test to remediation recipe internals. If the recipe is restructured the assertion breaks with no indication of whether serve-idempotence itself failed. Assert only that parsed_ok == parsed_lr (the actual invariant) instead of checking specific step values.

tests/server/test_serve_idempotence.py
```diff
@@ -0,0 +1,464 @@
...
+    )
+
+    lr_result = json.loads(await load_recipe(name=_RECIPE))
+    assert "content" in lr_result, f"load_recipe missing content: {lr_result}"
+    lr_content = lr_result["content"]
+
+    assert ok_content == lr_content, (
+        "load_recipe content diverges from open_kitchen content (no-override path) — "
+        "session_serve_overrides baseline not applied"
+    )
```
Trecek, line 133: [warning] tests: assert not gr_content.startswith('{') is a fragile error-detection heuristic — valid YAML starting with '{' would be falsely rejected, and non-JSON error strings would pass. Assert that gr_content equals ok_content (the actual invariant) and let the equality failure surface the error.

tests/server/test_serve_idempotence.py
```diff
@@ -0,0 +1,464 @@
...
+    tool_ctx_kitchen_open,
+    monkeypatch,
+    tmp_path,
+):
+    """Deferred-recall open_kitchen (called again while gate is open) produces
+    byte-identical content to the normal path first serving."""
+    monkeypatch.chdir(tmp_path)
+    from autoskillit.recipe import _api_cache
+    from autoskillit.recipe._api_cache import LoadCache
+
```
Trecek, line 146: [warning] tests: Hypothesis overrides strategy generates values from only 2 fixed strings (just(...)), so all non-empty overrides are identical. The test claims 'arbitrary overrides' but explores only key presence, not value variation. Use st.text() or st.from_regex() for values to get meaningful fuzz coverage.

tests/server/test_serve_idempotence.py
```diff
@@ -0,0 +1,464 @@
...
+    deferred_result = await _open_kitchen_patched(
+        _RECIPE,
+        None,
+        monkeypatch,
+    )
+    assert deferred_result.get("success") is True, (
+        f"deferred-recall open_kitchen failed: {deferred_result}"
+    )
+    deferred_content = deferred_result["content"]
+
```
Trecek, line 166: [warning] slop: Three consecutive # type: ignore[assignment] and # type: ignore[union-attr] suppressions indicate the fixture type is too narrow for the attributes being set. Type the fixture return as the concrete class (ToolContext) or use cast() — blanket ignores here will mask future type regressions.

src/autoskillit/server/tools/tools_kitchen.py
```diff
@@ -859,15 +861,24 @@ async def open_kitchen(
                         result["warnings"] = _override_warnings
                 if ingredients_only:
                     result = strip_ingredients_only_keys(result)
+                # When caller provides explicit overrides, update the snapshot so
+                # subsequent load_recipe/get_recipe calls see the new overrides.
+                # When overrides=None (replay previous context), leave the existing
+                # snapshot intact — the caller's intent is continuity, not reset.
+                if overrides is not None:
+                    tool_ctx.session_serve_overrides = dict(overrides) if overrides else {}
```
Trecek: [warning] arch: dict(overrides) if overrides else {} conflates overrides={} (explicitly empty dict) with None — both produce empty snapshot, but session_serve_defer_unresolved is set to True for both (not bool({}) is True). An explicit empty dict is semantically different from no overrides; consider: session_serve_overrides = dict(overrides); session_serve_defer_unresolved = overrides is None.

tests/arch/test_serve_surface_registry.py
```diff
@@ -0,0 +1,82 @@
...
+                break
+    return offenders
+
+
+def test_serve_surfaces_registry_is_exhaustive() -> None:
+    """load_and_validate must only be called from _serve_helpers.py in server/tools/.
+
+    tools_fleet_dispatch.py is exempt: its fleet-dispatch preflight call is Part B scope
+    (issue #4208 Part B — fleet dispatch resolved_defaults gap). It is intentionally
+    excluded from the Part A enforcement perimeter.
```
Trecek, line 44: [warning] slop: The inline comment at L44-45 duplicates information already in the docstring at L37-42 (both mention _serve_helpers.py as canonical and tools_fleet_dispatch.py as known Part B gap). Remove one to avoid drift between the two descriptions.

tests/server/test_serve_idempotence.py
```diff
@@ -0,0 +1,464 @@
...
+        "load_recipe content diverges from open_kitchen content (no-override path) — "
+        "session_serve_overrides baseline not applied"
+    )
+
+
+async def test_deferred_recall_open_kitchen_serves_identical_to_first_serving(
+    tool_ctx_kitchen_open,
+    monkeypatch,
+    tmp_path,
+):
```
Trecek, line 140: [warning] tests: HealthCheck.function_scoped_fixture suppression is undocumented in code — add a comment explaining why the manual state reset (L169-171) makes reuse safe. Future fixture changes may invalidate the assumption silently.

Trecek (commented):

src/autoskillit/server/tools/_serve_helpers.py
```diff
@@ -0,0 +1,103 @@
...
+        kwargs["ingredients_only"] = ingredients_only
```
Trecek, line 107: [warning] defense: serve_recipe() propagates all exceptions from load_and_validate without adding context about which recipe triggered them. Consider: except Exception as exc: raise RuntimeError(f'serve_recipe({name!r}) failed') from exc to preserve call-site identity in tracebacks.

Trecek (commented):
AutoSkillit review: 1 critical blocking issue found + 12 warnings.

**Verdict:** changes_requested (posted as COMMENT — own-PR guard active)

### Summary
- **[critical] defense** `src/autoskillit/server/tools/_serve_helpers.py:95` — `assert` stripped by Python `-O`, replace with explicit `RuntimeError`
- 12 warnings across defense, tests, arch, slop, and bugs dimensions — see inline comments

src/autoskillit/server/tools/_serve_helpers.py
Trecek: Agreed — fixed in e91e350. Replaced `assert ctx.recipes is not None` with an explicit `if ctx.recipes is None: raise RuntimeError(...)` guard that survives Python -O optimization.

src/autoskillit/core/types/_type_protocols_recipe.py
Trecek: Investigated — this is intentional. The docstring explicitly states 'minimum structural contract': only `__getitem__`, `keys`, `__iter__`, `__len__` are needed because every call site converts via `dict(ctx.session_serve_overrides)`, which only needs `__iter__` and `__getitem__`. No call site uses `__contains__`, `.get()`, `.items()`, or `.values()` on a `ServeOverridesSnapshot`. Adding unused protocol methods would expand beyond what the minimum-structural-contract pattern intends.

src/autoskillit/fleet/_api.py
Trecek: Investigated — this is intentional. Commit d2720fa41 (landed before this review) introduced `max(time.time(), started_at + 1e-6)` specifically to ensure `ended_at >= started_at` under NTP clock corrections in CI. The near-zero elapsed under severe clock skew is a known trade-off: a valid invariant (non-negative duration) is preserved at the cost of a misleading ≈0s elapsed in a corner case that only occurs when the clock jumps backward significantly during a dispatch. A warning log approach would not preserve the invariant.

tests/server/test_serve_idempotence.py
Trecek: Valid observation — flagged for design decision. Prior review round explicitly requested `assume(False)` for invalid combos (commit 1f5a656f0). Current review correctly notes the false-pass risk when `open_kitchen` fails due to a regression rather than an invalid input combo. The choice between `assume()` (silently discards) vs. a hard assertion (fails on regression) requires a human decision about the expected failure mode.

tests/server/test_serve_idempotence.py
Trecek: Valid observation — flagged for design decision. Same trade-off as the open_kitchen case: `assume(False)` silently discards examples where `load_recipe` fails, which would mask a regression. The previous review round introduced this pattern. Human decision needed on whether to switch to a hard assertion.

tests/server/test_serve_idempotence.py
Trecek: Agreed — fixed in b7c532d. Removed the `parsed_ok['steps']['clone']['on_success']` assertions that coupled the test to recipe internals. The content equality assertion at line 91 already covers routing correctness without depending on specific step names or transition values.

tests/server/test_serve_idempotence.py
Trecek: Agreed — fixed in b7c532d. Replaced all three `startswith('{')` heuristics with `json.loads` checks: if the content parses as JSON, it's an error envelope and we call `pytest.fail()`; if it raises `JSONDecodeError`, it's valid YAML. Applied to all three call sites in the test file.

tests/server/test_serve_idempotence.py
Trecek: Valid observation — flagged for design decision. The strategy uses `st.just()` with two fixed values, so all non-empty overrides resolve to a single pair of known-valid strings. The test varies the KEY combinations (4 subsets of {issue_url, task_description}) but not the values. Expanding to `st.text()` would be more thorough but may fail recipe validation that requires a valid URL format for `issue_url`. Human decision needed on whether broader values are safe.

tests/server/test_serve_idempotence.py
Trecek: Agreed — fixed in b7c532d. Changed `tool_ctx_kitchen_open: object` to `tool_ctx_kitchen_open: ToolContext` in the Hypothesis test and added the `ToolContext` import at module level. The three `# type: ignore[union-attr]` and `# type: ignore[assignment]` suppressions are now eliminated since `ToolContext` exposes `session_serve_overrides`, `session_serve_defer_unresolved`, and `recipe_name` directly.

src/autoskillit/server/tools/tools_kitchen.py
Trecek: Agreed — fixed in 77f58a7. Simplified `dict(overrides) if overrides else {}` to `dict(overrides)`. Inside the `if overrides is not None:` guard, both branches of the ternary produce the same result (`dict({})` == `{}`), so the conditional was purely redundant.

tests/arch/test_serve_surface_registry.py
Trecek: Agreed — fixed in e620e02. Removed the duplicate inline comment and moved the `TODO(#4208-B)` into the docstring where the rationale already lives. The docstring now reads: '...It is intentionally excluded from the Part A enforcement perimeter. TODO(#4208-B): remove tools_fleet_dispatch.py from the allowed set when Part B ships.'

tests/server/test_serve_idempotence.py
Trecek: Agreed — fixed in b7c532d. Expanded the `suppress_health_check` list to multi-line and added an inline comment: `HealthCheck.function_scoped_fixture,  # safe: manual state reset at test start`. The docstring already explains the rationale but the suppression site now self-documents why reuse is safe.

src/autoskillit/server/tools/_serve_helpers.py
Trecek: Investigated — this is intentional. Callers of `serve_recipe()` (e.g., `tools_kitchen.py:885-890`) already wrap the call in both `except ProcessStaleError` and `except Exception` with `_kitchen_failure_envelope(exc, stage='...')`. Adding an inner `try/except Exception as exc: raise RuntimeError(...) from exc` inside `serve_recipe()` would convert `ProcessStaleError` to `RuntimeError`, silently breaking the separate `except ProcessStaleError` catch that callers rely on for structured error handling.

src/autoskillit/server/tools/_serve_helpers.py
Trecek: Agreed — fixed in e91e350. Replaced `assert ctx.recipes is not None` with an explicit `if ctx.recipes is None: raise RuntimeError(...)` guard that survives Python -O optimization.

src/autoskillit/core/types/_type_protocols_recipe.py
Trecek: Investigated — this is intentional. The docstring explicitly states 'minimum structural contract': only `__getitem__`, `keys`, `__iter__`, `__len__` are needed because every call site converts via `dict(ctx.session_serve_overrides)`, which only needs `__iter__` and `__getitem__`. No call site uses `__contains__`, `.get()`, `.items()`, or `.values()` on a `ServeOverridesSnapshot`. Adding unused protocol methods would expand beyond what the minimum-structural-contract pattern intends.

src/autoskillit/fleet/_api.py
Trecek: Investigated — this is intentional. Commit d2720fa41 (landed before this review) introduced `max(time.time(), started_at + 1e-6)` specifically to ensure `ended_at >= started_at` under NTP clock corrections in CI. The near-zero elapsed under severe clock skew is a known trade-off: a valid invariant (non-negative duration) is preserved at the cost of a misleading ≈0s elapsed in a corner case that only occurs when the clock jumps backward significantly during a dispatch. A warning log approach would not preserve the invariant.

tests/server/test_serve_idempotence.py
Trecek: Valid observation — flagged for design decision. Prior review round explicitly requested `assume(False)` for invalid combos (commit 1f5a656f0). Current review correctly notes the false-pass risk when `open_kitchen` fails due to a regression rather than an invalid input combo. The choice between `assume()` (silently discards) vs. a hard assertion (fails on regression) requires a human decision about the expected failure mode.

tests/server/test_serve_idempotence.py
Trecek: Valid observation — flagged for design decision. Same trade-off as the open_kitchen case: `assume(False)` silently discards examples where `load_recipe` fails, which would mask a regression. The previous review round introduced this pattern. Human decision needed on whether to switch to a hard assertion.

tests/server/test_serve_idempotence.py
Trecek: Agreed — fixed in b7c532d. Removed the `parsed_ok['steps']['clone']['on_success']` assertions that coupled the test to recipe internals. The content equality assertion at line 91 already covers routing correctness without depending on specific step names or transition values.

tests/server/test_serve_idempotence.py
Trecek: Agreed — fixed in b7c532d. Replaced all three `startswith('{')` heuristics with `json.loads` checks: if the content parses as JSON, it's an error envelope and we call `pytest.fail()`; if it raises `JSONDecodeError`, it's valid YAML. Applied to all three call sites in the test file.

tests/server/test_serve_idempotence.py
Trecek: Valid observation — flagged for design decision. The strategy uses `st.just()` with two fixed values, so all non-empty overrides resolve to a single pair of known-valid strings. The test varies the KEY combinations (4 subsets of {issue_url, task_description}) but not the values. Expanding to `st.text()` would be more thorough but may fail recipe validation that requires a valid URL format for `issue_url`. Human decision needed on whether broader values are safe.

tests/server/test_serve_idempotence.py
Trecek: Agreed — fixed in b7c532d. Changed `tool_ctx_kitchen_open: object` to `tool_ctx_kitchen_open: ToolContext` in the Hypothesis test and added the `ToolContext` import at module level. The three `# type: ignore[union-attr]` and `# type: ignore[assignment]` suppressions are now eliminated since `ToolContext` exposes `session_serve_overrides`, `session_serve_defer_unresolved`, and `recipe_name` directly.

src/autoskillit/server/tools/tools_kitchen.py
Trecek: Agreed — fixed in 77f58a7. Simplified `dict(overrides) if overrides else {}` to `dict(overrides)`. Inside the `if overrides is not None:` guard, both branches of the ternary produce the same result (`dict({})` == `{}`), so the conditional was purely redundant.

tests/arch/test_serve_surface_registry.py
Trecek: Agreed — fixed in e620e02. Removed the duplicate inline comment and moved the `TODO(#4208-B)` into the docstring where the rationale already lives. The docstring now reads: '...It is intentionally excluded from the Part A enforcement perimeter. TODO(#4208-B): remove tools_fleet_dispatch.py from the allowed set when Part B ships.'

tests/server/test_serve_idempotence.py
Trecek: Agreed — fixed in b7c532d. Expanded the `suppress_health_check` list to multi-line and added an inline comment: `HealthCheck.function_scoped_fixture,  # safe: manual state reset at test start`. The docstring already explains the rationale but the suppression site now self-documents why reuse is safe.

src/autoskillit/server/tools/_serve_helpers.py
Trecek: Investigated — this is intentional. Callers of `serve_recipe()` (e.g., `tools_kitchen.py:885-890`) already wrap the call in both `except ProcessStaleError` and `except Exception` with `_kitchen_failure_envelope(exc, stage='...')`. Adding an inner `try/except Exception as exc: raise RuntimeError(...) from exc` inside `serve_recipe()` would convert `ProcessStaleError` to `RuntimeError`, silently breaking the separate `except ProcessStaleError` catch that callers rely on for structured error handling.

src/autoskillit/server/tools/tools_kitchen.py
```diff
@@ -957,6 +968,12 @@ async def open_kitchen(
                     await ctx.disable_components(tags={"kitchen"})
                     return _preflight_err
 
+            # Snapshot the caller-supplied values ONLY — NOT _merged_overrides.
+            # Storing _merged_overrides would inject stale kitchen_id/diagnostics_log_dir
+            # into subsequent load_recipe merges, silently overwriting fresh infra values.
+            tool_ctx.session_serve_overrides = dict(overrides) if overrides else {}
```
Trecek, line 974: [warning] bugs: If a caller explicitly passes `kitchen_id` or `diagnostics_log_dir` as ingredient overrides to `open_kitchen`, those keys are captured in `session_serve_overrides` (L974). On subsequent `get_recipe` calls, `_build_serve_override_stack` places `snapshot_baseline` above `session_overrides` in priority, so the stale user-supplied value wins over the freshly-computed `ctx.kitchen_id` / `resolve_log_dir(...)`. The comment at L971-973 warns against storing `_merged_overrides` for exactly this reason, but raw user overrides containing infra key names have the same effect. Consider stripping known infra keys (`kitchen_id`, `diagnostics_log_dir`) from the captured snapshot, or document that users must not pass these keys as overrides.

tests/server/test_serve_idempotence.py
```diff
@@ -0,0 +1,473 @@
...
+    """For arbitrary overrides, load_recipe content must match open_kitchen content.
+
+    Catches precedence bugs where specific key collisions in the override stack
+    cause snapshot_baseline values to be shadowed incorrectly.
+
+    The function-scoped fixture is not reset between Hypothesis examples, so we
+    manually reset serve-context state at the start of each example to ensure
+    open_kitchen always takes the normal path (not deferred-recall) and the
+    snapshot is always set from the current example's overrides.
+    """
```
Trecek, line 444: [warning] tests: The Hypothesis test mutates shared fixture state directly (`ctx.session_serve_overrides = None`, `ctx.session_serve_defer_unresolved = False`, `ctx.recipe_name = ""`) without using `monkeypatch.setattr` or a fixture teardown. If a future Hypothesis shrink or DB replay reorders or re-runs examples, partial state from a prior example that raised mid-test (before reaching the reset at L444-447) will leak into the next example. The manual reset only covers the case where the prior example completed normally. Consider using `monkeypatch` for all three assignments so pytest guarantees rollback on any exit path.

tests/server/test_serve_idempotence.py
```diff
@@ -0,0 +1,473 @@
...
+    monkeypatch.setattr(_api_cache, "_LOAD_CACHE", LoadCache())
+
+    ok_result = await _open_kitchen_patched(
+        _RECIPE,
+        {"issue_url": _ISSUE_URL, "task_description": _TASK_DESC},
+        monkeypatch,
+    )
+    assert ok_result.get("success") is True, f"open_kitchen failed: {ok_result}"
+    ok_content = ok_result["content"]
+
```
Trecek, line 274: [warning] tests: The error guard at L274-279 (`try: json.loads; pytest.fail except json.JSONDecodeError: pass`) will misfire if the production YAML content happens to also be valid JSON (e.g., a recipe containing only a single YAML mapping that starts with `{`). In that case `pytest.fail` fires and the test fails even when `get_recipe` returned correct content. The `assert ok_content == gr_content` at L281 already catches any real divergence — the JSON guard is redundant and should be removed, or replaced with a direct error-status field check if `get_recipe` error responses have a distinguishing field.

tests/server/test_serve_idempotence.py
```diff
@@ -0,0 +1,473 @@
...
+            st.just("test task"),
+        ),
+        max_size=2,
+    )
+)
+async def test_load_recipe_routing_matches_open_kitchen_for_arbitrary_overrides(
+    overrides: dict[str, str],
+    tool_ctx_kitchen_open: ToolContext,
+    monkeypatch: pytest.MonkeyPatch,
+    tmp_path: object,
```
Trecek, line 433: [warning] slop: The 4-line comment block at L433-436 (explaining why manual state reset is needed) is duplicated verbatim at L450-453 in the same function body. The docstring at L434-438 already explains the rationale. Remove the duplicate comment at L450-453 — the reset code at L454-457 is self-explanatory after reading the docstring once.

Trecek (commented):

Trecek (commented):

**Info findings (outside inline comments):**
- `tests/arch/test_serve_surface_registry.py:48` [info/tests]: The `allowed` set hard-codes `tools_fleet_dispatch.py` as an exemption but does not assert the file exists on disk. If the file is removed or renamed, the exemption silently remains. Consider asserting `all((TOOLS_DIR / name).exists() for name in allowed)`.
- `tests/arch/test_serve_surface_registry.py:19` [info/tests]: `_find_load_and_validate_call_files` uses `glob('*.py')` (non-recursive). Files nested in subdirectories of `server/tools/` would be silently missed. Use `rglob('*.py')` to be future-proof.

Issue #4215, opened 2026-07-09
Codex investigate child runs read-only, cannot write required report artifact, and returns adjudicated_failure with zero writes


Trecek:
## Problem

Codex-backed remediation runs can reach the first `run_skill` step (`investigate`) but fail before the recipe can continue because the child Codex session is read-only and cannot write the mandatory investigation report artifact under `.autoskillit/temp/investigate/`.

The child can complete useful analysis and emit `%%ORDER_UP`, but `run_skill` adjudicates the step as `adjudicated_failure` because no report file or other write evidence exists. This blocks the pipeline before implementation begins.

## Reproduction

Run the current queue self-test from `/home/talon/projects/generic_automation_mcp`:

```bash
env -u AUTOSKILLIT_SESSION_TYPE AUTOSKILLIT_CORE__DEFAULT_MODEL='opus[1m]' \
  AUTOSKILLIT_RUN_SKILL__STREAM_IDLE_TIMEOUT_MS=7200000 \
  AUTOSKILLIT_RUN_SKILL__MAX_SUPPRESSION_SECONDS=7200 \
  autoskillit fleet run remediation --backend codex \
  -t "Codex L2 food trucks killed by stream-idle watchdog while blocked on long MCP calls (fix GitHub issue #4202)" \
  -i issue_url=https://github.com/TalonT-Org/AutoSkillit/issues/4202 \
  -i source_dir=/home/talon/projects/generic_automation_mcp \
  -i post_run_diagnostics=true --timeout-sec 14400
```

## Evidence

Failed dispatch:

```json
{
  "success": false,
  "dispatch_status": "failure",
  "dispatch_id": "d6d87a72-f14a-4dc8-aca6-4c329ddfd1ec",
  "dispatched_session_id": "019f4498-1f77-70c2-ad4a-d80d40c234b2",
  "reason": "failed",
  "l3_payload": {
    "success": false,
    "reason": "failed",
    "summary": "Pipeline failed at investigate; issue claim was released, clone preserved, and diagnostics completed."
  }
}
```

Child investigate session:

- Session id: `019f4499-47bf-7ba3-b2e6-624dc4d486a2`
- Session summary: `~/.local/share/autoskillit/logs/sessions/019f4499-47bf-7ba3-b2e6-624dc4d486a2/summary.json`
[...]

PR #4216, opened 2026-07-09
[FIX] Fix Codex Investigate Report Sandbox

Trecek:
## Summary

Codex remediation runs were launching the `investigate` skill in a read-only sandbox even
though that skill must write a markdown report and emit an `investigation_path` token. The
fix updates the investigate skill contract to declare mandatory write behavior, gives the
remediation recipe a scoped investigate output directory, refreshes the compiled recipe and
diagram hash, and adds focused regression coverage for the server-side `run_skill` launch
parameters.

## Requirements

Codex-backed remediation runs can reach the first `run_skill` step (`investigate`) but fail before the recipe can continue because the child Codex session is read-only and cannot write the mandatory investigation report artifact under `.autoskillit/temp/investigate/`.

The child can complete useful analysis and emit `%%ORDER_UP`, but `run_skill` adjudicates the step as `adjudicated_failure` because no report file or other write evidence exists. This blocks the pipeline before implementation begins.

The remediation recipe's `investigate` step must be able to produce its required investigation artifact when run under the Codex backend, or the recipe must route that step to a backend/provider/sandbox mode capable of writing the artifact.

If a skill contract requires artifact writes, dispatch/admission should reject unsupported Codex read-only execution before launching a zero-progress child.

Closes #4215

## Implementation Plan

Plan file: `/home/talon/projects/generic_automation_mcp/.autoskillit/temp/codex-loop/issue-4215-narrow-implementation-plan.md`

### DIFF
--- a/src/autoskillit/skills_extended/investigate/SKILL.md
+++ b/src/autoskillit/skills_extended/investigate/SKILL.md
@@ -96,6 +96,14 @@ tool **before** beginning any analysis. Use the returned `content` field as the 
 - Check for similar existing patterns in codebase
 - Ensure approaches, solutions, and fixes are the appropriate long-term solutions with proper architecture
 
+## Context Limit Behavior
+
+If context pressure interrupts the investigation before the final report token is emitted,
+resume from the partial notes and continue toward the same mandatory report. A successful
+invocation must still write exactly one markdown report under `{{AUTOSKILLIT_TEMP}}/investigate/`
+and emit `investigation_path = {absolute_path_to_investigation_report_file}` as the final
+plain-text line. Do not emit the token until the report exists.
+
 ## Standard Mode Workflow (Steps 1–4)
 
 **Path-existence guard:** Before issuing a `Read` call on a path that is not guaranteed to
@@ -375,7 +383,7 @@ Spawn one adversarial subagent via `Agent(model="sonnet")` whose role is to disc
 
 **If counterevidence is found:** Return to D3 for one additional deepening batch focused on reconciling the contradiction.
 
-**If no counterevidence is found:** The hypothesis stands. Proceed to D5.
+**When the challenge finds no counterevidence:** The hypothesis stands. Proceed to D5.
 
 ### Step D5: Solution Convergence
 
--- a/tests/recipe/test_contracts.py
+++ b/tests/recipe/test_contracts.py
@@ -739,12 +739,17 @@ def test_write_behavior_conditional_loaded() -> None:
     assert any("conflict_report_path" in p for p in contract.write_expected_when)
 
 
-def test_write_behavior_defaults_to_none() -> None:
-    """investigate has no write_behavior — defaults to None."""
+def test_investigate_declares_report_write_contract() -> None:
+    """investigate writes a report artifact and must not run read-only."""
     manifest = load_bundled_manifest()
     contract = get_skill_contract("investigate", manifest)
     assert contract is not None
-    assert contract.write_behavior is None
+    assert contract.write_behavior == "always"
+    assert contract.read_only is False
+    assert contract.expected_output_patterns
+    assert any(
+        "/investigate/" in p and p.endswith(".md\n%%ORDER_UP%%") for p in contract.pattern_examples
+    )
 
 
 ALWAYS_WRITE_SKILLS = {
@@ -762,6 +767,7 @@ ALWAYS_WRITE_SKILLS = {
     "implement-experiment",
     "implement-worktree",
     "implement-worktree-no-merge",
+    "investigate",
     "make-campaign",
     "make-groups",
     "phoropter-null-synthesis",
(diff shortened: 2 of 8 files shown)


tests/server/test_tools_execution_write_prefix.py
```diff
@@ -45,3 +45,30 @@ async def test_allowed_write_prefix_uses_fallback_without_output_dir(
     expected = str(tmp_path / ".autoskillit" / "temp" / "test") + "/"
     assert executor.calls[0].allowed_write_prefix == expected
     assert executor.calls[0].allowed_write_prefixes == (expected,)
+
```
Trecek, line 48: [warning] tests: The assertion `assert call.write_behavior is not None` on the line at L65 is redundant with the immediately following `assert call.write_behavior.mode == "always"` — the first assertion provides no additional coverage and can be removed.

tests/recipe/test_contracts.py
```diff
@@ -739,12 +739,17 @@ def test_write_behavior_conditional_loaded() -> None:
     assert any("conflict_report_path" in p for p in contract.write_expected_when)
 
 
-def test_write_behavior_defaults_to_none() -> None:
-    """investigate has no write_behavior — defaults to None."""
+def test_investigate_declares_report_write_contract() -> None:
+    """investigate writes a report artifact and must not run read-only."""
     manifest = load_bundled_manifest()
     contract = get_skill_contract("investigate", manifest)
```
Trecek, line 745: [warning] tests: The pattern assertion `p.endswith(".md\n%%ORDER_UP%%")` couples the test to a precise byte-level tail. If the example is ever reformatted (trailing whitespace removed, CRLF, or a different ordering token), the test will fail for an unrelated reason. Consider a more semantic match such as `".md" in p and "/investigate/" in p`.

tests/recipe/test_contracts.py
```diff
@@ -739,12 +739,17 @@ def test_write_behavior_conditional_loaded() -> None:
     assert any("conflict_report_path" in p for p in contract.write_expected_when)
```
Trecek, line 741: [info] tests: `assert contract.write_behavior == "always"` and `assert contract.read_only is False` may be redundant if the contract has a single source of truth. One assertion plus a sanity check on the negation would suffice.

src/autoskillit/skills_extended/investigate/SKILL.md
```diff
@@ -96,6 +96,14 @@ tool **before** beginning any analysis. Use the returned `content` field as the
 - Check for similar existing patterns in codebase
```
Trecek, line 96: [info] slop: The "## Context Limit Behavior" section (lines L96-L102) restates contract requirements (one report under the temp dir, emit final token) likely already present in skill orchestration guidance. Consider tightening or moving to a shared conventions doc.

src/autoskillit/skills_extended/investigate/SKILL.md
```diff
@@ -375,7 +383,7 @@ Spawn one adversarial subagent via `Agent(model="sonnet")` whose role is to disc
```
Trecek, line 383: [info] slop: The renamed header "When the challenge finds no counterevidence" replaces "If no counterevidence is found" with no apparent semantic change — just a more verbose rewording. Revert to the terser wording unless there is a deliberate reason.

tests/server/test_tools_execution_write_prefix.py
```diff
@@ -45,3 +45,30 @@ async def test_allowed_write_prefix_uses_fallback_without_output_dir(
...
+
+@pytest.mark.anyio
+async def test_investigate_contract_runs_writable_with_report_watch_dir(
```
Trecek, line 51: [info] slop: Docstring "investigate must not be launched as a read-only skill because it writes a report." restates the test name. Consider removing or keeping only if it documents a non-obvious WHY.

Trecek (commented):

tests/server/test_tools_execution_write_prefix.py
Trecek: Investigated — this is intentional defensive style. The `is not None` check before `.mode == "always"` provides a clearer failure message if contract is misconfigured; removing it would yield an AttributeError instead. No change applied.

tests/recipe/test_contracts.py
Trecek: Investigated — the byte-level pattern is intentional. The test must enforce the exact contract literal (`%%ORDER_UP%%` token with newline before it) because downstream parsers consume that exact format. A semantic match (`".md" in p`) would allow contract drift. No change applied.

tests/recipe/test_contracts.py
Trecek: Acknowledged — minor suggestion noted.

src/autoskillit/skills_extended/investigate/SKILL.md
Trecek: Acknowledged — minor suggestion noted.

src/autoskillit/skills_extended/investigate/SKILL.md
Trecek: Acknowledged — minor suggestion noted.

tests/server/test_tools_execution_write_prefix.py
Trecek: Acknowledged — minor suggestion noted.

Issue #4218, opened 2026-07-09
Codex model aliases route default sessions to forbidden gpt-5.4


Trecek:
## Summary

Codex model aliasing still routes default `sonnet` and `haiku` class sessions to
forbidden 5.4-era model IDs. This makes fleet remediation dispatches invalid under the
current operator constraint that only `gpt-5.5` should be used.

## Evidence

- Fresh #4202 Codex self-test rerun dispatch
  `4e6fbbea-f4fc-4176-bfce-34cb09157a68` was started on 2026-07-09.
- The parent L2 process launched with `--model gpt-5.4`, even though the run was intended
  to exercise the Codex backend with the currently allowed model family.
- The process was intentionally interrupted before it could count as a valid #4202
  self-test.
- Source evidence: `src/autoskillit/core/types/_type_backend.py` mapped:
  - `sonnet` -> `gpt-5.4`
  - `opus` -> `gpt-5.5`
  - `haiku` -> `gpt-5.4-mini`
- Because recipes and fleet defaults frequently pass `sonnet`, the default-safe route was
  not actually safe for Codex during the current provider window.

## Expected Behavior

All Codex alias classes used by recipes and fleet dispatch should resolve to `gpt-5.5`.
Class distinctions may still be represented through `model_reasoning_effort`.

## Actual Behavior

Codex L2 orchestration can launch with `gpt-5.4`, and `haiku` class sessions can launch
with `gpt-5.4-mini`.

## Acceptance Criteria

- `CODEX_MODEL_ALIASES` contains no `gpt-5.4` or `gpt-5.4-mini` values.
- `sonnet`, `opus`, and `haiku` all resolve to `gpt-5.5` for Codex.
- Tests allow duplicate Codex model IDs when the class distinction is carried by
  `model_reasoning_effort`.
- Doctor/model alias allowlists reject 5.4-era Codex IDs.
- #4202 can be rerun without a Codex L2 process using `--model gpt-5.4`.

## Prior Attempt Context

This blocks #4202 after PR #4217 was reverted. The #4202 rerun must be fully autonomous
under Codex; manually completing the recipe tail cannot count. This issue only remediates
the model alias blocker so #4202 can be rerun correctly.

## Operator Requirement Update: GPT-5.6 Sol (2026-07-09)

PR #4219 completed the original 5.5 acceptance criteria and merged into `develop` at
[...]

PR #4219, opened 2026-07-09
Route Codex Aliases to GPT-5.5

Trecek:
## Summary

Codex model aliases now route `sonnet`, `opus`, and `haiku` to `gpt-5.5`, preventing
default Codex fleet sessions from launching on forbidden 5.4-era model IDs. The local
class distinction is preserved through `model_reasoning_effort`.

The doctor allowlist and model-alias regression tests now reject `gpt-5.4` and
`gpt-5.4-mini` as Codex alias targets.

Closes #4218

## Implementation Plan

Plan file: `/home/talon/projects/generic_automation_mcp/.autoskillit/temp/prepare-pr/plan_codex_model_alias_gpt55_4218.md`

Generated via AutoSkillit


Trecek (commented):

src/autoskillit/core/types/_type_backend.py
```diff
@@ -178,26 +178,26 @@ class BackendCapabilities:
...
 }
 
-_CODEX_MODEL_REVERSE: dict[str, str] = {v: k for k, v in CODEX_MODEL_ALIASES.items()}
-_codex_alias_values = list(CODEX_MODEL_ALIASES.values())
-assert len(_CODEX_MODEL_REVERSE) == len(CODEX_MODEL_ALIASES), (
-    "CODEX_MODEL_ALIASES values must be unique — duplicate makes an alias unreachable "
-    f"via model_class(). Duplicates: "
-    f"{[v for v in _codex_alias_values if _codex_alias_values.count(v) > 1]}"
-)
+_CODEX_MODEL_REVERSE: dict[str, str] = {}
```
Trecek: [critical] arch: Removing the uniqueness invariant on CODEX_MODEL_ALIASES values violates a structural contract of the alias layer. The loop makes the reverse mapping intentionally lossy: model_class('gpt-5.5') deterministically returns 'opus' (last iteration wins), erasing the distinction between sonnet/haiku/opus callers. Any downstream consumer relying on model_class() for telemetry, accounting, capability gating, or session diagnostics will silently misclassify sonnet and haiku dispatches as opus. The reverse lookup table is no longer a function from model_id to canonical class — it is now a destructive lossy projection.

src/autoskillit/cli/doctor/_doctor_runtime.py
```diff
@@ -28,7 +28,7 @@
 
 CODEX_MIN_VERSION: tuple[int, ...] = (0, 130, 0)
 
-VALID_CODEX_MODEL_IDS: frozenset[str] = frozenset({"gpt-5.4", "gpt-5.4-mini", "gpt-5.5"})
+VALID_CODEX_MODEL_IDS: frozenset[str] = frozenset({"gpt-5.5"})
```
Trecek: [warning] tests: VALID_CODEX_MODEL_IDS is hardcoded in two places: src/autoskillit/cli/doctor/_doctor_runtime.py:31 and tests/execution/test_model_alias_registry.py:9. There is no test asserting these two stay in sync. The bug fixed by this PR (stale 'gpt-5.4' entries) was possible precisely because two copies had to stay in sync.

tests/execution/test_model_alias_registry.py
```diff
@@ -54,3 +54,11 @@ def test_codex_alias_values_differ_from_keys() -> None:
             f"CODEX_MODEL_ALIASES[{key!r}] == {value!r}: identity mapping passes raw Anthropic "
             "alias to Codex CLI. Codex requires real OpenAI model IDs."
         )
+
+
+def test_codex_alias_values_do_not_route_to_forbidden_models() -> None:
```
Trecek: [warning] tests: test_codex_alias_values_do_not_route_to_forbidden_models overlaps with test_codex_alias_values_in_allowlist (line 29). The allowlist test already verifies every alias value is in VALID_CODEX_MODEL_IDS. The new test creates two sources of truth for what's allowed/forbidden that can drift apart.

tests/execution/backends/test_model_translation.py
```diff
@@ -200,12 +200,12 @@ def test_suffix_stripping(self) -> None:
         assert model_class("sonnet[1m]") == "sonnet"
         assert model_class("haiku[1m]") == "haiku"
 
-    def test_codex_reverse_map(self) -> None:
+    def test_codex_reverse_map_prefers_highest_class_for_shared_model_id(self) -> None:
```
Trecek: [warning] tests: test_codex_reverse_map_prefers_highest_class_for_shared_model_id — the test name says 'highest' but the loop walks ('haiku', 'sonnet', 'opus') so the answer is 'last wins', not 'highest'. Misleading test name.

Trecek (commented):

src/autoskillit/core/types/_type_backend.py
```diff
@@ -178,26 +179,33 @@ class BackendCapabilities:
...
 }
 
-_CODEX_MODEL_REVERSE: dict[str, str] = {v: k for k, v in CODEX_MODEL_ALIASES.items()}
-_codex_alias_values = list(CODEX_MODEL_ALIASES.values())
-assert len(_CODEX_MODEL_REVERSE) == len(CODEX_MODEL_ALIASES), (
-    "CODEX_MODEL_ALIASES values must be unique — duplicate makes an alias unreachable "
-    f"via model_class(). Duplicates: "
-    f"{[v for v in _codex_alias_values if _codex_alias_values.count(v) > 1]}"
-)
+# Reverse lookup is valid only for one-to-one native IDs. When multiple local
```
Trecek, line 217: [warning] arch: Module-level mutation loop (lines 199-208) executes at import time, mutating _CODEX_MODEL_REVERSE and _CODEX_MODEL_DUPLICATES via a multi-step imperative loop. For an IL-0 core types module, a comprehension-based approach would be more declarative.

src/autoskillit/core/types/_type_backend.py
```diff
@@ -178,26 +179,33 @@ class BackendCapabilities:
 }
 
 CODEX_MODEL_ALIASES: dict[str, str] = {
```
Trecek: [warning] arch: Routing all three aliases (sonnet, opus, haiku) to the same native ID gpt-5.5 means the reverse map will be empty after duplicate eviction. Callers relying on _CODEX_MODEL_REVERSE will silently return None/KeyError.

tests/cli/test_doctor_runtime.py
```diff
@@ -21,7 +21,7 @@ def test_ok_when_within_threshold_and_valid_aliases(
         monkeypatch.setattr(
             mod,
             "CODEX_MODEL_ALIASES",
-            {"sonnet": "gpt-5.4", "opus": "gpt-5.5", "haiku": "gpt-5.4-mini"},
+            {"sonnet": "gpt-5.5", "opus": "gpt-5.5", "haiku": "gpt-5.5"},
```
Trecek, line 24: [warning] tests: test_doctor_runtime hardcodes model id strings in monkeypatched CODEX_MODEL_ALIASES dicts. Same anti-pattern the PR fixes at source level.

tests/execution/backends/test_model_translation.py
```diff
@@ -200,12 +200,14 @@ def test_suffix_stripping(self) -> None:
         assert model_class("sonnet[1m]") == "sonnet"
         assert model_class("haiku[1m]") == "haiku"
 
-    def test_codex_reverse_map(self) -> None:
+    def test_codex_shared_model_id_stays_native_when_class_is_ambiguous(self) -> None:
```
Trecek, line 203: [warning] tests: test_codex_shared_model_id_stays_native_when_class_is_ambiguous only verifies model_class() returns the native id for the opus alias value. Should test all alias values.

src/autoskillit/core/types/_type_backend.py
Trecek, line 217: [warning] defense: The uniqueness assert was the only construction-time guard preventing ambiguous reverse-lookup. The new comment is documentation, not enforcement.

src/autoskillit/core/types/_type_backend.py
```diff
@@ -178,26 +179,33 @@ class BackendCapabilities:
...
-_CODEX_MODEL_REVERSE: dict[str, str] = {v: k for k, v in CODEX_MODEL_ALIASES.items()}
-_codex_alias_values = list(CODEX_MODEL_ALIASES.values())
-assert len(_CODEX_MODEL_REVERSE) == len(CODEX_MODEL_ALIASES), (
-    "CODEX_MODEL_ALIASES values must be unique — duplicate makes an alias unreachable "
-    f"via model_class(). Duplicates: "
-    f"{[v for v in _codex_alias_values if _codex_alias_values.count(v) > 1]}"
-)
+# Reverse lookup is valid only for one-to-one native IDs. When multiple local
+# classes share a Codex model, the class is carried by model_reasoning_effort.
+_CODEX_MODEL_REVERSE: dict[str, str] = {}
```
Trecek: [warning] defense: _CODEX_MODEL_REVERSE and _CODEX_MODEL_DUPLICATES are typed as plain dict/set with no marker that the post-loop result is bijection-safe.

src/autoskillit/core/types/_type_backend.py
```diff
@@ -178,26 +179,33 @@ class BackendCapabilities:
...
+    "sonnet": "gpt-5.5",
     "opus": "gpt-5.5",
-    "haiku": "gpt-5.4-mini",
+    "haiku": "gpt-5.5",
 }
 
-CODEX_MODEL_ALIASES_LAST_VERIFIED: str = "2026-06-11"
+CODEX_MODEL_ALIASES_LAST_VERIFIED: str = "2026-07-09"
+
+CODEX_VALID_MODEL_IDS: frozenset[str] = frozenset({"gpt-5.5"})
```
Trecek, line 193: [warning] defense: CODEX_VALID_MODEL_IDS is a frozenset that must contain every value of CODEX_MODEL_ALIASES, but no construction-time check enforces this. A module-level assert would catch this at import time.

src/autoskillit/core/types/_type_backend.py
```diff
@@ -178,26 +179,33 @@ class BackendCapabilities:
...
+_CODEX_MODEL_REVERSE: dict[str, str] = {}
+_CODEX_MODEL_DUPLICATES: set[str] = set()
+for _codex_class, _codex_model_id in CODEX_MODEL_ALIASES.items():
+    if _codex_model_id in _CODEX_MODEL_REVERSE:
+        _CODEX_MODEL_DUPLICATES.add(_codex_model_id)
+    else:
+        _CODEX_MODEL_REVERSE[_codex_model_id] = _codex_class
+for _codex_model_id in _CODEX_MODEL_DUPLICATES:
+    _CODEX_MODEL_REVERSE.pop(_codex_model_id, None)
+del _codex_class, _codex_model_id, _CODEX_MODEL_DUPLICATES
```
Trecek: [warning] slop: Use of del to clean up module-level loop variables is an unusual pattern. Module-level variables persist regardless, so del has no practical effect here.

src/autoskillit/core/types/_type_backend.py
Trecek: Replaced imperative duplicate-eviction loop with a comprehension-based reverse map that only retains bijection-safe entries (model_id -> alias only when the model_id maps to exactly one alias). Added assert that catches alias-value ambiguity at import time. model_class('gpt-5.5') now returns 'gpt-5.5' via the .get(base, base) fallback on line 232, not 'opus' — telemetry/analytics no longer silently misclassify sonnet/haiku calls as opus. See planned_fixes_4219.md for the consolidated diff.

src/autoskillit/cli/doctor/_doctor_runtime.py
Trecek: Verified the AFTER state of this PR. tests/execution/test_model_alias_registry.py no longer defines VALID_CODEX_MODEL_IDS locally — both call sites (doctor/_doctor_runtime.py and the test file) now import the single source-of-truth CODEX_VALID_MODEL_IDS from _type_backend.py. The duplication this review points to was the BEFORE state.

tests/execution/test_model_alias_registry.py
Trecek: The referenced test 'test_codex_alias_values_do_not_route_to_forbidden_models' does not exist in the codebase (grep returns zero results). Only test_codex_alias_values_in_allowlist exists at tests/execution/test_model_alias_registry.py:27, and it checks CODEX_VALID_MODEL_IDS membership — the single source of truth established by this PR. No two-source-of-truth drift.

tests/execution/backends/test_model_translation.py
Trecek: The referenced test 'test_codex_reverse_map_prefers_highest_class_for_shared_model_id' does not exist in the codebase. The test was renamed to test_codex_shared_model_id_stays_native_when_class_is_ambiguous in this PR and lives at tests/execution/backends/test_model_translation.py:203. The new name accurately describes what the test asserts (native id passthrough, not 'highest' or 'last wins').

src/autoskillit/core/types/_type_backend.py
Trecek: Replaced the imperative loop with a single comprehension: `{model_id: alias for alias, model_id in CODEX_MODEL_ALIASES.items() if sum(1 for v in CODEX_MODEL_ALIASES.values() if v == model_id) == 1}`. No multi-step mutation, no `del`-based cleanup. Module-level execution is now declarative.

src/autoskillit/core/types/_type_backend.py
Trecek: This is intentional design per the comment at lines 197-198 of _type_backend.py: 'When multiple local classes share a Codex model, the class is carried by model_reasoning_effort.' The reverse map deliberately excludes ambiguous entries; callers must use model_reasoning_effort to disambiguate. model_class() returns the base model name via the .get(base, base) fallback — no None/KeyError is raised.

tests/cli/test_doctor_runtime.py
Trecek: Monkeypatching specific alias values in test_doctor_runtime is the intended pattern for a staleness check — the test deliberately exercises a specific alias map to verify the doctor check returns OK vs WARNING based on alias validity. The PR updates the test monkeypatches to match the new production constants (gpt-5.5 everywhere); no drift.

tests/execution/backends/test_model_translation.py
Trecek: The test at lines 207-209 asserts `{CODEX_MODEL_ALIASES[key] for key in ('sonnet', 'opus', 'haiku')}` is a singleton set, then checks `model_class(shared_model_id) == shared_model_id`. This already validates ALL alias values collapse to one shared id — no per-key assertions needed.

src/autoskillit/core/types/_type_backend.py
Trecek: Added `assert set(CODEX_MODEL_ALIASES.values()).issubset(CODEX_VALID_MODEL_IDS)` at module scope in _type_backend.py. This is the new construction-time guard that catches any alias value not in the allowlist at import time, replacing the old uniqueness assert with one that fits the new design (where multiple aliases can share a native model).

src/autoskillit/core/types/_type_backend.py
Trecek: The comprehension-based replacement makes the post-loop state deterministic by construction — only bijection-safe entries survive. Adding type-level markers adds complexity without solving a real problem; the construction-time assert (Fix 1) provides the appropriate enforcement boundary.

src/autoskillit/core/types/_type_backend.py
Trecek: Added `assert set(CODEX_MODEL_ALIASES.values()).issubset(CODEX_VALID_MODEL_IDS)` at module scope in _type_backend.py. Any drift between alias values and the allowlist now fails at import time.

src/autoskillit/core/types/_type_backend.py
Trecek: Comprehension-based reverse map replaces the imperative loop. No `del` calls remain — module-level execution is declarative and idiomatic.

Trecek (commented):

src/autoskillit/cli/doctor/_doctor_runtime.py
```diff
@@ -14,6 +14,7 @@
 from autoskillit.core import (
```
Trecek, line 14: [warning] cohesion: Doctor runtime now imports CODEX_VALID_MODEL_IDS from core.types, coupling the CLI doctor to a constant defined alongside backend alias data. The allowlist is logically a property of the alias registry, not of the doctor. Consider exposing a validation helper (e.g., is_valid_codex_model_id(model_id) -> bool) on the core module so the doctor can use it without importing the raw allowlist.

Trecek (commented):

src/autoskillit/core/types/_type_backend.py
```diff
@@ -178,26 +180,35 @@ class BackendCapabilities:
 }
 
 CODEX_MODEL_ALIASES: dict[str, str] = {
```
Trecek: [warning] defense: CODEX_MODEL_ALIASES is a module-level mutable dict; the construction-time assert at L192 only runs at import. Any post-import mutation (e.g., .update(), .popitem(), reassignment of an entry's value) bypasses the assert and can leave alias values outside CODEX_VALID_MODEL_IDS, or introduce silent duplicates that _codex_unique_model_reverse will then silently drop. Defense hardening options: (a) freeze via MappingProxyType after the assert, (b) add a runtime validation helper that callers must invoke after mutation, or (c) document the dict as immutable and rely on type-checker enforcement.

src/autoskillit/core/types/_type_backend.py
```diff
@@ -178,26 +180,35 @@ class BackendCapabilities:
...
 
-_CODEX_MODEL_REVERSE: dict[str, str] = {v: k for k, v in CODEX_MODEL_ALIASES.items()}
-_codex_alias_values = list(CODEX_MODEL_ALIASES.values())
-assert len(_CODEX_MODEL_REVERSE) == len(CODEX_MODEL_ALIASES), (
-    "CODEX_MODEL_ALIASES values must be unique — duplicate makes an alias unreachable "
-    f"via model_class(). Duplicates: "
-    f"{[v for v in _codex_alias_values if _codex_alias_values.count(v) > 1]}"
-)
+
+def _codex_unique_model_reverse(aliases: Mapping[str, str]) -> dict[str, str]:
```
Trecek: [warning] defense: _codex_unique_model_reverse returns dict[str, str] but silently drops duplicate-value entries; the declared return type does not communicate that the result is a *strict subset* of the input's keys (only entries whose values are unique survive). Either annotate as Mapping[str, str] with a docstring noting the lossy semantics, or return a more specific type such as a TypedDict/NamedTuple flagging dropped entries, so callers cannot accidentally rely on every key being present.

Trecek (commented):

Warning-level findings detected: 1 mutation-after-import concern, 1 lossy return type concern. No blocking issues. See inline comments.

src/autoskillit/core/types/_type_backend.py
```diff
@@ -177,27 +180,39 @@ class BackendCapabilities:
...
 }
 
-_CODEX_MODEL_REVERSE: dict[str, str] = {v: k for k, v in CODEX_MODEL_ALIASES.items()}
-_codex_alias_values = list(CODEX_MODEL_ALIASES.values())
-assert len(_CODEX_MODEL_REVERSE) == len(CODEX_MODEL_ALIASES), (
-    "CODEX_MODEL_ALIASES values must be unique — duplicate makes an alias unreachable "
-    f"via model_class(). Duplicates: "
-    f"{[v for v in _codex_alias_values if _codex_alias_values.count(v) > 1]}"
-)
+
```
Trecek, line 206: [warning] defense: _codex_unique_model_reverse silently drops duplicate-value entries; with all three aliases now mapping to gpt-5.5 the reverse dict is empty, so model_class(gpt-5.5) cannot route by class alone — confirmed intentional by the comment at L209 (class is carried by model_reasoning_effort), but the empty result and the helper return type do not communicate the subset semantics to callers.

Trecek (commented):

src/autoskillit/cli/doctor/_doctor_runtime.py
```diff
@@ -19,6 +19,7 @@
     atomic_write,
     default_log_dir,
     get_logger,
+    is_valid_codex_model_id,
```
Trecek, line 22: [warning] cohesion: Doctor runtime (cli/IL-3) imports CODEX_VALID_MODEL_IDS and is_valid_codex_model_id from core.types (IL-0). Doctor is a leaf surface that now carries a direct dependency on a validator helper.

Trecek (commented):

### ⚠️ Outside Diff Range

**tests/execution/test_model_alias_registry.py**

- **L24** [warning/duplication]: Hardcoded model ID strings {'sonnet': 'gpt-5.5', 'opus': 'gpt-5.5', 'haiku': 'gpt-5.5'} in monkeypatch duplicate the sam
- **L48** [warning/duplication]: Same hardcoded model ID strings as L24. Centralizing the alias monkeypatch into a fixture/helper would reduce duplicatio

Trecek (commented):
AutoSkillit review passed. No blocking issues found — all new findings match prior resolved threads (within ±5 line tolerance) on the same files and are suppressed by the prior-resolution rule.

src/autoskillit/cli/doctor/_doctor_runtime.py
Trecek: Substantively addressed by prior commits. `is_valid_codex_model_id` helper was added to `src/autoskillit/core/types/_type_backend.py:239`, exported via `__all__`, re-exported through `src/autoskillit/core/__init__.pyi:372`, and adopted by the doctor at `src/autoskillit/cli/doctor/_doctor_runtime.py:392` (replacing the raw `CODEX_VALID_MODEL_IDS` membership check). The coupling this review flags is gone — the doctor now imports the predicate, not the underlying allowlist. Minor residual: error message at `_doctor_runtime.py:399` still names the constant — cosmetic text-only mismatch.

src/autoskillit/core/types/_type_backend.py
Trecek: Investigated — `CODEX_MODEL_ALIASES` is already wrapped in `MappingProxyType` at `_type_backend.py:183`, so post-import mutation is structurally prevented (not merely detected by the assert). The construction-time assert at lines 195-198 is a second layer of defense on top of an already-immutable structure. Option (a) from this comment is in place.

src/autoskillit/core/types/_type_backend.py
Trecek: Investigated — `_codex_unique_model_reverse` already returns `Mapping[str, str]` (broader than `dict`) and the docstring at `_type_backend.py:208-212` explicitly documents the lossy semantics: "Shared native IDs are intentionally omitted so model_class() falls back to the Codex model ID instead of projecting to an arbitrary local class." Subset semantics are communicated in-place. Narrow return type would be a stylistic preference rather than a defect fix.

src/autoskillit/core/types/_type_backend.py
Trecek: Investigated — overlap with 3553875482 (same function). The empty reverse dict is intentional per the design rationale documented in both the function docstring (`_type_backend.py:208-212`) and the inline comment at lines 217-218 ("Reverse lookup is valid only for one-to-one native IDs. When multiple local classes share a Codex model, the class is carried by model_reasoning_effort."). Subset semantics are communicated via two layers of in-place documentation.

src/autoskillit/cli/doctor/_doctor_runtime.py
Trecek: Investigated — the doctor import block at `_doctor_runtime.py:14-23` does NOT import `CODEX_VALID_MODEL_IDS`; it imports `is_valid_codex_model_id` (line 22). The cohesion violation does not exist in the current state of the code. Cohesion is satisfied: the doctor depends on the validator helper, not the raw allowlist constant.

PR #4250, opened 2026-07-14
Remediation Plan: command_guards_verb_position

### LINKED ISSUE #4222: Command guards execute-match quoted search patterns and block harmless reads
# Command guards execute-match quoted search patterns and block harmless reads

## Problem

Command policy guards classify forbidden substrings anywhere in a shell command as executable behavior, even when those substrings are data passed to a read-only search command.

The #4202 autonomous Codex self-test's final health analysis confirmed this in both `make-plan` and `dry-walkthrough`: read-only `rg` searches for plan text containing `pytest`, `python -m pytest`, or `pip install -e` were denied as test/install execution. Agents had to rewrite literals such as `pytest` to `py[t]est` to inspect the files.

The driver reproduced the same bug while doing duplicate detection. This read-only GitHub search was denied as an editable install:

```text
gh issue list --state all --search '"pip install -e" guard search in:title,body' --limit 30 --json number,state,title,labels,url
```

The hook response claimed the command was installing into system Python even though `pip install -e` appeared only inside the `--search` argument to `gh issue list`.

## Impact

- Planning and review agents cannot safely search documentation or issue bodies for prohibited command examples.
- Workarounds consume context and make searches less reliable.
- The behavior is broader than #4207: that fix addressed protected recipe/skill paths, while this recurrence is command-position classification of quoted argument data.

## Expected Behavior

Guards should classify executable command positions and shell structure. Forbidden test/install commands should still be denied when invoked, but inert quoted arguments to known read-only commands such as `rg`, `grep`, and `gh issue list --search` must not be treated as execution.

## Acceptance Criteria

- Literal forbidden-command text in read-only search arguments is allowed.
- The same text in an executable shell segment remains denied.
- Chaining, substitutions, eval-like constructs, and unsafe wrappers remain fail-closed.
- Hook-side and server-side command classification stay behaviorally aligned.
- Focused regressions cover `rg`, `grep`, and `gh issue list --search` with quoted test/install examples.

## Evidence

- Dispatch: `f6b3f866-1f83-487d-80e0-1ff00a4f74c5`
- Health session: `019f4926-9abc-74f0-bd84-effbf70e1f99`
- Health finding: `dry_walkthrough/make_plan: command guards falsely blocked harmless read-only searches`

This is nonblocking for the immediate #4202 retry because agents found an obfuscation workaround, but it is a genuine pipeline regression and should be processed after the queue blocker fixes.

Trecek:
## Summary

Complete the two missing Part A requirements identified by the audit while preserving all already-covered behavior. Limit production and test changes to:

- `src/autoskillit/core/bash_write_targets.py`
- `src/autoskillit/hooks/_command_classification.py`
- `src/autoskillit/hooks/guards/unsafe_install_guard.py`
- `tests/core/test_bash_write_targets.py`
- `tests/hooks/test_command_classification.py`
- `tests/infra/test_unsafe_install_guard.py`

Step 1 completes command-start parity in the IL-0 mirror by correcting `_command_start_index()` (mirroring hook-side env/wrapper/detached option arity, malformed-wrapper handling, `command_verb_and_args()` semantics) and wiring it into `_command_verb()`; extended parity tests cover wrapped write-target commands (FOO=bar, sudo, timeout, stdbuf, nohup) and malformed wrapper cases, asserting identical verbs at both core and hook layers.

Step 2 restores the narrow non-Python/raw interpreter detector: `has_interpreter_wrapped_command()` learns Node `child_process.exec/execSync/spawn/spawnSync` and Perl/Ruby `system(...)`; `_iter_install_segments()` normalizes segments via `command_verb_and_args()` + `shlex.join()` and applies the secondary helper to non-Python interpreter segments only; Python subprocesses remain routed through `extract_interpreter_command_payloads()`; `_is_unsafe_editable_install()` denies the new secondary-interpreter result without any `.venv` raw-text exemption. Tests cover real Node payload denials, inert Node `console.log` allowances, mixed-command denial, secondary denial without `--python .venv` exemption, and the existing fail-closed unresolved Python subprocess branch.

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260713-200935-309632/.autoskillit/temp/rectify/remediation_command_guards_verb_position_2026-07-13_224733.md`

Closes #4222

### DIFF
--- a/src/autoskillit/core/bash_write_targets.py
+++ b/src/autoskillit/core/bash_write_targets.py
@@ -65,6 +65,23 @@ _ENV_VALUE_FLAGS: frozenset[str] = frozenset(
     {"-u", "--unset", "-C", "--chdir", "-S", "--split-string", "--argv0"}
 )
 
+_WRAPPER_VALUE_FLAGS_DETACHED: frozenset[str] = frozenset(
+    {
+        "-u",
+        "--user",
+        "-n",
+        "--adjustment",
+        "-k",
+        "--kill-after",
+        "-s",
+        "--signal",
+        "-g",
+        "--group",
+        "-p",
+        "--priority",
+    }
+)
+
 
 def _resolve_write_target(path: str, cwd: str = "") -> str | None:
     if not path:
@@ -206,6 +223,31 @@ def _is_posix_assignment(token: str) -> bool:
     return all(c.isalnum() or c == "_" for c in name)
 
 
+def _consume_wrapper_options(start: int, segment: list[str]) -> int:
+    """Skip past a wrapper's attached/detached value options.
+
+    Wrappers (sudo, nice, nohup, time, command) accept option flags before
+    the inner command. Many take values either attached (--user=root) or as
+    the next detached token (-u root). Returns the index of the first
+    non-option token.
+    """
+    i = start
+    while i < len(segment):
+        token = segment[i]
+        if token == "--":
+            return i + 1
+        if not token.startswith("-"):
+            return i
+        if "=" in token:
+            i += 1
+            continue
+        if token in _WRAPPER_VALUE_FLAGS_DETACHED and i + 1 < len(segment):
+            i += 2
+            continue
+        i += 1
+    return i
+
+
 def _command_start_index(segment: list[str]) -> int | None:
     if not segment:
         return None
@@ -226,7 +268,11 @@ def _command_start_index(segment: list[str]) -> int | None:
                 start += 1
             continue
         if token in _COMMAND_WRAPPERS:
-            start += 1
+            new_start = _consume_wrapper_options(start + 1, segment)
+            if new_start <= start + 1:
+                start += 1
+                continue
+            start = new_start
             continue
         if token in _WRAPPERS_WITH_DURATION and start + 1 < len(segment):
             start += 2
--- a/src/autoskillit/core/bash_write_targets.py
+++ b/src/autoskillit/core/bash_write_targets.py
@@ -15,7 +15,7 @@ import os
 import re
 import shlex
 
-_SHELL_OPS: frozenset[str] = frozenset({"&&", "||", ";", "!", "|", "("})
+_SPLIT_TOKENS: frozenset[str] = frozenset({"&&", "||", ";", "|", "&"})
 
 _HEREDOC_BODY_RE = re.compile(
     r"(<<-?\s*['\"]?(\w+)['\"]?[^\n]*)\n.*?\n\t*(\2)(?=[ \t]*(?:\n|$))",
@@ -52,6 +52,20 @@ _WRITE_VERBS: frozenset[str] = frozenset(
 
 _GIT_FLAG_WITH_VALUE: frozenset[str] = frozenset({"-C", "--git-dir", "--work-tree", "-c"})
 
+_COMMAND_WRAPPERS: frozenset[str] = frozenset({"command", "nice", "time", "sudo", "nohup"})
+_WRAPPERS_WITH_DURATION: frozenset[str] = frozenset({"timeout"})
+_WRAPPERS_WITH_SHORT_FLAG: frozenset[str] = frozenset({"stdbuf"})
+
+_ENV_NO_VALUE_FLAGS: frozenset[str] = frozenset(
+    {"-i", "-0", "-v", "-V", "--help", "--version", "--ignore-environment", "--null", "--debug"}
+)
+_ENV_VALUE_FLAGS: frozenset[str] = frozenset(
+    {"-u", "--unset", "-C", "--chdir", "-S", "--split-string", "--argv0"}
+)
+_ENV_VALUE_FLAGS_ATTACHED: frozenset[str] = frozenset(
+    {"-u", "--unset", "-C", "--chdir", "--argv0"}
+)
+
 
 def _resolve_write_target(path: str, cwd: str = "") -> str | None:
     if not path:
@@ -73,15 +87,46 @@ def _strip_heredoc_bodies(command: str) -> str:
     return _HEREDOC_BODY_RE.sub(r"\1\n\3", command)
 
 
+def _normalize_newlines(command: str) -> str:
+    result: list[str] = []
+    in_single = False
+    in_double = False
+    i = 0
+    while i < len(command):
+        c = command[i]
+        if c == "\\" and not in_single and i + 1 < len(command):
+            result.append(c)
+            result.append(command[i + 1])
+            i += 2
+            continue
+        if c == "'" and not in_double:
+            in_single = not in_single
+        elif c == '"' and not in_single:
+            in_double = not in_double
+        elif c == "\n" and not in_single and not in_double:
+            result.append(" ; ")
+            i += 1
+            continue
+        result.append(c)
+        i += 1
+    return "".join(result)
+
+
 def _tokenize_command_segments(command: str) -> list[list[str]]:
     try:
-        tokens = shlex.split(_strip_heredoc_bodies(command))
+        lexer = shlex.shlex(
+            _normalize_newlines(_strip_heredoc_bodies(command)),
+            posix=True,
+            punctuation_chars=";&|",
+        )
+        lexer.whitespace_split = True
+        tokens = list(lexer)
     except (ValueError, TypeError):
         return []
     segments: list[list[str]] = []
     current: list[str] = []
     for token in tokens:
-        if token in _SHELL_OPS:
+        if token in _SPLIT_TOKENS:
             if current:
                 segments.append(current)
                 current = []
@@ -148,6 +193,53 @@ def _extract_redirect_targets(tokens: list[str], cwd: str = "") -> list[str]:
                     targets.append(resolved)
         i += 1
     return targets
+
+
+def _is_posix_assignment(token: str) -> bool:
+    if "=" not in token or token.startswith("="):
+        return False
+    name, _sep, _value = token.partition("=")
+    if not name:
+        return False
+    if not (name[0].isalpha() or name[0] == "_"):
+        return False
+    return all(c.isalnum() or c == "_" for c in name)
+
+
+def _command_start_index(segment: list[str]) -> int | None:
+    if not segment:
+        return None
+    start = 0
+    while start < len(segment):
+        token = segment[start]
+        if _is_posix_assignment(token):
+            start += 1
+            continue
+        if token == "env":
+            start += 1
+            while start < len(segment) and (
+                segment[start].startswith("-") or "=" in segment[start]
+            ):
+                if segment[start] in _ENV_VALUE_FLAGS and start + 1 < len(segment):
+                    start += 2
+                    continue
+                start += 1
+            continue
+        if token in _COMMAND_WRAPPERS:
+            start += 1
+            continue
+        if token in _WRAPPERS_WITH_DURATION and start + 1 < len(segment):
+            start += 2
+            continue
+        if (
+            token in _WRAPPERS_WITH_SHORT_FLAG
+            and start + 1 < len(segment)
+            and segment[start + 1].startswith("-")
+        ):
+            start += 2
+            continue
+        break
+    return start if start < len(segment) else None
 
 
 def _command_verb(segment: list[str]) -> str:
--- a/src/autoskillit/core/bash_write_targets.py
+++ b/src/autoskillit/core/bash_write_targets.py
@@ -22,6 +22,11 @@ _HEREDOC_BODY_RE = re.compile(
     re.DOTALL,
 )
 
+# After _strip_heredoc_bodies() a heredoc collapses to "<<WORD ...\nWORD".
+# This removes the marker and terminator, keeping the rest of the opening
+# line (real redirects), so segments carry only executable tokens.
+_HEREDOC_MARKER_RE = re.compile(r"<<-?\s*['\"]?(\w+)['\"]?([^\n]*)\n\t*\1(?=[ \t]*(?:\n|$))")
+
 _REDIRECT_TOKEN_RE = re.compile(r"^(\d*)>{1,2}(.+)$")
 _REDIRECT_OP_ONLY_RE = re.compile(r"^(\d*)>{1,2}$")
 _FD_REDIRECT_RE = re.compile(r"^\d*>{1,2}&")
@@ -114,8 +119,9 @@ def _normalize_newlines(command: str) -> str:
 
 def _tokenize_command_segments(command: str) -> list[list[str]]:
     try:
+        stripped = _HEREDOC_MARKER_RE.sub(r"\2", _strip_heredoc_bodies(command))
         lexer = shlex.shlex(
-            _normalize_newlines(_strip_heredoc_bodies(command)),
+            _normalize_newlines(stripped),
             posix=True,
             punctuation_chars=";&|",
         )
--- a/src/autoskillit/core/bash_write_targets.py
+++ b/src/autoskillit/core/bash_write_targets.py
@@ -61,14 +61,8 @@ _COMMAND_WRAPPERS: frozenset[str] = frozenset({"command", "nice", "time", "sudo"
 _WRAPPERS_WITH_DURATION: frozenset[str] = frozenset({"timeout"})
 _WRAPPERS_WITH_SHORT_FLAG: frozenset[str] = frozenset({"stdbuf"})
 
-_ENV_NO_VALUE_FLAGS: frozenset[str] = frozenset(
-    {"-i", "-0", "-v", "-V", "--help", "--version", "--ignore-environment", "--null", "--debug"}
-)
 _ENV_VALUE_FLAGS: frozenset[str] = frozenset(
     {"-u", "--unset", "-C", "--chdir", "-S", "--split-string", "--argv0"}
-)
-_ENV_VALUE_FLAGS_ATTACHED: frozenset[str] = frozenset(
-    {"-u", "--unset", "-C", "--chdir", "--argv0"}
 )
 
 
@@ -249,14 +243,10 @@ def _command_start_index(segment: list[str]) -> int | None:
 
 
 def _command_verb(segment: list[str]) -> str:
-    if not segment:
+    start = _command_start_index(segment)
+    if start is None:
         return ""
-    start = 0
-    if segment[0] == "env" and len(segment) > 1:
-        start = 1
-        while start < len(segment) and (segment[start].startswith("-") or "=" in segment[start]):
-            start += 1
-    return segment[start] if start < len(segment) else ""
+    return segment[start]
 
 
 def _is_gh_command(segment: list[str]) -> bool:
@@ -264,6 +254,11 @@ def _is_gh_command(segment: list[str]) -> bool:
 
 
 def _extract_segment_targets(segment: list[str], cwd: str) -> list[str] | None:
+    start = _command_start_index(segment)
+    if start is None:
+        return None
+    segment = segment[start:]
+
     if _is_gh_command(segment):
         return None
 
--- a/src/autoskillit/hooks/_command_classification.py
+++ b/src/autoskillit/hooks/_command_classification.py
@@ -98,7 +98,6 @@ _ENV_NO_VALUE_FLAGS: frozenset[str] = frozenset(
         "--ignore-environment",
         "--null",
         "--debug",
-        "--split-string",
     }
 )
 _ENV_VALUE_FLAGS: frozenset[str] = frozenset(
--- a/src/autoskillit/hooks/_command_classification.py
+++ b/src/autoskillit/hooks/_command_classification.py
@@ -997,7 +997,7 @@ def _literal_to_string(node: ast.AST) -> str | None:
     return None
 
 
-def extract_interpreter_command_payloads(command: str) -> tuple[list, bool]:
+def extract_interpreter_command_payloads(command: str) -> tuple[list[str | list[str]], bool]:
     """Return (literal command payloads, has_unresolved_subprocess).
 
     Each payload is either a shell-command string (for `shell=True` calls or
@@ -1005,7 +1005,7 @@ def extract_interpreter_command_payloads(command: str) -> tuple[list, bool]:
     calls). has_unresolved_subprocess is True when a matching process-launch
     call was present but its arguments could not be statically resolved.
     """
-    payloads: list = []
+    payloads: list[str | list[str]] = []
     has_unresolved = False
     if not _INTERPRETER_RE.search(command):
         return (payloads, has_unresolved)
--- a/src/autoskillit/hooks/_command_classification.py
+++ b/src/autoskillit/hooks/_command_classification.py
@@ -791,6 +791,42 @@ def extract_shell_command_payloads(command: str) -> list[str]:
     return payloads
 
 
+def tokenize_shell_payload_segments(command: str) -> list[list[str]] | None:
+    """Return tokenized segments for every evaluated shell payload in *command*.
+
+    Walks the outer command and every extracted payload recursively with a
+    ``seen`` set so nested ``bash -c``, ``eval``, and active-substitution
+    payloads are tokenized once without loops. Each successfully parsed
+    segment of every payload is appended to the result so callers can apply
+    verb-position policies like ``command_verb_and_args`` to each segment.
+
+    Returns ``None`` when the outer command or any non-empty evaluated
+    payload cannot be tokenized; callers interpret ``None`` as no deny
+    match (fail-open). Returns ``[]`` when the command has no evaluated
+    shell payload to traverse.
+    """
+    outer = tokenize_command_segments(command)
+    if not outer and command.strip():
+        return None
+
+    result: list[list[str]] = []
+    seen: set[str] = set()
+    queue: list[str] = list(extract_shell_command_payloads(command))
+    while queue:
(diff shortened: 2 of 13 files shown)


src/autoskillit/hooks/_command_classification.py
```diff
@@ -0,0 +1,1032 @@
...
+        "-i",
+        "-0",
+        "-v",
+        "-V",
+        "--help",
+        "--version",
+        "--ignore-environment",
+        "--null",
+        "--debug",
+        "--split-string",
```
Trecek: [critical] bugs: `--split-string` is listed in both `_ENV_NO_VALUE_FLAGS` (L101) and `_ENV_VALUE_FLAGS` (L111), but GNU `env -S/--split-string` requires a value argument. `_consume_env` checks `_ENV_NO_VALUE_FLAGS` first, so `--split-string` is always treated as taking zero arguments and its mandatory value token is misidentified as the command verb. Reproduced: `command_verb_and_args(['env', '--split-string', 'FOOBAR', 'pip', 'install', '-e', '.'])` returns verb='FOOBAR' instead of 'pip'. This lets `env --split-string <value> pip install -e .` (and any other guarded verb reached via `env --split-string`) bypass verb-position detection in unsafe_install_guard, pr_create_guard, planner_gh_discovery_guard, and any other guard using `command_verb`/`command_verb_and_args`. The parallel `_command_start_index` in bash_write_targets.py does not have this bug (it treats `-S`/`--split-string` as value-taking via `_ENV_VALUE_FLAGS` only). Fix: remove `--split-string` from `_ENV_NO_VALUE_FLAGS`.

src/autoskillit/hooks/_command_classification.py
```diff
@@ -0,0 +1,1032 @@
...
+                if command[j] == "`":
+                    inner_end = j + 1
+                    while inner_end < n and command[inner_end] != "`":
+                        inner_end += 1
+                    inner = command[j + 1 : inner_end]
+                    payloads.append(inner)
+                    payloads.extend(_extract_substitution_payloads(inner))
+                    j = inner_end + 1
+                    continue
+                if command[j] == "$" and j + 1 < n and command[j + 1] == "(":
```
Trecek, line 902: [warning] bugs: In `_extract_substitution_payloads`, the paren-depth scanners for `$(...)` (double-quote branch at L863-L878 and top-level branch at L894-L909) count `(`/`)` without skipping quoted spans. A literal `)` inside a quoted string within the substitution body prematurely closes the scan, truncating the extracted payload. Reproduced: `_extract_substitution_payloads('echo $(echo "a) b" foo bar)')` returns `['echo "a']` instead of the full substitution body, so command text after the false terminator (including a subsequent dangerous verb) is never scanned by `tokenize_shell_payload_segments`/guard logic — a false-negative bypass window in the nested-shell detection this PR adds.

Trecek (commented):

### ⚠️ Outside Diff Range

**src/autoskillit/hooks/_command_classification.py**
- **L167** [warning/defense]: `_iter_install_segments` has no return type annotation; generator tuple shape consumed by guard logic is not statically checked.
- **L183** [info/defense]: `_iter_install_segments` swallows (ValueError, TypeError) with a silent `segments = []` fallback and no signal.

src/autoskillit/hooks/_command_classification.py
```diff
@@ -0,0 +1,1032 @@
...
+    return out
+
+
+def _literal_to_string(node: ast.AST) -> str | None:
+    if isinstance(node, ast.Constant) and isinstance(node.value, str):
+        return node.value
+    return None
+
+
+def extract_interpreter_command_payloads(command: str) -> tuple[list, bool]:
```
Trecek: [warning] defense: `extract_interpreter_command_payloads` is declared to return `tuple[list, bool]` with an unparameterized `list`. The payload list actually mixes two element types (str for shell-command strings, list[str] for argv literals) and callers dispatch via isinstance checks. Every other public helper in this module uses precise generics. Annotate as `tuple[list[str | list[str]], bool]` to make the boundary type-safe and self-documenting.

src/autoskillit/core/bash_write_targets.py
```diff
@@ -0,0 +1,386 @@
...
+        return os.path.join(cwd, path)
+    return None
+
+
+def _strip_heredoc_bodies(command: str) -> str:
+    return _HEREDOC_BODY_RE.sub(r"\1\n\3", command)
+
+
+def _normalize_newlines(command: str) -> str:
+    result: list[str] = []
```
Trecek, line 107: [warning] slop: `_normalize_newlines`, `_is_posix_assignment`, `_HEREDOC_MARKER_RE`, `_COMMAND_WRAPPERS`, `_WRAPPERS_WITH_DURATION`, and `_WRAPPERS_WITH_SHORT_FLAG` are duplicated verbatim between this file and hooks/_command_classification.py, and `_ENV_NO_VALUE_FLAGS`/`_ENV_VALUE_FLAGS` are duplicated with partial drift (the hooks copies include `--default-signal`/`--ignore-signal`/`--block-signal`, missing here; and the hooks copy has the `--split-string` misclassification this file avoids). The drift has already produced a divergence within this PR. If the hooks import boundary genuinely prevents sharing, consider extracting the shared stdlib-only primitives into a common importable module (the HookDef-at-package-root pattern) rather than hand-syncing two copies.

tests/infra/test_unsafe_install_guard.py
```diff
@@ -178,3 +182,487 @@ def test_missing_cmd_field_fail_open(self):
...
+# --- Quoted/read-only false-positive cases ---
+# Each entry exercises a quoted reader/search that mentions an install pattern
+# without actually invoking one. Both _run_guard and _run_bash_guard must allow.
+
+
+_QUOTED_ALLOWED: list[tuple[str, str]] = [
+    ("rg 'pip install -e' docs/", "rg-quoted-single"),
+    ('rg "pip install -e" docs/', "rg-quoted-double"),
+    ('grep -r "pip install --editable" .', "grep-double-quoted"),
+    ('grep -r "pip install --editable" .', "grep-single-quoted-search"),
```
Trecek: [warning] tests: In `_QUOTED_ALLOWED`, the entry id 'grep-single-quoted-search' (L196) uses the exact same command string as the preceding 'grep-double-quoted' entry (L195): `grep -r "pip install --editable" .`. The case is a no-op duplicate of the previous parametrization and the single-quoted grep false-positive path is untested despite the id implying coverage. Change L196 to a single-quoted variant, e.g. `grep -r 'pip install --editable' .`.

tests/infra/test_command_guard_verb_position.py
```diff
@@ -0,0 +1,215 @@
...
+                tainted.add(target_name)
+                changed = True
+    return tainted
+
+
+def _expr_is_tainted(expr: ast.expr, tainted: set[str]) -> bool:
+    """Return True when *expr* is built entirely from tainted sub-expressions."""
+    if isinstance(expr, ast.Name):
+        return expr.id in tainted
+    if isinstance(expr, ast.BinOp) and isinstance(expr.op, ast.BitOr):
```
Trecek, line 97: [warning] tests: `_expr_is_tainted`'s ast.BinOp/ast.BitOr branch (handling `cmd | other` taint propagation, added alongside the ast.BoolOp/Or branch) has no synthetic self-test in TestDetectorSyntheticSelfTests, unlike the Or-join branch covered by test_rejects_or_joined_taint. Commit f7b7633ac fixed a silently-missing BoolOp/Or propagation — the untested BitOr branch is the same class of gap; if it regresses, the ratchet's own suite would not catch it.

src/autoskillit/core/bash_write_targets.py
Trecek: Valid observation — flagged for design decision. Duplication confirmed at symbol level between core/bash_write_targets.py and hooks/_command_classification.py, including real drift in the env flag tables. However, the bash_write_targets.py module docstring documents the independent re-implementation as deliberate: standalone hook scripts import siblings via sys.path and cannot import autoskillit.* (hooks/AGENTS.md), and the hook_registry.py package-root pattern does not transfer directly because that module imports from autoskillit.core. Anti-drift protection exists via tests/core/test_bash_write_targets.py::test_parity_with_command_classification, but its corpus lacks an `env --split-string` case — which is how this drift slipped through. Whether to extract a shared stdlib-only module serving both the subprocess hook scripts and IL-0 core is a structural design choice requiring a human decision; the narrower drift fix (env flag table alignment + parity-corpus extension) can proceed independently.

src/autoskillit/hooks/_command_classification.py
Trecek: Agreed — fixed in e4c431c096d8a8742169d16536449479f7e3c651. Removed `--split-string` from `_ENV_NO_VALUE_FLAGS`; `_consume_env` now consumes its mandatory value token, so `command_verb_and_args(["env", "--split-string", "FOOBAR", "pip", "install", "-e", "."])` returns verb=pip. Added regression test `test_env_split_string_consumes_value`.

src/autoskillit/hooks/_command_classification.py
Trecek: Agreed — fixed in 56b4aee101e93db459b2938cb26401f639848e68. Extracted a quote-aware `_find_substitution_end` helper used by both the double-quote branch and the top-level `$(` scanner, so a literal `)` inside a quoted span no longer truncates the payload (verified against real bash semantics). This also closes the fail-open path where the truncated unbalanced-quote payload made `tokenize_shell_payload_segments` return None. Added regression test `test_quoted_close_paren_does_not_truncate_substitution`.

src/autoskillit/hooks/_command_classification.py
Trecek: Agreed — fixed in 4003e5751d58d46be608107e9d5b88771e4a8089. Annotated as `tuple[list[str | list[str]], bool]` (and `payloads: list[str | list[str]]`); the module has `from __future__ import annotations`, so no new imports are needed and hooks stay stdlib-only.

tests/infra/test_unsafe_install_guard.py
Trecek: Agreed — fixed in 25e3e07075bec62cdb655600473670d667708e4c. L196 now uses the single-quoted grep variant so the id matches actual coverage (the previous entry duplicated the double-quoted command string verbatim).

tests/infra/test_command_guard_verb_position.py
Trecek: Agreed — fixed in da2e58577cdf1aec75a8c26722dbab4cf6213194. Added `test_rejects_bitor_joined_taint` to TestDetectorSyntheticSelfTests mirroring `test_rejects_or_joined_taint` with a `cmd | command` join, covering the ast.BinOp/BitOr branch of `_expr_is_tainted`.

src/autoskillit/core/bash_write_targets.py
```diff
@@ -0,0 +1,386 @@
...
+        return False
+    name, _sep, _value = token.partition("=")
+    if not name:
+        return False
+    if not (name[0].isalpha() or name[0] == "_"):
+        return False
+    return all(c.isalnum() or c == "_" for c in name)
+
+
+def _command_start_index(segment: list[str]) -> int | None:
```
Trecek, line 251: [critical] defense: `_command_start_index` (with `_is_posix_assignment` and the `_COMMAND_WRAPPERS`/`_WRAPPERS_WITH_DURATION`/`_WRAPPERS_WITH_SHORT_FLAG` constants) is added by this PR but never called — dead code. The live verb-resolution path (`_command_verb` at L251, used by `_is_gh_command`/`_extract_segment_targets`) still only strips a bare `env` prefix, so `extract_bash_write_targets('sudo git checkout -- ../outside/evil.txt', '/workspace')` and the `FOO=bar git checkout ...` form both return `[]` (verified by direct execution) — wrapper- or assignment-prefixed write commands silently under-report write targets on this security-relevant boundary. The sibling `hooks/_command_classification.py` correctly wires its equivalent helper into `command_verb_and_args`; this file does not. Either wire `_command_start_index` into `_command_verb`/callers, or remove the dead function and its supporting constants.

src/autoskillit/hooks/guards/unsafe_install_guard.py
```diff
@@ -7,44 +7,234 @@
...
 def _is_system_install(cmd: str) -> bool:
-    """Return True if cmd is a pip/uv install with --system flag."""
-    cmd_lower = cmd.lower()
-    if _SYSTEM_FLAG_PATTERN not in cmd_lower:
-        return False
-    tokens = cmd_lower.split()
-    return "--system" in tokens
+    """Return True if *cmd* runs a pip/uv-pip install with --system."""
+    for kind, install_args, post_install in _iter_install_segments(cmd):
+        if kind in ("pip", "uv-pip", "module-pip") and _has_system_flag(
```
Trecek, line 230: [critical] bugs: `_is_system_install` only checks kinds `("pip", "uv-pip", "module-pip")` and never handles the `"argv-payload"` kind yielded by `_iter_install_segments` for statically-resolved subprocess argv literals. Verified by direct execution: `python3 -c "import subprocess; subprocess.run(['pip', 'install', '--system', 'requests'])"` yields `('argv-payload', ['pip', 'install', '--system', 'requests'], [])` and `_is_system_install` returns False, while the equivalent shell-text form is blocked. The `--system` flag is never inspected on a code path the guard itself resolves — a bypass of the system-install block this PR hardens.

Trecek (commented):

src/autoskillit/hooks/guards/unsafe_install_guard.py
```diff
@@ -7,44 +7,234 @@
...
+
+def _is_unsafe_editable_install(cmd: str) -> bool:
+    """Return True if *cmd* runs an editable install not targeting .venv Python."""
+    for kind, _install_args, post_install in _iter_install_segments(cmd):
+        if kind == "maturin":
+            return True
+        if kind == "unresolved-subprocess":
+            # A matching process-launch call could not be statically resolved.
+            return True
+        if kind == "argv-payload":
```
Trecek: [critical] bugs: the `"argv-payload"` branch in `_is_unsafe_editable_install` only recognizes the direct pip shape (`_is_pip_executable(argv[0]) and argv[1] == "install"`). Argv literals like `['uv','pip','install','-e','.']`, `['python3','-m','pip','install','-e','.']`, and `['maturin','develop']` fall through to `continue` and are allowed (verified by direct execution), while their shell-text equivalents are blocked via `_classify_install_invocation`. Pass the argv list through `_classify_install_invocation(argv)` instead of re-implementing a narrow pip-only check, so the argv path recognizes the same pip/uv-pip/module-pip/maturin grammars.

src/autoskillit/hooks/_command_classification.py
```diff
@@ -0,0 +1,1053 @@
...
+    # Allow leading whitespace-like wrapper-only segment with no command.
+    while i < len(segment):
+        token = segment[i]
+        if token == "--":
+            i += 1
+            break
+        if token in _ENV_NO_VALUE_FLAGS:
+            i += 1
+            continue
+        if token in _ENV_VALUE_FLAGS_ATTACHED and "=" in token:
```
Trecek: [warning] bugs: in `_consume_env`, the branch `token in _ENV_VALUE_FLAGS_ATTACHED and "=" in token` is unreachable — `_ENV_VALUE_FLAGS_ATTACHED` contains only bare flag names (no `=` character) and the membership test is exact-string, so an attached-value token like `--chdir=/tmp` never matches any member. No observable misbehavior today only because the later generic `"=" in token` fallback independently skips the same tokens, but the dedicated branch contradicts the docstring's "value flags (detached and attached)" claim, and no test exercises the attached form (e.g. `env --chdir=/tmp cmd`). Match by prefix (e.g. `token.split("=", 1)[0] in _ENV_VALUE_FLAGS_ATTACHED`) or remove the branch and the set.

src/autoskillit/core/bash_write_targets.py
Trecek: Agreed — fixed in a7564600cd14b505f0e04c4344efa0201aae9bdd. Wired `_command_start_index` into the live path: `_command_verb` now delegates to it and `_extract_segment_targets` slices the segment at the resolved command start, so `sudo git checkout -- ../outside/evil.txt`, `FOO=bar git checkout ...`, `env -C /tmp git checkout ...`, and `timeout 30 tee ...` all report their write targets (runtime-verified). The now-dead `_ENV_NO_VALUE_FLAGS`/`_ENV_VALUE_FLAGS_ATTACHED` constants left by the incomplete mirror were removed; `_COMMAND_WRAPPERS`/`_WRAPPERS_WITH_DURATION`/`_WRAPPERS_WITH_SHORT_FLAG` are now live. Added direct unit tests for wrapper/assignment/env-prefixed write commands (kept out of the parity corpus, where the hooks write_guard scans args from index 1 and would diverge).

src/autoskillit/hooks/guards/unsafe_install_guard.py
Trecek: Agreed — fixed in 3bbbbbd1d900409398a5f1662b48f1c0643229d8. Argv-list payloads are now classified through `_classify_install_invocation` at yield time in `_iter_install_segments`, so `_is_system_install` sees canonical pip/uv-pip/module-pip kinds and inspects `--system` on the argv path. Runtime-verified: `subprocess.run(['pip','install','--system','requests'])` and the uv-pip argv form are both denied; `uv tool install demo --system` remains allowed. Added deny tests for both argv `--system` shapes.

src/autoskillit/hooks/guards/unsafe_install_guard.py
Trecek: Agreed — fixed in 3bbbbbd1d900409398a5f1662b48f1c0643229d8. The narrow pip-only argv branch was removed; argv payloads now route through `_classify_install_invocation` in `_iter_install_segments`, so the argv path recognizes the same pip/uv-pip/module-pip/maturin grammars as shell text — including the `--python .venv` exemption, which the old branch ignored. Runtime-verified: uv-pip, module-pip, and maturin argv literals are denied; the rg reader argv and the `.venv`-targeted pip argv are allowed. Added deny tests for all three grammars plus an allow test for the `.venv`-exempt argv form.

src/autoskillit/hooks/_command_classification.py
Trecek: Agreed — fixed in 4e10499d8ed346cfb47d0c3c26b3f9224bef372b. The branch now matches by prefix (`"=" in token and token.split("=", 1)[0] in _ENV_VALUE_FLAGS_ATTACHED`), so attached forms like `--chdir=/tmp` are consumed by the dedicated branch as the docstring claims, instead of relying on the generic `"=" in token` fallback. Added `test_env_attached_value_flag` exercising the attached forms (`env --chdir=/tmp --unset=FOO` prefixing an editable pip install) to pin the intended behavior.

src/autoskillit/hooks/guards/unsafe_install_guard.py
```diff
@@ -7,44 +7,230 @@
...
+        install_args, post_install = match
+        return ("module-pip", install_args, post_install)
+    if _is_maturin_executable(verb):
+        if not args or args[0] != "develop":
+            return None
+        return ("maturin", ["develop"], args[1:])
+    return None
+
+
+def _iter_install_segments(command: str):
```
Trecek: [warning] defense: `_iter_install_segments(command: str)` has no return type annotation. It is a generator whose 3-tuple shape (`kind`, `install_args`, `post_install`) is destructured by both `_is_unsafe_editable_install` and `_is_system_install` — the actual security gate logic for the unsafe-install guard. An untyped boundary here means a future edit to the yielded tuple shape (e.g. adding/reordering fields) would not be caught by mypy at either call site, silently breaking the install-safety classification. Should be annotated as `Iterator[tuple[str, list[str], list[str]]]`.

src/autoskillit/hooks/guards/unsafe_install_guard.py
```diff
@@ -7,44 +7,230 @@
...
-    tokens = cmd_lower.split()
-    for i, token in enumerate(tokens):
-        if token == "--python" and i + 1 < len(tokens):
-            python_arg = tokens[i + 1]
-            if python_arg.startswith(".venv") or "/.venv/" in python_arg:
-                return False
-    return True
+    if value == ".venv" or value.startswith(".venv/") or value.startswith(".venv\\"):
+        return True
+    if "/.venv/" in ("/" + value) or value.startswith("/.venv"):
```
Trecek: [critical] bugs: `_is_venv_path`'s absolute-path branch `value.startswith("/.venv")` matches any path beginning with the literal characters `/.venv`, not just a `.venv` path component — e.g. `/.venv-poison/bin/python` and even bare `/.venv-poison` satisfy this check and return True. This directly contradicts the function's own docstring guarantee ('`.venv-poison` and unrelated outer text do not qualify') and lets an editable install with `--python /.venv-poison/bin/python` bypass `_is_unsafe_editable_install`, defeating the guard. The relative-path case (`.venv-poison/bin/python`) is correctly rejected via the `parts` component check further down, but the absolute-path fast-path at this line short-circuits before that correct logic runs. Fix by requiring a path boundary after `.venv` (e.g. `value.startswith("/.venv/")` or `value == "/.venv"`) instead of a bare prefix match.

src/autoskillit/core/bash_write_targets.py
```diff
@@ -0,0 +1,381 @@
...
+            start += 1
+            while start < len(segment) and (
+                segment[start].startswith("-") or "=" in segment[start]
+            ):
+                if segment[start] in _ENV_VALUE_FLAGS and start + 1 < len(segment):
+                    start += 2
+                    continue
+                start += 1
+            continue
+        if token in _COMMAND_WRAPPERS:
```
Trecek, line 270: [critical] cohesion: `_command_start_index` in bash_write_targets.py diverges from its sibling `command_verb_and_args`/`_consume_wrapper_options` logic in hooks/_command_classification.py: the `_COMMAND_WRAPPERS` branch here just does `start += 1` with no option-consumption, so `sudo -u root git checkout -- file` misidentifies `-u` as the verb (write-target detection silently fails). The sibling file added `_consume_wrapper_options` (and `_WRAPPER_VALUE_FLAGS_ATTACHED`/`_WRAPPER_VALUE_FLAGS_DETACHED`) specifically to handle this case (see `test_sudo_wrapper` covering `sudo -u root pip install`), but bash_write_targets.py's mirrored implementation was not updated to match, and its own test (`test_sudo_wrapped_git_checkout`) only exercises unflagged `sudo git checkout`, masking the gap.

Trecek (commented):

Trecek (commented):
AutoSkillit review: blocking findings detected (2 critical, 1 warning). See inline comments — changes requested. (Posted as COMMENT because GitHub blocks REQUEST_CHANGES on self-authored PRs.)

src/autoskillit/hooks/guards/unsafe_install_guard.py
Trecek: Agreed — fixed in ff6cea1ab245e4c2c11be6b96fb9eca1cb9d0245. Replaced the bare `value.startswith("/.venv")` prefix match with an exact `value == "/.venv"` check; the boundary-safe `"/.venv/" in ("/" + value)` disjunct already covers `/.venv/...` interpreter paths. Verified `_is_venv_path("/.venv-poison/bin/python")` now returns False and `pip install -e . --python /.venv-poison/bin/python` is denied. Added absolute-path poison regression tests (run_cmd + Bash guard paths) plus a `/.venv/bin/python` still-exempt test in tests/infra/test_unsafe_install_guard.py.

src/autoskillit/core/bash_write_targets.py
Trecek: Agreed — fixed in 239420fb9cda8bca391d59cbbbbb9845fb47d2b6. Ported `_consume_wrapper_options` and `_WRAPPER_VALUE_FLAGS_DETACHED` from hooks/_command_classification.py into bash_write_targets.py and wired them into `_command_start_index`'s `_COMMAND_WRAPPERS` branch, mirroring `command_verb_and_args`. `sudo -u root git checkout -- /path/file.txt`, `sudo --user=root rm /path/file.txt`, and `nice -n 5 tee /path/out.txt` now extract write targets correctly; unit tests added in tests/core/test_bash_write_targets.py. Two scoping notes: `_WRAPPER_VALUE_FLAGS_ATTACHED` was not ported because it is unused in the sibling as well (attached `=` forms are consumed by the generic `"=" in token` branch — porting it would add dead code). The parity corpus intentionally gains no wrapper-with-flags entry: hooks/guards/write_guard.py's git-subcommand scan iterates the unsliced segment and has a separate pre-existing gap for flagged wrappers, so a corpus entry would diverge; core-only unit tests cover this fix.

src/autoskillit/hooks/guards/unsafe_install_guard.py
Trecek: Agreed — fixed in f29d60ce993820114ee1ec0455c5ad561f9d8415. Annotated `_iter_install_segments` as `Iterator[tuple[str, list[str], list[str]]]` via a stdlib `collections.abc.Iterator` import (hooks/ stdlib-only constraint respected), so mypy now checks the yielded 3-tuple shape at both destructuring call sites (`_is_unsafe_editable_install`, `_is_system_install`).

Issue #4229, opened 2026-07-10
audit-impl cumulative plan list is rejected by scalar path gate


Trecek:
## Summary

The remediation recipe deliberately accumulates `context.all_plan_paths` as a comma-separated list and passes that value as the first positional argument to `/autoskillit:audit-impl`. The audit skill explicitly accepts that representation. However, the server-side `run_skill` input-contract gate treats the whole comma-separated token as one `file_path` and rejects it because no file exists at the concatenated path.

This prevents the second or later cumulative audit in every remediation loop before the auditor can start.

## Production occurrence: #4150

The final cumulative audit for #4150 was invoked on 2026-07-10 with two valid stamped plans:

```text
/autoskillit:audit-impl /.../remediation_classifier_immunity_plan_2026-07-10_104500.md,/.../remediation_classifier_immunity_round3_plan_2026-07-10_140530.md impl-4150-20260710-142635 develop deviation_manifest_path=
```

Both files exist independently. `run_skill` rejected the call at the input gate because it attempted `Path.is_file()` on the comma-joined token.

A worker then guessed that the plans should be space-separated and retried:

```text
/audit-impl PLAN_1 PLAN_2 impl-4150-20260710-142635 develop deviation_manifest_path=
```

That invocation passed the server gate but shifted the audit skill's positional binding to:

- `plans_input=PLAN_1`
- `implementation_ref=PLAN_2`
- `base_branch=impl-4150-20260710-142635`
- `conflict_report_paths=develop`

The driver canceled that malformed session before accepting a verdict. Its session record has `subtype=cancelled`, `exit_code=-1`, and no usable audit result.

## Root cause

Three contracts disagree at the same boundary:

1. `remediation.yaml` documents and emits `all_plan_paths` as a comma-separated accumulation.
2. `audit-impl/SKILL.md` accepts a comma-separated list as its first `plans_input` argument.
3. `skill_contracts.yaml` declares both `plan_path` and `all_plan_paths` as scalar `file_path` inputs. `_check_input_contracts()` filters positional tokens with `is_path_like_token()` and validates each token directly with `Path.is_file()`. It has no collection-aware path type or normalization step.

The first plan works because a scalar path is valid. The first remediation re-entry adds a comma and makes the documented representation impossible to pass through the gate.

## Expected behavior

The input-contract layer must accept the same cumulative-plan representations that `audit-impl` accepts, while still validating every constituent path:

- scalar plan file;
- comma-separated plan files;
[...]

PR #4254, opened 2026-07-14
[FIX] Input Contract `file_path_list` Type Immunity — Part A

Trecek:
## Summary

The input contract validation gate (`_check_input_contracts` in `src/autoskillit/server/_guards.py`) rejects comma-separated plan path lists because the type system (`InputSpec.type`) only recognizes scalar `file_path` and `directory_path`. The `file_path_list` type already exists in `src/autoskillit/recipe/skill_contracts.yaml` for output fields and `parse_plan_paths()` already splits comma-joined strings — but neither is wired into the input contract resolution or validation pipeline. This is a half-finished feature where the type vocabulary, the splitter, and the gate each live in separate modules with no structural enforcement that they stay in sync.

## Requirements

The input-contract layer must accept the same cumulative-plan representations that `audit-impl` accepts, while still validating every constituent path:
- scalar plan file;
- comma-separated plan files;
- supported audit manifest file;
- supported plan directory, if retained by the skill contract.

Each constituent path must exist and have the expected file/directory type. A missing member must fail closed with an error naming that member. The gate must not require workers to alter positional syntax in ways that change skill argument binding.

Scope:
- Introduce collection-aware input validation for `audit-impl` cumulative plans, or replace the scalar contract with a typed representation that preserves the skill accepted forms.
- Remove the ambiguous duplicate scalar `plan_path`/`all_plan_paths` contract shape if it is not semantically valid.
- Add a direct server-gate regression for two comma-separated existing plan files.
- Add a negative regression for one existing plus one missing member.
- Add a bundled-remediation integration regression proving an audit-remediation re-entry reaches the auditor with all accumulated plans in order.
- Add a positional-binding regression preventing a space-separated workaround from silently shifting `implementation_ref` and `base_branch`.

Closes #4229

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260714-070103-362151/.autoskillit/temp/rectify/rectify_input_contract_list_type_immunity_2026-07-14_073000.md`

### DIFF
--- a/src/autoskillit/recipe/skill_contracts.yaml
+++ b/src/autoskillit/recipe/skill_contracts.yaml
@@ -388,11 +388,11 @@ skills:
     outputs: []
   audit-impl:
     inputs:
-    - name: plan_path
-      type: file_path
-      required: false
     - name: all_plan_paths
-      type: file_path
+      type: file_path_list
+      required: false
+    - name: all_conflict_report_paths
+      type: file_path_list
       required: false
     - name: branch_name
       type: string
--- a/src/autoskillit/recipe/skill_contracts.yaml
+++ b/src/autoskillit/recipe/skill_contracts.yaml
@@ -505,7 +505,7 @@ skills:
   prepare-pr:
     inputs:
     - name: plan_paths
-      type: string
+      type: file_path_list
       required: true
     - name: run_name
       type: string
@@ -599,6 +599,9 @@ skills:
       required: true
     - name: audit_verdict
       type: string
+      required: false
+    - name: conflict_report_paths
+      type: file_path_list
       required: false
     - name: domain_partitions_path
       type: file_path
--- a/src/autoskillit/server/_guards.py
+++ b/src/autoskillit/server/_guards.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import os
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, assert_never
 
 import regex as re
 
@@ -17,6 +17,7 @@ from autoskillit.core import (
     extract_skill_name,
     get_logger,
     is_path_like_token,
+    parse_plan_paths,
     session_type,
 )
 from autoskillit.hooks import command_has_blocked_protected_path_read
@@ -277,7 +278,11 @@ def _check_input_contracts(
     if not specs:
         return None
 
-    args = extract_positional_args(skill_command)
+    try:
+        args = extract_positional_args(skill_command)
+    except ValueError as e:
+        skill_label = extract_skill_name(skill_command) or skill_command
+        return gate_error_result(f"Malformed quoting in skill_command for {skill_label}: {e}")
     path_args = [a for a in args if is_path_like_token(a)]
 
     for spec in specs:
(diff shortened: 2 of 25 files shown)


src/autoskillit/server/_guards.py
```diff
@@ -277,7 +278,11 @@ def _check_input_contracts(
     if not specs:
         return None
 
-    args = extract_positional_args(skill_command)
+    try:
+        args = extract_positional_args(skill_command)
+    except ValueError as e:
+        skill_label = extract_skill_name(skill_command) or skill_command
+        return gate_error_result(f"Malformed quoting in skill_command for {skill_label}: {e}")
     path_args = [a for a in args if is_path_like_token(a)]
```
Trecek, line 286: [warning] arch: The binding model compacts path-looking tokens into path_args and indexes by path-spec position. With the newly declared multiple optional path specs, omitting an earlier optional input shifts a later value into the earlier spec. For example, open-integration-pr cannot reliably omit conflict_report_paths while supplying domain_partitions_path; that value is validated as the conflict list or skipped. Bind against actual contract arguments or named inputs instead of a filtered path-token sequence.

tests/arch/test_input_spec_type_exhaustive.py
```diff
@@ -0,0 +1,106 @@
...
+
+pytestmark = [pytest.mark.layer("arch"), pytest.mark.small]
+
+
+class _MatchWithAssertNeverVisitor(ast.NodeVisitor):
+    def __init__(self) -> None:
+        self.has_match = False
+        self.has_assert_never = False
+
+    def visit_Match(self, node: ast.Match) -> None:
```
Trecek, line 23: [warning] tests: This visitor records only whether any match and any assert_never occur anywhere in the function body. It does not verify that the match dispatches InputSpec.type, that every Literal member has a branch, or that assert_never is in the wildcard arm. Unrelated constructs would let the immunity tests pass after the intended dispatch became non-exhaustive; inspect the relevant ast.Match and its fallback directly.

tests/contracts/test_input_type_semantic_correctness.py
```diff
@@ -14,10 +14,74 @@
...
+        for placeholder in placeholder_variants:
+            mentions.extend(
+                m.start() for m in re.finditer(re.escape(placeholder), skillmd_text, re.IGNORECASE)
+            )
+        if mentions:
+            for pos in mentions:
+                window = skillmd_text[max(0, pos - 200) : pos + 200]
+                if _LIST_COLLECTION_RE.search(window):
+                    return "list"
+        else:
```
Trecek, line 81: [warning] tests: When the input name and aliases are absent, the classifier scans the entire SKILL.md and treats any unrelated collection wording as evidence for this input. That can make the semantic test accept an undocumented or incorrect file_path_list declaration. Keep the no-mention case ambiguous or require an explicit alias tied to a local mention window.

tests/contracts/test_input_type_semantic_correctness.py
```diff
@@ -93,12 +165,124 @@ def test_input_type_matches_skillmd(skill_name: str, input_name: str, declared_t
...
+        f"inferred {inferred!r} from SKILL.md (aliases={aliases!r})"
+    )
+
+
+def test_collect_path_input_params_includes_file_path_list() -> None:
+    """_collect_path_input_params() must surface file_path_list inputs and be nonempty."""
+    params = _collect_path_input_params()
+    list_types = [t for (_, _, t) in params if t == "file_path_list"]
+    assert list_types, "Expected at least one file_path_list entry in the YAML manifest"
+    assert len(list_types) >= 4, (
```
Trecek, line 225: [warning] tests: This checks only that four list inputs exist, although the failure text names four specific required entries. Removing one required declaration and adding an unrelated file_path_list would still pass. Assert that the exact expected (skill, input) pairs are a subset of the collected parameters.

tests/server/test_tools_execution_input_gates.py
```diff
@@ -565,6 +583,158 @@ def test_gate_rejects_wrong_path_type(
...
+        """When conflict-report is missing, the scalar spec at path position 1 fails."""
+        from autoskillit.recipe._contracts_manifest import resolve_input_specs
+        from autoskillit.server._guards import _check_input_contracts
+
+        plan_a = tmp_path / "plan_a.md"
+        plan_a.write_text("plan a")
+        plan_b = tmp_path / "plan_b.md"
+        plan_b.write_text("plan b")
+        joined = f"{plan_a},{plan_b}"
+        cmd = f"/autoskillit:prepare-pr {joined} my-run main 123 true /tmp/does-not-exist.md"
```
Trecek, line 681: [warning] tests: This test depends on the shared absolute path /tmp/does-not-exist.md remaining absent. Another xdist worker or external process can create it and make the test nondeterministic. Construct the missing path beneath this test instance tmp_path instead.

src/autoskillit/recipe/skill_contracts.yaml
```diff
@@ -505,7 +505,7 @@ skills:
   prepare-pr:
     inputs:
     - name: plan_paths
-      type: string
+      type: file_path_list
```
Trecek, line 508: [warning] cohesion: The Part A contract explicitly reserves the prepare-pr and open-integration-pr contract fixes, the semantic-classifier list bucket, and the existing parametrized-test widening for Part B. This line begins those separately scoped changes here (along with the open-integration-pr declaration and associated tests). Remove the Part B work from this PR so the implementation matches the authorized Part A boundary.

Trecek (commented):

Trecek (commented):

Issue #4272, opened 2026-07-15
Codex investigate deep mode can consume 60K+ parent tokens from two tool outputs


Trecek:
## Current coordination — 2026-07-22

> **Coordinator status:** Immediate per-call containment and recovery mechanisms have landed. The unresolved failure is **cumulative model-visible context admission across many individually bounded calls**. Keep this issue open as the parent tracker; do not treat `staged` or open/closed state alone as implementation truth.

### What has already landed

| Issue | Merged implementation | Delivered scope | Issue-state note |
|---|---|---|---|
| This issue | [PR #4277](https://github.com/TalonT-Org/AutoSkillit/pull/4277) | Output-budget protocol, response backstop, guards, and regression coverage | Parent remains open because cumulative admission is unresolved |
| [#4280](https://github.com/TalonT-Org/AutoSkillit/issues/4280) | [PR #4281](https://github.com/TalonT-Org/AutoSkillit/pull/4281) | Numeric read discipline, fresh-context subagents, lower-ceiling guidance | Closed and staged after implementation verification on 2026-07-22 |
| [#4286](https://github.com/TalonT-Org/AutoSkillit/issues/4286) | [PR #4288](https://github.com/TalonT-Org/AutoSkillit/pull/4288) | Lossless measured shell-output capture; deleted command-shape guard | Closed but missing `staged` |
| [#4291](https://github.com/TalonT-Org/AutoSkillit/issues/4291) | [PR #4298](https://github.com/TalonT-Org/AutoSkillit/pull/4298) | Python-owned cleanup and generated-content policy guard | Closed and staged |
| [#4300](https://github.com/TalonT-Org/AutoSkillit/issues/4300) | [PR #4302](https://github.com/TalonT-Org/AutoSkillit/pull/4302) | Backend-aware delivery-bound shaping and bounded summary | Closed and staged |
| [#4304](https://github.com/TalonT-Org/AutoSkillit/issues/4304) | [PR #4308](https://github.com/TalonT-Org/AutoSkillit/pull/4308) | Bounded recipe envelope and pull retrieval | Closed and staged |

[...]

Issue #4253, opened 2026-07-14
open_kitchen response exceeds Claude Code MCP output cap — step graph truncated from orchestrator context, causing out-of-order step execution (review_approach before rectify)


Trecek:
## Hard Requirements

1. **A working control over the harness output-persistence gate is required, verified — not assumed.** Live experiment (2026-07-14) proved `MAX_MCP_OUTPUT_TOKENS=50000` is delivered to wrapper-launched sessions yet does **not** govern Claude Code 2.1.197's disk-persistence gate — persistence still fired at 97.9KB with the override present. Split this requirement into: (1a) env-presence verification at launch (`required_env=` wiring — see Aggravating Defect); (1b) a behavioral canary (send an oversized synthetic MCP payload, assert inline delivery) run periodically/in CI; (1c) identify a real harness knob for the persistence gate or file an upstream Claude Code request — until one exists, the gate is an immovable ceiling.
2. **There can be no effective cap on `open_kitchen` delivery.** When the recipe payload is truncated, the orchestrator loses the step graph and everything downstream breaks silently.
3. **The ~100KB rendered `open_kitchen` response is itself suspicious and needs scrutiny.** Strategy resolution: the **primary** fix is keeping the payload under the gate with headroom (budget + slimming, enforced by regression test); **defense-in-depth** is restructuring the agent-visible output head so the step-flow chain survives any 2KB preview; the upstream knob (1c) is the long-term fix.

## Root Cause

Commit `2dc0ad819` (#4244, Hash-Bound Closure Authority, 2026-07-13 21:39:57 UTC) grew `remediation.yaml` by exactly +555 bytes (105,528 → 106,083), pushing the rendered `open_kitchen` payload (99,693 → 100,458 bytes) over Claude Code's output-persistence gate. The commit shipped in **v0.10.866** (between the v0.10.865 bump at 21:10:25 and the v0.10.866 bump at 21:46:56; the #4242/#4243 commit at 21:03:32 made zero byte change to the recipe). The gate sits empirically between 99,693 and ~100,250 bytes (~100KB; CLI 2.1.197 identical on both sides of the boundary — the payload crossed the gate, the gate did not move). The measured/persisted text is the **pretty-formatted** output produced by the `pretty_output_hook` PostToolUse formatter, not raw JSON.

[...]

PR #4277, opened 2026-07-17
Output Budget Protocol Audit Remediation

Trecek:
## Summary

Apply this remediation as a delta on the existing implementation branch at `093179c8f`, whose diff already contains the four-layer Output Budget Protocol. The remediation makes every model-visible reduction artifact-backed, makes `projected_utf8_bytes` describe the exact canonical handler response, enforces an explicit registered-schema/wire-schema/handler-type/post-conversion policy, ratchets both response exemptions to independent measured ceilings, and emits only bounded, path-free telemetry.

The implementation is split into four ordered phases, each independently gated by clean `pre-commit` and `task test-all` runs with evidence recorded in `.autoskillit/evidence/output-budget-remediation/manifest.json`:

- **Phase 1**: Lossless response formatting, exact `projected_utf8_bytes` canonical finalization, four-representation conformance middleware, closed `ResponseBackstopExemptionDef` registry with independent measured ceilings, privacy-safe bounded telemetry
- **Phase 2**: Close JSONL `wc -l` command guard bypass with literal operand validation, preserve both incident commands from issue #4272 byte-for-byte, exercise large JSONL fixture through the guard
- **Phase 3**: Installed Codex TOML config parse gate with Taskfile target and workflow ordering assertions
- **Phase 4**: Live generated-subagent delivery probe with spawn/wait/rollout linkage, deep-investigate E2E proving completed agent waves + inter-batch synthesis + D6 validators + Claude 200K provenance, fresh GraphQL closure postcondition for issues #4253/#3938 and PR #4259

## Implementation Plan

Plan file: `.autoskillit/temp/make-plan/output_budget_protocol_audit_remediation_plan_2026-07-15_194745.md`


Trecek (commented):

src/autoskillit/server/_notify.py
```diff
@@ -94,39 +102,84 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:
                         ),
                     }
                 )
-                logger.exception("Unhandled exception in tool %s", tool_name)
+                logger.error("track_response_size_handler_failed", tool_name=tool_name)
```
Trecek: [warning] bugs: The unhandled-exception branch of track_response_size's wrapper changed from `logger.exception("Unhandled exception in tool %s", tool_name)` to `logger.error("track_response_size_handler_failed", tool_name=tool_name)`. `logger.exception` implicitly attaches the exception traceback/type; the replacement `logger.error` call does not pass `exc_info=True`, so the original exception's type, message, and stack trace are now silently dropped. This is the last-resort catch-all for tool implementations, so losing the traceback here significantly hampers post-hoc debugging of crashes in MCP tools.

src/autoskillit/server/_response_conformance.py
```diff
@@ -0,0 +1,154 @@
...
+            registered_schema_is_wrapped_string = False
+            converted_result_conforms = False
+        decision = decide_response_conformance(
+            is_function_tool=is_function_tool,
+            handler_returns_exact_str=handler_returns_exact_str,
+            registered_schema_is_wrapped_string=registered_schema_is_wrapped_string,
+            converted_result_conforms=converted_result_conforms,
+        )
+        if decision is ResponseConformanceDecision.REWRITE_NONCONFORMING:
+            assert isinstance(registered_tool, FunctionTool)
```
Trecek: [warning] defense: `assert isinstance(registered_tool, FunctionTool)` is used as a runtime invariant guard inside `ResponseConformanceMiddleware.on_call_tool`. Assertions are stripped when Python runs with `-O`/`PYTHONOPTIMIZE`, so this check silently disappears outside the assert-enabled path the test suite normally exercises. If `decision is REWRITE_NONCONFORMING` is ever reached with `registered_tool` not a `FunctionTool`, an optimized run would raise an unguarded AttributeError from `registered_tool.convert_result(...)` a few lines later — a divergent, untested failure mode. Replace with an explicit `if not isinstance(...): raise` so behavior is identical in both configurations.

src/autoskillit/server/_response_budget.py
```diff
@@ -0,0 +1,508 @@
...
+
+
+def _artifact_path(artifact_dir: Path, tool_name: str) -> Path:
+    safe_name = "".join(c if c.isalnum() or c in "._-" else "_" for c in tool_name)
+    return artifact_dir / f"{safe_name or 'response'}_{uuid.uuid4().hex[:8]}.log"
+
+
+def _finalize_envelope(envelope: dict[str, Any], *, max_bytes: int) -> str:
+    metadata = envelope.get(RESPONSE_SPILL_METADATA_KEY)
+    if not isinstance(metadata, dict) or "projected_utf8_bytes" not in metadata:
```
Trecek, line 218: [warning] defense: `_finalize_envelope` raises `_ProjectionNonconvergentError` (a bare `RuntimeError` subclass with no chaining context) when the fixed-point loop for computing `projected_utf8_bytes` doesn't converge. This is caught purely by type in `enforce_response_budget` and converted to a generic bounded failure — acceptable given this is an internal invariant, but the raised exception carries no diagnostic payload (e.g., final measured/expected byte counts) that would help debug a future convergence bug in this novel iterative algorithm.

src/autoskillit/server/_response_budget.py
```diff
@@ -0,0 +1,508 @@
...
+                "response_budget_exemption",
+                tool_name=_bounded_tool_name(tool_name),
+                measurement_id=exemption.measurement_id,
+                original_chars=len(original),
+                original_utf8_bytes=original_size,
+                max_chars=exemption.max_chars,
+                max_utf8_bytes=exemption.max_utf8_bytes,
+            )
+            return result
+        return bounded_response_budget_failure(
```
Trecek, line 387: [warning] defense: Exemption ceiling breach drops the entire response with no artifact persisted: when an exempt tool (`open_kitchen`/`load_recipe`) exceeds its measured ceiling, this branch returns `bounded_response_budget_failure` WITHOUT writing a spill artifact — the only enforcement path that is not artifact-backed, contradicting the PR's stated 'every model-visible reduction artifact-backed' goal. The reused cause code `internal_invariant_failed` also obscures the actual condition (exemption ceiling exceeded). For `load_recipe`, ceilings were measured against bundled recipes (measurement_id `bundled-recipes-all-modes-2026-07-16`); a larger user-local recipe reaches this path at runtime and loses the recipe delivery entirely, in tension with ADR-0004's re-delivery obligation. Consider persisting the artifact before failing closed and/or introducing a distinct cause code.

src/autoskillit/core/types/_type_constants.py
```diff
@@ -44,6 +50,105 @@
     "CODEX_SESSIONS_SUBDIR",
 ]
 
+OUTPUT_DISCIPLINE_POLICY_VERSION = 1
+
+OUTPUT_DISCIPLINE_BLOCK = "\n".join(
```
Trecek, line 56: [warning] cohesion: OUTPUT_DISCIPLINE_BLOCK and OUTPUT_DISCIPLINE_DIGEST encode the same nine output-discipline policy bullets as two independently-worded strings. OUTPUT_DISCIPLINE_BLOCK is only used to derive OUTPUT_DISCIPLINE_BLOCK_SHA256 (feeding PROBE_POLICY_IDENTITY in _probe_cache.py) and is the text pasted verbatim into the four SKILL.md files; OUTPUT_DISCIPLINE_DIGEST is the text actually injected into agent prompts at runtime (_claude_prompt.py::_inject_output_discipline, codex.py's _generate_agent_tomls and build_interactive_cmd). Since the two blocks are authored independently, editing the runtime-injected DIGEST text does not change BLOCK_SHA256, so the probe-cache invalidation key can silently go stale relative to the policy text actually delivered to agents, and the same policy content must be hand-maintained in two differently-worded copies.

tests/server/test_wire_compat.py
```diff
@@ -193,3 +210,232 @@ async def call_next(context):
...
+                    decide_response_conformance(
+                        is_function_tool=True,
+                        handler_returns_exact_str=handler_returns_str,
+                        registered_schema_is_wrapped_string=schema_is_wrapped,
+                        converted_result_conforms=converted_conforms,
+                    )
+                    is ResponseConformanceDecision.CONFORMING
+                ), visible_tool.name
+        finally:
+            mcp._transforms.clear()
```
Trecek, line 298: [warning] tests: test_all_registered_string_tools_are_conforming_decision_entries mutates the shared module-level FastMCP singleton (autoskillit.server.mcp): it calls mcp.enable(tags=set(ALL_VISIBILITY_TAGS)) to make every tool visible, but the finally block only does mcp._transforms.clear() — it never restores the pre-test enabled/visible tag set. Since `mcp` is a process-wide singleton, this permanently widens tool visibility for the remainder of the pytest-xdist worker process, silently changing behavior for any other test that runs afterward in the same worker and depends on default tool visibility.

Trecek (commented):

src/autoskillit/server/_notify.py
Trecek: Agreed — fixed in f9797830b5194b0cc722401a4f4b585abfeff81c. Switched to `logger.exception("track_response_size_handler_failed", tool_name=tool_name)` — keeps the structured event name while restoring the traceback that `logger.error` dropped, consistent with the `logger.exception` convention used across `_lifespan.py` catch blocks.

src/autoskillit/server/_response_conformance.py
Trecek: Agreed — fixed in 2f460b6d33311199a6160490a2f6dd4b7697859c. Replaced the assert with an explicit `if not isinstance(registered_tool, FunctionTool): raise TypeError(...)`. Notably, `convert_result` is defined on the base `fastmcp.tools.base.Tool` class, so under `-O` an invariant violation would not even raise AttributeError — it would silently proceed with base-class conversion. The explicit guard keeps fail-closed behavior identical in both configurations.

src/autoskillit/server/_response_budget.py
Trecek: Agreed — fixed in 4f4d019d4fa8c2e05688c1ed3f28f79094cc3910. `_ProjectionNonconvergentError` now embeds the measured/projected byte counts, `max_bytes`, and attempted-state count at the raise site, and the catch site in `enforce_response_budget` surfaces `str(exc)` via `_emit_response_budget_event("response_budget_projection_nonconvergent", ...)` so the diagnostics are logged rather than swallowed. The caller-visible cause code (`projection_nonconvergent`) is unchanged.

src/autoskillit/server/_response_budget.py
Trecek: Valid observation — flagged for design decision, with the cause-code half applied in 0645ba6905be6cba269caad6b1dc09d6c85d74cf: exemption ceiling breaches now report a distinct `exemption_ceiling_exceeded` cause instead of reusing `internal_invariant_failed`. The artifact-persistence half is deliberately not applied: ADR-0005 "Accepted Gaps" item 3 documents fail-closed-without-echo as intended, and `test_exemption_overage_fails_closed_and_does_not_spill` (tests/server/test_response_backstop.py:302) explicitly encodes the no-spill behavior. Persisting a spill artifact on this path — and resolving the ADR-0004 re-delivery tension for user-local recipes — requires a human decision to amend the ADR.

src/autoskillit/core/types/_type_constants.py
Trecek: Agreed — fixed in 4ff4b667004e0f3dcf0aaf26789ea3a85b40eb6f (plus 4b02c9057 for the symbol-count pin). Added `OUTPUT_DISCIPLINE_COMBINED_SHA256` hashing BLOCK + NUL + DIGEST, and `PROBE_POLICY_IDENTITY` now composes from the combined digest, so editing either policy text invalidates the probe cache. `OUTPUT_DISCIPLINE_BLOCK_SHA256` remains for the SKILL.md byte-identity contract (`test_required_skill_contains_byte_identical_policy_block`). The two differently-worded texts are kept as-is — prose block for SKILL.md, compact digest for runtime injection.

tests/server/test_wire_compat.py
Trecek: Investigated — this is intentional. tests/conftest.py:324-340 defines the autouse fixture `_reset_mcp_visibility`, which runs after every test: it calls `mcp._transforms.clear()` and re-applies the baseline disabled state via `mcp.disable(tags={tag})` for all `ALL_VISIBILITY_TAGS`. tests/AGENTS.md (lines 14-21) documents this as the canonical restore pattern ("Never rely on inverse method calls for cleanup"). The arch guard `test_transforms_hygiene.py::test_inline_transforms_clear_has_finally_guard` only requires the inline try/finally (satisfied at lines 273-298) precisely because tag-set restoration is centralized in that fixture — visibility is not permanently widened for the xdist worker.

src/autoskillit/execution/backends/_codex_config.py
```diff
@@ -23,9 +24,14 @@
 
 CODEX_MCP_STARTUP_TIMEOUT_SEC: float = 30.0
 
-# Per-tool response size budget for Codex MCP tools. Sufficient for current
-# `open_kitchen` response sizes with 2x headroom.
-CODEX_TOOL_OUTPUT_TOKEN_LIMIT: int = 50_000
+# Damage bound for all tool/function output stored in Codex context, including
+# native shell and MCP output. Codex currently applies a coarse four-UTF-8-byte
+# truncation heuristic; the extra 8,000 tokens provide serialized-payload
+# headroom. Guard and spill layers remain the load-bearing mechanisms.
```
Trecek, line 30: [warning] defense: _MAX_RESPONSE_BACKSTOP_EXEMPTION_BYTES computes max() over RESPONSE_BACKSTOP_EXEMPTION_REGISTRY.values() at module import time with no emptiness guard. If the registry is ever emptied, this raises an unhandled ValueError('max() arg is an empty sequence') on import of _codex_config.py, taking down Codex backend registration with an opaque stack trace.

src/autoskillit/server/_response_budget.py
```diff
@@ -0,0 +1,519 @@
...
+
+    published = str(path.resolve())
+    metadata = {
+        "artifact_path": published,
+        "sha256": hashlib.sha256(original_bytes).hexdigest(),
+        "original_utf8_bytes": len(original_bytes),
+    }
+
+    parsed: Any = None
+    if isinstance(result, str):
```
Trecek, line 425: [warning] defense: enforce_response_budget: when isinstance(result, str), json.loads(result) failure is swallowed via except (TypeError, ValueError): parsed = None, silently discarding diagnostic context about why a string tool result wasn't valid JSON. Since this determines dict-projection vs plain-text-spill path, a malformed-but-almost-JSON payload silently degrades to plain-text preview with no error surfaced.

src/autoskillit/server/git.py
```diff
@@ -34,11 +38,46 @@
...
+    raw = json.dumps({"stdout": test_result.stdout, "stderr": test_result.stderr})
+    spilled = spill_output(
+        raw,
+        resolve_temp_dir(Path(worktree_path), config.workspace.temp_dir) / "merge_worktree",
+        "raw_test_output",
+        spec,
+    )
+    condensed_stdout, condensed_stderr = condense_test_output(test_result)
+
+    def preview(text: str) -> str:
```
Trecek, line 76: [critical] bugs: _shape_failed_test_output: spill decision uses raw combined JSON vs inline_max_chars, but truncation decision uses individual condensed fields vs inline_max_chars. When raw JSON is short but an individual condensed field exceeds inline_max_chars, spilled.artifact_path is None yet preview() truncates and emits marker f'[raw test output spilled -> {spilled.artifact_path}]' rendering literal string 'None' — pointing at a nonexistent artifact while discarding untruncated content that was never persisted to disk.

src/autoskillit/server/tools/tools_workspace.py
```diff
@@ -103,12 +112,33 @@ async def test_check(
                         extra={"worktree": worktree_path},
                     )
 
+                raw_artifact_path: str | None = None
```
Trecek, line 115: [critical] bugs: Same spill/truncation mismatch as git.py: raw_spill computed from combined JSON, but _bounded_test_stream independently checks each condensed field length. When raw JSON is short but a condensed field is long, _bounded_test_stream truncates it and embeds marker with artifact_path=None (literal 'None' in output), losing untruncated content.

src/autoskillit/hooks/guards/output_budget_guard.py
```diff
@@ -0,0 +1,343 @@
...
+    except (OSError, AttributeError, TypeError, json.JSONDecodeError):
+        section = {}
+    return (
+        section.get("disabled") is True,
+        _positive_int(section.get("small_file_max_bytes"), _DEFAULT_SMALL_FILE_MAX_BYTES),
+        _positive_int(section.get("shell_max_inline_bytes"), _DEFAULT_SHELL_MAX_INLINE_BYTES),
+    )
+
+
+def _has_flag(args: list[str], *names: str) -> bool:
```
Trecek, line 99: [critical] bugs: _has_flag() only special-cases combined short-flag bundling for -q/--quiet. It does not apply bundling logic to -r/-R, so _is_recursive_search() fails to detect recursive flag in combined forms like 'grep -rn TODO .' or 'grep -rl pattern .'. Verified: _classify_command('grep -rn TODO .') returns BOUNDED (passes) while 'grep -r -n TODO .' correctly returns HAZARDOUS (denied). This is an easily-triggered bypass of the guard's R1 purpose (blocking unbounded recursive search) via common idiomatic grep invocations.

Trecek (commented):

docs/decisions/0005-output-budget-protocol.md
```diff
@@ -0,0 +1,156 @@
...
+Codex CLI 0.144.1 uses `APPROX_BYTES_PER_TOKEN = 4` and integer byte arithmetic in
+[`codex-rs/utils/string/src/truncate.rs` at tag `rust-v0.144.1`, commit
+`44918ea10c0f99151c6710411b4322c2f5c96bea`](https://github.com/openai/codex/blob/44918ea10c0f99151c6710411b4322c2f5c96bea/codex-rs/utils/string/src/truncate.rs).
+This is a coarse one-token-per-four-UTF-8-bytes truncation heuristic. It performs no real
+tokenization and provides neither a tokenizer guarantee nor a cumulative-context
+estimate.
+
+The project therefore requires the stricter relationship
+`response_max_bytes // 3 < CODEX_TOOL_OUTPUT_TOKEN_LIMIT`. The three-byte divisor is
+deliberate margin: the 90,000-byte response backstop must fire before Codex's 53,000-token
```
Trecek: [warning] bugs: ADR states '53,000-token transport ceiling' but the actual CODEX_TOOL_OUTPUT_TOKEN_LIMIT constant is 54,500 computed as ((186_000 + 3) // 4) + 8_000. The 53,000 figure is internally inconsistent with the ADR's own formula and does not match any constant in the codebase.

docs/decisions/0005-output-budget-protocol.md
```diff
@@ -0,0 +1,156 @@
...
+transport ceiling can clip a producer-blind response. A static test pins the relationship,
+and the live large-output probe must pass before either side is retuned.
+
+The measured raw-text exemptions, `open_kitchen` and `load_recipe`, must each remain below
+their own registered character and UTF-8 byte ceilings, which in turn remain below the
+212,000-byte budget implied by the current 53,000-token, four-byte heuristic. Their
```
Trecek: [warning] bugs: '212,000-byte budget implied by the current 53,000-token, four-byte heuristic' repeats the wrong token value. Using actual CODEX_TOOL_OUTPUT_TOKEN_LIMIT of 54,500, the implied byte budget would be 218,000, not 212,000.

scripts/validate_output_budget_evidence.py
```diff
@@ -0,0 +1,408 @@
...
+
+
+def validate_manifest(manifest: dict[str, Any], repo_root: Path, *, mode: str) -> None:
+    """Validate one manifest in incremental or complete mode."""
+    if manifest.get("schema_version") != SCHEMA_VERSION:
+        raise EvidenceValidationError(f"schema_version must be {SCHEMA_VERSION}")
+    if mode not in {"incremental", "complete"}:
+        raise EvidenceValidationError("mode must be incremental or complete")
+
+    _validate_historical_context(manifest, repo_root)
```
Trecek: [warning] bugs: validate_manifest() validates gate_log_path references (e.g. 'temp/test-2026-07-15_170319.txt') that live under the gitignored temp/ directory. Running the validation script against the shipped manifest.json on a fresh clone fails immediately with 'gate_log_path does not exist'. The script ships broken against its own evidence fixture. Not wired into CI, so not currently gating anything.

tests/execution/test_headless_core.py
```diff
@@ -1250,8 +1250,8 @@ def test_stderr_included_in_response(self):
         )
         assert response["stderr"] == "queue contention"
 
-    def test_stderr_truncated(self):
-        """Stderr exceeding 5000 chars is truncated."""
+    def test_stderr_reaches_tool_boundary_losslessly(self):
```
Trecek, line 1253: [warning] tests: test_stderr_reaches_tool_boundary_losslessly replaces the old truncation assertion with a single equality check but doesn't test the new output-budget/spill boundary itself (e.g., stderr sized at threshold, or large enough to require spilling). It only proves old hardcoded truncation was removed.

tests/infra/test_release_sanity.py
```diff
@@ -8,6 +8,7 @@
 pytestmark = [pytest.mark.layer("infra"), pytest.mark.medium]
 
 REPO_ROOT = Path(__file__).parent.parent.parent
+_PERSONAL_HOME_PATH_FIXTURES = frozenset({Path("tests/hooks/test_output_budget_guard.py")})
```
Trecek: [warning] tests: _PERSONAL_HOME_PATH_FIXTURES exempts the entire file tests/hooks/test_output_budget_guard.py from the 'no personal home paths in tests' sanity check, rather than exempting only the specific line(s). This silently suppresses detection of any other accidental personal-path leak later added anywhere else in that file.

tests/skills_extended/test_audit_friction_output_budget.py
```diff
@@ -0,0 +1,95 @@
...
+        input=json.dumps(payload),
+        capture_output=True,
+        text=True,
+        cwd=tmp_path,
+        env=env,
+        timeout=5,
+    )
+
+    assert result.returncode == 0, result.stderr
+    if result.stdout.strip():
```
Trecek, line 91: [warning] tests: test_real_command_battery_passes_output_budget_guard only asserts on permissionDecision when result.stdout is non-empty. The guard exits 0 with empty stdout both when it correctly classifies a command AND when it silently bails out on malformed input. A regression making the guard bail out early on all input would still pass this test.

src/autoskillit/server/_response_budget.py
```diff
@@ -0,0 +1,519 @@
...
+    )
+    if isinstance(result, str):
+        return rendered
+    try:
+        return json.loads(rendered)
+    except ValueError:
+        return {"success": False, "error": "response_budget_projection_invalid"}
+
+
+def shape_json_response(
```
Trecek, line 498: [warning] tests: shape_json_response (used by shape_execution_response in _execution_helpers.py, wired into run_python/run_cmd tool responses) has zero direct/unit tests anywhere in the test suite. Only indirect e2e coverage exists.

src/autoskillit/server/_response_budget.py
```diff
@@ -0,0 +1,519 @@
...
+        metadata,
+        reason="minimal_projection",
+        omitted_chars=omitted_chars,
+        omitted_items=omitted_items,
+    )
+    rendered = _finalize_envelope(minimal, max_bytes=max_bytes)
+    return rendered if len(rendered.encode("utf-8")) <= max_bytes else None
+
+
+def _plain_spill_envelope(
```
Trecek, line 321: [warning] tests: _plain_spill_envelope's preview_limit==0 fail-to-None branch (driving 'irreducible_shape' failure for plain-text oversized responses) has no test. The equivalent JSON-object path IS tested but the plain-text variant is not.

Trecek (commented):
AutoSkillit review: 3 critical and 10 warning findings detected. See inline comments for details.

### Outside Diff Range

**tests/hooks/test_codex_hooks_format_contract.py**
- **L19** [info/tests]: test_output_budget_guard_is_wired_with_identical_matchers uses next() without a default over HOOK_REGISTRY; if the guard

src/autoskillit/server/git.py
Trecek: Agreed — fixed in b83d9d91f. Force-spill when any condensed field exceeds inline_max_chars but combined JSON did not trigger spill, eliminating the literal "None" in truncation markers.

src/autoskillit/server/tools/tools_workspace.py
Trecek: Agreed — fixed in b83d9d91f. Same force-spill pattern applied in tools_workspace.py, restructured to keep condensation and spill in the same scope.

src/autoskillit/hooks/guards/output_budget_guard.py
Trecek: Agreed — fixed in e63433ee9. Generalized _has_flag() to extract all single-char short flags from names and check bundled forms, so -rn, -rl, etc. are now detected alongside -r.

docs/decisions/0005-output-budget-protocol.md
Trecek: Agreed — fixed in daf1bf5d1. Commit 4a8d65a2b updated the numeric table and code to 54,500/186,000 but missed these two prose sentences in Ceiling and Backstop Reconciliation.

docs/decisions/0005-output-budget-protocol.md
Trecek: Agreed — fixed in daf1bf5d1. Corrected to 218,000-byte budget (54,500 × 4) and 54,500-token ceiling.

tests/infra/test_release_sanity.py
Trecek: Agreed — fixed in b64d751c0. Switched from file-level frozenset to (path, lineno) dict targeting only lines 17-18 (INCIDENT_LOG_SEARCH fixture).

tests/skills_extended/test_audit_friction_output_budget.py
Trecek: Agreed — fixed in b8bbc3a00. BOUNDED rg commands correctly produce no stdout, so instead of requiring non-empty output from all battery commands, added test_guard_denies_known_hazardous_command proving the guard classifies (grep -r → deny) rather than bailing.

src/autoskillit/server/_response_budget.py
Trecek: Agreed — fixed in 0cf458ebb. Added test_shape_json_response_under_threshold_is_passthrough (verifies small payloads bypass spill) and test_shape_json_response_over_threshold_spills_with_metadata (verifies spill + artifact creation).

src/autoskillit/server/_response_budget.py
Trecek: Agreed — fixed in 0cf458ebb + 4624fa9b2. Added test_plain_text_irreducible_shape_returns_failure that forces _plain_spill_envelope to exhaust its preview_limit halving loop and return None, triggering the irreducible_shape failure path.

src/autoskillit/execution/backends/_codex_config.py
Trecek: Investigated — this is intentional. RESPONSE_BACKSTOP_EXEMPTION_REGISTRY is a hardcoded 2-entry dict (core/types/_type_constants_registries.py:190-201), pinned by three independent tests: test_response_backstop_exemption_registry_is_closed_and_pinned (exact dict equality), test_response_backstop_exemption_registry_digest_is_canonical (SHA256 pin), and test_response_backstop_exemption_registry_public_gateways (non-empty gate). Emptying it would fail core-layer tests before reaching this import-time max().

src/autoskillit/server/_response_budget.py
Trecek: Investigated — this is intentional. When isinstance(result, str) and json.loads fails, parsed=None causes isinstance(parsed, dict) to be False, routing to _plain_spill_envelope(original, ...) which is the semantically correct path for non-JSON string content. No information is lost: the original string is persisted to the artifact and a plain-text preview is returned. This is type-discrimination, not error-suppression.

tests/execution/test_headless_core.py
Trecek: Investigated — this is intentional. Commit 014333e5b replaced the old 5000-char truncation with unconditional lossless passthrough to the MCP response boundary. There is no threshold/spill mechanism in this code path — grep for 'spill' in src/autoskillit/execution/ finds no such boundary. The renamed test's equality assertion (60,000-char stderr == response stderr) fully covers the implemented behavior.

scripts/validate_output_budget_evidence.py
Trecek: Valid observation — flagged for design decision. The 18 evidence references point to untracked temp/ files; the script only passes today because stale local gate-run logs remain on disk. Options: (a) exempt temp-rooted paths from existence/hash validation, (b) copy evidence files into the tracked evidence dir, or (c) accept this as a historical audit script and document the limitation. Each implies a different answer to what the validation script is for.

Trecek: Fixed in cb322c09a + af1a55a6e. The 18 untracked references were intentionally untracked — historical_context entries carry bound_to_commit=false and closure.historical_artifacts are marked non-reconstructible. The validator now checks structural metadata (field presence, SHA-256 format) without requiring the referenced files to exist on disk.

Issue #4282, opened 2026-07-18
resolve-review is not Codex-interchangeable: git_metadata_write pin hard-rejected, and historical Codex runs silently skipped fix application


Trecek:
## Problem

`resolve_review` is expected to be backend-interchangeable — any `run_skill` step routable to Codex the same way `review_pr` is (both are pinned identically in operator config). It is not:

1. **As of #4275, the pin is hard-rejected at kitchen open.** `open_kitchen(name="remediation")` fails recipe validation when `agent_backend.recipe_overrides.remediation.resolve_review: codex` is set.
2. **Before #4275, the pin was accepted but never functioned.** Session telemetry shows every Codex `resolve_review` run completed without applying or committing a single fix — zero fix-application across all 18 runs, reported as success in 17.

There is currently **no supported configuration** that runs review-resolution work on the Codex backend. This issue requests a codex-compatible resolution path.

## Operator-set design constraints (2026-07-17)

These constraints are set by the repo operator and bound the solution space — do not gold-plate beyond them:

1. **Threat model is worker siloing, not adversarial defense.** This repo is not defending against hackers, malware, or hostile prompt-injection. The value of `.git/` write protections here is keeping each dispatched worker siloed to its own worktree — not blocking a malicious actor.
2. **Parity standard: Codex is held to the same trust bar as Claude Code — no higher.** Claude Code worker sessions already run `pre-commit` + `git commit` in-session with full filesystem access. Any Codex-enablement design should match that trust level, not exceed it with adversarial-grade hardening (hook suppression, plumbing-only commits, confused-deputy defenses). Such hardening was evaluated during review and explicitly descoped; revisit only if the repo's threat model changes.
3. Consequence: for Option A, the tool-design requirements below are **correctness/siloing requirements** (stay inside the calling step's worktree), not security mitigations.

## Evidence ledger

Every claim below was verified first-hand during the investigation (report: `.autoskillit/temp/investigate/investigation_remediation_resolve_review_codex_pin_2026-07-17_224753.md`) and survived a subsequent 3-agent adversarial validation round (2026-07-17) in which every flagged discrepancy was re-checked against the primary source; corrections from that round are incorporated below.

**E1 — Rejection, reproduced live (2026-07-17):**
[...]

PR #4287, opened 2026-07-18
Rectify: resolve-review Codex interchangeability + silent-success immunity (#4282)

Trecek:
## Summary

Issue #4282 has two failure planes: the operator pin `resolve_review: codex` is hard-rejected at kitchen open (since #4275, because resolve-review declares `git_metadata_write`), and before that fix, 18 Codex runs reported success with zero fixes applied because the success adjudicator is infra-signal-only.

Three phases deliver architectural immunity:

- **Phase A — Outcome-contract adjudication.** Contract-declared output fields (`accept_count`, `fix_failures`) are parsed server-side, typed, recorded in telemetry, and evaluated against per-skill `outcome_invariants` in a consolidated post-session gate. Failed fix application (`fix_failures > 0`) demotes success. Legitimate all-skipped runs carry a success qualifier. Consolidates 4 duplicated zero-write demotion sites into one shared helper.
- **Phase B — Backend-neutral git mutation.** Server-side `commit_files(paths, message, cwd)` MCP tool executes staging, pre-commit hooks, and commit in the server process. resolve-review Step 4 calls it instead of in-session `git add/commit`. `git_metadata_write` dropped from `uses_capabilities`; Codex backend pin validates and functions.
- **Phase C — Generalization + drift closure.** Sibling skills (resolve-research-review, resolve-claims-review) migrated to `commit_files`. Frontmatter capability parsing fail-closed via `SkillInfo.invalid_reason` — unknown capability names mark the skill invalid with ERROR findings in recipe validation.

Closes #4282

## Implementation Plan

Plan file: `.autoskillit/temp/rectify/rectify_resolve_review_codex_immunity_2026-07-18_124500_combined.md`


src/autoskillit/server/tools/tools_git.py
```diff
@@ -627,3 +629,112 @@ async def create_and_publish_branch(
     except Exception as exc:
         logger.error("create_and_publish_branch unhandled exception", exc_info=True)
         return json.dumps({"error": f"{type(exc).__name__}: {exc}"})
+
+
+@mcp.tool(
+    tags={"autoskillit", "kitchen", "kitchen-core", "headless"},
+    annotations={"readOnlyHint": True},
```
Trecek, line 636: [critical] bugs: `readOnlyHint: True` on a tool that performs `git add` + `git commit` (mutating operations). This mislabels the tool's side effects to any MCP client or orchestration layer that relies on the hint to gate destructive operations. Should be `readOnlyHint: False` (or remove the annotation entirely, since False is the default).

Trecek (commented):

Trecek (commented):
AutoSkillit review: blocking issue found — `readOnlyHint: True` on a mutating tool. See inline comment on `tools_git.py:636`.

src/autoskillit/execution/headless/_headless_outcome.py
```diff
@@ -0,0 +1,132 @@
...
+if TYPE_CHECKING:
+    from autoskillit.recipe._contracts_types import (
+        OutcomeInvariantEntry,
+        SkillContract,
+        SuccessQualifierEntry,
+    )
+
+logger = get_logger(__name__)
+
+_FIELD_RE = re.compile(r"^(\w+)\s*=\s*(.+)$", re.MULTILINE)
```
Trecek, line 26: [warning] bugs: This regex (`^(\w+)\s*=\s*(.+)$`, MULTILINE) matches any line in the session result text that looks like `field_name = value`. Since the input is uncontrolled LLM free-text, an agent narrating something like `accept_count = 2 because there were two findings` would be captured as a field value (parsed to the raw string, failing int(), then causing the invariant evaluator to return None). The last match wins via finditer, so if structured output comes before prose, the prose overwrites the correct value. Consider anchoring extraction to only the final N lines of output, or requiring a structured block delimiter.

src/autoskillit/execution/headless/_headless_result.py
```diff
@@ -896,6 +886,24 @@ def _build_skill_result(
         )
         sr = _apply_budget_guard(sr, skill_command, audit, max_consecutive_retries)
 
+    if skill_contract is not None and skill_contract.outputs:
+        _parsed_fields = parse_outcome_fields(sr.result, skill_contract)
```
Trecek, line 890: [info] bugs: `parse_outcome_fields(sr.result, skill_contract)` is called here but was already computed inside `_apply_post_session_adjudication` a few lines above. The result is identical (sr.result is unchanged by dataclasses.replace). Consider threading the parsed fields through to avoid redundant re-parsing.

src/autoskillit/server/tools/tools_git.py
Trecek: Investigated — this is intentional. The test `test_all_tools_have_readonly_hint_true` enforces that ALL MCP tools must have `readOnlyHint: True` because all pipelines operate on independent branches/worktrees. The `readOnlyHint` in this project is a blanket "safe to call" signal for orchestration layers, not a literal mutability indicator. Changing it breaks the architectural constraint.

src/autoskillit/execution/headless/_headless_outcome.py
Trecek: Valid observation — flagged for design decision. The `if name not in declared: continue` guard limits matches to explicitly declared contract output fields, which provides significant protection. However, the last-match-wins behavior of `finditer` means a declared field name appearing in LLM prose before the structured output block could overwrite the correct value. The fix (anchoring to final N lines vs. requiring a structured delimiter) involves a contract-level design decision about how skills emit output tokens.

src/autoskillit/execution/headless/_headless_result.py
Trecek: Acknowledged — minor suggestion noted.

src/autoskillit/execution/headless/_headless_outcome.py
Trecek: Fixed in 7e71275cc — `parse_outcome_fields` now walks backward from the end, skips trailing gate tags, and only parses the contiguous block of `KEY = value` lines. Earlier prose matching a declared field name can no longer overwrite the authoritative final output. Added regression tests with adversarial prose fixtures.

Issue #4293, opened 2026-07-19
Stale pipeline tracker from unclosed kitchen silently disables step-completion marking — false DEPENDENCY UNMET aborts pipeline; _fmt_run_skill drops error field


Trecek:
## Incident

The remediation pipeline for #4291 (kitchen `c40ff8ac-2a21-4aed-bd23-f8c97c7c3563`, autoskillit v0.10.882, run dir `remediation-20260718-192507-037270`, 2026-07-18 ~19:25–19:45 local) aborted at the `review_approach` step with an apparent opaque "tool-level failure" — routed `on_failure: release_issue_failure` → `escalate_stop`, releasing the issue with the fail label. The `rectify` step had **succeeded** seven seconds earlier (session `b38c02c9`, 1214 s, `success: true, exit_code: 0`, 2-part plan produced).

No skill session was ever spawned for `review_approach`: no session record, no `no_session_*` directory, no output artifacts. The MCP server denied the `run_skill` call pre-flight. The raw envelope (preserved only in the transcript's `mcpMeta.structuredContent`, invisible to the orchestrator model):

```json
{"success": false, "is_error": true, "error": "DEPENDENCY UNMET: Step 'review_approach' requires ['rectify'] to complete first. Pipeline 'c40ff8ac-2a21-4aed-bd23-f8c97c7c3563': {'rectify': 'pending'}."}
```

### Defect A: stale pipeline tracker silently disables step-completion marking

- `open_kitchen` auto-creates `{project_dir}/.autoskillit/temp/pipeline_tracker/{kitchen_id}.json` (`_auto_init_pipeline_tracker`, `src/autoskillit/server/tools/tools_kitchen.py:330-383`; introduced in `1e2670ec7` #4259, 2026-07-14).
- Tracker cleanup exists **only** in `_close_kitchen_handler` (`tools_kitchen.py:533-540`). The previous kitchen that day (`fb9097f4`, 17:13–19:02, issues #4289/#4292) was never closed — its conversation ended via CLI `/exit` at 19:23:07 local with zero `close_kitchen` calls in the transcript (`18f10c72-8e73-40d8-9fd0-af8607319aaa.jsonl`), seconds before the new kitchen auto-initialized its own tracker — so its tracker file survived.
[...]

Issue #4291, opened 2026-07-19
shell_capture_hook injects prohibited rm cleanup and blocks every exec_command


Trecek:
## Summary

The AutoSkillit `shell_capture_hook v1` rewrites every normal Codex `exec_command` with a shell-output capture wrapper that contains:

```bash
if [ "$__as_sz" -le 12000 ]; then
  cat "$__as_f"
  rm -f -- "$__as_f"
fi
```

The command-safety layer correctly rejects the generated `rm -f`, so even harmless read-only commands such as `pwd` cannot execute. This creates a conflict between the capture hook and the destructive-command policy.

Cleanup design also needs stronger safety guarantees. A generic shell `rm` should not be injected into every user command. Cleanup must never be capable of resolving to a directory, repository root, home directory, empty path, symlink-swapped target, or other broad target. If a specific capture file must be removed, that should use a constrained file-only mechanism with canonical path and file-type validation rather than a shell command.

## Observed environment

- Date: 2026-07-18
- Session: Codex running under the AutoSkillit wrapper
- Working directory: `/home/talon/projects/esp-home-experiments`
- Shell capture directory generated by the hook: `/home/talon/projects/esp-home-experiments/.autoskillit/temp/shell_capture`
- Filesystem mode: unrestricted
- Approval policy: never
- No AutoSkillit kitchen was opened

## Reproduction

Run any normal shell command through Codex `exec_command`, including:

```bash
pwd
```

The PreToolUse hook rewrites it into a capture script containing `rm -f -- "$__as_f"`. The rewritten command is rejected before the requested command executes.

The same failure reproduced with:

```bash
find /home/talon/projects -mindepth 1 -maxdepth 2 -type d
mkdir -p /home/talon/projects/esp-home-experiments/.autoskillit/temp/session-search
sed -n '1,250p' /dev/shm/autoskillit-sessions/6f9eaadd44e94e38/skills/prepare-issue/SKILL.md
```

## Exact error

```text
Rejected: rm -f style commands are not permitted. Use a safer approach
```

The rejected generated wrapper identified itself as:

```text
# autoskillit-shell-capture v1
```

## Impact

- Every normal `exec_command` is blocked, including read-only diagnostics.
- The failure is caused by AutoSkillit-generated shell code, not the requested command.
- It pressures agents toward alternate execution paths such as the AutoSkillit MCP `run_cmd`, even when no kitchen is open and that execution path is inappropriate.
[...]

PR #4296, opened 2026-07-19
Pipeline Tracker Split-Brain — Server-Authoritative Step Completion

Trecek:
## Summary

The remediation pipeline for #4291 aborted on a false `DEPENDENCY UNMET` because three architectural weaknesses composed: split-brain state ownership between client-side hook and server-side enforcer, unmanaged tracker lifecycle allowing stale files to accumulate, and a lossy presentation layer that dropped the diagnostic `error` field.

This PR delivers three architectural immunities:

- **Phase A** — Step completion becomes server-authoritative, written at `run_skill`'s adjudication point through the same `resolve_tracker_order_id()` function the enforcement check uses. Enforcer and recorder become structurally unable to disagree. The fragile single-file discovery heuristic is deleted; an architectural guard prevents future client-side tracker writes.
- **Phase B** — Kitchen-scoped state gains a liveness-gated pruning lifecycle at every `open_kitchen`, using `active_kitchens.json` liveness checks with an any-entry-alive rule (fleet-safe). Close-only cleanup stops being load-bearing.
- **Phase C** — The `error` field becomes structurally indestructible through the formatter layer via a single-exit-point pass-through gate. All pre-flight deny sites use `deny_envelope()` with `stage`/`retriable` fields. DEPENDENCY UNMET becomes an actionable denial routed like QUOTA WAIT and INGREDIENT LOCK.

## Requirements

1. Make step-completion server-authoritative
2. Fix tracker lifecycle — prune/archive stale trackers at open_kitchen time
3. Guarantee failure-envelope fidelity through formatting
4. Make the deny actionable — include recovery guidance in DEPENDENCY UNMET
5. Regression tests covering two-tracker scenario, open-no-close-open lifecycle, deny-envelope fidelity, end-to-end handoff

Closes #4293

## Implementation Plan

Plan file: `.autoskillit/temp/rectify/rectify_pipeline_tracker_split_brain_2026-07-18_214000_combined.md`


tests/integration/test_pipeline_step_completion_flow.py
```diff
@@ -0,0 +1,254 @@
...
+        await run_skill("/autoskillit:rectify task", str(tmp_path), step_name="rectify")
+
+        result = json.loads(
+            await run_skill(
+                "/autoskillit:review-approach .autoskillit/temp/rectify/plan.md",
+                str(tmp_path),
+                step_name="review_approach",
+            )
+        )
+        assert "DEPENDENCY UNMET" not in result.get("error", "")
```
Trecek, line 100: [warning] tests: test_dependent_step_allowed_after_server_side_marking only asserts "DEPENDENCY UNMET" not in result.get("error", ""); it never asserts the call actually succeeded (e.g. result.get("success") is True). A regression that reworded the denial message would still pass.

src/autoskillit/server/tools/tools_execution.py
```diff
@@ -135,89 +133,57 @@ def _check_ingredient_locks(step_name: str, order_id: str) -> str | None:
...
-        # Kitchen-scoped fallback: interactive sessions have a kitchen_id but no
-        # order_id. Only resolve to the kitchen-scoped tracker when at most one
-        # pipeline is active under this kitchen.
-        if not ctx.kitchen_id:
-            return None
-        active_oids = _active_order_ids_for_kitchen(ctx)
-        if len(active_oids) > 1:
+    resolved = resolve_tracker_order_id(ctx, order_id)
+    if isinstance(resolved, ResolutionRefusal):
+        if "multiple pipelines" in resolved.reason:
```
Trecek: [warning] bugs: _check_pipeline_deps uses substring match "multiple pipelines" in resolved.reason to discriminate RefusalReason. This is a stringly-typed control-flow discriminant: if the reason wording changes, the branch silently breaks. No test pins the exact substring dependency.

Trecek (commented):

Trecek (commented):

tests/server/test_record_pipeline_step_complete.py
```diff
@@ -0,0 +1,100 @@
...
+    async def test_errors_on_unresolvable_pipeline(
+        self, tool_ctx_kitchen_open, tmp_path, monkeypatch
+    ):
+        monkeypatch.delenv("AUTOSKILLIT_DISPATCH_ID", raising=False)
+        tool_ctx_kitchen_open.project_dir = tmp_path
+        tool_ctx_kitchen_open.kitchen_id = ""
+        result = json.loads(
+            await record_pipeline_step(pipeline_id="", op="complete", step_name="rectify")
+        )
+        assert result["success"] is False
```
Trecek, line 85: [info] tests: test_errors_on_unresolvable_pipeline only asserts result["success"] is False, with no check on error content. Sibling tests assert on error message substring — this one is inconsistent.

src/autoskillit/server/tools/_types.py
```diff
@@ -258,6 +262,29 @@ def input_failure_envelope(
     )
 
 
+def deny_envelope(
```
Trecek, line 266: [info] arch: deny_envelope() is a good consolidation matching the existing envelope-factory pattern (server_failure_envelope, input_failure_envelope). No layering concern.

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -1,21 +1,91 @@
...
+@dataclass(frozen=True, slots=True)
+class ResolutionRefusal:
+    """Tracker resolution failed — carry a reason for the caller to wrap."""
+
+    reason: str
+
+
+def resolve_tracker_order_id(
+    tool_ctx: object, order_id: str
+) -> ResolvedTracker | ResolutionRefusal:
```
Trecek, line 57: [info] arch: resolve_tracker_order_id types its first parameter as object and accesses fields via getattr. ToolContext is a valid downward import from IL-3 to IL-1 — typing as object forfeits static type-checking without architectural justification.

src/autoskillit/server/tools/AGENTS.md
```diff
@@ -11,8 +11,8 @@ MCP `@mcp.tool()` handlers registered on import (20 tool modules).
...
 | `_cancellation_shield.py` | `_cancellation_shield` decorator — catches `asyncio.CancelledError` at MCP tool boundary, returns structured JSON |
 | `_backend_compat.py` | Shared target resolution and fail-closed backend compatibility gate for direct headless executor callers |
-| `_types.py` | TypedDict definitions for server tool JSON responses (RunSkillResult, RunCmdResult, ToolFailureEnvelope, etc.) and failure envelope factory helpers |
-| `tools_kitchen.py` | `open_kitchen`, `close_kitchen` (gate lifecycle), `recipe://` MCP resource |
+| `_types.py` | TypedDict definitions for server tool JSON responses (RunSkillResult, RunCmdResult, ToolFailureEnvelope, etc.), failure envelope factory helpers, and `deny_envelope()` canonical pre-flight deny constructor |
+| `tools_kitchen.py` | `open_kitchen`, `close_kitchen` (gate lifecycle), `recipe://` MCP resource, `prune_stale_kitchen_state` (liveness-gated tracker pruning at open), `kitchen_entry_alive` (registry liveness check) |
```
Trecek: [info] arch: AGENTS.md lists kitchen_entry_alive as belonging to tools_kitchen.py, but it is defined in core/_plugin_cache.py and merely imported. Conflates "used by" with "defined in".

src/autoskillit/server/tools/tools_execution.py
```diff
@@ -135,89 +133,57 @@ def _check_ingredient_locks(step_name: str, order_id: str) -> str | None:
...
-        if tracker.get("kitchen_id") == ctx.kitchen_id:
-            active.add(path.stem)
-    return active
-
-
 def _check_pipeline_deps(step_name: str, order_id: str) -> str | None:
     """Check if step_name's dependencies are satisfied. Returns deny JSON or None."""
-    effective_oid = order_id or os.environ.get(DISPATCH_ID_ENV_VAR, "")
     from autoskillit.server import _get_ctx  # circular-break
+    from autoskillit.server.tools.tools_pipeline_tracker import (  # circular-break
```
Trecek, line 166: [info] arch: _check_pipeline_deps, _has_active_deps, and _mark_step_complete_server_side each do their own local circular-break import from tools_pipeline_tracker. Consistent with existing convention but repeated boilerplate.

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -213,3 +294,135 @@ def _handle_status(tracker_path: Path, effective_pipeline_id: str) -> str:
...
+    """
+    canonical = _STEP_SUFFIX_RE.sub("", step_name)
+    lock_path = tracker_path.parent / ".pipeline_tracker.lock"
+
+    try:
+        lock_path.parent.mkdir(parents=True, exist_ok=True)
+        lock_fd = os.open(str(lock_path), os.O_RDWR | os.O_CREAT, 0o644)
+        import fcntl
+
+        fcntl.flock(lock_fd, fcntl.LOCK_EX)
```
Trecek, line 365: [info] bugs: In mark_step_complete, if os.open succeeds but fcntl.flock raises OSError, the except block returns early without closing lock_fd, leaking the file descriptor. Carried forward from the now-deleted pipeline_step_post_hook.py.

src/autoskillit/server/tools/tools_execution.py
```diff
@@ -1501,6 +1535,8 @@ async def run_skill(
                             retriable=True,
                         )
                     )
+                if _pipeline_marker is not None:
```
Trecek, line 1541: [info] bugs: _pipeline_marker assignment coverage requires reading code outside the diff hunk to confirm no early-return/exception path between assignment and use.

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -213,3 +294,135 @@ def _handle_status(tracker_path: Path, effective_pipeline_id: str) -> str:
...
+    Used by both ``op="complete"`` (operator repair) and the adjudication-point
+    marker in ``run_skill``. Returns a result dict (always includes ``success``).
+    """
+    canonical = _STEP_SUFFIX_RE.sub("", step_name)
+    lock_path = tracker_path.parent / ".pipeline_tracker.lock"
+
+    try:
+        lock_path.parent.mkdir(parents=True, exist_ok=True)
+        lock_fd = os.open(str(lock_path), os.O_RDWR | os.O_CREAT, 0o644)
+        import fcntl
```
Trecek: [info] slop: fcntl is imported locally inside both the try block (L352) and the finally block (L425) of mark_step_complete. Redundant duplicate local import — should be a single import at function entry.

tests/integration/test_pipeline_step_completion_flow.py
Trecek: Agreed — fixed in 4b7fdfaf. Added `assert result.get("success") is True` before the substring check so a differently-worded denial can't slip through.

src/autoskillit/server/tools/tools_execution.py
Trecek: Agreed — fixed in 9724c0f8. Added a typed `multi_pipeline: bool` field to `ResolutionRefusal` so the consumer branches on the field instead of substring-matching the reason text.

tests/server/test_record_pipeline_step_complete.py
Trecek: Agreed — fixed in d6b2aff3. Added `assert "pipeline_id is required" in result["error"]` for consistency with sibling tests.

src/autoskillit/server/tools/_types.py
Trecek: Acknowledged — minor suggestion noted.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — fixed in 9724c0f8. Added a `_TrackerCtx` Protocol with `kitchen_id: str` and `project_dir: Path` fields, and replaced `getattr` calls with direct attribute access.

src/autoskillit/server/tools/AGENTS.md
Trecek: Agreed — fixed in daf895da. Updated to clarify `kitchen_entry_alive` is imported from `core/_plugin_cache`, not defined in `tools_kitchen.py`.

src/autoskillit/server/tools/tools_execution.py
Trecek: Acknowledged — minor suggestion noted. Consistent with existing circular-break convention across the file.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — fixed in 9724c0f8. Split `os.open` and `fcntl.flock` into separate try blocks so a flock failure closes the fd before returning.

src/autoskillit/server/tools/tools_execution.py
Trecek: Acknowledged — minor suggestion noted. The assignment is guarded by the same try block that covers the use site; no early-return path between them.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — fixed in 9724c0f8. Moved fcntl import to function scope (after os.open, before flock) so the finally block reuses the same import.

Trecek (commented):

tests/infra/test_pretty_output_formatters.py
```diff
@@ -1008,3 +1013,73 @@ def test_fleet_error_through_formatter_renders_error_fields():
...
+        "retriable": True,
+    }
+    outer = json.dumps({"result": json.dumps(envelope)})
+    text = _format_response("mcp__plugin_autoskillit_autoskillit__run_skill", outer, pipeline=True)
+    assert text is not None
+    assert "DEPENDENCY UNMET" in text
+    assert "{'rectify': 'pending'}" in text
+
+
+def test_dependency_unmet_routing_documented():
```
Trecek, line 1076: [info] tests: `test_dependency_unmet_routing_documented` only asserts the literal string "DEPENDENCY UNMET" appears somewhere in three source files. This is a documentation-presence check, not a behavioral test — it passes even if the routing instructions are wrong, incomplete, or contradictory.

tests/hooks/test_pipeline_step_guard.py
```diff
@@ -120,21 +128,37 @@ def test_order_id_discovered_from_single_tracker_file(self, tmp_path, monkeypatc
...
+            "CD",
             {"a": {"status": "pending"}, "b": {"status": "pending"}},
             {"b": ["a"]},
+            kitchen_id="kitchen-1",
         )
         event = json.dumps({"tool_input": {"step_name": "b", "order_id": ""}})
         code, stdout = _run(event, cwd=tmp_path)
         assert code == 0
-        assert stdout.strip() == ""
+        output = json.loads(stdout)
```
Trecek, line 151: [info] tests: `test_advisory_on_ambiguous_tracker_state` and `test_advisory_when_no_kitchen_id_available` both only assert the same generic substring ("cannot resolve tracker") in additionalContext. These cover two distinct failure modes but neither test distinguishes them via message content, so a regression collapsing the two code paths would not be caught.

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -101,11 +184,17 @@ async def record_pipeline_step(
         if op == "status":
             return _handle_status(tracker_path, effective_pipeline_id)
 
+        if op == "complete":
+            return _handle_complete(ctx, effective_pipeline_id, step_name)
+
         return json.dumps(
             {
```
Trecek: [warning] cohesion: The "unknown op" error branch in `record_pipeline_step` still builds a raw `json.dumps({...})` dict literal, while this same PR refactored every equivalent deny path in `tools_execution.py` to use `deny_envelope()`. This leaves the pipeline_tracker module's error shape inconsistent with the rest of the `run_skill` pre-flight deny surface (missing `stage`/`retriable` fields).

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -213,3 +302,142 @@ def _handle_status(tracker_path: Path, effective_pipeline_id: str) -> str:
             "total": len(steps),
         }
     )
+
+
+def _handle_complete(ctx: _TrackerCtx, effective_pipeline_id: str, step_name: str) -> str:
```
Trecek, line 308: [warning] cohesion: `_handle_complete()` constructs three separate raw `json.dumps({...})` dicts for its error cases instead of using `deny_envelope()`, the canonical constructor introduced in this same PR. Since `_handle_complete` is a pre-flight-adjacent guard path reached from the same `run_skill`/`record_pipeline_step` surface, this is a structural asymmetry within the PR's own stated goal.

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -213,3 +302,142 @@ def _handle_status(tracker_path: Path, effective_pipeline_id: str) -> str:
...
+        lock_path.parent.mkdir(parents=True, exist_ok=True)
+        lock_fd = os.open(str(lock_path), os.O_RDWR | os.O_CREAT, 0o644)
+    except OSError as exc:
+        return {
+            "success": False,
+            "is_error": True,
+            "error": f"mark_step_complete: failed to open lock file: {exc}",
+        }
+
+    import fcntl
```
Trecek: [info] slop: `import fcntl` is done locally inside `mark_step_complete()` even though the module already imports `json` and `os` at top-level and is a regular package module (not a standalone hook script). Should be a top-level import for consistency.

src/autoskillit/server/tools/_types.py
```diff
@@ -258,6 +262,29 @@ def input_failure_envelope(
...
+    error: str,
+    *,
+    stage: str,
+    retriable: bool,
+    recovery: str | None = None,
+) -> dict[str, object]:
```
Trecek, line 272: [info] defense: `deny_envelope()` returns `dict[str, object]` rather than a TypedDict. Its output is merged ad-hoc into RunSkillResult-shaped payloads at call sites, but there is no static check tying the envelope keys to the RunSkillResult TypedDict fields. A typo in a future call site (e.g. wrong `stage` string) would not be caught by mypy.

src/autoskillit/hooks/guards/pipeline_step_guard.py
```diff
@@ -19,13 +22,32 @@
...
+def _resolve_order_id_from_kitchen(tracker_dir: Path, kitchen_id: str) -> str:
+    """Select tracker by internal kitchen_id field (same rule as the server)."""
+    if not tracker_dir.is_dir():
+        return ""
+    candidates = []
+    for f in tracker_dir.iterdir():
+        if f.suffix != ".json" or f.name.startswith("."):
+            continue
+        try:
+            data = json.loads(f.read_text())
```
Trecek, line 46: [info] defense: `_resolve_order_id_from_kitchen` swallows `json.JSONDecodeError` and `OSError` per-file with a bare `continue`, silently skipping malformed or unreadable tracker files without any logging. Both resolvers (hook and server) independently duplicate this silent-skip pattern rather than sharing one implementation, so error-handling behavior could drift.

tests/arch/test_tracker_write_provenance.py
```diff
@@ -0,0 +1,84 @@
...
+from pathlib import Path
+
+import pytest
+
+pytestmark = [pytest.mark.layer("arch"), pytest.mark.small]
+
+HOOKS_DIR = Path(__file__).resolve().parents[2] / "src" / "autoskillit" / "hooks"
+
+# Patterns indicating a file write operation
+_WRITE_FUNC_NAMES = frozenset(
```
Trecek: [warning] slop: `_WRITE_FUNC_NAMES` frozenset is defined but never used — `_has_file_write` hardcodes its own literal checks directly instead of referencing this constant. Dead code.

tests/arch/test_tracker_write_provenance.py
```diff
@@ -0,0 +1,84 @@
...
+    {"write_text", "open", "atomic_write", "_atomic_write", "os.replace"}
+)
+_TRACKER_PATH_SEGMENTS = frozenset({"pipeline_tracker"})
```
Trecek: [warning] slop: `_TRACKER_PATH_SEGMENTS` frozenset is defined but never used — `_has_tracker_reference` hardcodes the literal string "pipeline_tracker" instead of referencing this constant. Dead code.

src/autoskillit/hooks/guards/pipeline_step_guard.py
```diff
@@ -19,13 +22,32 @@
...
+        except (json.JSONDecodeError, OSError):
+            continue
+        if data.get("kitchen_id") == kitchen_id:
+            candidates.append(f.stem)
+    if kitchen_id in candidates:
```
Trecek: [warning] bugs: `_resolve_order_id_from_kitchen` resolution rule diverges from the server's `resolve_tracker_order_id` despite the docstring claiming "same rule as the server". The hook version does not skip the self-named file from candidates and can produce advisory context text (order_id) that disagrees with what the server-side enforcer will actually check.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — fixed in 392450d. Refactored the unknown-op error branch to use `deny_envelope()` with `stage="preflight:pipeline_tracker"`, matching the pattern established in `tools_execution.py`.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — fixed in 392450d. All three `_handle_complete()` error paths now use `deny_envelope()` with consistent `stage`/`retriable` fields.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — fixed in 392450d. Moved `import fcntl` to module top-level alongside `json` and `os`.

tests/arch/test_tracker_write_provenance.py
Trecek: Agreed — fixed in 38e80d7. Removed the unused `_WRITE_FUNC_NAMES` frozenset.

tests/arch/test_tracker_write_provenance.py
Trecek: Agreed — fixed in 38e80d7. Removed the unused `_TRACKER_PATH_SEGMENTS` frozenset.

src/autoskillit/hooks/guards/pipeline_step_guard.py
Trecek: Agreed — fixed in 7e0e8c5. Updated `_resolve_order_id_from_kitchen` to skip the self-named file from candidates, matching the server's `resolve_tracker_order_id` logic. When exactly one non-self candidate exists, returns that stem; when none, returns `kitchen_id` as the implicit default.

tests/infra/test_pretty_output_formatters.py
Trecek: Acknowledged — minor suggestion noted. The test is intentionally a documentation-presence guard, not a behavioral contract.

tests/hooks/test_pipeline_step_guard.py
Trecek: Acknowledged — minor suggestion noted. The generic substring is intentional — the advisory message may evolve, and the test validates the guard fires, not the exact wording.

src/autoskillit/server/tools/_types.py
Trecek: Acknowledged — minor suggestion noted. A TypedDict would add static safety but also coupling between the envelope factory and callers. Worth considering in a future cohesion pass.

src/autoskillit/hooks/guards/pipeline_step_guard.py
Trecek: Acknowledged — minor suggestion noted. The hook is stdlib-only and advisory; logging would require either importing the logger (breaking the stdlib-only constraint) or writing to stderr (which could confuse the hook protocol). The silent continue is intentional for robustness.

Trecek (commented):

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -1,21 +1,101 @@
...
 from autoskillit.server._notify import track_response_size
 from autoskillit.server.tools._cancellation_shield import _cancellation_shield
+from autoskillit.server.tools._types import deny_envelope
 
 logger = get_logger(__name__)
 
+_STEP_SUFFIX_RE = re.compile(r"-\d+$")
+
+
+class _TrackerCtx(Protocol):
```
Trecek, line 32: [info] arch: `_TrackerCtx` is a new local Protocol (kitchen_id, project_dir) duck-typing the real `ToolContext` rather than importing it directly. This appears to be deliberate circular-import avoidance (tools_execution.py locally imports resolve_tracker_order_id from this module). Structurally sound, but creates a second unenforced contract for ToolContext shape — if ToolContext relevant fields are ever renamed, this Protocol won't be caught by the same refactor tooling. Worth a one-line comment noting why the Protocol exists instead of a direct import.

tests/server/test_kitchen_lifecycle.py
```diff
@@ -168,3 +174,277 @@ async def test_back_to_back_open_close_open_resets_infrastructure(monkeypatch, t
...
+    """A tracker whose registered PID is dead must be reaped on next open."""
+    tracker_dir = tmp_path / ".autoskillit" / "temp" / "pipeline_tracker"
+    _write_tracker(tracker_dir, "K1")
+    _write_registry(
+        monkeypatch,
+        tmp_path,
+        [
+            {
+                "kitchen_id": "K1",
+                "pid": 99999,
```
Trecek, line 202: [info] tests: test_open_without_close_prunes_dead_kitchen_tracker (and siblings) hardcode pid=99999 and assume it is dead. Latent flakiness source under PID reuse/namespaced containers; consider a shared helper that picks a guaranteed-invalid pid.

tests/integration/test_pipeline_step_completion_flow.py
```diff
@@ -0,0 +1,255 @@
...
+    }
+    tool_ctx_kitchen_open.input_contract_resolver = None
+
+
+def _read_tracker(tmp_path, kitchen_id="test-kitchen"):
+    tracker_path = tmp_path / ".autoskillit" / "temp" / "pipeline_tracker" / f"{kitchen_id}.json"
+    return json.loads(tracker_path.read_text())
+
+
+class TestServerSideStepCompletionMarking:
```
Trecek, line 50: [info] tests: Every test in this new file is wrapped in its own single-method class (7 classes / 7 tests). Sibling file test_run_skill_pipeline_deps.py groups multiple related tests per class. Consider consolidating into one class or using bare test functions to match convention.

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -101,12 +186,15 @@ async def record_pipeline_step(
         if op == "status":
             return _handle_status(tracker_path, effective_pipeline_id)
 
+        if op == "complete":
+            return _handle_complete(ctx, effective_pipeline_id, step_name)
```
Trecek: [warning] bugs: record_pipeline_step top-level gate rejects with "pipeline_id is required" when both pipeline_id and AUTOSKILLIT_DISPATCH_ID are absent, BEFORE op=complete dispatches to _handle_complete. This makes the kitchen_id-only fallback tier inside resolve_tracker_order_id unreachable for the op=complete path — an interactive session (kitchen_id set, no pipeline_id/dispatch env) cannot use record_pipeline_step(op=complete) to repair its own kitchen-scoped tracker, even though the shared resolver was built to support that case.

src/autoskillit/server/tools/tools_execution.py
```diff
@@ -803,40 +830,40 @@ async def run_skill(
...
-                        "is_error": True,
-                        "error": (
+                    deny_envelope(
+                        (
                             f"{DEPENDENCY_DENY_PREFIX}: step_name is empty and could "
                             "not be resolved from the recipe. Cannot verify dependency "
                             "status. Pass step_name explicitly."
                         ),
-                    }
+                        stage="preflight:ambiguous_step",
```
Trecek: [info] bugs: The third empty-step_name deny branch is tagged stage="preflight:ambiguous_step", same stage value used by the actual ambiguity branch at line 840. The error text in this branch does not mention ambiguity, so the stage label is misleading for anything triaging by stage.

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -213,3 +301,136 @@ def _handle_status(tracker_path: Path, effective_pipeline_id: str) -> str:
...
+    """
+    canonical = _STEP_SUFFIX_RE.sub("", step_name)
+    lock_path = tracker_path.parent / ".pipeline_tracker.lock"
+
+    try:
+        lock_path.parent.mkdir(parents=True, exist_ok=True)
+        lock_fd = os.open(str(lock_path), os.O_RDWR | os.O_CREAT, 0o644)
+    except OSError as exc:
+        return {
+            "success": False,
```
Trecek: [warning] cohesion: mark_step_complete() builds four separate raw success/error dict literals instead of using deny_envelope() that this same PR introduces. _handle_complete() in the same file and all preflight sites in tools_execution.py now use deny_envelope() consistently — this function is the remaining holdout.

src/autoskillit/hooks/guards/pipeline_step_guard.py
```diff
@@ -19,13 +22,41 @@
...
 
-from _hook_utils import (  # type: ignore[import-not-found]  # noqa: E402
-    discover_single_tracker_order_id,
-)
+from _hook_settings import read_merged_hook_config  # type: ignore[import-not-found]  # noqa: E402
 
 _SUFFIX_RE = re.compile(r"-\d+$")
 
 
+def _resolve_order_id_from_kitchen(tracker_dir: Path, kitchen_id: str) -> str:
```
Trecek, line 28: [warning] cohesion: _resolve_order_id_from_kitchen() duplicates the kitchen-scoped tracker resolution logic (scan tracker_dir, skip self-named file, match internal kitchen_id field, refuse on >1 candidate) that resolve_tracker_order_id() in tools_pipeline_tracker.py already implements. The hook is stdlib-only so cannot import the server module, but no shared stdlib-only helper was factored out for this matching rule.

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -213,3 +301,136 @@ def _handle_status(tracker_path: Path, effective_pipeline_id: str) -> str:
...
+        return json.dumps(
+            deny_envelope(
+                "record_pipeline_step: step_name is required for op='complete'.",
+                stage="preflight:pipeline_tracker",
+                retriable=False,
+            )
+        )
+
+    resolved = resolve_tracker_order_id(ctx, effective_pipeline_id)
+    if isinstance(resolved, ResolutionRefusal):
```
Trecek, line 319: [info] slop: mark_step_complete() returns raw dicts on its four error paths instead of using deny_envelope(). The docstring on deny_envelope says all pre-flight guards must use this constructor — mark_step_complete failures are pre-flight-adjacent but lack the stage/retriable fields the rest of the diff standardizes on.

tests/infra/test_schema_version_convention.py
```diff
@@ -119,12 +119,14 @@ def _is_yaml_dump(node: ast.expr) -> bool:
...
-    # tools_pipeline_tracker.py — tracker_data dict
-    ("src/autoskillit/server/tools/tools_pipeline_tracker.py", 166),
+    ("src/autoskillit/server/tools/tools_kitchen.py", 276),
+    ("src/autoskillit/server/tools/tools_kitchen.py", 295),
+    ("src/autoskillit/server/tools/tools_kitchen.py", 329),
+    ("src/autoskillit/server/tools/tools_kitchen.py", 1391),
+    # tools_pipeline_tracker.py — tracker_data dict (init) and mark_step_complete write
+    # (same tracker file schema as init — not a new format, grandfathered alongside it)
+    ("src/autoskillit/server/tools/tools_pipeline_tracker.py", 254),
+    ("src/autoskillit/server/tools/tools_pipeline_tracker.py", 402),
```
Trecek, line 129: [info] slop: Comment justifies adding a new unversioned-dict allowlist entry rather than giving the new dict a schema version — scope justification embedded as comment rather than addressed structurally, though acceptable as rationale.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — fixed in fbe5a134. Moved op=complete dispatch above the pipeline_id gate so resolve_tracker_order_id's kitchen_id-only fallback tier is reachable for interactive sessions.

src/autoskillit/server/tools/tools_execution.py
Trecek: Agreed — fixed in fbe5a134. Renamed to `preflight:unresolved_step` so it's distinguishable from the actual ambiguity branch at line 840.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — fixed in fbe5a134. All four error paths in mark_step_complete() now use deny_envelope() with stage="mark_step_complete" and appropriate retriable values.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — fixed in fbe5a134. Covered by the same change as the cohesion finding above — all mark_step_complete error paths now use deny_envelope().

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Acknowledged — minor suggestion noted. The Protocol exists for circular-import avoidance (tools_execution.py imports from this module); a comment would help but the pattern is well-established in this codebase.

tests/server/test_kitchen_lifecycle.py
Trecek: Acknowledged — minor suggestion noted. PID 99999 is a common test pattern; the flakiness risk is acceptable for these unit tests.

src/autoskillit/hooks/guards/pipeline_step_guard.py
Trecek: Acknowledged — valid observation. The hook is stdlib-only (cannot import server modules), so the duplication is architectural — a shared stdlib-only helper would need to live in hooks/ and be imported by both. Worth considering but a separate cleanup scope.

tests/infra/test_schema_version_convention.py
Trecek: Acknowledged — minor suggestion noted. The allowlist entry with comment is the established pattern for grandfathered formats.

tests/integration/test_pipeline_step_completion_flow.py
Trecek: Acknowledged — minor suggestion noted. The single-class-per-test structure provides clear fixture scoping and matches the independent-scenario intent.

Trecek (commented):

src/autoskillit/server/tools/tools_execution.py
```diff
@@ -351,6 +327,55 @@ def _derive_run_cmd_write_prefixes() -> tuple[str, ...]:
...
+                order_id=result["order_id"],
+                done=result["done"],
+                total=result["total"],
+            )
+        else:
+            logger.warning(
+                "pipeline_marker_failed",
+                error=result.get("error", "unknown"),
+            )
+        return {
```
Trecek: [warning] cohesion: _mark_step_complete_server_side() builds its returned dict from only three keys of mark_step_complete()'s result (step, order_id, status), dropping the 'advisory' key that mark_step_complete() populates when dependent steps still show unmet prerequisites (tools_pipeline_tracker.py line ~429). The op='complete' operator-repair path (_handle_complete) returns the full result including advisory, but the automatic server-side marker invoked at the run_skill adjudication point silently discards it — even though this is exactly the staleness/corruption signal the advisory field exists to surface to the agent.

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -1,21 +1,103 @@
...
+    _pipeline_tracker_dir,
+    _pipeline_tracker_path,
+)
 from autoskillit.server._notify import track_response_size
 from autoskillit.server.tools._cancellation_shield import _cancellation_shield
+from autoskillit.server.tools._types import deny_envelope
 
 logger = get_logger(__name__)
 
+_STEP_SUFFIX_RE = re.compile(r"-\d+$")
```
Trecek, line 29: [info] cohesion: New constant _STEP_SUFFIX_RE duplicates the identical regex (r"-\d+$") and purpose already defined as _SUFFIX_RE in the sibling hooks/guards/pipeline_step_guard.py (and hooks/token_summary_hook.py), but under a different name. No shared constant was factored out despite this PR already building a shared resolve_tracker_order_id() to unify tracker-resolution logic between the hook and server sides.

tests/server/test_kitchen_lifecycle.py
```diff
@@ -168,3 +174,277 @@ async def test_back_to_back_open_close_open_resets_infrastructure(monkeypatch, t
...
+            }
+        ],
+    )
+
+    prune_stale_kitchen_state(tmp_path, "K2")
+
+    assert not (tracker_dir / "K1.json").exists()
+
+
+def test_open_kitchen_sweeps_stale_kitchen_state_markers(monkeypatch, tmp_path):
```
Trecek: [warning] tests: test_open_kitchen_sweeps_stale_kitchen_state_markers claims (via name and docstring) to verify that open_kitchen wires sweep_stale_markers(), but the test body never calls open_kitchen() or _open_kitchen_handler() — it only calls sweep_stale_markers() directly. This duplicates existing coverage in tests/core/test_kitchen_state.py::test_sweep_stale_markers_scoped_to_namespace and leaves the actual new wiring added in tools_kitchen.py (the open_kitchen_sweep_markers_failed try/except block) completely untested.

tests/integration/test_pipeline_step_completion_flow.py
```diff
@@ -0,0 +1,255 @@
...
+_FAIL_RESULT = dataclasses.replace(
+    _SUCCESS_RESULT,
+    success=False,
+    exit_code=1,
+    is_error=True,
+    subtype="error",
+)
+
+
+def _write_tracker(tmp_path, pipeline_id, steps, dependencies, kitchen_id="test-kitchen"):
```
Trecek: [info] tests: _write_tracker and _setup_project are duplicated (near-verbatim) across this file, tests/server/test_run_skill_pipeline_deps.py, and tests/server/test_record_pipeline_step_complete.py instead of being factored into a shared conftest fixture or test utility module.

tests/server/test_kitchen_state_pruning.py
```diff
@@ -0,0 +1,82 @@
...
+from datetime import UTC, datetime
+
+import pytest
+
+from autoskillit.server.tools.tools_kitchen import prune_stale_kitchen_state
+
+pytestmark = [pytest.mark.layer("server"), pytest.mark.small]
+
+
+def _write_registry(monkeypatch, tmp_path, entries):
```
Trecek: [info] tests: _write_registry is duplicated verbatim from tests/server/test_kitchen_lifecycle.py (added in this same PR) rather than shared via conftest — both files test prune_stale_kitchen_state-related behavior and could share this helper.

src/autoskillit/server/tools/tools_pipeline_tracker.py
```diff
@@ -1,21 +1,103 @@
...
+        if len(active) > 1:
+            return ResolutionRefusal(
+                reason=(
+                    f"multiple pipelines are active under this kitchen "
+                    f"({sorted(active)}). Pass order_id explicitly to scope "
+                    "the dependency check."
+                ),
+                multi_pipeline=True,
+            )
+        effective_oid = kitchen_id
```
Trecek: [warning] bugs: resolve_tracker_order_id's kitchen-scoped fallback builds an `active` set of non-self tracker files matching kitchen_id, but only uses it to detect ambiguity (len > 1). When exactly one non-self candidate exists, the function still falls through to `effective_oid = kitchen_id` at this line rather than aliasing to the single found candidate. This diverges from `_resolve_order_id_from_kitchen` in `pipeline_step_guard.py` (same PR), whose docstring explicitly claims to match 'the same rule the server uses' but which does return the single candidate's stem (`next(iter(active))`) in that case. The hook's advisory order_id can therefore disagree with the tracker file the server-side enforcer actually resolves to.

src/autoskillit/hooks/guards/pipeline_step_guard.py
```diff
@@ -19,13 +22,41 @@
...
+            data = json.loads(f.read_text())
+        except (json.JSONDecodeError, OSError):
+            continue
+        if data.get("kitchen_id") == kitchen_id:
+            active.add(f.stem)
+    if len(active) > 1:
+        return ""
+    if len(active) == 1:
+        return next(iter(active))
+    return kitchen_id
```
Trecek, line 55: [warning] bugs: _resolve_order_id_from_kitchen aliases to the single non-self candidate found via kitchen_id match (`return next(iter(active))`), but the server's `resolve_tracker_order_id` (tools_pipeline_tracker.py) never does this — it always falls through to `effective_oid = kitchen_id` regardless of whether 0 or 1 non-self candidates were found. The docstring here claims equivalence to the server's rule ('matching resolve_tracker_order_id') but the behavior differs whenever exactly one non-self candidate exists, producing an advisory order_id that disagrees with what the server will actually enforce/mark against.

Trecek (commented):

src/autoskillit/server/tools/tools_execution.py
Trecek: Agreed — fixed in 7e4d5171a. `_mark_step_complete_server_side()` now forwards the `advisory` key when present. Added test `TestAdvisorySurfacedOnUnmetDependents` to verify end-to-end propagation.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed within the hooks layer — consolidated into `_hook_utils.STEP_SUFFIX_RE` in 10a8cdcc8. The server-layer copy (`_STEP_SUFFIX_RE` in `tools_pipeline_tracker.py`) is intentionally kept separate due to the import layer boundary (hooks are stdlib-only, can't import from IL-3 server).

tests/server/test_kitchen_lifecycle.py
Trecek: Agreed — rewrote the test in 7e4d5171a to actually call `_open_kitchen_handler()` instead of testing `sweep_stale_markers()` directly. Now genuinely validates the wiring path.

tests/integration/test_pipeline_step_completion_flow.py
Trecek: Agreed — extracted into `tests/server/_pipeline_test_helpers.py` in 7e4d5171a. All three consumer files now import from the shared module.

tests/server/test_kitchen_state_pruning.py
Trecek: Agreed — moved to `tests/server/_helpers.py` in 7e4d5171a. Both consumer files now import from the shared location.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — real bug. Fixed in 5b8edb88: `resolve_tracker_order_id` now uses `next(iter(active)) if active else kitchen_id`, matching the hook's single-candidate behavior. Added `TestResolveTrackerOrderIdSingleCandidate` to prevent regression.

src/autoskillit/hooks/guards/pipeline_step_guard.py
Trecek: Same fix as the sibling comment — 5b8edb88 aligns the server-side `resolve_tracker_order_id` to match this hook's behavior. Both now return the single candidate's stem when exactly one non-self candidate exists.

src/autoskillit/server/tools/tools_execution.py
Trecek: Agreed — fixed in 7e4d5171a. _mark_step_complete_server_side() now copies the advisory key from mark_step_complete()'s result into the returned marker dict when present, so it reaches _parsed['pipeline_tracker'] in the run_skill response. Added a regression test (test_advisory_surfaced_when_dependent_still_unmet).

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — fixed in 10a8cdcc8. Consolidated the two hook-side duplicates (pipeline_step_guard.py, token_summary_hook.py) into a shared STEP_SUFFIX_RE constant in hooks/_hook_utils.py. The server-side copy in tools_pipeline_tracker.py is intentionally kept separate: hooks/ scripts are stdlib-only standalone subprocesses that cannot import from autoskillit.server (see hooks/AGENTS.md), so full 3-way unification isn't achievable without violating that layering boundary.

tests/server/test_kitchen_lifecycle.py
Trecek: Agreed — fixed in 7e4d5171a. Rewrote test_open_kitchen_sweeps_stale_kitchen_state_markers to actually call _open_kitchen_handler() (via make_context + monkeypatched server._state._ctx) instead of calling sweep_stale_markers() directly, so it genuinely exercises the wiring the name/docstring claim to cover.

tests/integration/test_pipeline_step_completion_flow.py
Trecek: Agreed — fixed in 7e4d5171a. Extracted _write_tracker and _setup_project into shared tests/server/_pipeline_test_helpers.py; all three consumer test files (this one, test_run_skill_pipeline_deps.py, test_record_pipeline_step_complete.py) now import from there.

tests/server/test_kitchen_state_pruning.py
Trecek: Agreed — fixed in 7e4d5171a. Moved _write_registry into tests/server/_helpers.py; both this file and test_kitchen_lifecycle.py now import the shared helper.

src/autoskillit/server/tools/tools_pipeline_tracker.py
Trecek: Agreed — fixed in 5b8edb883. resolve_tracker_order_id now aliases to the single non-self candidate's stem (next(iter(active))) instead of falling through to kitchen_id, matching _resolve_order_id_from_kitchen in pipeline_step_guard.py. Added TestResolveTrackerOrderIdSingleCandidate regression test.

src/autoskillit/hooks/guards/pipeline_step_guard.py
Trecek: Agreed — fixed in 5b8edb883, on the server side rather than here. This file's _resolve_order_id_from_kitchen behavior was already correct (returns the single candidate's stem); resolve_tracker_order_id on the server was the one diverging. Both now agree, so the docstring's equivalence claim is accurate.

PR #4330, opened 2026-07-22
Rectify: Host-Attested Codex Recipe Delivery and Cumulative-Insertion Immunity

Trecek:
## Summary

Codex recipe delivery currently treats four different quantities as though they were one limit:

- the 195,000-byte measured `open_kitchen` / `load_recipe` source exemption;
- Code Mode's 10,000-token default when an outer maximum is omitted;
- the 56,750-token damage-bounded maximum that can be explicitly requested for the measured recipe;
- `tool_output_token_limit`, which controls later history retention rather than the outer Code Mode result.

## Implementation Plan

Plan file: `/home/talon/projects/generic_automation_mcp/.autoskillit/temp/rectify/rectify_codex_host_attested_recipe_delivery_2026-07-22_094943.md`


tests/server/test_tools_recipe_pull.py
```diff
@@ -1,1262 +1,342 @@
...
+    generation = persist_recipe_artifact(
+        tool_ctx_kitchen_open.temp_dir,
+        kitchen_id=tool_ctx_kitchen_open.kitchen_id,
+        producer_tool="open_kitchen",
+        recipe_name="remediation",
+        payload=_payload("héllo\n" * 2_000),
     )
+    kwargs = generation.pull_identity()
+    kwargs.pop("pull_tool")
+    response = json.loads(await get_recipe_section(section="content", part=0, **kwargs))
```
Trecek: [warning] tests: This byte-offset test exercises only part=0 and merely checks byte_end <= byte_total. It never validates the offsets against the returned UTF-8 content or reconstructs all next_part chunks, so cursor, multibyte-boundary, termination, and lossless-round-trip regressions can pass.

tests/contracts/test_delivery_bound_fitness.py
```diff
@@ -97,28 +101,24 @@ def test_bundled_recipe_open_kitchen_envelope_fits_per_backend(
...
+    bound_tokens = resolve_general_output_token_limit(caps)
     bound_bytes = _effective_bound_bytes(bound_tokens)
     envelope = build_recipe_envelope(
         payload,
         recipe_name=recipe_name,
-        artifact_path=str(artifact_path),
-        artifact_sha256=sha256,
-        skeleton=skeleton,
+        generation=generation,
+        skeleton_source=cast(
```
Trecek, line 133: [warning] tests: The fitness test passes recipe_name="" and active_recipe_steps=None, so build_recipe_envelope never exercises production routing edges or step summaries. That understates the real envelope size and leaves the new skeleton/summary truncation path untested.

src/autoskillit/server/tools/tools_recipe.py
```diff
@@ -481,30 +490,39 @@ async def get_recipe_section(
...
-            if (
-                artifact_path is not None
-                and Path(artifact_path).resolve() != resolved_artifact_path.resolve()
-            ):
+            if descriptor_version <= 0 or schema_version <= 0:
                 return json.dumps({"success": False, "error": "invalid_recipe_artifact_identity"})
 
-            if not resolved_artifact_path.exists():
-                if producer_tool != "open_kitchen":
+            try:
```
Trecek, line 519: [warning] tests: The rewritten missing-generation recreation path has no replacement regression coverage after the prior recreation tests were removed. Successful recreation, invalid recreation, and the post-recreation identity check can now regress unnoticed.

src/autoskillit/execution/backends/_codex_config.py
```diff
@@ -24,14 +31,105 @@
...
+
+# A protected evidence identity is enabled only with its passing conformance
+# report. Writable rollout and trace formats are intentionally absent.
+SUPPORTED_CODEX_RECIPE_EVIDENCE_REGISTRY: dict[str, CodexRecipeDeliveryEvidenceDef] = {}
+
+
+def codex_recipe_delivery_calling_contract(*, mcp_prefix: str = "") -> str:
+    """Generate the only permitted Codex high-budget recipe calling contract."""
+    budget = CODEX_RECIPE_DELIVERY_BUDGET
+    full_recipe_tools = ", ".join(
```
Trecek, line 96: [warning] cohesion: The generated calling contract hardcodes open_kitchen/load_recipe rather than deriving distinct negotiation-eligible producers from RECIPE_DELIVERY_SURFACE_REGISTRY. A surface addition or rename can update the registry and digest while leaving the model-visible eligible-tool list stale.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,562 @@
...
+    """Return public producers authorized to resolve immutable pull generations."""
+    return frozenset(
+        definition.producer_tool
+        for definition in RECIPE_DELIVERY_SURFACE_REGISTRY.values()
+        if definition.pull_eligible
+    )
+
+
+def _estimated_tokens(rendered: str) -> int:
+    return (len(rendered.encode("utf-8")) + 3) // 4
```
Trecek: [warning] defense: ceil(UTF-8 bytes / 4) is not a conservative token bound. Token-dense JSON or Unicode can be classified as fitting the ordinary or attested result limit when it does not, allowing downstream truncation. Use an authoritative tokenizer or a genuinely conservative upper bound for mode selection.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,562 @@
...
+        producer_tool=producer_tool,
+        recipe_name=recipe_name,
+        payload_sha256=generation.payload_sha256,
+    )
+    descriptor = _canonical_payload(generation.pull_identity()).decode("utf-8")
+    with _generation_lock(temp_dir, exclusive=True):
+        directory.mkdir(parents=True, exist_ok=True)
+        blob_path = directory / "payload.json"
+        descriptor_path = directory / "descriptor.json"
+        if blob_path.exists() and blob_path.read_bytes() != blob:
```
Trecek: [warning] defense: Collision checking reads an existing payload.json fully into memory, and descriptor.json is handled similarly on line 205. Because this generation tree is same-user-writable, a precreated or replaced file can force an unbounded allocation during persistence. Compare via server-capped reads and reject oversized files.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,562 @@
...
+        producer_tool=identity.producer_tool,
+        recipe_name=identity.recipe_name,
+        payload_sha256=identity.payload_sha256,
+    )
+    with _generation_lock(temp_dir, exclusive=False):
+        blob_path = directory / "payload.json"
+        descriptor_path = directory / "descriptor.json"
+        try:
+            with blob_path.open("rb") as blob_file:
+                blob = blob_file.read(identity.artifact_blob_size_bytes + 1)
```
Trecek: [warning] defense: artifact_blob_size_bytes comes from the public pull request and directly controls the read size; descriptor.json is also read without a cap. A forged identity can request a huge allocation before integrity checks fail. Validate identity sizes and enforce independent server-owned read caps.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
```diff
@@ -0,0 +1,638 @@
...
+        return base / campaign_id
+    return base
+
+
+def _fresh_marker_session_id(path: Path, *, now_unix: int, ttl_seconds: int) -> str | None:
+    try:
+        stat = path.stat()
+        if not path.is_file() or stat.st_size <= 0 or stat.st_size > _MARKER_MAX_BYTES:
+            return None
+        record = json.loads(path.read_text(encoding="utf-8"))
```
Trecek: [warning] defense: The marker size check and read are separate path operations. A same-user replacement or growth race after stat() can defeat _MARKER_MAX_BYTES and cause an unbounded read. Open once, validate that descriptor, and read at most max+1 bytes.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
```diff
@@ -0,0 +1,638 @@
...
+def enumerate_fresh_codex_marker_ids(
+    project_dir: Path,
+    *,
+    now_unix: int,
+    ttl_seconds: int = 24 * 60 * 60,
+) -> tuple[tuple[str, Path], ...]:
+    """Return every fresh marker; ordering never grants authority."""
+    state_dir = _marker_state_dir(project_dir)
+    try:
+        candidates = tuple(state_dir.glob("*.json"))
```
Trecek: [warning] defense: Diagnostic correlation materializes every *.json marker before processing and has no candidate-count limit. Many small markers can cause unbounded memory use and rollout lookups despite the per-file cap. Iterate under a strict maximum and fail closed when exceeded.

src/autoskillit/core/types/_type_constants_registries.py
```diff
@@ -189,20 +182,88 @@ class ResponseBackstopExemptionDef(NamedTuple):
...
+    route_kind: Literal["tool", "deferred_tool", "resource"]
+    negotiation_eligible: bool
+    pull_eligible: bool
+    response_exemption_tool: str | None
+    response_exemption: ResponseBackstopExemptionDef | None
+
+
+_RECIPE_RESPONSE_MAX_UTF8_BYTES = 195_000
+
+RECIPE_DELIVERY_SURFACE_REGISTRY: dict[str, RecipeDeliverySurfaceDef] = {
```
Trecek: [warning] defense: This policy authority is a mutable dict, while its contract digest and derived exemption registry are computed only once. A later mutation changes negotiation or pull authorization without changing the attested digest. Validate the completed registry and expose an immutable mapping.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,562 @@
...
+            else:
+                receipt_handle = reservation.handle
+                decision = replace(decision, receipt_status="pending")
+
+    if decision.mode is RecipeDeliveryMode.ORDINARY_INLINE:
+        rendered = ordinary_rendered
+    elif decision.mode is RecipeDeliveryMode.ATTESTED_INLINE:
+        rendered = high_rendered
+    else:
+        from autoskillit.server.tools._serve_helpers import (  # circular-break
```
Trecek: [warning] arch: The delivery finalizer reaches into server.tools._serve_helpers for envelope construction and needs a local circular-break import. This reverses the dependency between delivery infrastructure and tool presentation; move the builder into this module or a neutral lower-level module used by both.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
```diff
@@ -0,0 +1,638 @@
...
+        return self._path
+
+    @classmethod
+    def initialize_protected(
+        cls, authority: ProtectedStoreAuthority
+    ) -> RecipeDeliveryReceiptLedger:
+        """Initialize a store only after the protected host attests its boundary."""
+        ledger = cls(authority)
+        authority.root.mkdir(mode=0o700, parents=True, exist_ok=True)
+        if ledger._path.exists():
```
Trecek, line 332: [warning] bugs: Any existing database path is treated as an initialized protected store. A crash after sqlite3.connect creates the file but before schema/metadata setup leaves a file that every later startup rejects permanently. Initialize a temporary database and publish atomically, or explicitly recover incomplete first-time initialization.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,562 @@
...
+        recipe_name=identity.recipe_name,
+        payload_sha256=identity.payload_sha256,
+    )
+    with _generation_lock(temp_dir, exclusive=False):
+        blob_path = directory / "payload.json"
+        descriptor_path = directory / "descriptor.json"
+        try:
+            with blob_path.open("rb") as blob_file:
+                blob = blob_file.read(identity.artifact_blob_size_bytes + 1)
+            descriptor_raw = descriptor_path.read_text(encoding="utf-8")
```
Trecek: [warning] bugs: Non-UTF-8 artifact bytes are not normalized to RecipeArtifactError: read_text() and json.loads(blob) may raise UnicodeDecodeError, but the handlers catch only OSError or JSONDecodeError. The pull tool then misses its corruption/recreation path and returns a generic unhandled error.

Trecek (commented):

Trecek (commented):

tests/server/test_tools_recipe_pull.py
Trecek: Agreed — fixed in 5a7073dfc8524f1cadb18c6ce79692cc646568fa. The test now follows every continuation, validates contiguous UTF-8 byte offsets and terminal byte totals, and losslessly reconstructs multibyte content.

tests/contracts/test_delivery_bound_fitness.py
Trecek: Agreed — fixed in 9729c7f92362e237ad328acf9c13a1d09b69fbba. The fitness path now supplies real recipe names and active steps, verifies production skeleton metadata, resets shared recipe cache state, and measures exact compact wire serialization.

src/autoskillit/server/tools/tools_recipe.py
Trecek: Agreed — fixed in b8cbd6743fec408513de9ce63cea52bcf8a9486d. Focused tests now cover successful missing-generation recreation, invalid recreated payloads, and rejection when the recreated generation does not match the requested identity.

src/autoskillit/execution/backends/_codex_config.py
Trecek: Agreed — fixed in 2487cd5032b7b9f449b47740ab7a1f317111b736. The Codex calling contract derives its sorted distinct producer list from negotiation-eligible entries in RECIPE_DELIVERY_SURFACE_REGISTRY, with a registry-derived contract test.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in e134980eb5c18c6a0843ee46e08c43d412421ea8. Admission now uses UTF-8 byte count as a genuinely conservative token upper bound, so token-dense JSON or Unicode fails closed to the immutable pull envelope instead of risking downstream truncation.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in 8c9be5e915f46bddc9e295f05e89f73341987c5c. Collision checks use server-capped reads at the expected canonical payload and descriptor lengths plus one byte, rejecting oversized pre-created files without unbounded allocation.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in ec591855bc8466b1d1ab0520253c46b513c06cb3. Caller-supplied sizes are validated against independent server ceilings, and both artifact blobs and descriptors are read through bounded helpers before integrity validation.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
Trecek: Agreed — fixed in 4a94119ba35160037b4ee50a3b4a7a7522267081, with path-guard test isolation refined in 6be9b3f38cb60e45db408c252714a1960f733ecb. Marker handling opens once, validates that descriptor with fstat, and reads at most the ceiling plus one byte.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
Trecek: Agreed — fixed in 45d77496f28b8df2e324347acc127bb926e0b97b. Diagnostic correlation lazily iterates marker candidates, caps the set at 256, and fails closed on candidate 257 before further marker processing or rollout lookup.

Trecek: Agreed — fixed in 45d77496f28b8df2e324347acc127bb926e0b97b. Diagnostic correlation lazily iterates marker candidates, caps the set at 256, and fails closed on candidate 257 before further marker processing or rollout lookup.

src/autoskillit/core/types/_type_constants_registries.py
Trecek: Agreed — fixed in 7195e2eec2efd5e08783b25b7800b90ce347ff01. The delivery-surface and derived exemption authorities are immutable MappingProxyType mappings over immutable definitions, with mutation guards and canonical digest tests.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in ee4719a85358efd211fe132e375742a2faa424a5. Envelope construction and its supporting helpers now live in the recipe-delivery domain; the reverse import from server.tools is removed and tool presentation consumes the lower-level builder.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
Trecek: Agreed — fixed in e37d412a4c360b26a5ba6046b3aeeb7d4e0bed72. Protected SQLite stores are fully initialized at a unique temporary path and atomically published; interrupted first-time initialization leaves no poisoned final database and a retry succeeds.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in 535e7603767ce3b5aadd7e54b2ca8e0cdc1deac1. Descriptor and payload decoding now normalize UnicodeDecodeError alongside JSON decoding failures into RecipeArtifactError, preserving the corruption/recreation path.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
```diff
@@ -0,0 +1,638 @@
...
+                "INSERT OR REPLACE INTO metadata(key, value) VALUES('schema_version', ?)",
+                (str(_LEDGER_SCHEMA_VERSION),),
+            )
+            connection.execute(
+                "INSERT OR REPLACE INTO metadata(key, value) VALUES('security_identity', ?)",
+                (authority.security_identity,),
+            )
+        finally:
+            connection.close()
+        try:
```
Trecek: [warning] defense: Protected-store initialization swallows chmod failure and still returns a usable ledger. That accepts a database whose required 0600 protection was never established; initialization should fail closed or verify an equivalent protected permission boundary.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
```diff
@@ -0,0 +1,638 @@
...
+        self,
+        *,
+        request: RecipeDeliveryRequest,
+        attestation: RecipeDeliveryAttestation,
+        producer: str,
+        payload_sha256: str,
+        now_unix: int,
+    ) -> RecipeReservationResult:
+        """Consume one host call and create its pending receipt in one transaction."""
+        if (
```
Trecek: [warning] defense: This evidence-consumption boundary accepts raw request and attestation objects but revalidates only delivery_call_id, token count, and code_digest. Audience, contract, expiry, request digest, evidence identity, and selected limit remain caller prerequisites, so a direct caller can permanently consume unvalidated evidence. Require a validated capability type or perform the complete validation here.

src/autoskillit/server/_notify.py
```diff
@@ -108,9 +112,15 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:
                     }
                 )
                 logger.exception("track_response_size_handler_failed", tool_name=tool_name)
+            finalized = result if isinstance(result, FinalizedRecipeResponse) else None
```
Trecek, line 115: [warning] tests: The new FinalizedRecipeResponse branch is not tested through the track_response_size decorator used by open_kitchen/load_recipe. Direct finalizer tests cannot catch regressions in wrapper unwrapping, selected-limit handoff, or commit/abort sequencing. Add decorated-handler coverage for exact commit, transformed/spilled abort, and enforcement failure.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,562 @@
...
+        payload,
+        ensure_ascii=False,
+        sort_keys=True,
+        separators=(",", ":"),
+    ).encode("utf-8")
+
+
+def _safe_component(value: str) -> str:
+    safe = "".join(
+        character if character.isalnum() or character in "._-" else "_" for character in value
```
Trecek: [critical] defense: _safe_component preserves the special path components '.' and '..'. retire_recipe_artifacts then passes that component to shutil.rmtree, so a malformed kitchen_id such as '..' resolves outside the kitchen namespace and can delete unrelated temp data. Reject or rewrite dot-only components before constructing paths.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,562 @@
...
+            handle,
+            now_unix=int(time.time()) if now_unix is None else now_unix,
+        ):
+            return enforced
+        enforced = json.dumps(
+            {"success": False, "error": "recipe_delivery_receipt_commit_failed"},
+            separators=(",", ":"),
+        )
+    if handle is not None and ledger is not None:
+        ledger.abort(handle)
```
Trecek: [warning] defense: The return value from receipt abort is ignored. A transient store failure can leave a pending receipt that blocks later attested delivery, while the caller receives no indication cleanup failed; the stale-recovery method is not integrated into a production path in this diff.

Trecek (commented):

tests/infra/test_pretty_output_recipe.py
```diff
@@ -441,7 +446,21 @@ def test_fmt_open_kitchen_plain_text():
     )
     assert formatted is not None
```
Trecek, line 449: [warning] tests: This edit removes the only assertion that the plain-text open_kitchen body survives formatting. The remaining assertion passes for a header-only result, so payload loss is no longer detected. Restore an assertion for the original 'Kitchen is open' content.

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments — changes required. The formal request-changes event is unavailable because this PR is self-authored.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
Trecek: Agreed — fixed in eed4e5588. Protected-store initialization now fails closed when chmod fails or the resulting file is not a regular file with mode 0600; reopening also rejects stores without that protection.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
Trecek: Agreed — fixed in d0598c8dd. The durable reservation boundary now reruns the complete recipe-delivery decision validation, including audience, contract, expiry, request digest, selected limit, and evidence identity, before consuming evidence.

src/autoskillit/server/_notify.py
Trecek: Agreed — fixed in 910a0d405. Decorated-handler tests now cover exact-response commit and selected-limit handoff, transformed and spilled-response aborts, and enforcement-failure abort sequencing for FinalizedRecipeResponse.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in c26751133 with the fail-closed retirement boundary completed in f783b2d8f. Dot-only path components are rejected before namespace construction, and retirement returns False without touching unrelated temp data.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in 4e0a86603. Receipt abort failures now return a structured recipe_delivery_receipt_abort_failed response instead of silently leaving a pending receipt while reporting the transformed result.

tests/infra/test_pretty_output_recipe.py
Trecek: Agreed — fixed in a4be986dc. The plain-text formatter test again asserts that the original Kitchen is open payload survives formatting, so a header-only result cannot pass.

src/autoskillit/core/_delivery_bounds.py
```diff
@@ -1,19 +1,198 @@
...
+def resolve_general_output_token_limit(caps: BackendCapabilities) -> int:
+    """Return the backend's conservative limit for an unnegotiated tool result."""
+    return caps.unnegotiated_tool_result_token_limit
+
+
+def resolve_recipe_delivery_decision(
+    *,
+    capabilities: BackendCapabilities,
+    required_serialized_tokens: int,
+    budget: CodexRecipeDeliveryBudgetDef,
```
Trecek: [warning] arch: The backend-neutral resolver is typed to CodexRecipeDeliveryBudgetDef, while the generic server finalizer always supplies CODEX_RECIPE_DELIVERY_BUDGET for every backend. A future protected-capable backend would therefore be evaluated against Codex limits and evidence versions. Supply a backend-neutral budget from the selected backend, or make the flow explicitly Codex-only.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
```diff
@@ -0,0 +1,638 @@
...
+
+
+def _marker_state_dir(project_dir: Path) -> Path:
+    override = os.environ.get("AUTOSKILLIT_STATE_DIR")
+    if override:
+        return Path(override) / "kitchen_state"
+    base = project_dir / ".autoskillit" / "temp" / "kitchen_state"
+    campaign_id = os.environ.get("AUTOSKILLIT_CAMPAIGN_ID", "")
+    if campaign_id:
+        if "/" in campaign_id or os.sep in campaign_id:
```
Trecek: [warning] defense: Campaign validation rejects separators but accepts the special component `..`; _marker_state_dir then escapes kitchen_state and scans the parent temp directory for marker JSON. Reject `.` and `..` before joining so a campaign cannot broaden correlation discovery or force ambiguity.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
```diff
@@ -0,0 +1,638 @@
...
+        now_unix: int,
+    ) -> CodexAttestationResult:
+        correlation = resolve_unique_codex_host_correlation(
+            project_dir,
+            locate_rollout=self.locate_rollout,
+            now_unix=now_unix,
+        )
+        if correlation is None:
+            return CodexAttestationResult(None, None, None, "host_correlation_ambiguous")
+        try:
```
Trecek, line 277: [warning] defense: The protected-provider boundary catches operational and validation exceptions and discards all error context without logging, collapsing provider corruption, configuration faults, and transient I/O into the same generic reason. Preserve the exception in protected diagnostics while still failing closed.

src/autoskillit/server/_notify.py
```diff
@@ -142,19 +152,23 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:
             artifact_dir = (
                 temp_dir / "responses" / tool_name if isinstance(temp_dir, Path) else None
             )
-            effective_delivery_token_limit: int | None = None
-            if ctx is not None:
+            unnegotiated_tool_result_token_limit: int | None = None
+            if finalized is not None:
+                unnegotiated_tool_result_token_limit = (
+                    finalized.decision.selected_result_token_limit
```
Trecek: [warning] cohesion: The variable and enforce_response_budget parameter named unnegotiated_tool_result_token_limit receive finalized.decision.selected_result_token_limit here; enforce_recipe_resource_response repeats the mismatch. One API consequently means the ordinary bound for general callers and the negotiated bound for recipe callers, obscuring the authority distinction. Use a neutral selected-result limit parameter or separate enforcement entry points.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,562 @@
...
+) -> RecipeArtifactGeneration:
+    """Publish an immutable canonical payload and its generation descriptor."""
+    blob = _canonical_payload(payload)
+    generation = _generation_from_payload(
+        producer_tool=producer_tool,
+        recipe_name=recipe_name,
+        blob=blob,
+        payload=payload,
+    )
+    directory = _generation_dir(
```
Trecek, line 246: [warning] bugs: Generation paths omit descriptor_version and schema_version. If either version is incremented while payload bytes remain unchanged, the new generation resolves to the old directory and the descriptor comparison reports a content-addressed collision, making recipe delivery fail until that kitchen namespace is removed. Include the version domain in the generation key or publish versioned descriptors separately.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,562 @@
...
+        raise RecipeArtifactError("recipe body identity mismatch")
+    return payload
+
+
+def retire_recipe_artifacts(temp_dir: Path, *, kitchen_id: str) -> bool:
+    """Retire one kitchen namespace after all shared-lock readers finish."""
+    namespace = _artifact_root(temp_dir) / _safe_component(kitchen_id)
+    try:
+        with _generation_lock(temp_dir, exclusive=True):
+            if namespace.exists():
```
Trecek, line 353: [warning] bugs: The exclusive lock only prevents persistence while retirement is executing; it does not mark the kitchen namespace retired. A recipe call that passed its kitchen checks before close but reaches persist_recipe_artifact after this lock is released can recreate the deleted namespace, leaving artifacts behind after close. Coordinate close with in-flight producers or record retirement under the same lock and reject later writes.

src/autoskillit/server/tools/tools_recipe.py
```diff
@@ -481,30 +490,39 @@ async def get_recipe_section(
...
+                persisted = load_recipe_artifact(
+                    artifact_dir,
+                    kitchen_id=tool_ctx.kitchen_id,
+                    identity=identity,
+                )
+            except RecipeArtifactError:
+                if producer_tool not in {"open_kitchen", "get_recipe"}:
```
Trecek: [warning] cohesion: Artifact authorization is derived from RECIPE_DELIVERY_SURFACE_REGISTRY via recipe_pull_producers(), but missing-generation recreation uses a second hard-coded producer set. Adding or changing a surface can authorize pulls while silently leaving recreation behavior stale. Encode recreation eligibility in the registry and consume it through the delivery service.

tests/server/test_tools_recipe_pull.py
```diff
@@ -1,1262 +1,342 @@
...
-    ``recipe_artifact_unavailable`` envelope (driven here via a
-    ``serve_recipe`` monkeypatch, since the literal "no active recipe"
-    condition is checked earlier and produces a different error)."""
-    from autoskillit.server.tools.tools_recipe import get_recipe_section
 
-    monkeypatch.chdir(tmp_path)
 
-    ok_result = await _open_kitchen_patched(_RECIPE_FOR_PULL, None, monkeypatch)
-    assert ok_result.get("success") is True, f"open_kitchen failed: {ok_result}"
+def test_codex_without_supported_host_evidence_uses_bounded_envelope(tool_ctx) -> None:
```
Trecek, line 335: [warning] tests: This single synthetic 40,000-byte envelope check replaces the removed tight and extreme budget, multibyte truncation, priority field, and bundled recipe/backend matrix tests. It never exercises those build_recipe_envelope branches, so envelope overflow or loss of routing/control fields can regress unnoticed.

Trecek (commented):

Trecek (commented):

src/autoskillit/core/_delivery_bounds.py
Trecek: Agreed — fixed in f96c98a24. RecipeDeliveryBudgetDef and RecipeDeliveryEvidenceDef are backend-neutral, BackendCapabilities now owns recipe_delivery_budget, and finalization selects that backend-provided budget instead of injecting the Codex contract. Added fail-closed missing-budget and selected-budget regression coverage.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
Trecek: Agreed — fixed in 53680b64e. _marker_state_dir now rejects the special components `.` and `..` before joining, with a regression proving neither component can broaden marker discovery outside its campaign namespace.

src/autoskillit/execution/backends/_codex_recipe_delivery.py
Trecek: Agreed — fixed in 1be1594c6. Protected-provider exceptions now emit the structured `codex_protected_provider_failed` diagnostic with the exception type and traceback (`exc_info=True`) while preserving the same fail-closed result.

src/autoskillit/server/_notify.py
Trecek: Agreed — fixed in 489ca3f3f. The response enforcement API now uses the neutral `selected_result_token_limit` name end to end: ordinary callers supply the backend-resolved limit, while finalized recipe responses supply the selected decision limit.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in 1c459c118 (following the versioned-path change in 051814364). Generation paths include descriptor and schema versions, and historical generations are revalidated against their own recorded versions rather than current constants.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in a0c64637e. Retirement writes a kitchen tombstone under the same exclusive generation lock before deleting the namespace; persistence checks that tombstone after acquiring the lock and rejects any producer that resumes after close.

src/autoskillit/server/tools/tools_recipe.py
Trecek: Agreed — fixed in 36954ec9f. RecipeDeliverySurfaceDef now carries `recreation_eligible`; recipe_recreation_producers derives policy from RECIPE_DELIVERY_SURFACE_REGISTRY, and get_recipe_section consumes that helper instead of a hard-coded producer set.

tests/server/test_tools_recipe_pull.py
Trecek: Agreed — fixed in 92c95f7e2. Direct tests now cover multibyte priority-field truncation plus pull, error-only, empty-object, and irreducible fallback bounds. The existing contract-level bundled-recipe × backend matrix continues to exercise production envelope metadata. Full gate: 29,673 passed.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,947 @@
...
+    kitchen_id: str,
+    producer_tool: str,
+    recipe_name: str,
+    descriptor_version: int,
+    schema_version: int,
+    payload_sha256: str,
+) -> Path:
+    return (
+        _artifact_root(temp_dir)
+        / _safe_component(kitchen_id)
```
Trecek, line 174: [critical] defense: _safe_component is non-injective: distinct kitchen IDs such as `a/b` and `a?b` both become `a_b`. Persistence and retirement use that normalized component, so retiring one malformed ID can delete another kitchen's immutable generations. Reject invalid components or encode them collision-free.

Trecek (commented):

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,947 @@
...
+            RECIPE_ARTIFACT_DESCRIPTOR_VERSION
+            if descriptor_version is None
+            else descriptor_version
+        ),
+        schema_version=(
+            RECIPE_ARTIFACT_SCHEMA_VERSION if schema_version is None else schema_version
+        ),
+        payload_sha256=_domain_sha256("autoskillit.recipe-payload.v1", blob),
+        artifact_blob_sha256=_qualified_sha256(blob),
+        artifact_blob_size_bytes=len(blob),
```
Trecek, line 222: [critical] bugs: Persistence accepts blobs above RECIPE_ARTIFACT_MAX_BLOB_BYTES and returns their generation identity successfully. The success envelope then advertises a recipe_pull identity that every subsequent load rejects, making the returned recipe unpullable. Enforce the server-owned ceiling before publishing the generation.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,947 @@
...
+                shutil.rmtree(namespace)
+    except (OSError, RecipeArtifactError, TypeError):
+        return False
+    return True
+
+
+def recipe_pull_producers() -> frozenset[str]:
+    """Return public producers authorized to resolve immutable pull generations."""
+    return frozenset(
+        definition.producer_tool
```
Trecek, line 375: [warning] arch: Pull and recreation authorization collapse the route-level registry to producer names. Because multiple surfaces already share producer_tool="open_kitchen", divergent policy on those surfaces would be unenforceable: one permissive route authorizes every route with that producer. Preserve the surface in the artifact identity or enforce producer-level policy consistency in the registry.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -0,0 +1,947 @@
...
+            else:
+                receipt_handle = reservation.handle
+                decision = replace(decision, receipt_status="pending")
+
+    if decision.mode is RecipeDeliveryMode.ORDINARY_INLINE:
+        rendered = ordinary_rendered
+    elif decision.mode is RecipeDeliveryMode.ATTESTED_INLINE:
+        rendered = high_rendered
+    else:
+        envelope_bound_bytes = ordinary_limit * 4
```
Trecek: [critical] bugs: Envelope sizing multiplies the ordinary token limit by four even though this finalizer now treats UTF-8 byte count as the conservative token upper bound. A token-dense envelope can be emitted above the selected outer limit and clipped downstream, losing routing or pull metadata. Size the envelope with the same conservative admission bound.

src/autoskillit/execution/backends/_codex_config.py
```diff
@@ -24,14 +32,112 @@
...
+    evidence_version=1,
+    contract_digest=_CODEX_RECIPE_DELIVERY_CONTRACT_DIGEST,
+)
+CODEX_HISTORY_RETENTION_TOKEN_LIMIT: int = (
+    CODEX_RECIPE_DELIVERY_BUDGET.history_retention_token_limit
+)
+
+# A protected evidence identity is enabled only with its passing conformance
+# report. Writable rollout and trace formats are intentionally absent.
+SUPPORTED_CODEX_RECIPE_EVIDENCE_REGISTRY: dict[str, RecipeDeliveryEvidenceDef] = {}
```
Trecek: [warning] defense: SUPPORTED_CODEX_RECIPE_EVIDENCE_REGISTRY is mutable while _probe_cache snapshots its digest at import time. Runtime mutation can therefore change accepted evidence without changing probe-policy identity, unlike the immutable sibling surface registry. Expose an immutable mapping and derive the digest from that authority.

src/autoskillit/execution/backends/codex.py
```diff
@@ -639,7 +640,9 @@ def capabilities(self) -> BackendCapabilities:
             skill_sigil="$",
             session_dir_persistent=True,
             supports_model_invocation_gating=False,
-            effective_delivery_token_limit=10_000,
+            unnegotiated_tool_result_token_limit=10_000,
```
Trecek: [warning] cohesion: The Codex ordinary limit is independently hardcoded here even though CODEX_RECIPE_DELIVERY_BUDGET owns ordinary_omitted_result_token_limit and _codex_config repeats the literal in the contract digest. A one-sided retune can split model-visible policy from enforcement. Derive all three from one constant or budget field.

src/autoskillit/hooks/formatters/_fmt_recipe.py
```diff
@@ -253,6 +253,13 @@ def _fmt_open_kitchen(data: OpenKitchenResult, pipeline: bool) -> str:
 
 def _fmt_open_kitchen_plain_text(text: str, _pipeline: bool) -> str:
     """Format open_kitchen plain-text response (no recipe attached)."""
+    if (
+        text.startswith('{"recipe_delivery":')
```
Trecek: [warning] cohesion: The formatter recognizes attested delivery by duplicating all three wire-marker literals owned by server._recipe_delivery, and its test duplicates the same literals rather than exercising the producer. A framing change can make real attested responses get wrapped while formatter tests still pass. Share lightweight framing constants or a predicate.

src/autoskillit/server/tools/tools_recipe.py
```diff
@@ -669,16 +688,27 @@ async def get_recipe_section(
                 else None
             )
             bound_tokens = (
-                resolve_effective_delivery_bound(_backend_caps)
+                resolve_general_output_token_limit(_backend_caps)
                 if _backend_caps is not None
                 else 10_000
             )
             bound_bytes = bound_tokens * 4
```
Trecek: [critical] bugs: get_recipe_section sizes chunks at four bytes per ordinary-result token, and the changed response-budget guard uses the same four-byte estimate, so no later layer corrects this. A token-dense chunk can exceed the backend limit and be clipped despite being reported as complete. Use the same conservative byte-as-token bound as finalization.

tests/server/_helpers.py
```diff
@@ -11,6 +12,38 @@
 _HOOK_CONFIG_OVERLAY_RELPATH = (".autoskillit", "temp", ".hook_config_overlay.json")
 
 
+async def _resolve_recipe_content(result: dict[str, Any]) -> str:
+    """Return exact recipe content from either inline or pull delivery."""
+    inline_content = result.get("content")
```
Trecek, line 18: [warning] tests: _resolve_recipe_content returns inline content before validating success. A response such as {"success": false, "content": "stale"} would satisfy tests using this helper and mask a failed delivery surface. Assert success before accepting either representation.

tests/execution/backends/test_codex_recipe_delivery.py
```diff
@@ -0,0 +1,690 @@
...
+    )
+    rollout = project_dir / f"rollout-{thread_id}.jsonl"
+    rollout.write_text(
+        json.dumps({"type": "thread.started", "thread_id": thread_id}) + "\n",
+        encoding="utf-8",
+    )
+    return rollout
+
+
+def test_marker_validation_uses_one_bounded_file_descriptor(tmp_path: Path) -> None:
```
Trecek, line 188: [warning] tests: This test claims to verify one bounded file descriptor but never counts open calls and does not forbid read-through-another-open. An implementation that opens the marker twice, recreating the replacement race this test is meant to prevent, would still pass.

tests/execution/backends/test_codex_recipe_delivery_conformance.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+@pytest.mark.smoke
+@pytest.mark.timeout(900)
+@_skip_unless_live
+def test_live_codex_envelope_pull_and_next_request_retention(tmp_path: Path) -> None:
+    observation, transcript = _run_live_probe(tmp_path)
+    body_sha256 = observation.payload["body_sha256"]
+    assert isinstance(body_sha256, str) and body_sha256.startswith("sha256:")
+    assert observation.retained["body_sha256"] == body_sha256
```
Trecek, line 387: [warning] tests: The live pull oracle compares only body_sha256 metadata echoed by the envelope and pull result; it never validates or reconstructs the returned content. A pull returning empty or corrupted content with the copied digest field would still pass.

tests/server/test_tools_recipe_pull.py
```diff
@@ -1,1262 +1,725 @@
...
-        default_edges = build_routing_edges_by_step(
-            active_recipe_steps,
-            edge_extractor=_extract_routing_edges,
+    kwargs = generation.pull_identity()
+    kwargs.pop("pull_tool")
+    chunks: list[str] = []
+    expected_byte_start = 0
+    part = 0
+    while True:
+        response = json.loads(await get_recipe_section(section="content", part=part, **kwargs))
```
Trecek: [warning] tests: Every added successful pull test requests section="content", while this diff deletes coverage for named-step extraction, unknown sections, malformed recipe YAML, and oversized step continuation. The non-content branch of get_recipe_section is left without regression coverage.

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments — changes required. GitHub does not permit the PR author to submit a REQUEST_CHANGES review on their own PR, so this summary is posted as a comment.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in 9a5321974. Artifact path components now use injective UTF-8 byte escapes, and the regression test proves retiring a/b cannot remove a?b.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in d833202fc. Persistence now rejects canonical blobs above RECIPE_ARTIFACT_MAX_BLOB_BYTES before publishing an artifact identity.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in 66fe0129f. Authorization now fails closed if surfaces sharing a producer diverge on pull or recreation policy, with regression coverage.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in c80d810af. Envelope admission now uses the conservative ordinary-limit byte bound that matches the byte/token admission contract.

src/autoskillit/execution/backends/_codex_config.py
Trecek: Agreed — fixed in 95dde7f35. The Codex evidence registry is now immutable via MappingProxyType, and a regression test prevents digest authority from drifting at runtime.

src/autoskillit/execution/backends/codex.py
Trecek: Agreed — fixed in cfcc14e65. Codex digest and backend capabilities now share one ordinary-result token-limit authority, with parity coverage.

src/autoskillit/hooks/formatters/_fmt_recipe.py
Trecek: Agreed — fixed in 542adaf5a. A shared stdlib framing module now owns the wire markers and predicate, and the formatter test consumes producer-rendered attested output.

src/autoskillit/server/tools/tools_recipe.py
Trecek: Agreed — fixed in c0eda28db. Pull chunks now use bound_tokens as the conservative byte ceiling rather than multiplying by four; serialized responses are asserted within the ordinary limit.

tests/server/_helpers.py
Trecek: Agreed — fixed in bcd893835. The recipe helper now asserts success before inspecting content, and a regression test rejects false responses carrying stale content.

tests/execution/backends/test_codex_recipe_delivery.py
Trecek: Agreed — fixed in 2b8cfc81a. The descriptor test now overrides Path.open and asserts that the marker descriptor is opened exactly once.

tests/execution/backends/test_codex_recipe_delivery_conformance.py
Trecek: Agreed — fixed in d3602862f. The live probe follows every continuation part, reconstructs the content, recomputes SHA-256, and compares it with the envelope.

tests/server/test_tools_recipe_pull.py
Trecek: Agreed — fixed in de53a38d6. Public recipe-section coverage now includes named steps, unknown names, malformed YAML, and oversized continuation behavior.

Issue #4319, opened 2026-07-22
shell_capture_hook follows symlinked capture roots and cleanup can delete outside the project


Trecek:
## Problem

The Codex shell-capture mechanism does not establish a stable filesystem trust
anchor for `<cwd>/.autoskillit/temp/shell_capture`.

If the capture root or one of its parent components is a symlink, the harness can
create, truncate, and later read an artifact outside the intended working tree.
The registered SessionStart cleanup follows a symlinked capture root and can
unlink matching stale files in its external target. The separate cleanup helper
has the same root-containment defect.

This conflicts with the documented mechanism-owned `<cwd>` artifact location and
is broken existing behavior, so it is a remediation rather than a new feature.

## Evidence

- `docs/decisions/0006-output-containment.md:28-30,91-92` and
  `docs/safety/hooks.md:90-103` describe a mechanism-owned artifact beneath
  `<cwd>`.
- `src/autoskillit/hooks/shell_capture_hook.py:60-63` constructs the capture
  directory and file lexically from the hook-supplied `cwd`; `main()` only checks
  that `cwd` is a nonempty absolute string at lines 135-148.
- `src/autoskillit/hooks/shell_capture_hook.py:102-117` uses `mkdir -p`, ordinary
  shell redirection, and later pathname-based reads without rejecting a symlinked
  capture root or preserving directory/file identity across those operations.
- The registered cleanup in `src/autoskillit/hooks/session_start_hook.py:83-98`
  follows the capture root through `is_dir()` and `iterdir()`. It rejects
  symlinked entries but not a symlinked root before unlinking matching files.
- `src/autoskillit/hooks/_capture_cleanup.py:33-62,76-97` is a separate helper.
  It resolves a symlinked root and consequently treats the external target as
  the accepted containment root before unlinking matching stale files.
- `tests/hooks/test_shell_capture_conformance.py:308-325,417-435` covers
  symlinked entries beneath a real capture directory, but not a symlinked capture
  root or ancestor.
- A bounded reproduction confirmed that a symlinked capture directory caused a
  capture to be written outside the intended working tree.

## Expected Behavior

Capture creation, post-execution measurement/readback, and cleanup must operate
relative to a defined, verified physical trust anchor. Path-component
substitution must not redirect any capture operation outside that anchor. If a
secure capture cannot be established, the wrapped user command must not execute,
matching the current capture-setup failure contract.

## Acceptance Criteria

[...]

Issue #4327, opened 2026-07-22
Add enforceable storage quotas and exhaustion semantics for shell captures


Trecek:
## Problem

The managed command writes directly to its capture file, and AutoSkillit measures
the artifact only after execution. Stored output is unbounded by AutoSkillit until
filesystem capacity, an external OS quota, tool timeout, process termination, or
another environmental limit intervenes.

There is no per-artifact bound, aggregate reservation/accounting, minimum-free-
space policy, or truthful capture state for `ENOSPC`, `EDQUOT`, short writes, or
an exceeded application quota. A pre/post size check cannot enforce a concurrent
bound. These controls are explicitly absent, so this is an implementation issue.

The capture harness has been refactored since the issue was filed. The
`shell_capture_hook.py` shim now only rewrites the managed command into an
isolated-Python runner invocation (`python -I -B _capture_artifacts.py <capture_request>`);
real I/O, lifecycle, and capacity admission live in
`src/autoskillit/hooks/_capture_artifacts.py` and the `_capture/` subpackage. The
original `OutputBudgetConfig` reference and several `shell_capture_hook.py`
line numbers in this issue's evidence section are stale and have been corrected
below.

## Evidence (current)

- `src/autoskillit/hooks/shell_capture_hook.py:70-133` is the runner-dispatch
  shim: it sets `_HARNESS_SENTINEL`, `_RUNNER_BASENAME = "_capture_artifacts.py"`,
  resolves `_runner_path()` to the sibling runner, and emits the rewritten
  command via `_runner_argv(request)` + `_render_harness(...)` (`:123-152`). The
  original `:106-110` write-to-regular-file site no longer exists — the actual
  capture write path is `os.write(fd, ...)` inside
  `src/autoskillit/hooks/_capture/_runner.py:110-116` (`_write_all`) and
  `_capture/_capacity.py` admission.
- `src/autoskillit/hooks/_capture/_runner.py:201-261` — `run_capture(...)`
  policy read + `CaptureLifecycleStore.from_open_authorities(anchor, root, lock_wait=HOT_PATH_LOCK_WAIT, capacity=policy.capacity)` is where
  capacity admission gates a new artifact. Failure at `:245-261` constructs a
  `_capture_replay.failure_transport(...)` with
  `_capture_failure_policy.CAPACITY_FAILURE_REASONS` so capacity exhaustion is
  classified distinctly from generic `OSError`.
[...]

Issue #4320, opened 2026-07-22
Codex shell-capture artifacts lack a guaranteed cleanup lifecycle


Trecek:
## Problem

Codex native-shell commands retain their shell-capture artifacts, but Codex has
no installed, guaranteed lifecycle path that reaps them in a persistent
workspace. The only registered targeted sweep belongs to an interactive
SessionStart hook that Codex hook generation excludes, and the separate
`_capture_cleanup.py` helper is tested but not production-wired.

An interactive Claude SessionStart in the same directory or removal of an entire
worktree can delete captures incidentally; neither is a Codex lifecycle owner.
This conflicts with the documented retention-and-reclamation lifecycle and is
broken existing behavior, so it is a remediation.

## Evidence

- `src/autoskillit/hook_registry.py:204-213` registers the native `Bash` producer
  with `session_scope="any"` and Codex support.
- `src/autoskillit/hooks/shell_capture_hook.py:36-37,123-155` gates the producer
  to Codex sessions when enabled and rewrites native shell commands.
- `src/autoskillit/hooks/shell_capture_hook.py:102-119` creates an artifact and
  only reads it in both size branches; it does not delete it.
- `src/autoskillit/hook_registry.py:366-371` registers `session_start_hook.py` as
  `interactive_only`, while
  `src/autoskillit/execution/backends/_codex_hooks.py:42-55` skips that scope when
  generating Codex configuration.
- `src/autoskillit/hooks/session_start_hook.py:21-23` exits for headless sessions
  independently of registration; its inline capture sweep is at lines 83-100.
- Repository references to `sweep_stale_captures()` in
  `src/autoskillit/hooks/_capture_cleanup.py:65-100` are confined to direct calls
  from `tests/hooks/test_shell_capture_conformance.py:23-27,303-360`.
- `docs/safety/hooks.md:99-109` says artifacts are retained and stale captures
  older than one hour are reclaimed at SessionStart. The stated trigger is not
  installed for Codex.

## Expected Behavior

Codex must have an explicit production lifecycle owner that reaps an eligible
capture at the next supported Codex lifecycle trigger after the one-hour TTL.
The contract must cover persistent interactive and headless Codex use, or clearly
define which trigger owns cleanup for each.

## Dependency

Blocked by #4319 and #4321. Cleanup must not be wired until the shared
implementation has both a verified filesystem trust anchor and authoritative
writer-liveness handling; otherwise wiring would promote known deletion defects.

[...]

PR #4342, opened 2026-07-23
Rectify: Shell Capture Filesystem Trust Anchor

Trecek:
## Summary

Issue #4319 is not only a missing symlink check. The shell-capture lifecycle lacks a stable filesystem authority. `shell_capture_hook.py` accepts an absolute `cwd`, builds `<cwd>/.autoskillit/temp/shell_capture` by pathname, and the generated Bash harness creates, redirects to, measures, hashes, replays, and reports the artifact by reopening that path. `session_start_hook.py` then performs cleanup through a separate pathname-only implementation. `_capture_cleanup.py` is tested, but it is not used by the registered SessionStart hook and it trusts a resolved symlinked root.

Closes #4319

## Implementation Plan

Plan file: `/home/talon/projects/generic_automation_mcp/.autoskillit/temp/rectify/rectify_shell_capture_trust_anchor_2026-07-22_185019.md`


docs/safety/hooks.md
```diff
@@ -88,25 +88,33 @@ in `.hook_config.json` for future recipes that legitimately need these operation
...
+minimal isolated-Python runner. The runner opens the supplied `cwd` as a
+`ProjectAnchor` directory descriptor before deriving a physical path, then opens or
+creates `.autoskillit`, `temp`, and `shell_capture` relative to retained directory
+descriptors without following symlinks. It reads output policy through the verified
+`temp` descriptor and creates `shell_<uuid16>.log` exclusively with no-follow
+semantics. The child sends combined stdout+stderr through a pipe; the runner writes,
+measures, hashes, and replays bytes through owned descriptors. Only a bounded inline
+slice enters context: full content when small, otherwise head + provenance marker
+(bytes, sha256, verified path or `unavailable`) + tail. Set
+`output_budget.guard_enabled: false` to disable capture after verified policy loading;
```
Trecek: [warning] cohesion: [naming consistency] The docs use `output_budget.guard_enabled` and `output_budget.shell_max_inline_bytes`, while the runner in this diff reads `output_budget_policy.disabled` and `output_budget_policy.shell_max_inline_bytes`. That leaves two names for the same policy surface and points maintainers at the wrong config shape.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,893 @@
...
+            raise CaptureSetupError(f"cannot create capture path component: {name}") from exc
+        try:
+            fd = os.open(name, _DIRECTORY_FLAGS, dir_fd=parent_fd)
+        except OSError as exc:
+            raise CaptureSetupError(f"cannot open created capture component: {name}") from exc
+    except OSError as exc:
+        raise CaptureSetupError(f"unsafe capture path component: {name}") from exc
+
+    try:
+        if not stat.S_ISDIR(os.fstat(fd).st_mode):
```
Trecek: [warning] defense: [validation at construction] Existing capture path components are accepted after only an S_ISDIR check. A preexisting group/world-writable .autoskillit/temp/shell_capture directory can let another local actor rename or replace artifacts during capture, weakening the descriptor-owned artifact guarantee. Validate owner/mode when opening existing components.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,893 @@
...
+    artifact_path: str | None,
+    inline_bytes: int,
+) -> None:
+    if result.total_bytes <= inline_bytes:
+        payload = result.inline
+    else:
+        prefix = render_capture_marker(_capture_event("SHELL_OUTPUT_CAPTURED", "input rewrite"))
+        path = artifact_path if artifact_path is not None else "unavailable"
+        marker = (
+            f"\n{prefix} full output {result.total_bytes} bytes -> {path} "
```
Trecek, line 703: [warning] defense: [validation at construction] artifact_path is derived from cwd and inserted raw into the provenance marker, while open_project_anchor only rejects NUL. Newlines, ']', or other marker-control characters in the project path can split or spoof the marker; encode/sanitize the path or reject those characters when constructing the anchor.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,893 @@
...
+    if running:
+        try:
+            process.terminate()
+        except _CAPTURE_RUNTIME_ERRORS:
+            try:
+                process.kill()
+            except _CAPTURE_RUNTIME_ERRORS:
+                pass
+    try:
+        return _normalized_returncode(process.wait())
```
Trecek: [warning] bugs: [process_lifecycle] _settle_failed_capture() sends SIGTERM and then waits without a timeout. If the child ignores or traps SIGTERM, capture failure handling can hang indefinitely instead of escalating to kill().

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,893 @@
...
+            failure_stage = "capture process wait"
+            returncode = _normalized_returncode(process.wait())
+            if result.write_error is not None:
+                return _capture_failure_return("capture artifact write failed", returncode)
+            failure_stage = "capture marker verification"
+            artifact_path = current_artifact_path_if_bound(anchor, root, artifact)
+            failure_stage = "capture replay emission"
+            _emit_capture(result, artifact_path, policy.inline_bytes)
+            return returncode
+        except _CAPTURE_RUNTIME_ERRORS:
```
Trecek: [warning] defense: [error context preservation] The broad _CAPTURE_RUNTIME_ERRORS handler discards the caught exception and emits only the stage name. Capture failures become generic messages like 'capture readback failed' without exception class or message, making root-cause diagnosis much harder. Include sanitized exception context in the failure marker.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,893 @@
...
+        fd=fd,
+        device=value.st_dev,
+        inode=value.st_ino,
+        mode=value.st_mode,
+        nlink=value.st_nlink,
+        mtime=value.st_mtime,
+    )
+
+
+def sweep_stale_captures(
```
Trecek: [warning] cohesion: [naming consistency] `sweep_stale_captures` now only classifies stale captures and always retains them, but the public function/export/call sites still use sweep/delete terminology. The cleanup API name should match the retention/classification behavior.

src/autoskillit/hooks/shell_capture_hook.py
```diff
@@ -1,123 +1,94 @@
...
-from _hook_settings import read_merged_hook_config  # type: ignore[import-not-found]  # noqa: E402
-from _policy_event import (  # type: ignore[import-not-found]  # noqa: E402
-    PolicyEvent,
-    render_capture_marker,
-)
-
 _HARNESS_SENTINEL = "# autoskillit-shell-capture v1"
-_DEFAULT_SHELL_MAX_INLINE_BYTES = 12_000
-_CAPTURE_SUBDIR = ".autoskillit/temp/shell_capture"
+_CAPTURE_ID_RE = re.compile(r"^[0-9a-f]{16}$")
```
Trecek: [warning] cohesion: [feature locality] The wrapper duplicates runner-owned transport constraints (`_CAPTURE_ID_RE`, `_MAX_COMMAND_BYTES`) that `_capture_artifacts.py` also validates. With execution delegated to the runner, the capture-id and command-size contract should have one stdlib-only home to avoid drift.

src/autoskillit/hooks/shell_capture_hook.py
```diff
@@ -1,123 +1,94 @@
...
+    if not _CAPTURE_ID_RE.fullmatch(capture_id):
+        raise ValueError("capture_id must contain exactly 16 lowercase hex characters")
+    try:
+        command_bytes = command.encode("utf-8")
+    except UnicodeEncodeError:
+        command_bytes = b""
+    if "\x00" in command or not command_bytes or len(command_bytes) > _MAX_COMMAND_BYTES:
+        argv = [sys.executable, "-I", str(_runner_path()), "reject", capture_id]
+    else:
+        encoded = base64.b64encode(command_bytes).decode("ascii")
```
Trecek, line 90: [critical] defense: [late validation] The hook base64-encodes the original shell command and returns allow with only a generic Python runner in updatedInput. Any policy check that evaluates the rewritten command can no longer inspect dangerous shell text before execution; keep the original command policy-visible or run equivalent validation before returning allow.

Trecek (commented):

tests/arch/test_shell_capture_trust_anchor.py
```diff
@@ -0,0 +1,177 @@
...
+    source = (
+        "capture_root = CAPTURE_PATH_COMPONENTS\n"
+        "output_path = capture_root\n"
+        'harness = f"cmd > {output_path}/shell.log"\n'
+    )
+    assert _capture_path_redirections(source)
+    assert not _capture_path_redirections('harness = "cmd > ordinary.log"')
+
+
+def test_project_temp_cleanup_debt_is_narrow_and_owned() -> None:
```
Trecek: [warning] tests: [weak assertions] This debt test only checks a hard-coded test-local dict against another hard-coded literal plus file existence, so it cannot detect new or removed pathname cleanup debt in production. Derive the debt list from an actual scan or remove the self-referential assertion.

tests/execution/backends/test_hook_strength_matrix.py
```diff
@@ -27,6 +27,19 @@
 )
 
 
+@pytest.fixture(scope="module", autouse=True)
```
Trecek: [warning] tests: [xdist safety] This autouse fixture skips the entire strength-matrix module whenever test_hook_deny_efficacy_probe.py is collected. In normal full-suite/xdist runs that file is collected, so these assertions silently stop running instead of being ordered after the probe output.

tests/hooks/test_shell_capture_conformance.py
```diff
@@ -260,6 +275,209 @@ def test_capture_dir_uncreatable_fail_stops(tmp_path: Path) -> None:
...
+        (external / ".hook_config.json").write_text(
+            json.dumps({"output_budget_policy": {"disabled": True}})
+        )
+
+    wrapper = _main_generated_wrapper(
+        "printf ran > command_ran",
+        project,
+        monkeypatch,
+    )
+    completed = subprocess.run(
```
Trecek, line 314: [warning] tests: [xdist safety] The new generated-wrapper integration subprocesses run without a timeout. If the capture runner or pipe drain hangs, an xdist worker can hang indefinitely; add bounded timeouts to these subprocess.run calls.

tests/hooks/test_shell_capture_conformance.py
```diff
@@ -337,9 +554,9 @@ def test_sweep_filename_allowlist(tmp_path: Path) -> None:
     invalid_ext = capture / "evil.sh"
     invalid_ext.write_text("x")
 
-    deleted = sweep_stale_captures(capture, max_age_seconds=0)
-    assert deleted == 1
-    assert not valid.exists()
+    deleted = sweep_stale_captures(tmp_path, max_age_seconds=0)
```
Trecek: [warning] tests: [weak assertions] After changing sweep to retain files and return 0, this filename-allowlist test no longer exercises the allowlist; it would pass if sweep ignored names entirely. Assert classifier decisions for valid versus invalid names instead.

tests/hooks/test_shell_capture_conformance.py
```diff
@@ -357,89 +574,13 @@ def test_sweep_stale_captures_uses_wall_clock_not_directory_mtime(tmp_path: Path
...
-    assert _is_safe_capture_file(path, capture) is expect_safe
-    if expect_safe:
-        try:
-            resolve_contained_path(path, capture)
-        except ContainmentError as exc:
-            pytest.fail(f"core containment rejected a valid capture file: {exc}")
-    else:
-        with pytest.raises(ContainmentError):
-            resolve_contained_path(path, capture)
+    deleted = sweep_stale_captures(tmp_path, max_age_seconds=100)
```
Trecek: [warning] tests: [weak assertions] The wall-clock mtime test now only asserts deleted == 0 and that the stale file still exists, so it no longer proves wall-clock age is consulted. Move the age-threshold assertion to the classifier level or rename the test to retained-artifact behavior.

tests/hooks/test_shell_capture_hook.py
```diff
@@ -68,34 +90,58 @@ def test_rewrites_on_turn_id_payload(monkeypatch):
...
+
+    assert argv[0] == sys.executable
+    assert argv[1] == "-I"
+    assert Path(argv[2]).name == "_capture_artifacts.py"
+    assert argv[3] == "run"
+    assert base64.b64decode(argv[4], validate=True).decode() == "echo hello"
+    assert argv[5] == str(tmp_path)
+    assert re.fullmatch(r"[0-9a-f]{16}", argv[6])
+
+    completed = subprocess.run(
```
Trecek, line 136: [warning] tests: [xdist safety] This new runner subprocess has no timeout. A regression in the generated wrapper can hang the worker indefinitely; pass a bounded timeout to subprocess.run.

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments. GitHub does not permit the PR author to submit a request-changes review on their own PR, so this was submitted as a comment review; verdict remains changes_requested.

docs/safety/hooks.md
Trecek: Agreed — fixed in c6244be3cd06f50a10faf47f4a5aad942281c0fa. The names represent two layers of one intentional bridge, but docs/safety/hooks.md:100-101 does not identify them as public versus internal. OutputBudgetConfig exposes guard_enabled and shell_max_inline_bytes at src/autoskillit/config/_config_dataclasses.py:345,357-358; _output_budget_policy_hook_payload() in src/autoskillit/server/tools/tools_kitchen.py:245-255 inverts guard_enabled into disabled, and _write_hook_config() serializes that payload under output_budget_policy at line 275. read_capture_policy() then reads the internal section at src/autoskillit/hooks/_capture_artifacts.py:386,404. Clarifying this mapping is a useful, architecture-safe documentation improvement while retaining output_budget.* as the user-facing configuration.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in a43f052a015ebaedbaaaec375c111ca4b64d802d. `_open_directory_component()` lines 230-254 creates missing components with mode 0700 at line 237, but for an existing component line 250 checks only `S_ISDIR`; it never checks `st_uid` or `_UNTRUSTED_WRITE_BITS` (defined at line 75). Because `open_capture_root()` applies this to every `CAPTURE_PATH_COMPONENTS` directory, owner/write-mode validation is a concrete trust-anchor improvement and violates no catalog constraint.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in 7644b449613288eaefb0112ee6a42d851c2dc3ab. `open_project_anchor()` lines 258-276 accepts any nonempty absolute string without NUL and retains that spelling; `current_artifact_path_if_bound()` resolves `anchor.supplied_path` at line 446, and `_emit_capture()` line 639 interpolates the resulting path directly into a bracketed marker. Unlike `_emit_failure()` line 616, this path receives no control-character sanitization, so newline or `]` can corrupt marker framing. Sanitizing/encoding it is architecture-safe.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in e23886742ed61cae5fdf2aa0852bb522de1a59e0. `_settle_failed_capture()` calls `process.terminate()` at line 683 and then `process.wait()` with no timeout at line 690. `kill()` is attempted only when `terminate()` itself raises (lines 684-688), not when a child survives SIGTERM. A bounded wait followed by kill and a second bounded reap directly prevents the documented failure-path hang and conflicts with no catalog rule.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in dbbbe5d909492854cf3a5e10799d7ec4d5eb0d6d. `run_capture()` tracks only a stage string at lines 726-735; its line-738 handler does not bind the exception and line 741 emits only `"{failure_stage} failed"`. `_emit_failure()` already normalizes whitespace and bounds detail at line 616, so adding safely sanitized exception type/context would materially improve diagnostics without violating an architectural constraint.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in 68deb53712d4f777c04ce858dcfb56eb1afd0f98. `sweep_stale_captures()` lines 817-855 explicitly says validated stale files are retained and the deleted count is always zero (lines 821-826), yet its public name and session-start call retain cleanup/sweep terminology. Renaming the function/export/call sites to classification/retention terminology is a concrete cohesion improvement; the catalog's hook-rename constraint concerns script basenames, not this function.

src/autoskillit/hooks/shell_capture_hook.py
Trecek: Agreed — fixed in b4305b126fc38f9525d4028caf5618aa3056ac75. `shell_capture_hook.py` defines `_CAPTURE_ID_RE` and `_MAX_COMMAND_BYTES` at lines 25-26 and applies them at lines 68-74; `_capture_artifacts.py` independently defines the same contract at lines 64 and 69 and applies it at runner ingress. A small stdlib-only shared contract module can single-source the values while preserving validation at both trust boundaries, matching the hooks package's standalone/stdlib architecture and violating no catalog constraint.

src/autoskillit/hooks/shell_capture_hook.py
Trecek: Agreed — fixed in 761b132d46c8f91799500ed82b45264d71ff2ff3. `_build_harness()` base64-encodes the original command at line 77, while `main()` returns explicit `permissionDecision: allow` and only the generic runner command in `updatedInput` at lines 120-121. Codex launches peer matching hooks concurrently, so they still inspect the original payload, but history commit 3d7829d14a379caf999a8fcf03cdbaa33d3d8850 records the native exec-policy evaluating the generated harness; that downstream policy therefore loses the original shell text. Preserving or equivalently validating the command before allow is a real defense improvement and violates no catalog constraint.

tests/arch/test_shell_capture_trust_anchor.py
Trecek: Agreed — fixed in 1199bb73810d41bfcb5d2abce6b334dc0562b19c. In test_project_temp_cleanup_debt_is_narrow_and_owned, lines 165-171 compare set(_PROJECT_TEMP_CLEANUP_DEBT) with a second hard-coded copy of the same paths, while lines 172-177 only verify that those listed paths and metadata fields exist. Nothing in the test discovers pathname cleanup sites from production, so adding or removing production debt is invisible unless both literals are manually updated. History confirms commit 94ea1c553ff25059395fa16f07fd0f47ae216f83 introduced the manifest and duplicate assertion together, and f3b5a05ef5eec9c9c91d0614036ea03dbdbf7aba later extended both in lockstep.

tests/execution/backends/test_hook_strength_matrix.py
Trecek: Agreed — fixed in f55812dae29c16aabe5f250242c25e64b277adc1. _require_completed_probe_session at lines 30-40 unconditionally module-skips when test_hook_deny_efficacy_probe.py appears in session collection, so lines 50 onward never execute in a normal combined collection. The producer cannot make the matrix available earlier: tests/execution/backends/conftest.py:126-140 serializes hook_deny_strength_matrix.json only in pytest_sessionfinish, after tests have run. Commit 326b94f54928b4c531aa6ec18459f7f433e11396 added this skip, whereas the matrix assertions originated as merge-blocking checks in earlier history.

tests/hooks/test_shell_capture_conformance.py
Trecek: Agreed — fixed in bf624ae33ad4b4daa604b0ed1314af45b8f54261. The generated-wrapper integration call at lines 309-315 invokes subprocess.run without timeout=. The same newly added wrapper tests repeat the unbounded pattern at lines 335-340, 379-384, 414-419, and 462-467, so a wrapper or pipe-drain regression can indefinitely occupy a pytest/xdist worker. Commit 94ea1c553ff25059395fa16f07fd0f47ae216f83 introduced all five calls without timeouts; adding bounded timeouts violates none of the Architectural Constraint Catalog rules.

tests/hooks/test_shell_capture_conformance.py
Trecek: Agreed — fixed in 16d9808f11f582bf27016ed10e0a37141738a783. test_sweep_filename_allowlist at lines 545-561 now observes the same retained-file and deleted == 0 outcome for the valid, legacy-short, and invalid-extension names, so it cannot prove that the filename gate ran. The actual gate is _open_stale_candidate in src/autoskillit/hooks/_capture_artifacts.py:769-815, specifically _CAPTURE_FILENAME_RE.fullmatch at lines 775-776. History shows commit 94ea1c553ff25059395fa16f07fd0f47ae216f83 replaced the formerly differentiating deletion assertions with uniform retention; directly asserting classifier decisions would restore coverage without violating a catalog constraint.

tests/hooks/test_shell_capture_conformance.py
Trecek: Agreed — fixed in 606522c0a542fe943b3eefb98ac0055a384f9473. test_sweep_stale_captures_uses_wall_clock_not_directory_mtime at lines 564-579 only checks the API's intentional always-zero/retain behavior, which is identical whether the backdated mtime is consulted or ignored. The age decision actually occurs in _open_stale_candidate at src/autoskillit/hooks/_capture_artifacts.py:773-786, with sweep_stale_captures computing time.time() - max_age_seconds at line 834. Commit 94ea1c553ff25059395fa16f07fd0f47ae216f83 removed the prior observable deletion assertion, so classifier-level threshold coverage or an accurate retained-artifact test name is a concrete, architecture-safe improvement.

tests/hooks/test_shell_capture_hook.py
Trecek: Agreed — fixed in a3c99bc3c13dad539461dbbf1df868ef738b3828. test_verified_disabled_policy_is_runner_owned executes the generated Bash wrapper with subprocess.run at lines 133-139 using capture_output=True but no timeout=. A wrapper or child-process hang therefore has no test-side bound. Git blame and the required path history identify commit 94ea1c553ff25059395fa16f07fd0f47ae216f83 as introducing this call; commit 326b94f54928b4c531aa6ec18459f7f433e11396 later strengthened argv assertions but left execution unbounded. Adding timeout= is architecture-safe under the catalog.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,893 @@
...
+class FileIdentity:
+    device: int
+    inode: int
+
+    @classmethod
+    def from_stat(cls, value: os.stat_result) -> FileIdentity:
+        return cls(device=value.st_dev, inode=value.st_ino)
+
+
+@dataclass(slots=True)
```
Trecek: [warning] defense: The exported authority types ProjectAnchor, CaptureRoot, and CaptureArtifact are directly constructible mutable records of raw file descriptors, names, and identities. Downstream functions treat instances as proof that descriptor and identity validation already occurred, but callers can construct mismatched or invalid authority objects. Make construction opaque/factory-only or validate invariants at construction.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,893 @@
...
+        original_cwd_fd = os.open(".", _DIRECTORY_FLAGS & ~getattr(os, "O_NOFOLLOW", 0))
+    except OSError as exc:
+        raise CaptureSetupError("cannot preserve runner cwd") from exc
+
+    process: subprocess.Popen[bytes] | None = None
+    restore_error: OSError | None = None
+    try:
+        os.fchdir(anchor.fd)
+        process = subprocess.Popen(
+            [bash_path, "-c", _wrap_user_command(command), "autoskillit-capture"],
```
Trecek: [critical] bugs: Passing "autoskillit-capture" as Bash's command-name argument changes $0 observed by every wrapped command; native bash -c commands ordinarily see the Bash executable as $0. Commands that inspect or dispatch on $0 therefore produce different results under capture. Preserve the native Bash argv[0] semantics.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,893 @@
...
+
+    while True:
+        chunk = stream.read(_DRAIN_CHUNK_BYTES)
+        if not chunk:
+            break
+        total += len(chunk)
+        digest.update(chunk)
+        if write_error is None:
+            try:
+                _write_all(artifact_writer_fd, chunk)
```
Trecek, line 605: [critical] bugs: The artifact remains writable at a discoverable project path while the user command runs, while the digest is computed only from pipe bytes. The command can truncate or modify shell_*.log without changing its inode, after which path verification still emits complete=true for corrupted content. Verify the artifact's bytes against the captured digest after the child exits before publishing its path.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,893 @@
...
+    safe_detail = " ".join(detail.split())[:240]
+    sys.stderr.write(f"{prefix} {safe_detail}]\n")
+
+
+def _capture_failure_return(detail: str, returncode: int | None) -> int:
+    try:
+        _emit_failure(detail)
+    except _CAPTURE_RUNTIME_ERRORS:
+        return _CAPTURE_FAILURE_RETURN_CODE
+    return returncode if returncode is not None else _CAPTURE_FAILURE_RETURN_CODE
```
Trecek: [critical] defense: Capture-mechanism failures reuse the user command's return code, so a successful command still exits 0 when artifact writing, readback, verification, or replay fails. The failure marker then replaces the command output, producing a successful tool result with lost or inaccessible output. Report capture integrity separately or fail the rewritten command when capture cannot complete.

tests/arch/test_shell_capture_trust_anchor.py
```diff
@@ -0,0 +1,177 @@
...
+        for match in _REDIRECTION_RE.finditer(rendered):
+            target = match.group("target")
+            if _CAPTURE_PATH_TERM_RE.search(target) or any(
+                re.search(rf"\b{re.escape(name)}\b", target) for name in capture_path_names
+            ):
+                violations.append(match.group(0).strip())
+    return violations
+
+
+def test_session_start_calls_canonical_capture_cleanup() -> None:
```
Trecek: [warning] tests: The architecture guard verifies canonical wiring and several prohibited operations with raw substring checks. A call in dead code or a comment can satisfy the cleanup assertion, while aliases or getattr-based pathname operations can evade the forbidden-operation checks; use AST-based call and import analysis for this invariant.

tests/hooks/test_capture_artifacts.py
```diff
@@ -0,0 +1,1171 @@
...
+    (temp_dir / ".hook_config.json").symlink_to(external_config)
+
+    anchor = open_project_anchor(str(project))
+    try:
+        assert read_capture_policy(anchor) == CapturePolicy()
+    finally:
+        anchor.close()
+
+
+def test_verified_policy_merges_overlay_and_bounds_inline_bytes(tmp_path: Path) -> None:
```
Trecek: [warning] tests: The test claims to verify bounded inline policy values but exercises only the valid in-range value 31. It would not catch removal of the 1,000,000-byte clamp or regressions for zero, negative, bool, and non-integer values; parameterize boundary and fallback cases.

Trecek (commented):

Trecek (commented):
AutoSkillit review: blocking findings detected. Changes are requested; see the inline comments.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in dc07baa0d47416a6a7db1299f9c6d6538890cb4c. ProjectAnchor, CaptureRoot, and CaptureArtifact now require module-factory construction and are frozen slotted dataclasses; internal close() methods retain idempotent descriptor release. Regression coverage rejects direct construction and external mutation.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in c382f446c9a76f475bdecfc18a8c4d50eb9ec0d9. _spawn_bash() no longer supplies a synthetic Bash command-name argument, so wrapped commands observe the native bash -c $0 value. A regression test compares the captured value with the trusted Bash path.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in de7d8e013351e58144f2ead1e8a16503d9b7c061. _verify_capture_artifact() now checks retained-descriptor identity, type, link count, exact size, and a pread SHA-256 digest before path verification or replay publication. A tampering regression test proves corrupted same-inode bytes cannot be published.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in 9c5ebf4ee5986e57969d5a7e699e9539550ccc59. Capture-mechanism failures now convert a successful or unavailable user status to the dedicated nonzero capture-failure code while preserving an already-nonzero user status. Readback, digest, write, verification, and replay-failure tests now require failure status.

tests/arch/test_shell_capture_trust_anchor.py
Trecek: Agreed — fixed in a9e01aa5b3e844341a49925d3ab31d717625aac4. The guard now parses imports and calls with AST, detects aliased and getattr-based unlink/rmtree calls, scans shell mkdir commands as string nodes, and requires the canonical classifier call directly in main() rather than in a dead nested branch.

tests/hooks/test_capture_artifacts.py
Trecek: Agreed — fixed in f7ef69118d8ca822dc5687d7745c07de5a12d525. The policy test now covers minimum and ordinary valid values, the exact maximum, above-maximum clamping, and fallback behavior for zero, negative, boolean, and non-integer inputs.

docs/decisions/0006-output-containment.md
```diff
@@ -79,6 +81,16 @@ hook can be retired in favor of that mechanism.
...
+7. A supplied symlink spelling of `cwd` is accepted only by opening it first as
+   the `ProjectAnchor`; `.autoskillit`, `temp`, and `shell_capture` symlinks are
+   rejected. Physical path strings are display hints, not filesystem authority.
+8. Same-user code inside the command can rename a verified directory entry after
+   it is opened. Output, hashing, and replay remain fd-bound; if the live pathname
+   chain no longer matches, the marker reports its path as `unavailable`.
+9. Portable Linux/macOS Python exposes descriptor-relative unlink but no
+   expected-inode conditional unlink. SessionStart therefore retains stale
+   candidates rather than making a security claim across a validation/deletion
+   race. Artifact quota and lifecycle reclamation remain follow-up work.
```
Trecek: [warning] slop: Artifact quota and lifecycle reclamation are deferred as follow-up work without a tracking reference, despite Codex artifacts having no automatic deletion guarantee. Link a concrete issue so this unbounded-retention debt is not orphaned.

tests/hooks/test_shell_capture_conformance.py
```diff
@@ -120,10 +124,15 @@ def _run_wrapped(command: str, tmp_path: Path) -> subprocess.CompletedProcess[by
...
-# nested_wrap wraps a harness-of-"echo hi" and feeds it back through the
-# harness a second time, proving the sentinel/idempotency path is also
-# byte-safe under double-wrapping.
-_CORPUS[-1] = ("nested_wrap", _build_harness(_NESTED_WRAP_INNER, "/tmp", _INLINE_BYTES))
+def _main_generated_wrapper(command: str, cwd: Path, monkeypatch: pytest.MonkeyPatch) -> str:
+    event = {"cwd": str(cwd), "tool_input": {"command": command}}
+    monkeypatch.setenv("AUTOSKILLIT_AGENT_BACKEND", "codex")
+    monkeypatch.setattr("sys.stdin", io.StringIO(json.dumps(event)))
+    output = io.StringIO()
+    with pytest.raises(SystemExit), redirect_stdout(output):
```
Trecek: [warning] tests: The generated-wrapper helper accepts any SystemExit code. All integration cases can remain green if the hook emits a payload but exits nonzero, which may cause the host to discard the rewrite. Capture the exception and assert code == 0.

tests/hooks/test_shell_capture_hook.py
```diff
@@ -115,25 +161,64 @@ def test_non_string_command_fails_open(monkeypatch):
...
-    import re
+    assert argv[3] == "reject"
+    assert command not in harness
 
-    marker_matches = re.findall(r"printf '\\n%s\\n' (.+)", command)
-    assert marker_matches, "expected a printf-embedded capture marker in the harness"
-    marker = marker_matches[0]
 
-    assert "`" not in marker
+def test_arg_max_exhaustion_builds_nonexecuting_rejection(
```
Trecek, line 191: [warning] tests: The ARG_MAX test forces SC_ARG_MAX to 1, so any nonempty argv is rejected and the test never exercises the new environment-byte, pointer-byte, or headroom accounting. Use a boundary where argv alone fits but the complete exec footprint does not.

Trecek (commented):

Trecek (commented):

docs/decisions/0006-output-containment.md
Trecek: Agreed — fixed in 7b8321b7f85c0c29d4e2f53aa4be105f262aaea5. The ADR now links quota debt to #4327 and lifecycle reclamation to #4320, keeping both deferred obligations actionable.

tests/hooks/test_shell_capture_conformance.py
Trecek: Agreed — fixed in 3b9e1f0b2e2434f9870cbda854bc2439d49fabd4. The generated-wrapper helper now captures SystemExit and asserts code == 0 before parsing output, preventing a nonzero hook exit from producing a false-green integration result.

tests/hooks/test_shell_capture_hook.py
Trecek: Agreed — fixed in c5a2503a06e9230cc3c2bc12dd94fb6410c4f02a. The test now proves both raw argv vectors fit, verifies the exact environment and pointer contribution to _exec_footprint, and sets SC_ARG_MAX one byte below the complete footprint plus headroom before asserting rejection.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,971 @@
...
+def _read_bounded_file_at(directory_fd: int, name: str) -> dict:
+    try:
+        fd = os.open(name, _READ_FLAGS, dir_fd=directory_fd)
+    except FileNotFoundError:
+        return {}
+    except OSError:
+        return {}
+    try:
+        value = os.fstat(fd)
+        if not stat.S_ISREG(value.st_mode) or value.st_nlink != 1 or value.st_mode & stat.S_IWOTH:
```
Trecek, line 390: [warning] defense: Regular-file trust checks reject only world-writable files, unlike directory and executable checks that use `_UNTRUSTED_WRITE_BITS`. A group-writable policy file in a traversable verified temp directory can be modified by another group member to disable capture or change its bound; the same world-only check also permits group-writable artifacts during publication/classification. Validate owner and both group/world write bits consistently.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,971 @@
...
+    tail_limit = inline_bytes - head_limit
+    total = 0
+    digest = hashlib.sha256()
+    inline = bytearray()
+    head = bytearray()
+    tail = bytearray()
+    write_error: OSError | None = None
+
+    while True:
+        chunk = stream.read(_DRAIN_CHUNK_BYTES)
```
Trecek, line 598: [warning] bugs: The pipe drain waits for EOF with no post-shell cutoff. A disowned or otherwise detached descendant that retains inherited stdout/stderr keeps the pipe open after Bash exits, so the rewritten tool call hangs until that descendant closes the descriptors, contradicting the documented best-effort detached-output behavior.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -0,0 +1,971 @@
...
+    anchor = open_project_anchor(cwd)
+    root: CaptureRoot | None = None
+    artifact: CaptureArtifact | None = None
+    artifact_writer_fd = -1
+    process: subprocess.Popen[bytes] | None = None
+    try:
+        policy = read_capture_policy(anchor)
+        bash_path = _resolve_bash()
+        if policy.disabled:
+            process = _spawn_bash(anchor, bash_path, command, capture_output=False)
```
Trecek, line 791: [warning] bugs: A disabled capture policy still executes the command through `_spawn_bash`, which always adds the EXIT trap that waits for background jobs. Disabling the guard previously left the command unchanged; commands such as `sleep 60 &` now block instead of returning promptly even though capture is disabled.

Trecek (commented):

src/autoskillit/hooks/shell_capture_hook.py
```diff
@@ -1,123 +1,108 @@
...
+            "-I",
+            str(_runner_path()),
+            "run",
+            encoded,
+            cwd,
+            capture_id,
+        ]
+        harness = _render_harness(argv, policy_command=command)
+        outer_argv = ["bash", "-c", harness]
+        if not _fits_arg_max(argv) or not _fits_arg_max(outer_argv):
```
Trecek, line 102: [warning] bugs: The ARG_MAX preflight checks only aggregate exec footprint and misses Linux's per-argument MAX_ARG_STRLEN limit. A valid command near the 64 KiB transport limit expands to roughly 150 KiB when duplicated as base64 and policy text in the single `bash -c` argument, so `_fits_arg_max` accepts it but exec fails with E2BIG before the rejection runner can emit a controlled failure.

tests/arch/test_shell_capture_trust_anchor.py
```diff
@@ -0,0 +1,295 @@
...
+    source = (
+        "capture_root = CAPTURE_PATH_COMPONENTS\n"
+        "output_path = capture_root\n"
+        'harness = f"cmd > {output_path}/shell.log"\n'
+    )
+    assert _capture_path_redirections(source)
+    assert not _capture_path_redirections('harness = "cmd > ordinary.log"')
+
+
+def test_project_temp_cleanup_debt_entries_exist_and_are_owned() -> None:
```
Trecek, line 289: [warning] tests: This cleanup-debt guard remains self-referential: it iterates a hard-coded test-local mapping and verifies only that each named file and metadata string exists. It cannot detect newly introduced pathname cleanup debt or removal of the listed cleanup behavior while the file remains, so the architectural invariant can regress without failing. Derive the debt set from production AST scanning and compare it to the allowlist.

Trecek (commented):

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in d7fd02262. Regular-file trust now requires current-euid ownership, one link, and no group/world write bits across policy reads, artifact publication and integrity verification, and stale classification. Regression coverage exercises unsafe modes and ownership.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in d4f61404e and 352cbb0d5. The drain now polls the pipe, detects shell exit, consumes buffered output, and enforces a bounded post-shell cutoff. The regression test keeps a descendant writer open and proves bounded return.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in b041977ee. Disabled capture now passes the original command directly to Bash instead of installing the capture-only EXIT wait trap, restoring native background-job behavior. A regression test asserts the unwrapped Bash command.

src/autoskillit/hooks/shell_capture_hook.py
Trecek: Agreed — fixed in 6617c01f3. Exec preflight now enforces the Linux per-string ceiling derived from page size as well as aggregate ARG_MAX accounting, including the terminating NUL. Regression coverage proves an oversized single bash -c argument routes to the controlled rejection runner.

tests/arch/test_shell_capture_trust_anchor.py
Trecek: Agreed — fixed in 71edf3a08. The guard now discovers pathname cleanup debt from production Python ASTs and shell mutations, then checks both unregistered discoveries and stale allowlist entries. It also self-tests the detector against synthetic debt and catalogs the existing fleet cleanup path found by the scan.

Issue #4341, opened 2026-07-23
Preserve JSON validity when paging list-valued recipe sections


Trecek:
## Problem

`get_recipe_section` serializes the list-valued `errors`, `warnings`, and
`ingredients_table` sections as JSON and then applies character-offset
chunking. When a section exceeds the delivery bound, a returned `content`
fragment can split a JSON token or list element and cannot be parsed
independently.

This is residual class 1 from #4314, verified on `develop` at `ac8f653a0`
after PR #4330.

## Current behavior

- `_bounded_recipe_section_response` in
  `src/autoskillit/server/tools/tools_recipe.py` accepts a string and slices it
  at character offsets.
- `get_recipe_section` passes JSON-serialized strings for `errors`, `warnings`,
  and `ingredients_table`.
- Existing continuation tests cover reconstructed raw/YAML content, not the
  parsing contract for these list-valued sections.
- Current tool and sous-chef guidance does not explicitly define
  concatenate-before-parse behavior.

## Required outcome

- Define one explicit, machine-consumable pagination contract for list-valued
  recipe sections.
- Preserve list-element boundaries so every successful page is independently
  valid JSON and retains its original element values and ordering.
- Keep every serialized tool response within the effective delivery byte bound,
  including multibyte data and a single element that cannot fit on an ordinary
  page.
- Return an explicit bounded failure or continuation representation for an
  individually oversized element; do not silently drop it or return an invalid
  fragment.
- Preserve the existing continuation metadata contract (`part`, `total_parts`,
  `has_more`, and `next_part`) and the behavior of string-valued sections.
- Document the list-page consumption contract in the public tool guidance used
  by agents.

## Verification

Add regression coverage for all three list-valued sections:

1. Multiple pages round-trip through JSON parsing and reconstruct the original
   ordered list.
2. Each complete tool response respects the effective byte bound.
3. Multibyte values and a single oversized element have deterministic,
   lossless behavior.
4. Invalid `part` values fail explicitly.
5. Existing string/YAML continuation behavior remains unchanged.

## Provenance and boundaries

- Parent reconciliation: #4314
- Reference implementation from the closed competing branch: PR #4312 commit
  `8c00afaa1` (`paginate list-typed sections by element, not JSON-string
  offset`). Use it as design evidence, not as a wholesale cherry-pick, because
  PR #4330 changed the surrounding delivery architecture.
- The absent-backend delivery-limit fallback drift is a separate residual and
  is out of scope for this issue.

PR #4344, opened 2026-07-24
Rectify: Issue #4341 — Schema-Driven Recipe Section Pagination

Trecek:
## Summary

`get_recipe_section` currently destroys a section's type before pagination. The architectural correction retains and validates the typed value, resolves an explicit `RecipeSectionDef`, constructs an immutable grammar-aware page plan, and renders and measures each complete final response.

The implementation adds schema-driven section definitions, deterministic grammar-aware pagination, fixed-point metadata convergence, immutable plan identities, strict consumer validation, bounded failure handling, cache retirement, and exhaustive verification for raw text, scalar JSON, arrays, and oversized element fragments.

Closes #4341

## Implementation Plan

Plan file: `/home/talon/projects/generic_automation_mcp/.autoskillit/temp/rectify/rectify_issue_4341_schema_driven_recipe_section_pagination_2026-07-23_095431.md`


src/autoskillit/core/types/_type_constants_registries.py
```diff
@@ -266,6 +274,211 @@ def _recipe_delivery_surface_registry_digest() -> str:
 
 RECIPE_DELIVERY_SURFACE_REGISTRY_DIGEST: str = _recipe_delivery_surface_registry_digest()
 
+
+@dataclass(frozen=True, slots=True)
+class RecipeSectionDef:
```
Trecek, line 281: [warning] defense: Public `RecipeSectionDef` construction accepts contradictory strategy/format/default combinations; validity is checked only later while validating an artifact. Enforce these invariants (including name/default consistency) in `__post_init__` so invalid registry definitions fail at construction instead of breaking artifact persistence/loading at runtime.

src/autoskillit/core/types/_type_constants_registries.py
```diff
@@ -266,6 +274,211 @@ def _recipe_delivery_surface_registry_digest() -> str:
...
+    """Static schema and pagination definition for one pullable recipe section."""
+
+    name: str
+    value_kind: Literal["string", "array"]
+    element_kind: Literal["string"] | None
+    missing_behavior: Literal["invalid", "absent", "default"]
+    none_behavior: Literal["invalid", "absent"]
+    section_strategy: Literal["raw", "scalar", "array"]
+    ordinary_content_format: Literal["raw-text", "json-scalar-page", "json-array-page"]
```
Trecek, line 291: [warning] cohesion: `RecipeSectionDef` declares ordinary and oversized content formats as schema data, but pagination never consumes them; planner branches hard-code the same strings while the policy manifest duplicates another mapping. The schema, policy digest, and renderer can drift independently. Derive emitted formats and policy metadata from the registry.

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -354,6 +371,18 @@ def retire_recipe_artifacts(temp_dir: Path, *, kitchen_id: str) -> bool:
                 shutil.rmtree(namespace)
     except (OSError, RecipeArtifactError, TypeError):
         return False
+    try:
+        from autoskillit.server._recipe_section_pagination import (  # circular-break
```
Trecek: [warning] arch: This late import creates a bidirectional dependency: `_recipe_section_pagination` imports `_recipe_delivery` for generation types and limits, while delivery imports pagination for cache eviction. Move lifecycle notification behind a lower-level hook/registry or shared recipe-section component so dependencies remain one-way.

src/autoskillit/server/_recipe_section_pagination.py
```diff
@@ -0,0 +1,914 @@
...
+    get_logger,
+    recipe_section_digest,
+    recipe_section_element_digest,
+    recipe_section_plan_digest,
+)
+from autoskillit.server._recipe_delivery import (
+    RECIPE_ARTIFACT_MAX_BLOB_BYTES,
+    RecipeArtifactGeneration,
+)
+from autoskillit.server.recipe_section._verification import (
```
Trecek, line 43: [warning] arch: The recipe-section domain is split across a parent module and child package with a deliberate cycle: the planner imports `_verification`, while `_verification` references the planner's private `_PlannedPage`/selection types and late-imports its error class. Co-locate the planner contracts and verifier behind one package boundary with one-way imports.

src/autoskillit/server/_recipe_section_pagination.py
```diff
@@ -0,0 +1,914 @@
...
+    """A registry-selected section whose value remains typed."""
+
+    section: str
+    definition: RecipeSectionDef
+    value: object
+    present: bool
+
+
+@dataclass(frozen=True, slots=True)
+class RecipeSectionPageDescriptor:
```
Trecek: [warning] defense: Public page descriptors are stringly typed and every mutually exclusive range family is optional, so callers can construct unknown formats or mixed/incomplete descriptor states. Use Literal/discriminated descriptor variants or construction-time validation; the late final verifier should be defense-in-depth, not the first invariant gate.

src/autoskillit/server/_recipe_section_pagination.py
```diff
@@ -0,0 +1,914 @@
...
+                descriptor = RecipeSectionPageDescriptor(
+                    content_format="raw-text",
+                    page_content_sha256=_qualified_content_digest(content),
+                    byte_start=offsets[start],
+                    byte_end=offsets[end],
+                    byte_total=offsets[-1],
+                )
+            return _PlannedPage(descriptor, content)
+
+        terminal_candidate = candidate_for(len(value))
```
Trecek: [warning] bugs: Each planning loop first serializes and hashes the entire remaining value before searching for a fitting prefix; the raw/scalar, array, and fragment paths repeat this synchronously, making pagination quadratic in section size. CI already hits the 60-second timeout in `test_page_and_fragment_indices_cross_two_digit_boundaries`. Avoid materializing the whole remainder on every page.

src/autoskillit/server/tools/tools_recipe.py
```diff
@@ -616,103 +642,62 @@ async def get_recipe_section(
...
+            try:
+                page_plan = get_or_build_recipe_section_page_plan(
+                    kitchen_id=tool_ctx.kitchen_id,
+                    generation=identity,
+                    selected=selected,
+                    recipe_section_bound_bytes=request_state.recipe_section_bound_bytes,
+                )
+            except RecipeSectionBoundError:
+                return _recipe_section_failure("recipe_section_bound_too_small")
+            except RecipeSectionPaginationError:
```
Trecek: [warning] defense: Every `RecipeSectionPaginationError`—including final verification/invariant failures, not only convergence exhaustion—is silently mapped to `recipe_section_pagination_nonconvergent`. No exception context is logged or retained, so distinct planner defects become misclassified and undiagnosable; split the exception classes or log the caught cause before returning the bounded code.

tests/arch/test_recipe_section_registry.py
```diff
@@ -0,0 +1,178 @@
...
+    assert (
+        definition.value_kind,
+        definition.element_kind,
+        definition.section_strategy,
+        definition.ordinary_content_format,
+        definition.oversized_content_format,
+    ) == (value_kind, element_kind, strategy, ordinary, oversized)
+
+
+def test_recipe_section_presence_and_default_semantics_are_explicit() -> None:
```
Trecek, line 159: [warning] tests: The presence/default-semantics test covers only `ingredients_table`, `errors`, and `warnings`. It omits required `content`, `orchestration_rules`, `stop_step_semantics`, and the dynamic definition, leaving their missing/none behavior contracts free to drift despite those flags controlling artifact acceptance and section visibility.

tests/arch/test_subpackage_isolation.py
```diff
@@ -884,11 +885,11 @@ def test_no_subpackage_exceeds_10_files() -> None:
             logic on DispatchRecord.from_dict. Exempt at 15 files.
     """
     EXEMPTIONS: dict[str, int] = {
-        "server": 15,  # +_recipe_delivery unified finalizer and immutable generation store
+        "server": 16,  # +_recipe_section_pagination deterministic bounded planner
```
Trecek, line 890: [warning] cohesion: The PR raises the server root file-count exemption while simultaneously creating `server/recipe_section` for this feature: the 914-line planner/cache/renderer remains in the parent and only verification is placed in the subpackage. This weakens the architectural guard and splits one private feature asymmetrically; localize the implementation under the new package instead of increasing the exemption.

tests/server/test_recipe_section_pagination.py
```diff
@@ -0,0 +1,736 @@
...
+            "RECIPE_SECTION_PAGINATION_VERSION",
+            pagination.RECIPE_SECTION_PAGINATION_VERSION + 1,
+        )
+
+    changed = get_or_build_recipe_section_page_plan(**kwargs)
+    assert changed is not baseline
+    assert get_or_build_recipe_section_page_plan(**kwargs) is changed
+
+
+def test_cache_entry_limit_evicts_oldest_plan() -> None:
```
Trecek, line 1004: [warning] tests: Cache coverage exercises only the entry-count eviction path. `PAGE_PLAN_CACHE_MAX_BYTES`, single-overweight rejection, cumulative byte eviction, replacement weight accounting, and byte accounting after kitchen eviction are untested, so regressions can silently defeat the cache's advertised memory bound.

Trecek (commented):

Trecek (commented):

src/autoskillit/core/types/_type_constants_registries.py
Trecek: Agreed — fixed in 2e74b6b46. `RecipeSectionDef` now rejects invalid layout/default combinations in `__post_init__`, and registry keys are validated against definition names.

src/autoskillit/core/types/_type_constants_registries.py
Trecek: Agreed — fixed in b0065ebdf. The planner and verifier now consume formats from `selected.definition`; the policy digest is derived from declared definitions and centralized content-format metadata.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in 35d49cf49. Artifact retirement now notifies a lower `recipe_section` lifecycle registry, so delivery no longer imports pagination while success-only, post-lock eviction semantics remain intact.

src/autoskillit/server/_recipe_section_pagination.py
Trecek: Agreed — fixed in 40f6d4ead. Shared errors, selections, descriptors, manifests, and planned-page contracts now live in `recipe_section/_contracts.py`; verification no longer imports back from the planner.

src/autoskillit/server/_recipe_section_pagination.py
Trecek: Agreed — fixed in 6bf9f98cb. Descriptor formats are now `Literal`-typed and construction requires exactly one complete range family matching the declared format.

src/autoskillit/server/_recipe_section_pagination.py
Trecek: Agreed — fixed in db76eda60. Raw, scalar, array, and fragment probes now cap candidate ends from byte-prefix indices; a whole remainder is materialized only when its content bytes can fit the response bound.

src/autoskillit/server/tools/tools_recipe.py
Trecek: Agreed — fixed in 5e65593c9. A dedicated nonconvergence subtype maps to `recipe_section_pagination_nonconvergent`; other pagination invariant failures are logged and return the bounded internal-error code.

tests/arch/test_recipe_section_registry.py
Trecek: Agreed — fixed in 2e74b6b46. The architecture test now pins missing, null, and default semantics for every fixed section plus the dynamic definition.

tests/arch/test_subpackage_isolation.py
Trecek: Valid observation — flagged for design decision. The asymmetry is real, but the originating plan and current `server/AGENTS.md` deliberately place planner/cache/renderer at the server root and verification in the supporting package. Moving the full feature is an architectural migration beyond a targeted review fix.

tests/server/test_recipe_section_pagination.py
Trecek: Agreed — fixed in 18b6cd5ef. Added coverage for single-overweight rejection, cumulative byte eviction, replacement weight accounting, and byte accounting after kitchen eviction.

Trecek (commented):

src/autoskillit/server/_recipe_section_pagination.py
```diff
@@ -0,0 +1,914 @@
...
+            len(canonical_element) - 1
+            if element_index + 1 == element_total
+            else len(canonical_element)
+        )
+        accepted: _PlannedPage | None = None
+        accepted_end = start
+        while low <= high:
+            end = (low + high) // 2
+            candidate = candidate_for(end)
+            if _fits(
```
Trecek, line 627: [critical] bugs: The fragment planner repeatedly slices, JSON-serializes, and fully renders each binary-search candidate for every fragment. The current general CI job times out after 60 seconds in this `_fits` call while running `test_page_and_fragment_indices_cross_two_digit_boundaries`, so a sufficiently large warning element can monopolize the request path instead of returning a bounded page. Rework candidate sizing to avoid repeated whole-prefix serialization.

src/autoskillit/server/tools/tools_recipe.py
```diff
@@ -514,17 +548,21 @@ async def get_recipe_section(
...
+            if part < 0:
+                return _recipe_section_failure("invalid_recipe_section_part")
 
             try:
                 persisted = load_recipe_artifact(
                     artifact_dir,
                     kitchen_id=tool_ctx.kitchen_id,
                     identity=identity,
                 )
+            except RecipeArtifactSchemaError:
```
Trecek: [warning] defense: The `RecipeArtifactSchemaError` catches discard the exception without logging it, even though the lower layer produces a value-free code/path summary. Persisted corruption and recreation schema failures therefore become indistinguishable generic responses with no retained diagnostic context; log the safe exception details before returning the bounded public code.

tests/arch/test_recipe_section_registry.py
```diff
@@ -0,0 +1,178 @@
...
+
+    for section in ("errors", "warnings"):
+        definition = RECIPE_SECTION_REGISTRY[section]
+        assert definition.missing_behavior == "default"
+        assert definition.none_behavior == "invalid"
+        assert definition.has_default is True
+        assert definition.default_value == ()
+
+
+def test_recipe_section_registry_identity_is_stable_and_qualified() -> None:
```
Trecek, line 185: [warning] tests: This registry-identity test checks only SHA-256 syntax and inequality. An arbitrary or stale digest disconnected from the registry or pagination policy would still pass, so the test does not enforce the promised stable/versioned identity. Pin or independently derive the expected digests so definition changes require an intentional version update.

tests/server/test_recipe_section_pagination.py
```diff
@@ -0,0 +1,736 @@
...
+def test_convergence_ceiling_is_derived_from_artifact_policy() -> None:
+    max_count_digits = len(str(RECIPE_ARTIFACT_MAX_BLOB_BYTES))
+
+    assert pagination._convergence_iteration_ceiling() == 1 + (
+        (RECIPE_ARTIFACT_MAX_BLOB_BYTES + 1) * (max_count_digits - 1)
+    )
+    assert pagination._convergence_iteration_ceiling() > 32
+
+
+def test_final_digest_injection_revalidates_descriptor_boundaries(
```
Trecek, line 650: [warning] tests: The only corruption test changes a rendered range after descriptor creation, exercising the final response-versus-descriptor comparison but none of the descriptor/content invariants. Add negative cases for fragment count/digest/ranges, reconstruction mismatch, and page-content digest rejection so the fail-closed verifier is actually covered.

tests/server/test_recipe_section_pagination.py
```diff
@@ -0,0 +1,736 @@
...
+    plan = _build(
+        _payload(warnings=["a", "b", "c"] * 20),
+        "warnings",
+        bound=bound,
+        generation=generation,
+    )
+    pages = _decoded_pages(plan, bound=bound)
+    manifest = dataclasses.asdict(plan.manifest)
+
+    assert {
```
Trecek: [warning] tests: The manifest completeness assertion uses subset inclusion, so accidental extra fields pass while changing the `page_plan_sha256` preimage without requiring a pagination-version update. Assert the exact field set for this versioned wire identity.

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments. GitHub does not permit an author to request changes on their own pull request.

src/autoskillit/server/_recipe_section_pagination.py
Trecek: Agreed — fixed in db76eda60. Candidate search is now bounded by precomputed UTF-8/array prefix offsets before slicing or serialization, eliminating whole-remainder probes; the two-digit boundary regression now completes within the full local gate.

src/autoskillit/server/tools/tools_recipe.py
Trecek: Agreed — fixed in ada933b00. All three RecipeArtifactSchemaError paths now retain the safe structural code/path diagnostic with an explicit load, recreate_persist, or reload stage while preserving the bounded public response.

tests/arch/test_recipe_section_registry.py
Trecek: Agreed — fixed in d292620b6. The architectural test now pins both independently derived qualified digests, so registry or policy identity changes require an intentional test/version update.

tests/server/test_recipe_section_pagination.py
Trecek: Agreed — fixed in 137f4035b. Added direct negative coverage for fragment count, element digest, fragment ranges, reconstruction, and page-content digest corruption through the finalized verifier boundary.

tests/server/test_recipe_section_pagination.py
Trecek: Agreed — fixed in b0c969b91. The versioned manifest contract now asserts exact field equality, preventing accidental preimage expansion from passing unnoticed.

Trecek (commented):

src/autoskillit/core/types/_type_constants_registries.py
```diff
@@ -266,6 +274,308 @@ def _recipe_delivery_surface_registry_digest() -> str:
...
+                "string",
+                "array",
+                "elements",
+                "json-array-page",
+                "json-element-fragment",
+            ),
+        }
+        if layout not in valid_layouts:
+            raise ValueError("invalid recipe section strategy and content-format combination")
+        if self.has_default != (self.missing_behavior == "default"):
```
Trecek, line 333: [warning] defense: `RecipeSectionDef.__post_init__` validates the layout but never validates runtime values for `missing_behavior` or `none_behavior`. Arbitrary strings can therefore survive construction; validation later treats unknown behaviors as permissive/absent and can silently weaken the artifact schema. Reject values outside the declared Literal sets at construction.

src/autoskillit/core/types/_type_recipe_sections.py
```diff
@@ -0,0 +1,193 @@
...
+    "RecipeSectionValidationFinding",
+    "canonical_recipe_section_json",
+    "recipe_section_digest",
+    "recipe_section_element_digest",
+    "recipe_section_plan_digest",
+    "validate_recipe_artifact_sections",
+]
+
+_MISSING = object()
+_RAW_SECTION_DOMAIN = "autoskillit.recipe-section.raw.v1"
```
Trecek: [warning] cohesion: Digest-domain strings and canonical-JSON settings are implemented here but independently repeated in `_RECIPE_SECTION_PAGINATION_POLICY`. A change to hashing behavior can therefore leave `RECIPE_SECTION_PAGINATION_POLICY_DIGEST` unchanged even though the attested protocol changed. Build the policy manifest from the implementation constants, or drive both from one shared definition.

src/autoskillit/core/types/_type_recipe_sections.py
```diff
@@ -0,0 +1,193 @@
...
+                        definition.value_kind,
+                        value,
+                    )
+                )
+            continue
+        if definition.value_kind == "string":
+            if type(value) is not str:
+                findings.append(
+                    _finding(
+                        section,
```
Trecek, line 157: [warning] bugs: Schema validation accumulates every invalid array element and then joins every finding into one exception string. Because persistence invokes this before enforcing the artifact blob-size limit, a large malformed payload can allocate an enormous diagnostic and later log it, turning the validation error path into a memory/log-volume failure. Cap the reported findings and include a bounded omitted-count summary.

src/autoskillit/server/_recipe_section_pagination.py
```diff
@@ -0,0 +1,896 @@
...
+class PagePlanCache:
+    """Bounded thread-safe LRU storing only verified immutable plans."""
+
+    def __init__(
+        self,
+        *,
+        max_entries: int = PAGE_PLAN_CACHE_MAX_ENTRIES,
+        max_bytes: int = PAGE_PLAN_CACHE_MAX_BYTES,
+    ) -> None:
+        self._max_entries = max_entries
```
Trecek, line 119: [warning] defense: `PagePlanCache` accepts negative capacity limits. With `max_entries=-1`, the first `put()` evicts the inserted entry and then loops again because `0 > -1`, raising `KeyError` from an empty `popitem`; a negative byte limit silently disables admission. Validate cache limits at construction.

src/autoskillit/server/recipe_section/_contracts.py
```diff
@@ -0,0 +1,160 @@
...
+from autoskillit.core import RecipeSectionDef
+from autoskillit.server._recipe_delivery import RecipeArtifactGeneration
+
+RecipeSectionContentFormat = Literal[
+    "raw-text",
+    "json-scalar-page",
+    "json-array-page",
+    "json-element-fragment",
+]
+_RANGE_FIELDS_BY_FORMAT: dict[RecipeSectionContentFormat, frozenset[str]] = {
```
Trecek: [warning] cohesion: The content-format/range-field registry duplicates the format metadata in core's `_RECIPE_SECTION_CONTENT_FORMAT_REGISTRY`; the core consistency check compares only format names, so descriptor validation and the advertised pagination policy can silently diverge on required range fields. Define this metadata once and derive both the wire policy and descriptor validation from it.

src/autoskillit/server/recipe_section/_lifecycle.py
```diff
@@ -0,0 +1,24 @@
...
+    with _KITCHEN_RETIREMENT_LOCK:
+        _KITCHEN_RETIREMENT_CALLBACKS.add(callback)
+
+
+def notify_kitchen_retired(kitchen_id: str) -> None:
+    """Notify registered supporting state after artifact retirement succeeds."""
+    with _KITCHEN_RETIREMENT_LOCK:
+        callbacks = tuple(_KITCHEN_RETIREMENT_CALLBACKS)
+    for callback in callbacks:
+        callback(kitchen_id)
```
Trecek: [warning] bugs: A failing retirement callback aborts the loop, so later registered cleanup callbacks are never notified; the caller catches the exception and still completes retirement. Isolate exceptions per callback so one failed cleanup cannot leave other supporting state stale.

src/autoskillit/server/tools/_cancellation_shield.py
```diff
@@ -21,29 +22,85 @@
...
-    result_type controls the response schema:
-    - "fleet_error": fleet_error() envelope (dispatch_food_truck, record_gate_dispatch)
-    - "run_cmd": {"success": False, "exit_code": -1, "stdout": "", "stderr": ...}
-    - "run_python": {"success": False, "exit_code": -1, "stdout": "", "stderr": ...}
-    - "generic" (default): {"success": False, "error": "cancelled", "subtype": "cancelled"}
+    Legacy result modes retain their existing response schemas. Typed mode captures
+    one request state before the handler, publishes that exact object through a
+    ContextVar, and passes it to the cancellation response factory.
     """
+    typed_arguments = (state_factory, state_context_var, response_factory)
```
Trecek, line 64: [warning] defense: Typed cancellation arguments are checked only for all-or-none presence; callable and `ContextVar` validation is deferred until the first request and implemented with `assert`. Misconfigured decorators therefore import successfully and fail on traffic, while optimized Python removes the checks entirely. Validate all three argument types when constructing the decorator and raise `TypeError` there.

tests/server/test_tools_recipe_pull.py
```diff
@@ -71,16 +90,122 @@ def _persist(
...
+        "ingredients_table",
+        "orchestration_rules",
+        "stop_step_semantics",
+        "errors",
+        "warnings",
+    ):
+        assert type(fixture[section]) is type(producer[section]), section
+
+
+@pytest.mark.parametrize(
```
Trecek, line 175: [warning] tests: The malformed-artifact matrix tests wrong concrete types but omits the schema's key absence/null branches: required `content` is never removed, and `None` is never supplied for sections whose `none_behavior` is `invalid`. The tool-state test exercises only allowed `ingredients_table=None`, so regressions that accept missing required content or invalid nulls remain uncovered.

Trecek (commented):

src/autoskillit/core/types/_type_constants_registries.py
Trecek: Agreed — fixed in 114117d65ab763ed8b2e22c0e8f89447bb76f28f. `RecipeSectionDef.__post_init__` now rejects unknown `missing_behavior` and `none_behavior` values at construction, with regression coverage for both fields.

src/autoskillit/core/types/_type_recipe_sections.py
Trecek: Agreed — fixed in 453af474a95d2a925b866f6266911882ba845eed. Canonical-JSON settings and digest-domain identifiers now come from shared IL-0 definitions consumed by both the digest implementation and the attested pagination-policy manifest.

src/autoskillit/core/types/_type_recipe_sections.py
Trecek: Agreed — fixed in 89aa120401e75f7c309f1618de8ebb822fe68189. Recipe-section validation now retains at most 100 concrete findings, counts the remainder without allocating more finding objects, and emits a bounded omitted-count diagnostic; the regression test verifies the 100+30 case.

src/autoskillit/server/_recipe_section_pagination.py
Trecek: Agreed — fixed in bf87a935f26f63fc670c607a743c2a64f6f689ba. `PagePlanCache` now rejects negative entry and byte capacities in its constructor, and both invalid configurations are covered by tests.

src/autoskillit/server/recipe_section/_contracts.py
Trecek: Agreed — fixed in eaa00de1a68e0dddaeac4785be8437f4b308032b. Core now exposes one immutable content-format metadata registry, and server descriptor validation derives its required range-field sets from that same registry used to build the wire policy.

src/autoskillit/server/recipe_section/_lifecycle.py
Trecek: Agreed — fixed in 9c30efbf271b41f4e2b1ae343e0f3dd3f95e821b. Retirement notification now catches and logs failures per callback, allowing all later cleanup callbacks to run; the regression test orders a failing callback before a successful one.

src/autoskillit/server/tools/_cancellation_shield.py
Trecek: Agreed — fixed in 7f05cb6464138f96866a38e31da9dda56d5336cc. Typed cancellation mode now validates both callables and the `ContextVar` eagerly while constructing the decorator and raises `TypeError`; request-time assertions were removed and all three invalid argument kinds are tested.

tests/server/test_tools_recipe_pull.py
Trecek: Agreed — fixed in 9d53f27bf24656559e83b9e205e3df8442f4fe25. The persistence tests now cover missing required `content` and `None` for every registered section whose `none_behavior` is `invalid`, while retaining the allowed `ingredients_table=None` behavior.

src/autoskillit/core/types/_type_constants_registries.py
```diff
@@ -266,6 +276,322 @@ def _recipe_delivery_surface_registry_digest() -> str:
...
+    "reconstruction": {
+        content_format: RECIPE_SECTION_CONTENT_FORMAT_REGISTRY[content_format].reconstruction
+        for content_format in _DECLARED_RECIPE_SECTION_CONTENT_FORMATS
+    },
+}
+RECIPE_SECTION_PAGINATION_POLICY_DIGEST = _qualified_registry_digest(
+    _RECIPE_SECTION_PAGINATION_POLICY
+)
+
+RECIPE_SECTION_MANDATORY_FAILURE_CODES: tuple[str, ...] = (
```
Trecek, line 569: [warning] cohesion: The configured response floor can drift from the actual failure renderer: this manually maintained code tuple and its compact-JSON settings duplicate the arbitrary string codes and serialization in render_recipe_section_failure. A newly added code or renderer change could make RECIPE_SECTION_RESPONSE_FLOOR_BYTES too small. Make failure definitions/rendering one enforced source of truth and derive the floor from it.

src/autoskillit/server/_recipe_section_pagination.py
```diff
@@ -0,0 +1,900 @@
...
+        kitchen_id=kitchen_id,
+        generation=generation,
+        selected=selected,
+        recipe_section_bound_bytes=recipe_section_bound_bytes,
+    )
+    cache = _page_plan_cache()
+    cached = cache.get(key)
+    if cached is not None:
+        return cached
+    plan = build_recipe_section_page_plan(
```
Trecek: [warning] tests: The cache miss/build/put sequence is outside the cache lock, so a concurrent retirement can evict the kitchen while a build is in flight and that build can then reinsert the retired plan; same-key requests can also duplicate expensive planning. Add barrier-controlled concurrency coverage and synchronize admission with retirement.

src/autoskillit/server/tools/tools_recipe.py
```diff
@@ -616,103 +653,70 @@ async def get_recipe_section(
...
-                content = json.dumps(_it) if _it is not None else ""
-            elif section == "orchestration_rules":
-                content = persisted.get("orchestration_rules", "") or ""
-            elif section == "stop_step_semantics":
-                content = persisted.get("stop_step_semantics", "") or ""
-            elif section == "errors":
-                content = json.dumps(persisted.get("errors", []) or [])
-            elif section == "warnings":
-                content = json.dumps(persisted.get("warnings", []) or [])
+            if section in RECIPE_SECTION_REGISTRY:
```
Trecek: [warning] cohesion: Section classification is split between this fixed-registry branch, _is_post_prune_step, and select_recipe_section, even though select_recipe_section repeats fixed and dynamic recognition. These parallel paths can drift as registry or dynamic-step semantics evolve; make one resolver own classification and let the handler only extract/render the selected value.

tests/execution/backends/test_codex_recipe_delivery_conformance.py
```diff
@@ -129,33 +163,256 @@ def _isolated_environment(tmp_path: Path, workspace: Path) -> tuple[dict[str, st
...
+        and candidate.get("body_sha256") == recipe_pull.get("body_sha256")
+        and candidate.get("payload_sha256") == recipe_pull.get("payload_sha256")
+    ]
+    by_part: dict[int, dict[str, object]] = {}
+    for candidate in matching:
+        part = candidate.get("part")
+        if not isinstance(part, int):
+            continue
+        previous = by_part.get(part)
+        if previous is not None:
```
Trecek: [warning] tests: The conformance oracle accepts identical duplicate parts by collapsing them into by_part, although the added continuation contract requires consumers to reject duplicates. Fail on every repeated part so an unnecessary repeated pull cannot pass conformance.

tests/execution/backends/test_codex_recipe_delivery_conformance.py
```diff
@@ -193,58 +450,30 @@ def _run_live_probe(tmp_path: Path) -> tuple[_RecipeProbeObservation, str]:
...
-    assert pulls, "live Codex probe did not retain a get_recipe_section result"
+    envelopes: list[dict[str, object]] = []
+    for candidate in candidates:
+        candidate_pull = candidate.get("recipe_pull")
+        if (
+            isinstance(candidate_pull, dict)
+            and candidate_pull.get("producer_tool") == "open_kitchen"
+        ):
+            envelopes.append(candidate)
+    assert envelopes, "live Codex probe did not retain the open_kitchen pull envelope"
```
Trecek: [warning] tests: The live oracle only requires at least one open_kitchen envelope and then selects the last one, so it still passes if the agent calls the envelope producer multiple times despite the new prompt contract requiring exactly one call. Assert the observed producer-call count in the live oracle.

Trecek (commented):

Trecek (commented):

src/autoskillit/core/types/_type_constants_registries.py
Trecek: Agreed — fixed in b94cd8d8ee99f39140c6a6f1c3dee5eff908901e. `render_recipe_section_failure()` now sits beside `RECIPE_SECTION_MANDATORY_FAILURE_CODES`, rejects unregistered codes, and uses the canonical compact-JSON policy; `RECIPE_SECTION_RESPONSE_FLOOR_BYTES` invokes that same renderer for every registered code, while the server imports it instead of maintaining a duplicate implementation.

src/autoskillit/server/_recipe_section_pagination.py
Trecek: Agreed — fixed in 3b6373bde9cd8defcc1b4f59fec550072bd4e3d3. `PagePlanCache.get_or_build()` now shares one in-flight build per cache key. Retirement records a kitchen tombstone before eviction, so a build completing after retirement is returned to its caller but cannot be re-admitted; direct later puts are also blocked. Barrier/event tests cover both concurrent same-key requests and retirement during a blocked build.

src/autoskillit/server/tools/tools_recipe.py
Trecek: Agreed — fixed in 2a76a7b64c13bcfb3bb72a750ad24af90bd48898. `resolve_recipe_section_definition()` is now the single fixed/dynamic recognition authority. `select_recipe_section()` invokes it and calls the dynamic-content loader only for a recognized dynamic section; the tool handler no longer branches on the registry or maintains `_is_post_prune_step`.

tests/execution/backends/test_codex_recipe_delivery_conformance.py
Trecek: Agreed — fixed in 240f70dc8bea8a5687edc83f2af1a729bbf584a7. `_validated_section_pages()` now fails whenever a part number has already been observed, including byte-for-byte identical duplicates, before updating the reconstruction map.

tests/execution/backends/test_codex_recipe_delivery_conformance.py
Trecek: Agreed — fixed in b5d7e2784a244b9e6aac9386795435c7d891bafa. The live oracle now requires `len(envelopes) == 1` and consumes that sole envelope, so zero or repeated `open_kitchen` producer envelopes both fail the exactly-once contract.

src/autoskillit/core/types/_type_constants_registries.py
```diff
@@ -266,6 +277,352 @@ def _recipe_delivery_surface_registry_digest() -> str:
...
+    "recipe_section_bound_too_small",
+    "recipe_section_cancelled",
+    "recipe_section_internal_error",
+    "recipe_section_pagination_nonconvergent",
+    "recipe_section_serialization_failed",
+    "section_not_found",
+)
+
+
+def render_recipe_section_failure(
```
Trecek: [warning] arch: The IL-0 core/types registry now owns JSON wire-response rendering and byte-bound fallback behavior for a server tool. This places transport protocol business logic in the foundational type layer and makes configuration and server behavior depend on a type-registry module. Keep static definitions in core, but move response rendering/bound enforcement behind the server recipe-section boundary.

src/autoskillit/server/_recipe_section_pagination.py
```diff
@@ -0,0 +1,955 @@
...
+        if value is None and definition.none_behavior == "absent":
+            return SelectedRecipeSection(section, definition, None, False)
+        return SelectedRecipeSection(section, definition, value, True)
+
+    if dynamic_content_loader is not None:
+        dynamic_content = dynamic_content_loader(section)
+    return SelectedRecipeSection(
+        section=section,
+        definition=DYNAMIC_RECIPE_SECTION_DEF,
+        value=dynamic_content,
```
Trecek, line 300: [critical] bugs: Dynamic sections treat any non-None loader result as present, including the empty string. `_extract_step_body_from_persisted` explicitly returns `""` when a named YAML step is null, so such an invalid step now produces a successful empty `raw-text` page instead of `section_not_found`/serialization failure, contradicting the dynamic definition's invalid-null contract.

src/autoskillit/server/recipe_section/_contracts.py
```diff
@@ -0,0 +1,148 @@
...
+class RecipeSectionNonConvergenceError(RecipeSectionPaginationError):
+    """Pagination width planning could not reach a stable state."""
+
+
+class RecipeSectionBoundError(RecipeSectionPaginationError):
+    """The captured request bound cannot fit one progress-making page."""
+
+
+@dataclass(frozen=True, slots=True)
+class RecipeSectionRequestState:
```
Trecek, line 64: [warning] defense: RecipeSectionRequestState does not validate that its captured bound is positive and at least RECIPE_SECTION_RESPONSE_FLOOR_BYTES. Because this state authorizes both pagination and cancellation rendering, an invalid or backend-derived below-floor bound can reach render_recipe_section_failure, which returns its base envelope even when that envelope exceeds the bound. Enforce the bound invariant at construction.

src/autoskillit/server/recipe_section/_contracts.py
```diff
@@ -0,0 +1,148 @@
...
+    scalar_byte_total: int | None = None
+    element_index: int | None = None
+    element_sha256: str | None = None
+    fragment_index: int | None = None
+    fragment_count: int | None = None
+    fragment_byte_start: int | None = None
+    fragment_byte_end: int | None = None
+    fragment_byte_total: int | None = None
+
+    def __post_init__(self) -> None:
```
Trecek, line 115: [warning] defense: RecipeSectionPageDescriptor validates only which range fields are populated, not their runtime types or numeric invariants. Values such as false for element_start or 1.0 for a total can survive final verification through Python's bool/int and int/float equality and be emitted as malformed wire metadata. Validate exact integer types, non-negativity, ordering, totals, and digest syntax in __post_init__.

src/autoskillit/server/tools/tools_recipe.py
```diff
@@ -616,111 +652,58 @@ async def get_recipe_section(
...
-                    content = _extract_step_body_from_persisted(persisted, section)
-                except _RecipeSectionError as exc:
-                    return json.dumps(
-                        {
-                            "success": False,
-                            "error": exc.code,
-                            "detail": str(exc),
-                        }
+                    return _recipe_section_failure("recipe_artifact_schema_mismatch")
+                except RecipeArtifactError:
```
Trecek: [warning] defense: A failed post-recreation artifact reload discards RecipeArtifactError without logging its exception context and returns only a static public detail. Checksum, descriptor, bounds, and filesystem failures therefore become operationally indistinguishable. Keep the bounded public response, but log the caught exception with the reload stage and exc_info.

Trecek (commented):

Trecek (commented):
AutoSkillit review: blocking findings detected. See the inline comments; changes are required before approval.

src/autoskillit/core/types/_type_constants_registries.py
Trecek: Agreed — fixed in caf327261. Failure-envelope JSON rendering and bound-aware context fallback now live in autoskillit.server.recipe_section._rendering; core retains only the static registered failure codes and floor metadata, and the export-count guard was updated in 27aff8a0d.

src/autoskillit/server/_recipe_section_pagination.py
Trecek: Agreed — fixed in 35d302d2c. Dynamic loader output now uses truthy presence semantics, so the empty string emitted for a null named YAML step is absent and follows the existing section_not_found path; a regression test covers the empty-loader result.

src/autoskillit/server/recipe_section/_contracts.py
Trecek: Agreed — fixed in 8c57b80f6. RecipeSectionRequestState now rejects non-exact integers and any captured bound below RECIPE_SECTION_RESPONSE_FLOOR_BYTES in __post_init__, with parameterized coverage for bool, float, zero, and below-floor values.

src/autoskillit/server/recipe_section/_contracts.py
Trecek: Agreed — fixed in fd38bf432, with provisional fragment-planning integration finalized in d6bf85c82. RecipeSectionPageDescriptor now validates lowercase SHA-256 digests, exact non-negative integer fields, ordered/progress-making ranges within totals, and fragment index/count bounds; malformed raw and fragment cases are covered.

src/autoskillit/server/tools/tools_recipe.py
Trecek: Agreed — fixed in db465455c. The post-recreation RecipeArtifactError branch now logs get_recipe_section_artifact_unavailable with stage=reload, the internal detail, and exc_info=True while preserving the bounded public post-recreation reload failed response; a regression test verifies both surfaces.

Issue #4332, opened 2026-07-23
run_skill capability leaks L2 orchestration into L1 skill contracts


Trecek:
## Problem

`run_skill` is an L2 orchestration operation, but skill frontmatter exposes it as a capability to the L1 session that executes the skill. This collapses two orchestration levels and makes the invocation transport look like an operation the invoked skill should perform.

`review-pr/SKILL.md`, for example, declares:

```yaml
uses_capabilities: [agent_model, github_api_write, run_skill]
```

The `review-pr` workflow does not need to invoke itself or launch another skill session. During a direct `$review-pr` invocation for PR #4330, the L1 caller interpreted this declaration as a reason to delegate `/review-pr` through `mcp__autoskillit__run_skill`. That created an unnecessary nested headless invocation and consumed provider quota before the direct workflow resumed.

## Architectural boundary violation

The intended level boundary is:

- L2 owns orchestration and invokes a skill through `run_skill`.
- L1 receives and executes the selected skill workflow.
- The invoked L1 session should not need context or awareness that L2 used `run_skill` to launch it.

`uses_capabilities` is part of the skill contract visible at L1. Listing `run_skill` there incorrectly grants or advertises an L2-only orchestration primitive inside the L1 contract. It also makes these two statements indistinguishable:

1. “L2 invoked this skill through `run_skill`.”
2. “This skill itself uses `run_skill`.”

Those are different architectural facts and must not share one capability marker.

## Expected invariants

- Capability metadata must be scoped to the orchestration level where the capability is valid.
- `uses_capabilities` must describe operations initiated by the skill, never the transport used by its parent to invoke it.
- L1 skill sessions must remain invocation-agnostic unless an explicit cross-level protocol requires otherwise.
- L2-only tools such as `run_skill` must not appear in ordinary L1 skill contracts.
- A skill that does not intentionally orchestrate a separate child skill must not declare `run_skill`.
- Validation must reject level-incompatible capability declarations.

## Required investigation

- Determine why `review-pr` and any other leaf skills declare `run_skill`.
- Inventory every `uses_capabilities: [..., run_skill, ...]` declaration and classify whether it represents a real nested orchestration operation or merely the skill's invocation transport.
- Define capability ownership by orchestration level and enforce it in the capability registry and architectural tests.
- Ensure headless sessions launched by `run_skill` do not receive transport metadata that encourages recursive dispatch.
[...]

PR #4345, opened 2026-07-24
Enforce level-aware skill capability contracts and agent-safe projections

Trecek:
## Summary

This change separates machine-authoritative skill capability contracts from model-facing skill documents, validates exact execution-role ownership, resolves project-local overrides consistently, and persists the effective contract and projected document across resumed sessions. It also closes the follow-up audit findings by using effective project-root resolution during LLM triage and doctor checks, aligning projection and subprocess `cwd`, and adding regression coverage for override identity, role validation, fresh-versus-resume guards, and contract persistence.

## Requirements

- Parse and validate capabilities as typed, exact-role contracts.
- Keep machine authority metadata out of model-facing projections.
- Preserve effective source identity across routing, projection, execution, and resume.
- Resolve project-local overrides consistently in triage and diagnostics.
- Keep fresh-only validation out of resumed execution paths.

Closes #4332

## Implementation Plan

Plan file: `/home/talon/projects/generic_automation_mcp/.autoskillit/temp/rectify/rectify_run_skill_capability_contract_immunity_2026-07-22_201946.md`


Trecek (commented):
AutoSkillit PR Review — Verdict: approved

Trecek (commented):

Trecek (commented):

src/autoskillit/execution/session/_skill_session_contract_store.py
```diff
@@ -0,0 +1,632 @@
...
+@dataclass(frozen=True, slots=True)
+class SkillSessionContract:
+    """Immutable execution contract bound to a projected skill snapshot."""
+
+    root_name: str
+    execution_role: SkillExecutionRole
+    source_refs: Mapping[str, SkillSourceRef]
+    closure: tuple[str, ...]
+    capability_union: frozenset[str]
+    canonical_digests: Mapping[str, str]
```
Trecek: [warning] defense: The frozen contract leaves `canonical_digests` and `projected_digests` as caller-owned mutable mappings, while the other mapping fields are copied into `MappingProxyType`. A loaded, validated contract can therefore be mutated before rehydration, invalidating the claimed immutable digest authority. Copy and freeze both digest maps in `__post_init__`, and add a mutation-rejection test.

src/autoskillit/execution/session/_skill_session_contract_store.py
```diff
@@ -0,0 +1,632 @@
...
+                    str(data.get("write_behavior", {}).get("mode"))
+                    if data.get("write_behavior", {}).get("mode") is not None
+                    else None
+                ),
+                expected_when=tuple(
+                    str(pattern)
+                    for pattern in data.get("write_behavior", {}).get("expected_when", [])
+                ),
+            ),
+            read_only=bool(data.get("read_only", False)),
```
Trecek: [warning] defense: The persistence boundary coerces arbitrary JSON values with `bool(...)`; for example, a malformed persisted `"read_only": "false"` is silently accepted as `True` instead of being rejected. `completion_required` has the same issue, and `projection_gating` is also accepted without a boolean-or-null type check. Validate these serialized fields explicitly before constructing the contract.

src/autoskillit/server/tools/tools_execution.py
```diff
@@ -958,29 +920,14 @@ async def run_skill(
...
-            target_name: str | None = None
-            if tool_ctx.skill_resolver is not None:
-                resolved_command, target_name = resolve_target_skill(
-                    skill_command, tool_ctx.skill_resolver
-                )
-            _skill_info = (
-                tool_ctx.skill_resolver.resolve(target_name)
-                if tool_ctx.skill_resolver and target_name
-                else None
+            _stored_contract = (
```
Trecek, line 926: [critical] bugs: The resume branch loads the persisted contract but leaves `resolved_command` initialized from the caller-supplied `skill_command`; it never restores `stored_contract.resolved_command`. The executor and newly built capability contract can therefore run/describe a different command under the old session's persisted skill authority. Restore the persisted command on resume or reject a mismatch.

src/autoskillit/server/tools/tools_execution.py
```diff
@@ -1312,65 +1296,45 @@ async def run_skill(
...
-                    # Replay-path sessions inherit write scope from their original
-                    # snapshot — they don't need re-augmentation because the snapshot
-                    # was built from a live session that already had the full prefix set.
-                    if closure and tool_ctx.skill_resolver is not None:
-                        write_watch_dirs.extend(
-                            resolve_closure_write_dirs(
-                                closure, tool_ctx.skill_resolver, cwd, write_watch_dirs
-                            )
+                # Fresh invocations extend scope from their validated closure.
+                if invocation is not None and _effective_skill_resolver is not None:
```
Trecek: [warning] bugs: Closure write paths are extended only when `_effective_skill_resolver` is non-None, but that variable is deliberately left None on resume even though `invocation` is rehydrated. Resumed write-capable skills consequently lose their declared closure write directories and may be denied legitimate writes. Resolve paths from `invocation.closure` for both fresh and resumed executions.

src/autoskillit/server/tools/tools_execution.py
```diff
@@ -1575,6 +1579,31 @@ async def run_skill(
             order_id=_oid,  # type: ignore[arg-type]
         ).to_json()
     finally:
+        _contract_store_for_cleanup = locals().get("_contract_store")
```
Trecek: [warning] cohesion: The new provisional/finalized contract lifecycle is spread across several local variables and cleaned up through `locals()` name lookups. This makes essential failure-path cleanup silently dependent on variable spelling and control-flow placement. Encapsulate the lifecycle in a typed state object or context manager owned by the contract store.

src/autoskillit/workspace/skill_capabilities.py
```diff
@@ -0,0 +1,547 @@
...
+            heading = stripped.lstrip("#").strip()
+            heading_level = len(stripped) - len(stripped.lstrip("#"))
+            is_step_heading = bool(_STEP_HEADING_RE.match(heading))
+            if is_step_heading or heading_level <= 3:
+                in_step_section = is_step_heading
+            artifact_section = bool(_ARTIFACT_HEADING_RE.search(heading))
+            result.append(_SourceLine(number, text, False))
+            continue
+
+        if stripped.startswith("```"):
```
Trecek: [warning] defense: Fence tracking recognizes only backtick fences. Valid Markdown tilde fences (`~~~bash ... ~~~`) remain marked executable, so documentary commands such as `git commit` or `run_skill(...)` become genuine capability evidence and can incorrectly invalidate an otherwise safe skill contract. Recognize both Markdown fence forms and preserve the opening fence delimiter through closing.

src/autoskillit/workspace/skills.py
```diff
@@ -2,61 +2,239 @@
...
+                )
+
+
+@dataclass(frozen=True, slots=True)
+class EffectiveSkillInvocation:
+    """Validated root plus its complete executable dependency closure."""
+
+    root: SkillInfo
+    closure: tuple[SkillInfo, ...]
+    capability_union: frozenset[str]
```
Trecek, line 226: [warning] arch: `EffectiveSkillInvocation` is described and exported as validated authority, but it has no construction-time invariants. Backend admission trusts `capability_union`, while dispatch projection recomputes capabilities from `closure`; a directly constructed invocation can therefore make routing and dispatch disagree. Add a `__post_init__` validating root membership, member roles, and the closure-derived capability union.

src/autoskillit/workspace/skills.py
```diff
@@ -220,6 +584,258 @@ def list_all(self) -> list[SkillInfo]:
         _LIST_ALL_CACHE_KEY = key
         return list(combined)
 
+    def resolve_effective(self, name: str, project_root: Path | None) -> SkillInfo | None:
+        """Resolve the current highest-precedence source without caching overrides."""
+        normalized_root = project_root.resolve() if project_root is not None else None
+        if normalized_root is not None:
+            for precedence, search_dir in enumerate(_OVERRIDE_SEARCH_DIRS):
+                skill_path = normalized_root / search_dir / name / "SKILL.md"
+                if not skill_path.is_file():
```
Trecek: [warning] defense: Project-local resolution accepts `SKILL.md` symlinks because `is_file()` follows them, without requiring the resolved file to remain beneath the selected project search directory. A repository can therefore project an external SKILL document into a session as a trusted project-local override. Reject symlinked entries/files and verify the resolved path is contained by `search_root`; the unfiltered catalog path at line 624 needs the same guard.

src/autoskillit/workspace/skills.py
```diff
@@ -220,6 +584,258 @@ def list_all(self) -> list[SkillInfo]:
...
+                    features=features,
+                    experimental_enabled=experimental_enabled,
+                    allow_only=allow_only,
+                )
+            )
+            skills = tuple(sorted((*skills, *internal), key=lambda skill: skill.name))
+        return EffectiveSkillCatalog(
+            skills=skills,
+            execution_role=execution_role,
+            namespace_sources={skill.name: skill.source for skill in effective_skills},
```
Trecek: [warning] cohesion: `EffectiveSkillCatalog.skills` is role/visibility-filtered, but `namespace_sources` is populated from every unfiltered effective skill. Projection consults this map for targets absent from the catalog, so references to disabled or differently scoped skills can be rewritten as though those skills were available. Build namespace metadata from the admitted catalog, or explicitly separate available external targets from excluded skills.

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments.

src/autoskillit/execution/session/_skill_session_contract_store.py
Trecek: Agreed — fixed in acb3e9384fc7d59ca06da45c45f8fd2ed1d129b3. Both digest maps are now defensively copied into immutable mapping proxies, with tests covering caller mutation and item assignment.

src/autoskillit/execution/session/_skill_session_contract_store.py
Trecek: Agreed — fixed in 034b6dc4dff188919c264f67911415f0876c5323. Persisted read_only and completion_required values now require booleans, and projection_gating requires boolean-or-null; malformed manifest tests cover all three fields.

src/autoskillit/server/tools/tools_execution.py
Trecek: Agreed — fixed in e045505ca0ebf7288225bf353ae4980c481d1b48. Resumed execution now restores resolved_command from the validated stored contract, and the resume test verifies the executor receives the persisted command.

src/autoskillit/server/tools/tools_execution.py
Trecek: Agreed — fixed in 0dcf63d8fafba2c717dfc2ca12c7ca8c6bc9bec7. Closure write paths are now derived from every validated invocation, including rehydrated and snapshot-replayed resumes; the resume test verifies the restored watch directory.

src/autoskillit/server/tools/tools_execution.py
Trecek: Agreed — fixed in 741043f434db04b03cff6b3d2ca05f355fd099c6. Contract-store lifecycle state is now encapsulated in a typed lifecycle object with explicit observe, finalize, retention, and cleanup transitions, plus cleanup coverage.

src/autoskillit/workspace/skill_capabilities.py
Trecek: Agreed — fixed in ded150b775b8d0ab372c2ba34dd86a264020efa0. Fence tracking now supports backtick and tilde delimiters, preserves delimiter type and length through closing, and tests documentary and executable cases.

src/autoskillit/workspace/skills.py
Trecek: Agreed — fixed in 66630bb259e00c79874526e003ff7fb735acc845. EffectiveSkillInvocation now validates root membership, unique closure members, member roles, role-capability compatibility, and the closure-derived capability union at construction.

src/autoskillit/workspace/skills.py
Trecek: Agreed — fixed in 23f514df5185870729c2930221def0cbf439fc2b. Project-local resolution now rejects symlinked search roots, entries, and SKILL.md files and verifies resolved containment; directory and file symlink tests cover both resolution paths.

src/autoskillit/workspace/skills.py
Trecek: Agreed — fixed in fb88e0160e4961dde800aaa37dc122a0c471da3c. Namespace metadata now contains only same-role, visibility-admitted external targets while remaining separate from allow-only delivery filtering; visibility and projection contract tests were aligned.

Trecek (commented):

docs/verification/review-pr-immunity.md
```diff
@@ -0,0 +1,77 @@
...
+- Projection: `project_agent_skill_document` with a bound immutable catalog.
+- Provider/backend context: Claude Code 2.1.197 with requested model `sonnet`.
+- Live agent output: none; all four requests failed with HTTP 429 before inference.
+- Causal scope: the deterministic result establishes projection behavior only. It does
+  not claim a causal effect on a stochastic reviewer or CI outcome.
+
+## Worktree verification evidence
+
+Verification ran on 2026-07-23 (America/Los_Angeles) in the feature worktree at
+`HEAD=8c4a4ff925077139516cc22faa37bd1a583a3323`, with the remediation represented by
```
Trecek: [warning] slop: The checked-in verification record says the tested remediation existed only in an uncommitted working-tree diff, and its evidence table points to ephemeral local-only files. The committed document therefore preserves neither the tested source state nor durable evidence; record a commit containing the remediation and reproducible evidence instead.

src/autoskillit/core/types/_type_protocols_execution.py
```diff
@@ -38,6 +99,23 @@ def check_infrastructure(self, cwd: Path) -> str | None: ...
     async def run(self, cwd: Path) -> TestResult: ...
 
 
+@runtime_checkable
+class SkillSessionContractStore(Protocol):
+    """Persistence boundary for skill-session contract ownership."""
+
+    def create_provisional(self, *, contract: Any, snapshot: Any) -> str: ...
```
Trecek: [warning] arch: The new IL-0 SkillSessionContractStore erases the types of its core persistence boundary (`contract`, `snapshot`, and `load()` are all Any) because SkillSessionContract/StoredSkillSessionContract are defined in the IL-1 execution implementation. This defeats static enforcement at every caller and lets incompatible store implementations satisfy the protocol; move the immutable contract DTOs to core and type these methods concretely.

src/autoskillit/workspace/skill_projection.py
```diff
@@ -0,0 +1,799 @@
...
+
+@dataclass(frozen=True, slots=True)
+class SkillProjectionContext:
+    """Execution-local inputs that may affect an agent-visible projection."""
+
+    cwd: Path
+    project_root: Path | None = None
+    catalog: EffectiveSkillCatalog | None = None
+    invocation: EffectiveSkillInvocation | None = None
+    backend: Any | None = None
```
Trecek: [warning] arch: SkillProjectionContext imports CodingAgentBackend for type checking but declares both `backend` and `conventions` as Any and does not assert that the conventions belong to the backend. These fields control model-visible projection and command rendering, so mismatched values can enter the supposedly immutable dispatch authority without type-checker or runtime rejection. Use the concrete core types and derive or validate conventions from backend.

src/autoskillit/workspace/skill_projection.py
```diff
@@ -0,0 +1,799 @@
...
+    if isinstance(skills_or_catalog, EffectiveSkillCatalog):
+        return skills_or_catalog.skills
+    return tuple(skills_or_catalog)
+
+
+def _replace_directory(staging: Path, destination: Path) -> None:
+    if destination.is_symlink() or (destination.exists() and not destination.is_dir()):
+        destination.unlink()
+    elif destination.exists():
+        shutil.rmtree(destination)
```
Trecek, line 373: [warning] defense: `_replace_directory` deletes the destination before publishing staging, while `project_direct_install` sends every identical source/backend/catalog to one shared cache directory without locking. Concurrent dispatch preparation can remove or replace another session's CODEX_HOME or plugin directory while it is being discovered, causing intermittent missing-skill failures. Serialize publication per cache key or use immutable generation directories.

src/autoskillit/workspace/skill_projection.py
```diff
@@ -0,0 +1,799 @@
...
+    destination = Path.home() / ".autoskillit" / "plugin-projections" / cache_key
+    context = SkillProjectionContext(
+        cwd=Path(cwd).resolve(),
+        catalog=catalog,
+        backend=backend,
+        conventions=backend.conventions,
+        substitutions={
+            "{{AUTOSKILLIT_TEMP}}": temp_dir_display_str(None),
+            "{{AUTOSKILLIT_SCRIPTS}}": str(destination / "recipes" / "scripts"),
+            "{{DEFAULT_BASE_BRANCH}}": "main",
```
Trecek: [critical] bugs: The sanitized plugin projection hardcodes {{DEFAULT_BASE_BRANCH}} to `main` even though the normal projection path uses the configured default branch. Orchestrator skills launched through this path can therefore operate on the wrong branch when a project configures `develop` or another default; thread the configured default branch into this projection instead.

tests/cli/test_doctor_standing_pins.py
```diff
@@ -55,6 +55,35 @@ def test_feasible_pin_reports_ok(
...
+            "---\n"
+            "name: resolve-failures\n"
+            "description: Project-local capability contract.\n"
+            "---\n"
+            "Resolve the supplied test failures.\n"
+        )
+
+        results = _check_standing_backend_pins_feasibility(project_dir=project_dir)
+
+        assert all(result.severity == Severity.OK for result in results)
```
Trecek: [warning] tests: This assertion is vacuous when the doctor returns no results, so the test can pass even if the project override prevents the pinned-backend check from running. Assert the expected result count and check identity before checking severities.

tests/server/test_tools_execution_routing.py
```diff
@@ -71,49 +65,34 @@ async def test_run_skill_calls_session_skill_manager_init_session(
...
-    # Use a tier3 skill name
-    await run_skill("/open-pr", "/tmp")
+    await run_skill("/autoskillit:investigate issue", "/tmp")
 
-    # activate_skill_deps must have been called regardless of tier
-    mock_ssm.activate_skill_deps.assert_called_once()
+    mock_ssm.materialize_invocation.assert_called_once()
+    invocation = mock_ssm.materialize_invocation.call_args.args[1]
+    assert invocation.root.name == "investigate"
+    assert {member.name for member in invocation.closure}
```
Trecek: [warning] tests: The test claims to verify the resolver's exact dependency closure but only asserts that the closure is nonempty; the parallel test at L153 has the same weakness. Either test would pass with missing or unrelated members. Assert the expected member-name set.

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments — changes required.

docs/verification/review-pr-immunity.md
Trecek: Agreed — fixed in 4bb2616afedb7b7952d237bd7606d7ab914a1067. The verification record now identifies committed tree 9562bf47fa2987c7108fbb40ffb07817d185c985 and repository-owned reproduction commands/tests instead of ignored .autoskillit/temp artifacts.

src/autoskillit/core/types/_type_protocols_execution.py
Trecek: Agreed — fixed in a1adad9d4f92e68c7e987918da2aca53eb6b8d1c. SkillSessionContract and StoredSkillSessionContract now live in IL-0 core, and SkillSessionContractStore concretely types contract, snapshot, and load() without Any erasure.

src/autoskillit/workspace/skill_projection.py
Trecek: Agreed — fixed in 255760d12985e65d3c585fb8e5db19e1877cbf9f. SkillProjectionContext now uses CodingAgentBackend and BackendConventions, derives conventions from the backend when omitted, and rejects mismatched values.

src/autoskillit/workspace/skill_projection.py
Trecek: Agreed — fixed in 66d06f9c9d3559f8dd5cf8462aae4f66b67bf2e0. Content-addressed direct-install projections are now published once under the established cross-process lock and subsequently validated/reused without deleting a live shared directory.

src/autoskillit/workspace/skill_projection.py
Trecek: Agreed — fixed in f8030ff0bf005bfbf12e5f7f9cbfc8f0f0587817. The configured default base branch is threaded through direct, marketplace, default-plugin, catalog, and effective dispatch projection paths and participates in the immutable cache identity.

tests/cli/test_doctor_standing_pins.py
Trecek: Agreed — fixed in 3545e8de7f92c28664c763717f2e63bb1314c581. The test now asserts one result, its standing_backend_pins_feasibility identity, and Severity.OK before accepting success.

tests/server/test_tools_execution_routing.py
Trecek: Agreed — fixed in 0f02f8da066c213da852961ba5dc4ead9e61679a. Both routing tests now assert the exact resolver closure member-name set {investigate} rather than mere non-emptiness.

src/autoskillit/core/types/_type_protocols_workspace.py
```diff
@@ -61,22 +62,19 @@ def push_to_remote(
...
         session_id: str,
-        *,
-        cook_session: bool = False,
-        config: Any | None = None,
-        project_dir: Path | None = None,
-        recipe_packs: frozenset[str] | None = None,
-        recipe_features: frozenset[str] | None = None,
-        allow_only: frozenset[str] | None = None,
-        backend: CodingAgentBackend | None = None,
+        invocation: Any,
```
Trecek: [warning] arch: The new core workspace boundary erases every skill authority object as `Any`: `materialize_invocation` accepts an untyped invocation/context, `init_session` accepts an untyped catalog/context, and the added resolver methods return `Any`. Consequently, incompatible implementations satisfy these runtime-checkable protocols and callers receive no static protection for the role, closure, or projection contracts. Define core structural protocols/DTOs for these values or parameterize the boundary instead of exposing workspace-owned types through `Any`.

src/autoskillit/execution/session/_skill_session_contract_store.py
```diff
@@ -0,0 +1,565 @@
...
+        """Bind a provisional entry only to the final backend session ID."""
+        _validate_raw_session_id(session_id)
+        provisional = self._provisional_path(correlation_key)
+        destination = self._session_path(session_id)
+        with self._lock:
+            manifest = self._read_manifest(provisional)
+            self._validate_entry(provisional, manifest, expected_raw_session_id=None)
+            manifest["raw_session_id"] = session_id
+            self._write_manifest(provisional, manifest)
+            if destination.exists():
```
Trecek, line 122: [critical] bugs: Finalization uses an instance-local `RLock` around a check-then-`os.replace` sequence. Two store instances finalizing different provisional contracts to the same session ID can both observe no destination, after which the second `os.replace` silently overwrites the first contract. Use a cross-instance atomic no-replace operation or shared filesystem lock so finalized session authority cannot be replaced.

src/autoskillit/execution/session/_skill_session_contract_store.py
```diff
@@ -0,0 +1,565 @@
...
+                expected_when=tuple(
+                    str(pattern)
+                    for pattern in data.get("write_behavior", {}).get("expected_when", [])
+                ),
+            ),
+            read_only=read_only,
+            completion_required=completion_required,
+            skill_contract_json=str(data.get("skill_contract_json", "")),
+            projection_substitutions=tuple(
+                (str(item[0]), str(item[1])) for item in data.get("projection_substitutions", [])
```
Trecek: [warning] bugs: A persisted `projection_substitutions` entry is indexed without validating that it is a two-element sequence. For `projection_substitutions=[["x"]]`, `item[1]` raises `IndexError`, which the surrounding exception handler does not catch, so malformed retained state escapes the documented `ValueError` boundary instead of being rejected as an invalid contract.

src/autoskillit/server/tools/_backend_compat.py
```diff
@@ -3,27 +3,54 @@
...
-def _is_backend_incompatible(skill_info: object, effective_backend: str) -> bool:
-    """Return True if skill's backend_requirements exclude effective_backend."""
-    reqs = getattr(skill_info, "backend_requirements", None)
+@dataclass(frozen=True, slots=True)
+class DirectSkillDispatch:
+    """Projected materialization retained for one direct headless dispatch."""
+
+    add_dirs: tuple[ValidatedAddDir, ...]
+    session_id: str
+    resolved_command: str
```
Trecek: [warning] cohesion: `DirectSkillDispatch` stores `resolved_command`, `invocation`, and `projection_context` independently even though `capability_contract` already owns those same values. Downstream code executes the wrapper's command/CWD while separately passing the embedded contract, and this dataclass has no invariant requiring the copies to agree. Derive these properties from the capability contract or validate identity/equality in `__post_init__` so one dispatch cannot carry two conflicting authorities.

src/autoskillit/server/tools/_execution_helpers.py
```diff
@@ -36,6 +67,435 @@
...
+    try:
+        data = json.loads(payload)
+        return SkillContract(
+            inputs=[SkillInput(**item) for item in data["inputs"]],
+            outputs=[SkillOutput(**item) for item in data["outputs"]],
+            expected_output_patterns=list(data.get("expected_output_patterns", [])),
+            pattern_examples=list(data.get("pattern_examples", [])),
+            write_behavior=data.get("write_behavior"),
+            write_expected_when=list(data.get("write_expected_when", [])),
+            read_only=bool(data.get("read_only", False)),
```
Trecek: [warning] defense: The persisted recipe-contract boundary coerces `read_only` and `completion_required` with `bool(...)`. Malformed values such as `"false"` therefore become `True` and silently change resumed execution authority instead of being rejected. Require actual JSON booleans before constructing `SkillContract`.

src/autoskillit/workspace/skill_projection.py
```diff
@@ -0,0 +1,831 @@
...
+        tempfile.mkdtemp(
+            prefix=f".{destination.name}.plugin-",
+            dir=destination.parent,
+        )
+    )
+    try:
+        _copy_non_skill_plugin_assets(source_root, staging)
+        skill_infos = _skill_sequence(catalog)
+        documents = materialize_agent_skill_tree(staging / "skills", skill_infos, context)
+        _replace_directory(staging, destination)
```
Trecek, line 501: [warning] bugs: The public plugin tree is published before its sibling integrity manifest is written. If manifest construction or `write_versioned_json` fails after this replacement, the cache contains a new tree with a missing or stale manifest; later `project_direct_install` calls see the destination already exists, refuse to rebuild it, and repeatedly fail validation. Publish the tree and manifest as one recoverable transaction or rebuild invalid cache entries.

src/autoskillit/workspace/skill_projection.py
```diff
@@ -0,0 +1,831 @@
...
+            "{{DEFAULT_BASE_BRANCH}}": default_base_branch,
+        },
+    )
+    manifest_path = destination.parent / f".{destination.name}.autoskillit-projection.json"
+    with _InstallLock():
+        if destination.is_symlink() or (destination.exists() and not destination.is_dir()):
+            raise SkillContractError(
+                f"direct plugin projection cache is not a directory: {destination}"
+            )
+        if not destination.exists():
```
Trecek, line 730: [warning] bugs: An existing projection is reused solely from a cache key built from member names/digests, backend, and branch. Projection also depends on `catalog.namespace_sources`: two catalogs with identical members but different external target origins produce the same key, skip rematerialization here, and validation merely hashes the stale files instead of recomputing the current projection. Cross-skill references can consequently retain the wrong `/autoskillit:` namespace.

src/autoskillit/workspace/skills.py
```diff
@@ -2,61 +2,279 @@
...
+                )
+            except (OSError, UnicodeDecodeError):
+                pass
+        if self.canonical_content and not self.canonical_digest:
+            object.__setattr__(
+                self,
+                "canonical_digest",
+                hashlib.sha256(self.canonical_content.encode()).hexdigest(),
+            )
+        if self.frontmatter is None and self.canonical_content:
```
Trecek, line 113: [warning] defense: `SkillInfo.__post_init__` parses caller-supplied canonical content but does not reject an invalid parse or set `invalid_reason`. A directly constructed `SkillInfo` with malformed frontmatter can consequently pass `EffectiveSkillInvocation` and `EffectiveSkillCatalog` validation as trusted authority, failing only during later projection. Enforce frontmatter validity at construction.

Trecek (commented):

Trecek (commented):
AutoSkillit review — Verdict: changes_requested. Blocking and warning findings were detected; see the inline comments for required changes.

src/autoskillit/core/types/_type_protocols_workspace.py
Trecek: Agreed — fixed in 177cfbcd9e391d968afc098102ebcd4b3ae89bdd. Skill authority values crossing the core boundary now use structural SkillAuthority and SkillSourceRefAuthority protocols instead of Any, while the concrete workspace types satisfy those protocols.

src/autoskillit/execution/session/_skill_session_contract_store.py
Trecek: Investigated — this is intentional. _finalize_pending_projection uses os.replace(pending_dir, final_dir), and a rename cannot replace an existing non-empty finalized directory. The winning writer publishes its manifest and snapshot before the rename returns; a losing writer therefore receives an error instead of overwriting the winner's authority data.

src/autoskillit/execution/session/_skill_session_contract_store.py
Trecek: Agreed — fixed in fc0b6312789e5ab9887bde0c50781d0e771e0e15. Contract restoration now rejects projection_substitutions unless it is a list whose entries are exactly two-element lists, preventing malformed persisted data from reaching tuple indexing.

src/autoskillit/server/tools/_backend_compat.py
Trecek: Agreed — fixed in 643fb763ca23e0cd099fc34e190dcb1c41f977f3. DirectSkillDispatch now stores the capability contract as the single authority and derives command, invocation, and projection context through properties.

src/autoskillit/server/tools/_execution_helpers.py
Trecek: Agreed — fixed in 960a005a1dda8f766704086d472403ba8e8a04c2. Persisted capability authorities now require actual JSON booleans; strings and integers are rejected instead of being coerced by bool().

src/autoskillit/workspace/skill_projection.py
Trecek: Agreed — fixed in e5ba2c775d8b3207538fa245312cfa903f59f2f2. Projection cache hits are revalidated under the install lock, and an entry missing its manifest is rebuilt before it can be reused.

src/autoskillit/workspace/skill_projection.py
Trecek: Agreed — fixed in c5554369bb65dd5a4658be5465809946e4def91e. The projection cache identity now incorporates the sorted namespace_sources mapping, so different namespace authorities cannot alias the same cache entry.

src/autoskillit/workspace/skills.py
Trecek: Agreed — fixed in 5a9198e8fe6625ae90be930c816596b249615eac. SkillInfo now records malformed frontmatter as an invalid authority, and EffectiveSkillInvocation rejects any invalid member before dispatch.

Trecek (commented):

src/autoskillit/workspace/skills.py
```diff
@@ -220,6 +649,306 @@ def list_all(self) -> list[SkillInfo]:
...
+                if _skill_is_visible(
+                    skill,
+                    disabled=disabled,
+                    custom_tags=custom_tags,
+                    features=features,
+                    experimental_enabled=experimental_enabled,
+                    allow_only=allow_only,
+                )
+            )
+            skills = tuple(sorted((*skills, *internal), key=lambda skill: skill.name))
```
Trecek, line 812: [critical] bugs: For ORCHESTRATOR catalogs, `skills` already contains the visible bundled `sous-chef`, then `internal` appends the same entry again. `prepare_effective_skill_dispatch` uses `allow_only=None`, so the real catalog reaches `materialize_agent_skill_tree` with duplicate names and raises `duplicate projected skill name`, preventing food-truck/orchestrator launch. Exclude internal names from the first set or deduplicate before constructing the catalog.

src/autoskillit/workspace/skill_projection.py
```diff
@@ -0,0 +1,845 @@
...
+    """Project a visible catalog and bind the artifact to executor authority."""
+    plugin_source = project_default_plugin_source(
+        cwd=cwd,
+        backend=backend,
+        default_base_branch=default_base_branch,
+        skill_catalog=catalog,
+    )
+    contract = build_effective_skill_dispatch_contract(
+        resolved_command,
+        SkillProjectionContext(
```
Trecek: [critical] bugs: The plugin is materialized by `project_direct_install` with `AUTOSKILLIT_*`/`DEFAULT_BASE_BRANCH` substitutions, but the dispatch contract is recomputed from this new context with no substitutions. Its `projected_artifacts` and digests therefore attest different bytes from `plugin_source` whenever a skill contains a placeholder; executor validation checks only map shape, so the integrity mismatch is silent. Build both artifacts from the same projection context or materialized documents.

src/autoskillit/workspace/skill_projection.py
```diff
@@ -0,0 +1,845 @@
...
+        source = (
+            target.source
+            if target is not None
+            else context.catalog.namespace_sources.get(match.group(1))
+            if context.catalog is not None
+            else None
+        )
+        if source is None:
+            return match.group(0)
+        return f"/{_agent_skill_namespace(source)}{match.group(1)}"
```
Trecek: [warning] bugs: Cross-skill references are always rendered with a literal `/`, ignoring `context.conventions.skill_sigil`. Codex and any backend whose native sigil is not `/` receive unusable commands even though top-level command rendering honors the backend convention. Render this reference with the bound convention's sigil.

src/autoskillit/workspace/skills.py
```diff
@@ -2,61 +2,304 @@
...
 
 logger = get_logger(__name__)
 
 
-@dataclass
+def _contained_project_skill_path(search_root: Path, name: str) -> Path | None:
+    """Return a non-symlinked SKILL.md contained by its project search root."""
+    entry = search_root / name
+    skill_path = entry / "SKILL.md"
+    if search_root.is_symlink() or entry.is_symlink() or skill_path.is_symlink():
```
Trecek: [warning] defense: This rejects symlinks only at `search_root`, the skill directory, and `SKILL.md`; it does not reject a symlinked intermediate component such as `project/.claude -> /external/.claude`. In that case `search_root.is_symlink()` is false and the resolved skill is tautologically beneath the resolved external search root, so an outside capability contract is accepted as project-local. Verify the resolved search root remains beneath the resolved project root or reject symlinks in every component.

src/autoskillit/workspace/skills.py
```diff
@@ -2,61 +2,304 @@
...
+        return None
+    if not entry.is_dir() or not skill_path.is_file():
+        return None
+    try:
+        resolved_root = search_root.resolve(strict=True)
+        resolved_skill = skill_path.resolve(strict=True)
+    except OSError:
```
Trecek: [warning] defense: An `OSError` while resolving a project-local override is collapsed to `None`, so `resolve_effective` silently falls through to the lower-precedence bundled skill. Permission errors, transient I/O failures, or a broken mount can therefore execute a different skill than the repository-selected override. Preserve this as an invalid effective contract and fail closed.

src/autoskillit/workspace/skill_capabilities.py
```diff
@@ -0,0 +1,569 @@
...
+
+
+_STATIC_PATTERNS: dict[str, tuple[re.Pattern[str], ...]] = {
+    "agent_subagent": (re.compile(r"Agent\(\s*subagent_type\s*="),),
+    "agent_model": (re.compile(r"Agent\(\s*model\s*="),),
+    "claude_dir": (re.compile(r"\.claude/"),),
+    "commit_files": (re.compile(r"\bcommit_files\s*\("),),
+    "git_metadata_write": (
+        re.compile(r"create_impl_worktree\.sh|git worktree add\b[ \t]+\S|git checkout -b"),
+        re.compile(r"git\s+(?:-C\s+\S+\s+)?commit\s+-m"),
```
Trecek: [warning] defense: The metadata-write classifier recognizes only `git commit -m`. Equivalent writes such as bare `git commit`, `git commit --amend`, `git commit -F file`, and `git commit --message=...` are missed, so executable instructions can mutate Git metadata without declaring `git_metadata_write` and bypass the backend/sandbox policy derived from that capability.

src/autoskillit/workspace/skill_capabilities.py
```diff
@@ -0,0 +1,569 @@
...
+        re.compile(r'\bgit\s+(?:-C\s+\S+\s+)?rebase\s+(?:--\w|[$"\{])'),
+    ),
+    "github_api_write": (
+        re.compile(
+            r"gh api[^\n]*(?:--method\s+(?:POST|PATCH|PUT|DELETE))"
```
Trecek: [warning] defense: The GitHub-write classifier accepts only `gh api --method ...` and a narrow command list. It misses the documented `gh api -X POST|PATCH|PUT|DELETE` form and mutating commands such as `gh pr edit`, allowing those instructions to pass authenticity checks without `github_api_write` and its network/sandbox admission policy.

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments — automated verdict: changes_requested. GitHub does not allow an author to request changes on their own PR.

src/autoskillit/workspace/skills.py
Trecek: Agreed — fixed in 97cf9b84e. The orchestrator catalog now removes internal skill names from the ordinary visible set before appending the authoritative internal entry. A project-local `sous-chef` regression verifies that the catalog contains exactly one entry.

src/autoskillit/workspace/skill_projection.py
Trecek: Agreed — fixed in a8ab87c68. Direct-install materialization and dispatch-contract construction now share the same projection-context builder, including all placeholder substitutions. The regression compares the contract artifact byte-for-byte with the materialized `SKILL.md`.

src/autoskillit/workspace/skill_projection.py
Trecek: Agreed — fixed in 7111d6007. Cross-skill namespace rewriting now uses `context.conventions.skill_sigil`, with regression coverage for Codex's `$` sigil.

src/autoskillit/workspace/skills.py
Trecek: Agreed — fixed in d9cb3b80f. Project-local skill validation now binds the project root and rejects any resolved search root outside it, including a symlinked `.claude` ancestor. The regression confirms fallback to the trusted bundled source.

src/autoskillit/workspace/skills.py
Trecek: Agreed — fixed in b90885c7f. Override validation now distinguishes absence from operational I/O failure: missing paths remain absent, while other `OSError` cases raise `SkillContractError` and search-root enumeration also fails closed. A permission-error regression verifies that no bundled fallback executes.

src/autoskillit/workspace/skill_capabilities.py
Trecek: Agreed — fixed in fc21f5ba9. The classifier now treats every `git commit` form as `git_metadata_write`, including bare commits, `--amend`, `-F`, `--message=`, and `git -C ... commit`; non-mutating git commands remain excluded.

src/autoskillit/workspace/skill_capabilities.py
Trecek: Agreed — fixed in 8725ca4ec, with documentary-context hardening in 27aae5d6c. GitHub write detection now covers `gh api -X`, `--method=...`, and mutating PR/issue/release commands, while commands listed under `NEVER` remain non-executable evidence.

src/autoskillit/server/tools/_execution_helpers.py
```diff
@@ -36,6 +68,441 @@
...
+        dataclasses.asdict(typing.cast(typing.Any, skill_contract))
+    )
+
+
+def deserialize_skill_contract(payload: str) -> SkillContract | None:
+    """Reconstruct a recipe skill contract without consulting current metadata."""
+    if not payload:
+        return None
+    try:
+        data = json.loads(payload)
```
Trecek, line 362: [warning] defense: `deserialize_skill_contract()` assumes the decoded JSON is a mapping before calling `.get()`. A persisted payload such as `[]` raises an uncaught `AttributeError`, escaping the intended `SkillContractError` boundary instead of being reported as invalid retained state. Validate that `data` is a dict or include `AttributeError` in the normalization path.

src/autoskillit/workspace/skills.py
```diff
@@ -121,61 +375,238 @@ def detect_project_local_overrides(
...
+        ):
+            enabled_categories.update(feature_def.skill_categories)
+        else:
+            disabled_categories.update(feature_def.skill_categories)
+    gated_categories = disabled_categories - enabled_categories
+    return allow_only_member or not bool(skill.categories & gated_categories)
+
+
+def _visibility_policy(
+    config: Any | None,
```
Trecek: [warning] arch: The workspace layer now depends on the config package's internal schema through duck typing (`subsets`, `features`, `experimental_enabled`, `packs`, and later `skills` tiers). Avoiding an import does not preserve IL-1 subpackage isolation: config changes can silently break workspace policy without type checking. Convert configuration into a core-owned visibility/tier specification at the composition boundary.

src/autoskillit/server/tools/_execution_helpers.py
```diff
@@ -36,6 +68,441 @@
...
+def get_routing_caps(skill_caps: frozenset[str]) -> list[str]:
+    """Return the sorted worker-routable capabilities."""
+    return sorted(
+        cap
+        for cap in skill_caps
+        if SKILL_CAPABILITY_REGISTRY.get(cap) and SKILL_CAPABILITY_REGISTRY[cap].worker_routable
+    )
+
+
+def build_skill_session_contract(
```
Trecek, line 271: [warning] arch: The server adapter now owns the skill-session domain lifecycle: it constructs persisted contracts, serializes recipe contracts, rehydrates workspace invocations, and validates resume authority. This couples the MCP layer directly to recipe, workspace, and execution persistence models and makes resumability policy unavailable outside server tools. Move this lifecycle behind an execution/session service and leave the server responsible for request/response orchestration.

src/autoskillit/workspace/skills.py
```diff
@@ -2,61 +2,315 @@
...
-    backend_requirements: frozenset[str] = frozenset()
     uses_capabilities: frozenset[str] = frozenset()
+    execution_role: SkillExecutionRole | None = SkillExecutionRole.SESSION
+    activate_deps: tuple[str, ...] = ()
+    canonical_content: str = ""
+    canonical_digest: str = ""
+    frontmatter: SkillFrontmatterParseResult | None = None
     invalid_reason: str | None = None
 
+    def __post_init__(self) -> None:
```
Trecek, line 87: [warning] cohesion: `SkillInfo` stores duplicate source authority (`name`/`source`/`path` and `source_ref`) but validates nothing when a `source_ref` is supplied. Downstream projection and manifests read identity from `source_ref`, while other paths read the direct fields, so one contract can expose conflicting logical names, origins, or paths. Derive one representation from the other or enforce equality in `__post_init__`.

Trecek (commented):

Trecek (commented):

src/autoskillit/server/tools/_execution_helpers.py
Trecek: Agreed — fixed in f91fc091944c2b7ef379ba0ea19c7d40b0258049. `deserialize_skill_contract()` now rejects non-object JSON before any mapping access, so persisted arrays are normalized to `SkillContractError`; regression coverage exercises `[]`.

src/autoskillit/workspace/skills.py
Trecek: Agreed — fixed in 33d8bbc6bc884331fcd04844296799780392454d with integration follow-up 0799ffc8d92ac9bc45b24d15a6fb04ce4ac9b0a7. Config now projects an immutable core `SkillVisibilitySpec` at composition boundaries; workspace protocols and implementations consume that typed contract instead of probing config internals through `Any`.

src/autoskillit/server/tools/_execution_helpers.py
Trecek: Investigated — this ownership is intentional at the IL-3 composition/session-dispatch boundary. Persistence already lives behind the core `SkillSessionContractStore` protocol with an execution-owned implementation injected through `ToolContext`. Moving construction and rehydration into execution would require execution to import workspace/recipe authority, violating enforced IL-004 and IL-008 import boundaries.

src/autoskillit/workspace/skills.py
Trecek: Agreed — fixed in ee274263e45820e0119f9b505e46c5393fd3d13e with integration follow-up 0799ffc8d92ac9bc45b24d15a6fb04ce4ac9b0a7. Supplied `SkillSourceRef` values are now validated against `SkillInfo` origin, logical name, and path before downstream projection or persistence can observe contradictory authority; mismatch cases are covered.

Issue #4333, opened 2026-07-23
Define cumulative context-admission contract and Codex coverage matrix (from #4272)


Trecek:
## Parent

- Tracking issue: [#4272](https://github.com/TalonT-Org/AutoSkillit/issues/4272)
- Work package: C1

## Problem

AutoSkillit bounds individual outputs but has no authoritative contract for cumulative model-visible context admission. Before implementing a ledger, the project must establish what is counted, which component owns admission, how identities and compaction windows are represented, and which Codex surfaces can actually be intercepted.

Current Codex 0.145 has an experimental, disabled-by-default `token_budget` feature and internally computes `base_window_tokens_remaining`, but that value is not documented as an admission-time hook field. Current hooks expose session/turn identity, compaction events, input rewriting, and some result replacement behavior; coverage and authority must be proven rather than assumed.

## Investigation scope

1. Define the exact serialized/tokenized boundary that counts as model-visible admission, including tool arguments, result envelopes, hook feedback, retrieval, and nested/code-mode calls.
2. Define stable identities for session, turn, tool use, subagent/fork, and compaction/window epoch.
3. Specify an atomic reserve/commit/release protocol for concurrent, failed, retried, resumed, and crashed calls.
4. Build a version-pinned coverage matrix for shell, unified exec, apply-patch, MCP, local functions, code-mode nested calls, hosted/specialized tools, and headless/subagent delivery.
5. Evaluate Codex 0.145's `token_budget`/`get_context_remaining` implementation and determine whether AutoSkillit can consume a trustworthy measurement today.
6. Define the narrow upstream Codex contract, if required, for hook-visible active usage, remaining capacity, window limits, and compaction epoch.
7. Define privacy-safe observability and explicit behavior when no authoritative watermark is available.

## Required outputs

- A versioned admission protocol and state machine.
- A producer/control-point coverage matrix with VERIFIED, PARTIAL, and UPSTREAM-GATED states.
- Exact accounting and identity invariants.
- Failure and reconciliation semantics for concurrency, crash, resume, fork, compaction, and model/window changes.
- A decision on authoritative token accounting versus conservative raw-byte emergency ceilings.
- A concrete upstream Codex request tied to existing internal context-window status, if local enforcement cannot be authoritative.
- An implementation dependency graph for the remaining #4272 child issues.

## Non-goals

- Do not implement enforcement or select numeric budget defaults.
- Do not remove existing raw per-producer ceilings.
[...]

PR #4346, opened 2026-07-24
Rectify #4333: Freeze the Cumulative Context-Admission Contract

Trecek:
## Summary

Freeze the versioned, implementation-independent contract required by #4333 before any cumulative ledger or enforcement work begins. The deliverable defines the exact model-visible admission boundary, stable identities and window-epoch rules, atomic lifecycle semantics, a version-pinned coverage matrix, authoritative accounting decisions, the upstream provider contract, privacy-safe observability, and the dependency graph for the existing C2–C8 issues.

The normative machine contract is a pure IL-0 state transition model plus frozen value types and a static coverage registry. ADR-0007 explains that contract and its evidence. This work does not create a journal, wire a producer, or change a model-visible result.

Closes #4333

## Implementation Plan

Plan file: `/home/talon/projects/generic_automation_mcp/.autoskillit/temp/rectify/rectify_cumulative_context_admission_immunity_2026-07-22_223942.md`


Trecek (commented):

src/autoskillit/core/context_admission.py
```diff
@@ -0,0 +1,2927 @@
...
+        policy_snapshot.window_epoch_id == state.snapshot.window_epoch_id
+        and policy_snapshot.window_epoch_number == state.snapshot.window_epoch_number
+    ):
+        protected_pools = state.protected_pools
+    else:
+        open_event = next(
+            (
+                record.event
+                for record in state.processed_events
+                if isinstance(record.event, OpenEpochEvent)
```
Trecek: [warning] bugs: Closed protected batches from epoch 2 or later cannot be resolved after another rollover. The old-policy lookup searches only OpenEpochEvent records, but those epochs were created by RolloverEpochEvent; it therefore finds no protected pool, rejects every otherwise valid release/rollback witness, and retains the charge indefinitely.

src/autoskillit/core/context_admission.py
```diff
@@ -0,0 +1,2927 @@
...
+        reservations=tuple(
+            sorted(
+                state.reservations + event.input_reservations,
+                key=lambda item: item.reservation_id.value,
+            )
+        ),
+        generation_reservations=tuple(
+            sorted(
+                state.generation_reservations
+                + ((generation,) if generation is not None and generation_count > 0 else ()),
```
Trecek, line 1071: [warning] bugs: A reserve event can reuse an active generation_reservation_id because _reserve does not check for a collision before appending the record. ActiveContextAdmissionState then raises duplicate_generation_owner during replace(), so the reducer throws instead of returning a typed conflict or rejection.

src/autoskillit/core/context_admission.py
```diff
@@ -0,0 +1,2927 @@
...
+        reservation_key=record.reservation_key,
+        original_descriptor=record.original_descriptor,
+        expiry_witness=event.expiry_witness,
+        original_terminal_decision=record.original_reserve_decision,
+    )
+    next_state = replace(
+        state,
+        expired_idempotency_tombstones=tuple(
+            sorted(
+                state.expired_idempotency_tombstones + (tombstone,),
```
Trecek, line 2544: [warning] bugs: Expiring an already-expired reservation key with a new event ID appends a duplicate tombstone while retaining the original idempotency record. State validation raises duplicate_idempotency_tombstone, so this foreseeable retry crashes the reducer instead of returning an idempotent or expired decision.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+    raise ContextAdmissionValidationError(reason_code)
+
+
+def _validate_protocol_version(protocol_version: int) -> None:
+    if protocol_version != CONTEXT_ADMISSION_PROTOCOL_VERSION:
+        raise UnsupportedContextAdmissionProtocolError("unsupported_protocol_version")
+
+
+def _validate_non_negative(value: int, reason_code: str) -> None:
+    if isinstance(value, bool) or not isinstance(value, int) or value < 0:
```
Trecek: [warning] defense: The shared numeric validator enforces only non-negativity, not the ADR's 64-bit upper bound. Counts, revisions, and sequences therefore accept arbitrarily large persisted values; reject values above the declared 64-bit range at construction.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+    if isinstance(value, list):
+        return tuple(_decode(item) for item in value)
+    if not isinstance(value, Mapping):
+        return value
+    if set(value) == {"dispatch_id"}:
+        dispatch_id = value["dispatch_id"]
+        if not isinstance(dispatch_id, str):
+            _raise_invalid("invalid_dispatch_identity")
+        return DispatchIdentity.from_dispatch_id(dispatch_id)
+    if "__enum__" in value:
```
Trecek, line 114: [warning] defense: Deserialization selects enum, tuple, and frozenset variants when a marker key is merely present and ignores additional keys. Noncanonical or tampered serialized mappings are silently accepted instead of failing closed; require each tagged representation's exact key set.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+        if (
+            not isinstance(enum_name, str)
+            or enum_name not in _ENUM_REGISTRY
+            or not isinstance(enum_value, str)
+        ):
+            _raise_invalid("unknown_serialized_enum")
+        try:
+            return _ENUM_REGISTRY[enum_name](enum_value)
+        except (TypeError, ValueError) as exc:
+            raise ContextAdmissionValidationError("invalid_serialized_enum") from exc
```
Trecek: [warning] defense: Invalid enum values are wrapped with an explicit exception cause, but the original ValueError includes the attacker-controlled value. The sanitized outer message can therefore still leak forbidden serialized content through __cause__ and tracebacks; suppress or sanitize the cause.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+    if "__frozenset__" in value:
+        raw = value["__frozenset__"]
+        if not isinstance(raw, list):
+            _raise_invalid("invalid_serialized_frozenset")
+        return frozenset(_decode(item) for item in raw)
+    type_name = value.get("__type__")
+    if type_name == "ModelIdentity":
+        try:
+            return ModelIdentity(
+                configured_model=str(value["configured_model"]),
```
Trecek: [warning] defense: ModelIdentity decoding coerces arbitrary values with str() rather than requiring strings, so malformed typed input such as an integer model identifier is accepted as authoritative metadata. Validate the raw field types before construction.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+        ):
+            _raise_invalid("invalid_contract_field_type")
+
+
+@dataclass(frozen=True, slots=True)
+class _OpaqueString(_ContractValue):
+    value: str
+
+    def __post_init__(self) -> None:
+        _validate_bounded_text(self.value, "invalid_opaque_identifier")
```
Trecek: [warning] defense: Opaque protocol identifiers use the generic 128-character/content-free validator, while the ADR limits these IDs to 96 ASCII characters with a narrower character and segment contract. Out-of-contract identifiers can therefore enter persisted state.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+    batch_id: AdmissionBatchId
+    representation_revision: RepresentationRevision
+    representation_binding_id: RepresentationBindingId
+    occurrence_ids: tuple[AdmissionOccurrenceId, ...]
+    authority_source_id: AuthoritySourceId
+
+    def __post_init__(self) -> None:
+        _validate_non_negative(self.window_epoch_number, "invalid_window_epoch_number")
+        _validate_non_negative(self.snapshot_sequence, "invalid_snapshot_sequence")
+        _validate_canonical_tuple(
```
Trecek, line 654: [warning] defense: AdmissionWitness only checks that occurrence_ids are sorted; it accepts an empty tuple and duplicate IDs. Invalid authority witnesses can therefore be constructed and serialized before contextual reducer checks, contrary to the construction-time validation boundary.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+    reserve_class: ReserveClass
+    capability_owner_id: ProtectedPoolOwnerId
+    injected_count: int
+    priority: int
+    required_release_witness_kind: WitnessKind
+
+    def __post_init__(self) -> None:
+        _validate_non_negative(self.injected_count, "invalid_protected_pool_count")
+        _validate_non_negative(self.priority, "invalid_protected_pool_priority")
+        if self.reserve_class is ReserveClass.ORDINARY:
```
Trecek, line 706: [warning] defense: ProtectedPoolSpec permits any WitnessKind as required_release_witness_kind, although release logic can satisfy only NON_ADMISSION or ROLLBACK. Another enum member creates a pool whose capacity can never be released; reject unsupported kinds during construction.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+    kind: AdmissionDecisionKind
+    reason_code: str
+    window_epoch_id: WindowEpochId | None
+    snapshot_sequence: int | None
+    requested_count: int
+    available_ordinary_count: int
+    available_protected_count: int
+
+    def __post_init__(self) -> None:
+        _validate_bounded_text(self.reason_code, "invalid_reason_code", maximum=64)
```
Trecek: [warning] defense: Reason codes use the permissive generic text validator, which accepts uppercase, underscores, dots, colons, and other non-kebab values despite the declared ^[a-z][a-z0-9-]*$ provenance contract. This defeats the closed telemetry/cardinality boundary.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+@dataclass(frozen=True, slots=True)
+class AdmissionBatchRecord(_ContractValue):
+    batch: AdmissionBatch
+    state: AdmissionState
+    reservation_id: AdmissionReservationId | None
+    witness_ids: tuple[AdmissionWitnessId, ...]
+    committed_input_count: int
+    unresolved_input_count: int
+
+    def __post_init__(self) -> None:
```
Trecek, line 793: [warning] defense: AdmissionBatchRecord does not validate counts against lifecycle state; for example, a RESERVED record with nonzero committed_input_count is accepted and persisted even though capacity calculation ignores that contradictory field. Enforce state-dependent record invariants at construction.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+    configuration_mode: str
+    verifier: str
+    source_locator: str
+    tested_version: str
+    tested_revision: str
+    checked_at: str
+    freshness_policy: str
+
+    def __post_init__(self) -> None:
+        for value, reason, maximum in (
```
Trecek, line 1036: [warning] defense: CoverageEvidence applies only generic length checks to semantic provenance fields. Invalid dates, non-40-hex tested revisions, and arbitrary freshness-policy strings are accepted despite their declared ISO-date, git-SHA, and closed-literal contracts.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+    | IdempotencyExpiredEffect
+    | ReservationInvalidatedEffect
+    | EpochClosedEffect
+    | QuarantineRecordedEffect
+    | AuthorityUnavailableEffect
+)
+
+
+@dataclass(frozen=True, slots=True)
+class ProcessedEventRecord(_ContractValue):
```
Trecek, line 1675: [warning] defense: ProcessedEventRecord does not validate event_id == event.event_id, and the adjacent IdempotencyRecord similarly leaves duplicated namespace/event identities unchecked. Deserialized journal state can therefore contain inconsistent identities that preflight replay logic trusts.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2268 @@
...
+                    _raise_invalid("orphan_protected_charge_owner")
+                if generation_record.state in {
+                    GenerationState.RESERVED,
+                    GenerationState.STREAMING,
+                    GenerationState.INDETERMINATE,
+                }:
+                    protected_charges[key] = (
+                        protected_charges.get(key, 0) + generation_record.maximum_allowance
+                    )
+        if any(
```
Trecek, line 2027: [warning] defense: Active-state construction checks protected-pool allocation only; it never rejects nonterminal ordinary input/generation allocations whose global total exceeds snapshot.remaining_count. Deserialization can admit an impossible overreserved state, which _capacity then masks by clamping availability to zero.

tests/core/test_context_admission_state_machine.py
```diff
@@ -0,0 +1,1098 @@
...
+    def _fields(self, operation_kind: str) -> dict[str, object]:
+        self.event_sequence += 1
+        return {
+            "event_id": AdmissionEventId(f"event-{self.event_sequence}"),
+            "protocol_version": CONTEXT_ADMISSION_PROTOCOL_VERSION,
+            "idempotency_namespace": _namespace(operation_kind),
+            "expected_aggregate_revision": self.state.aggregate_revision,
+        }
+
+    def _accept_publication(self, transition: Any, event: object) -> None:
```
Trecek, line 471: [warning] tests: _accept_publication never inspects transition.effects, while most lifecycle paths check only decisions and state. Missing or incorrect reserve, commit/release, reconciliation, rollover, or replay effects could pass despite effects being part of the atomic publication contract; add exact effect assertions for each transition family.

tests/core/test_context_admission_coverage.py
```diff
@@ -0,0 +1,385 @@
...
+
+
+def test_claim_ids_are_stable_and_unique() -> None:
+    expected = tuple(f"COV-{surface.replace('_', '-')}" for surface in EXPECTED_SURFACES)
+    actual = tuple(row.evidence[0].claim_id for row in CONTEXT_ADMISSION_COVERAGE)
+    assert actual == expected
+    assert len(actual) == len(set(actual))
+
+
+def test_runtime_version_or_configuration_mismatch_degrades_deterministically() -> None:
```
Trecek, line 339: [warning] tests: This mismatch test varies only source_version and configuration_mode, but the resolver also gates on backend and as_of and emits coverage_runtime_mismatch. Regressions that ignore backend/freshness or preserve the original reason code would pass; cover every selector and assert the fallback reason.

tests/docs/test_context_admission_decision.py
```diff
@@ -0,0 +1,458 @@
...
+        "watermark_unavailable",
+        "upstream_gated",
+        "raw-byte",
+        "no numeric",
+        "independent",
+    ]:
+        assert required in section
+
+
+def test_decision_freezes_the_complete_producer_matrix(decision_text: str) -> None:
```
Trecek, line 212: [warning] tests: The test named as freezing the complete producer matrix checks only that each surface and the three state words occur somewhere. Swapping a row's control-point owner or observation/authority state would still pass, allowing the ADR projection to drift from CONTEXT_ADMISSION_COVERAGE; parse and compare complete rows.

Trecek (commented):

src/autoskillit/core/context_admission.py
Trecek: Agreed — fixed in 812779b7731f5ef480edd2ae15107ca25fd211f2. The reducer now persists and consults the protected-pool release policy for rollover-created epochs.

src/autoskillit/core/context_admission.py
Trecek: Agreed — fixed in da4e7179c06ad4fdd1cf5836988c3f566c7dbb90. Active generation reservation ID reuse is rejected before any existing reservation can be replaced.

src/autoskillit/core/context_admission.py
Trecek: Agreed — fixed in aa39a0207626afbf5a06fcc674128848f658131a. Repeated idempotency expiry now returns the typed expiry decision without appending a duplicate tombstone.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in 5dbc178ad1040edeb95a480a4731ce0c22d95c25. Protocol integer validation now enforces the unsigned 64-bit upper bound as well as non-negativity.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in 3b4c15252bece6fbeb7d860e71e19be8b3232425. Tagged enum, tuple, and frozenset decoding now requires the exact canonical key set.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in c8b12b77486d86bc6b04c63aa6fd57a035c09359. Invalid enum decoding now suppresses the underlying ValueError cause so serialized input is not exposed.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in 91696975b86683d4852a9114913254ecaab6ae92. ModelIdentity decoding now requires correctly typed fields instead of coercing malformed values with str().

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in 240ccba60a4a0d5d0ff50aaecada361362e537e5. Opaque identifiers now enforce the protocol ASCII segment grammar and 96-character limit.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in ce5e3cef57cb8a9f554b6f1cbf644c5edf77b60d. Batch-scoped witnesses now require non-empty unique occurrence IDs, while rollover keeps its intentional empty exception.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in 10bf5ba17b0831213631cfc0b56b6bdee993245d. ProtectedPoolSpec now limits release witness kinds to values the reducer can actually satisfy.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in 7786e09e56e4e74bbb0016bc8da4429b436fbd61. Reason-bearing protocol values now enforce canonical kebab-case reason codes.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in 28c879af4fe37e7627eafebf0ebb436ce3fc574f. AdmissionBatchRecord now enforces lifecycle-specific counter invariants.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in e5d206908dc6ecd5adda6be0e0fea5c49f40aeeb. Coverage evidence now validates ISO dates, full Git revisions, and the declared freshness-policy vocabulary.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in dab4b87713d783cabef1c20733dfe941d25ca9ea. Processed-event and idempotency records now cross-check their duplicated event and key identities.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in 152109beef2ac4b5d313f2e7ce3cfa8746e2abdd. Active state validation now rejects nonterminal input and generation allocations that exceed remaining context capacity.

tests/core/test_context_admission_state_machine.py
Trecek: Agreed — fixed in d9bc9a34e146ea067380307f5868421e30e69f6a. The state-machine oracle now asserts the exact publication effect family and its source and coordinate fields.

tests/core/test_context_admission_coverage.py
Trecek: Agreed — fixed in 8e45ebbfce8b908a461a52632e6877946935bae4. Coverage mismatch tests now exercise backend and as-of selectors and assert the exact fallback reason.

tests/docs/test_context_admission_decision.py
Trecek: Agreed — fixed in f98dd8a722418b8da6653260f1d5148e059c3fbc. The ADR coverage test now compares every matrix row and state exactly.

src/autoskillit/core/context_admission.py
```diff
@@ -0,0 +1,2933 @@
...
+    receiver_fence_valid = proof is not None and (
+        proof.old_window_epoch_id == state.snapshot.window_epoch_id
+        and proof.old_window_epoch_number == state.snapshot.window_epoch_number
+        and proof.new_window_epoch_id == event.new_snapshot.window_epoch_id
+        and proof.new_window_epoch_number == event.new_snapshot.window_epoch_number
+        and proof.receiver_authority_source_id == event.witness.authority_source_id
+        and proof.highest_admitted_dispatch_sequence == _highest_dispatch_sequence(state)
+    )
+    fully_resolved = retained_total == 0
+    snapshot_deducts_unresolved = (
```
Trecek, line 2637: [critical] bugs: The no-fence rollover alternative treats `new_snapshot.active_count >= retained_total` as proof that the snapshot deducted every unresolved old-epoch charge. An unchanged snapshot (for example active=60, remaining=40) with retained=10 passes; rollover then clears the current records and `_capacity` exposes all 40, double-spending the retained 10. Require evidence that remaining capacity was actually reduced by the retained charge, or typed authority evidence, rather than comparing against total active context.

src/autoskillit/core/context_admission.py
```diff
@@ -0,0 +1,2933 @@
...
+                snapshot_sequence=reservation.snapshot_sequence,
+                witness_ids=(event.witness.witness_id,),
+            ),
+            *effects,
+        )
+    generation_effects: tuple[AdmissionEffect, ...] = ()
+    generation_records: list[GenerationReservationRecord] = []
+    revision, sequence = _effect_coordinates(state, capacity_changed=True)
+    for generation in next_state.generation_reservations:
+        if generation.request_id == record.batch.request_id and generation.state in {
```
Trecek: [critical] bugs: Generation invalidation is keyed by `request_id` even though each reservation has a typed `batch_id` (and the state-machine effect oracle matches by batch). The state contract does not forbid two batches from sharing a request ID, so releasing one batch removes the other batch’s live output allowance and publishes a false invalidation. Match `generation.batch_id` to `record.batch.batch_id`; the closed-epoch mirror has the same predicate.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2279 @@
...
+        if self.state in {
+            GenerationState.RESERVED,
+            GenerationState.STREAMING,
+            GenerationState.INDETERMINATE,
+            GenerationState.QUARANTINED,
+        }:
+            return self.maximum_allowance
+        return 0
+
+    def __post_init__(self) -> None:
```
Trecek, line 844: [critical] defense: `GenerationReservationRecord` validates numeric and ownership shape but not lifecycle/terminal-field consistency, and reserve-event validation accepts the record as-is. A caller can reserve a positive allowance with `state=RECONCILED`; it passes the initial capacity check and is appended, but `charged_output_count()` immediately returns zero, allowing the next reservation to reuse that capacity. Require a new reservation to be `RESERVED` with empty terminal fields, or use a separate construction type for reserve input.

src/autoskillit/core/context_admission.py
```diff
@@ -0,0 +1,2933 @@
...
+
+def _reserve(
+    state: ContextAdmissionState,
+    event: ReserveRequestEvent,
+) -> AdmissionTransition:
+    if not isinstance(state, ActiveContextAdmissionState):
+        return _reject(state, event, "epoch-uninitialized")
+    if event.snapshot_sequence != state.snapshot.snapshot_sequence:
+        return _reject(state, event, "snapshot-sequence-mismatch")
+    if _batch_record(state, event.batch.batch_id) is not None:
```
Trecek: [warning] bugs: This collision check considers only the active epoch even though batch IDs are retained for the session and `_closed_batch_location` returns the first closed match. Reusing a closed batch ID is accepted; after the new batch rolls over, accept/release resolution targets the older audit record and can leave the newer retained charge unresolved. Reject IDs already present in `closed_epochs`.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2279 @@
...
+            linked_batch = batch_records_by_id.get(occurrence_record.batch_id)
+            if (
+                linked_batch is None
+                or occurrence_record.occurrence.occurrence_id
+                not in linked_batch.batch.occurrence_ids
+                or occurrence_record.reservation_id != linked_batch.reservation_id
+                or occurrence_record.state is not linked_batch.state
+            ):
+                _raise_invalid("inconsistent_occurrence_link")
+        for batch_record in self.batch_records:
```
Trecek, line 1890: [warning] defense: Active-state construction validates each reservation forward to a batch, but never requires a nonterminal batch to own a reservation. A `REQUEST_DISPATCHED` batch and matching occurrence records with `reservation_id=None` therefore construct successfully; capacity charges zero and acceptance later quarantines with `missing-reservation`. Require charge-owning lifecycle states to reference a matching reservation.

tests/core/test_context_admission_state_machine.py
```diff
@@ -0,0 +1,1255 @@
...
+            self.latest_replayable_event = event
+        self.events.append(event)
+        self.published_effects.append(transition.effects)
+
+    def _expected_effect_types(
+        self,
+        transition: Any,
+        event: object,
+    ) -> tuple[type[object], ...]:
+        if transition.decision.kind not in {
```
Trecek, line 501: [warning] tests: The effect oracle returns `()` for every decision outside `WOULD_ADMIT`/`QUARANTINED` before it can recognize `AuthorityUnavailableEvent`. An active authority-unavailable transition publishes `AuthorityUnavailableEffect`, but the machine has no rule that generates this event and the focused reducer test does not assert effects, so a missing or malformed effect would pass. Add an authority-unavailable rule and its expected-effect case.

tests/core/test_context_admission_state_machine.py
```diff
@@ -0,0 +1,1255 @@
...
+        other_slot = 1 if slot == 0 else 0
+        if other_slot not in self.occurrences or slot not in self.occurrences:
+            return
+        if slot in self.charges or other_slot in self.charges:
+            return
+        first = self.occurrences[slot]
+        second = self.occurrences[other_slot]
+        if first.reserve_class is not second.reserve_class:
+            return
+        batch = _multi_batch((first, second))
```
Trecek, line 796: [warning] tests: This rule reserves a `_multi_batch`, but every later lifecycle rule reconstructs a singleton `_batch(occurrence)` with a different batch ID. Successful multi-member batches are therefore never prepared, staged, dispatched, accepted, or released; deterministic coverage exercises only atomic reservation/rejection. Keep the actual reserved batch in the model and drive it through the full lifecycle.

Trecek (commented):

Trecek (commented):
AutoSkillit review found 3 critical and 4 warning findings. See the inline comments for exact code anchors. GitHub does not permit the PR author to submit a REQUEST_CHANGES review, so this self-review is recorded as COMMENT; the AutoSkillit verdict remains changes_requested.

src/autoskillit/core/context_admission.py
Trecek: Agreed — fixed in 3860c3748. The no-fence alternative now requires the new snapshot remaining capacity to be reduced by at least the retained unresolved charge; unchanged snapshots are rejected, and focused rollover tests cover both paths.

src/autoskillit/core/context_admission.py
Trecek: Agreed — fixed in 86f6c5673. Both active and closed-epoch release paths now match generation.batch_id to the exact batch. Regression coverage verifies that a second batch sharing the request ID retains its live generation allowance.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in f02e9ffe9. ReserveRequestEvent now requires a new generation reservation to be RESERVED with empty terminal fields, preventing terminal records from reopening capacity after admission.

src/autoskillit/core/context_admission.py
Trecek: Agreed — fixed in c649dbe7a. Reservation now rejects batch IDs retained in closed epoch audits; 60ac52d0d adds a non-conflicting regression event that proves the closed-history guard is reached.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in 15ac1e746. Active-state validation now requires RESERVED, PREPARED, HISTORY_STAGED, and REQUEST_DISPATCHED batches to resolve to an actual reservation owner, with a dispatched-state regression test.

tests/core/test_context_admission_state_machine.py
Trecek: Agreed — fixed in dda57e50c. The state machine now generates active AuthorityUnavailableEvent transitions and requires the exact AuthorityUnavailableEffect before the general decision-kind early return.

tests/core/test_context_admission_state_machine.py
Trecek: Agreed — fixed in 95bc49d21 and completed in 1903234f8. The model retains the actual reserved batch per member slot, drives it through prepare, stage, dispatch, accept, release, reconciliation, and rollover paths, and preserves each member revision in the atomic reservation descriptor.

src/autoskillit/core/context_admission.py
```diff
@@ -0,0 +1,2937 @@
...
+            reservation.reserved_count
+            if (reservation := _reservation_for(state, record)) is not None
+            else 0
+        )
+        for record in unresolved_batch_records
+    )
+    retained_generation_count = sum(
+        generation.maximum_allowance
+        for generation in state.generation_reservations
+        if generation.request_id in retained_generation_request_ids
```
Trecek: [critical] bugs: Rollover identifies retained generation allowances by request_id, but request IDs are not unique per batch. If batch A is still RESERVED while batch B with the same request ID is COMMITTED, B causes A allowance to be retained instead of invalidated, corrupting retained charge and audit/effect accounting. Match retained generations by batch_id, and add the shared-request rollover case to the oracle.

src/autoskillit/core/context_admission.py
```diff
@@ -0,0 +1,2937 @@
...
+            record.occurrence.occurrence_id,
+            record.occurrence.representation_revision,
+        )
+        for record in member_records
+    )
+    if reservation.key.occurrence_revisions != expected_revisions:
+        return _reject(state, event, "reservation-revision-mismatch")
+    generation = event.generation_reservation
+    generation_count = generation.maximum_allowance if generation is not None else 0
+    if generation is not None and any(
```
Trecek: [warning] bugs: Generation-reservation ID reuse is checked only in the active epoch. A retained ID from an older closed epoch can be reused, and after another rollover _closed_generation_location returns the older audit first, so reconciliation targets the wrong epoch. Reject IDs already present in closed_epochs.

src/autoskillit/core/context_admission.py
```diff
@@ -0,0 +1,2937 @@
...
+    as_of: str,
+) -> ProducerCoverageDef:
+    """Resolve one static coverage row against runtime lineage inputs."""
+    row = next(
+        (item for item in CONTEXT_ADMISSION_COVERAGE if item.surface is surface),
+        None,
+    )
+    if row is None:
+        raise ContextAdmissionValidationError("unknown_producer_surface")
+    evidence = row.evidence[0]
```
Trecek, line 2936: [warning] cohesion: The schema permits multiple evidence records and three freshness policies, but this resolver always reads evidence[0] and compares a fixed backend/configuration/version/date set; it cannot implement revision-based or infer-only semantics. Please either define and implement evidence-selection/policy behavior or constrain the v1 schema to the one supported shape.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2297 @@
...
+            self.witness_ids,
+            "noncanonical_witness_ids",
+            key=lambda witness_id: witness_id.value,
+        )
+        if len(self.witness_ids) != len(set(self.witness_ids)):
+            _raise_invalid("duplicate_witness_id")
+
+
+@dataclass(frozen=True, slots=True)
+class ExpiredIdempotencyTombstone(_ContractValue):
```
Trecek, line 872: [warning] defense: ExpiredIdempotencyTombstone has no construction-time identity or witness validation. Deserialized state can pair reservation key K with an unrelated namespace, descriptor, batch, or witness, while _preflight trusts K alone and permanently returns IDEMPOTENCY_EXPIRED for valid work. Cross-check the duplicated identities and require a correctly bound IDEMPOTENCY_EXPIRY witness.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2297 @@
...
+                tombstone.reservation_key.window_epoch_number,
+                tombstone.reservation_key.batch_id.value,
+            ),
+        )
+        _validate_canonical_tuple(
+            self.closed_epochs,
+            "noncanonical_closed_epochs",
+            key=lambda audit: audit.snapshot.window_epoch_number,
+        )
+        _validate_state_metadata_uniqueness(
```
Trecek: [warning] defense: State construction checks journal uniqueness but not journal coordinates against aggregate_revision and admission_sequence. For example, a state at revision 0 can contain a processed record at revision 10; the next publication writes revision 1 and silently regresses the journal. Bound processed and idempotency coordinates to the current state and enforce their consistency.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2297 @@
...
+                if record.reservation_id is not None
+                else None
+            )
+            if (
+                record.state
+                in {
+                    AdmissionState.RESERVED,
+                    AdmissionState.PREPARED,
+                    AdmissionState.HISTORY_STAGED,
+                    AdmissionState.REQUEST_DISPATCHED,
```
Trecek, line 1957: [warning] defense: The missing-reservation guard covers RESERVED through REQUEST_DISPATCHED but omits INDETERMINATE. An INDETERMINATE batch with reservation_id=None and unresolved_input_count=0 therefore constructs successfully, and charged_input_count returns zero, dropping its conservative charge. Include INDETERMINATE among reservation-owning states.

tests/core/test_context_admission_state_machine.py
```diff
@@ -0,0 +1,1348 @@
...
+                    AdmissionState.RESERVED,
+                    AdmissionState.PREPARED,
+                    AdmissionState.HISTORY_STAGED,
+                }
+            }
+            input_invalidation_count = sum(
+                reservation.key.batch_id in invalidated_batch_ids
+                for reservation in self.state.reservations
+            )
+            retained_request_ids = {
```
Trecek: [warning] tests: The rollover oracle mirrors the reducer by grouping generation reservations on request_id. Distinct batches may share a request ID, so both implementation and oracle can retain the wrong allowance without a failing state-machine test. Derive expected retention from batch_id and add a two-batch shared-request rollover case.

Trecek (commented):

Trecek (commented):
AutoSkillit review: blocking findings detected. See the inline comments. Verdict: changes_requested. GitHub does not allow an author to submit REQUEST_CHANGES on their own pull request, so this verdict is recorded as a comment review.

src/autoskillit/core/context_admission.py
Trecek: Agreed — fixed in 104141ea8 by deriving rollover generation retention and invalidation from batch_id. State-machine oracle coverage landed in 8c9f9650c, with the shared-request lifecycle fixture completed in 324b36642. Distinct batches sharing one request ID can no longer retain each other’s generation allowance.

src/autoskillit/core/context_admission.py
Trecek: Agreed — fixed in 2b463fb42. Reserve admission now rejects generation-reservation IDs found in active state or any retained closed-epoch audit, preventing later reconciliation from selecting an older same-ID record.

src/autoskillit/core/context_admission.py
Trecek: Agreed — fixed in fa215b0c5 by constraining protocol-v1 coverage evidence to the resolver’s supported shape: exactly one evidence record using verify_on_version_or_configuration_change. Tests reject additional evidence and the unimplemented revision/infer-only policies.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in c7dedc196. ExpiredIdempotencyTombstone now cross-checks namespace, reservation key, descriptor batch, terminal decision coordinates, and a fully bound IDEMPOTENCY_EXPIRY witness before state construction can succeed.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed across 19acc8797 and 4e423d17d, with valid rejection-coordinate semantics finalized in 56e9143a6. Processed-event and idempotency coordinates are bounded to state counters, idempotency publications must match their processed event, and reserve metadata is installed atomically.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in c2769653e. INDETERMINATE is now included among charge-owning batch states that require a live reservation owner, preventing zero unresolved count plus a missing reservation from dropping conservative charge.

tests/core/test_context_admission_state_machine.py
Trecek: Agreed — fixed in 8c9f9650c and completed in 324b36642. The rollover effect oracle now derives retained generations from batch_id, and a two-batch shared-request lifecycle test proves that only the committed batch’s allowance survives rollover.

src/autoskillit/core/context_admission.py
```diff
@@ -0,0 +1,2942 @@
...
+        or event.final_manifest != record.batch.manifest
+        or event.final_manifest.representation_revision != expected_revision
+        or event.final_manifest.request_id != record.batch.request_id
+        or binding.counted_representation_revision != expected_revision
+        or binding.dispatched_representation_revision != expected_revision
+        or binding.final_manifest_revision != expected_revision
+        or binding.representation_binding_id != record.batch.manifest.representation_binding_id
+        or binding.request_id != record.batch.request_id
+        or binding.batch_id != record.batch.batch_id
+        or event.authority_source != event.witness.authority_source_id
```
Trecek: [warning] bugs: An AcceptInputEvent with mismatched authority IDs is quarantined and conservatively charged while its batch is active, but the same event arriving after rollover is rejected here without recording the exact charge. Event timing relative to rollover therefore changes accounting semantics and can leave the closed batch permanently unresolved.

src/autoskillit/core/context_admission.py
```diff
@@ -0,0 +1,2942 @@
...
+        )
+        if item.occurrence.occurrence_id in member_ids
+        else item
+        for item in audit.terminal_occurrence_records
+    )
+    updated_audit = replace(
+        audit,
+        terminal_occurrence_records=occurrence_records,
+        terminal_batch_records=batch_records,
+        retained_unresolved_count=_closed_retained_input_count(
```
Trecek: [warning] bugs: Closed-epoch resolutions reduce retained_unresolved_count, but _capacity ignores closed_epochs and the active snapshot.remaining_count is unchanged. After a no-fence rollover whose snapshot deducted the retained charge, a valid release/rollback never restores that capacity; closed generation reconciliation has the same flaw, so future requests can be rejected despite the reservation being resolved.

Trecek (commented):

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2283 @@
...
+            profile_name=profile_name,
+        )
+    if not isinstance(type_name, str) or type_name not in _TYPE_REGISTRY:
+        _raise_invalid("unknown_serialized_contract_type")
+    contract_type = _TYPE_REGISTRY[type_name]
+    kwargs = {key: _decode(item) for key, item in value.items() if key != "__type__"}
+    try:
+        return contract_type(**kwargs)
+    except TypeError as exc:
+        raise ContextAdmissionValidationError("invalid_serialized_contract") from exc
```
Trecek: [warning] defense: Malformed serialized mappings with attacker-controlled field names reach the constructor, and the resulting TypeError is retained as __cause__. Tracebacks can therefore expose forbidden input despite the sanitized outer error; suppress the cause or validate exact contract field sets before construction.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2283 @@
...
+    terminal_batch_records: tuple[AdmissionBatchRecord, ...]
+    terminal_reservations: tuple[AdmissionReservation, ...]
+    terminal_generation_reservations: tuple[GenerationReservationRecord, ...]
+    closure_witness_id: AdmissionWitnessId
+    fence_proof: EpochFenceProof | None
+    processed_event_tombstones: tuple[AdmissionEventId, ...]
+    retained_unresolved_count: int
+    retained_generation_count: int
+
+    def __post_init__(self) -> None:
```
Trecek, line 919: [warning] defense: ClosedEpochAudit construction does not enforce the ADR's 10^4-occurrence-per-audit cardinality bound. Deserialization accepts arbitrarily large terminal collections and immediately performs multiple sorting, set-building, and nested ownership passes, leaving the persistence boundary open to memory/CPU exhaustion.

src/autoskillit/core/types/_type_context_admission.py
```diff
@@ -0,0 +1,2283 @@
...
+        _AdmissionEffectBase.__post_init__(self)
+        if not isinstance(self.target_id, AdmissionBatchId | GenerationReservationId):
+            _raise_invalid("invalid_effect_target")
+        _validate_reason_code(self.reason_code)
+
+
+@dataclass(frozen=True, slots=True)
+class ConflictRejectedEffect(_AdmissionEffectBase):
```
Trecek, line 1565: [warning] cohesion: ConflictRejectedEffect is publicly exported and included in the closed AdmissionEffect union, but no reducer path constructs it: both event-ID and reservation-key conflicts return effects=(). This leaves a dead protocol variant whose relationship to CONFLICT decisions is undefined. Decide whether conflicts must publish this effect and wire both paths consistently, or remove the unused type from the protocol surface.

tests/core/test_context_admission_reducer.py
```diff
@@ -0,0 +1,2504 @@
...
+        effect.target_id
+        for effect in released.effects
+        if isinstance(effect, ReservationInvalidatedEffect)
+    }
+    assert invalidated_targets == {
+        GenerationReservationId(f"generation-{first_batch.batch_id.value}")
+    }
+
+
+def test_rollover_moves_old_work_to_resolvable_closed_epoch_audit() -> None:
```
Trecek, line 2290: [warning] tests: This is the only focused exercise of the separate closed-epoch acceptance and generation-reconciliation implementations, but it covers only their normal success paths. The independent closed-audit quarantine branches for input overage/incomplete span ownership and output usage above allowance remain untested, including retained-count mutation and QuarantineRecordedEffect publication. Add closed-epoch quarantine cases rather than relying on active-epoch analogues.

tests/core/test_context_admission_state_machine.py
```diff
@@ -0,0 +1,1398 @@
...
+                    )
+                ),
+                None,
+            )
+            if slot_and_occurrence is None:
+                continue
+            slot, occurrence = slot_and_occurrence
+            batch = self.batches[slot]
+            record = self._find_batch(batch.batch_id)
+            if record is None or record.state not in {
```
Trecek: [warning] tests: The generation rule treats HISTORY_STAGED, COMMITTED, and QUARANTINED batches as eligible, then unconditionally expects StartGenerationEvent to return WOULD_ADMIT. The reducer admits generation start only from REQUEST_DISPATCHED, and these other states are reachable while the generation reservation remains RESERVED, so Hypothesis can produce a false test failure. Restrict the success path to REQUEST_DISPATCHED or model the expected rejections.

Trecek (commented):

src/autoskillit/core/context_admission.py
Trecek: Agreed — fixed in 61e944bf571b4a0e278e3720edb3594ef1f2c118. Closed-epoch acceptance now quarantines authority-source mismatches, records the exact charge, clears retained unresolved accounting, and publishes charge, occurrence, and quarantine effects.

src/autoskillit/core/context_admission.py
Trecek: Agreed — fixed in c6c81d93f46ccdd0ca6cc28b910a22ab72682c44. No-fence closed resolutions now reconcile deducted charge into snapshot active and remaining counts for accepted input, release or rollback, and generation reconciliation; fence-backed rollovers remain authoritative and unchanged.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in af2b0fa134d9d8f8d1c3c2ac48ccb41a65d1de5f. Contract construction now suppresses constructor exception causes at the metaclass boundary, so attacker-controlled serialized field names cannot survive in the exception cause.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Agreed — fixed in 38f173a62dd283410fd0258bcd060e1cc2a2f9ee. ClosedEpochAudit now rejects more than 10,000 terminal occurrence records before canonical sorting, set construction, or ownership traversal, matching the ADR cardinality limit.

src/autoskillit/core/types/_type_context_admission.py
Trecek: Valid observation — flagged for design decision. ConflictRejectedEffect is public but has no established publication semantics. Emitting it changes the closed AdmissionEffect protocol, while removing it changes the public protocol surface, so either direction needs an explicit contract decision.

tests/core/test_context_admission_reducer.py
Trecek: Agreed — fixed in b613e3517962e033d7ffdd57fa6596030dbe3636. Focused tests now cover closed input overage, incomplete canonical span ownership, and output-generation overage, including retained-count clearing and QuarantineRecordedEffect publication.

tests/core/test_context_admission_state_machine.py
Trecek: Agreed — fixed in e81199e67d38d1e4bf39a64aa275417c06178796. The Hypothesis success path now selects only REQUEST_DISPATCHED batches, matching the reducer precondition and preventing false failures for other reachable lifecycle states.

Issue #4267, opened 2026-07-15
Prevent Codex cook startup from re-indexing persistent rollout history


Trecek:
## Problem

`autoskillit cook` with the Codex backend is substantially slower than Claude Code before it reaches Codex's interactive **Review hooks** screen. The delay is already about 20–40 seconds on a profile with accumulated rollout history and grows with that history.

This is not primarily hook generation or rendering latency. AutoSkillit currently combines a fresh per-cook Codex home with the complete persistent rollout archive but without the matching persistent SQLite state. Codex must rebuild that state before the TUI can display startup hook review.

The implementation should remove history-dependent startup work while preserving the original goals of session skill isolation, durable rollout logs, safe configuration validation, and an explicit hook-trust policy.

## Confirmed root cause

The interactive cook launch composes the following behavior:

1. [`_session_cook.py`](https://github.com/TalonT-Org/AutoSkillit/blob/9d029bd4a694fe8abd4be463993e0cd1c7ede15b/src/autoskillit/cli/session/_session_cook.py#L182-L193) creates a new session ID and initializes a fresh session directory.
2. [`session_skills.py`](https://github.com/TalonT-Org/AutoSkillit/blob/9d029bd4a694fe8abd4be463993e0cd1c7ede15b/src/autoskillit/workspace/session_skills.py#L633-L645) runs the backend prelaunch and session-directory setup.
3. [`CodexBackend.setup_session_dir()`](https://github.com/TalonT-Org/AutoSkillit/blob/9d029bd4a694fe8abd4be463993e0cd1c7ede15b/src/autoskillit/execution/backends/codex.py#L1052-L1103) copies config, links auth, and links the new home's `sessions/` directory to `default_log_dir()/codex-sessions`.
4. [`CodexBackend.build_interactive_cmd()`](https://github.com/TalonT-Org/AutoSkillit/blob/9d029bd4a694fe8abd4be463993e0cd1c7ede15b/src/autoskillit/execution/backends/codex.py#L927-L994) sets `CODEX_HOME` to that new session directory.
5. The setup does not copy or link `state_5.sqlite`, so Codex sees a fresh state database alongside the full historical sessions tree.

The effective layout is:

```text
/dev/shm/autoskillit-sessions/<new-id>/
├── config.toml
├── auth.json -> ~/.codex/auth.json
├── sessions -> ~/.local/share/autoskillit/logs/codex-sessions
└── state_5.sqlite  # created fresh and backfilled by Codex
[...]

PR #4347, opened 2026-07-24
Rectify: Codex Cook Startup History Isolation and Lifecycle Immunity

Trecek:
## Summary

- isolate fresh Codex cook attempts from retained rollout history using managed generated homes and durable per-attempt views
- serialize configuration and rollout lifecycle mutations, validate recovery state, and preserve canonical active/archive representations
- centralize cook process-group ownership, PTY descriptor inheritance, startup tracing, and hard readiness budgets
- add contract, lifecycle, race, recovery, environment, observability, and installed-Codex canary coverage

## Verification

- `AUTOSKILLIT_TEST_FILTER=conservative AUTOSKILLIT_TEST_BASE_REF=develop task test-check` — 28,622 passed
- `AUTOSKILLIT_CODEX_STARTUP_CANARY=1 PYTEST_TEST_PATHS="tests/integration/test_codex_startup_canary.py" task test-all` — 2 passed
- `pre-commit run --all-files` — passed

Closes #4267


Trecek (commented):

src/autoskillit/cli/_init_helpers.py
```diff
@@ -485,12 +487,17 @@ def _register_all(
             backend = get_backend("claude-code")
 
     if backend.capabilities.mcp_config_capable:
-        from autoskillit.cli._hooks_codex import sync_hooks_to_codex_config
-
-        sync_hooks_to_codex_config(
-            hook_config_format=backend.capabilities.hook_config_format,
-        )
+        codex_backend = cast("CodexBackend", backend)
```
Trecek: [warning] arch: Registration repeats the abstraction bypass: a generic mcp_config_capable backend is unconditionally cast to CodexBackend and driven through a Codex transaction. This capability path should dispatch through a backend-owned prelaunch/configuration operation.

src/autoskillit/cli/session/_session_cook.py
```diff
@@ -171,91 +145,267 @@ def cook(
 
     configure_logging()
 
+    project_dir = Path.cwd().resolve(strict=True)
+    launch_id = uuid.uuid4().hex[:16]
     resume_spec = resume_spec_from_cli(resume=resume, session_id=session_id)
+    trace_setting = os.environ.pop(CODEX_STARTUP_TRACE_ENV_VAR, None)
+    if trace_setting not in {None, "1"}:
+        raise ValueError(f"{CODEX_STARTUP_TRACE_ENV_VAR} must be absent or exactly '1'")
+    trace_enabled = trace_setting == "1" and backend.capabilities.session_dir_persistent
```
Trecek: [warning] arch: Generic cook orchestration uses session_dir_persistent as a Codex discriminator: the same path later injects sqlite_home command syntax and constructs a CodexStateReadinessProbe. A persistent non-Codex backend would inherit Codex-specific state and CLI behavior; this should dispatch through a backend-owned hook or dedicated capability.

src/autoskillit/cli/session/_session_cook.py
```diff
@@ -171,91 +145,267 @@ def cook(
...
+                    result = run_cook_attempt(
+                        spec,
+                        pass_fds=pass_fds,
+                        on_spawn=attempt_handle.record_spawn,
+                        on_reaped=attempt_handle.record_reaped,
+                        trace=trace,
+                        observer=observer,
+                    )
+                    reload_session_id = consume_reload_sentinel(project_dir)
+                    _require_observer_ready(observer)
```
Trecek, line 316: [warning] tests: No added test exercises the enabled cook observer path through _startup_observer and _require_observer_ready. The cook test doubles omit version(), causing the compatibility fallback to return no observer, so the fail-closed production wiring can regress without a test failure.

src/autoskillit/cli/session/_session_cook.py
```diff
@@ -171,91 +145,267 @@ def cook(
...
+    """Build the optional Codex PTY observer without leaking trace state to children."""
+    if not enabled:
+        return None
+    from autoskillit.cli.session.pty._observer import (
+        CodexStateReadinessProbe,
+        ObserverStatus,
+        PtyObserver,
+    )
+
+    version_getter = getattr(backend, "version", None)
```
Trecek: [warning] slop: CodingAgentBackend now requires version(), but this getattr/callable fallback silently accepts incomplete backend implementations and disables the readiness observer. That backward-compatibility escape hatch weakens a fail-closed path; call the protocol method directly and let nonconforming backends fail.

src/autoskillit/cli/session/pty/_observer.py
```diff
@@ -0,0 +1,347 @@
...
+            if self.on_hook_review is not None:
+                self.on_hook_review()
+        return chunk
+
+    def check_readiness(self) -> ObserverStatus | None:
+        """Observe readiness once and emit each changed status at most once."""
+        if self.readiness_probe is None:
+            return None
+        status = self.readiness_probe.check()
+        if status is not self.readiness_status:
```
Trecek, line 91: [critical] bugs: Readiness is stored as the latest poll instead of being latched once READY, and relay exits on PTY EIO/EOF without a final poll. If Codex completes the state database between the last poll and process exit, or a transient later poll replaces READY, _require_observer_ready rejects an otherwise successful launch.

src/autoskillit/cli/session/pty/_observer.py
```diff
@@ -0,0 +1,347 @@
...
+            timeout_seconds=timeout_seconds,
+            cancelled=cancelled,
+        )
+        if status is not self.readiness_status:
+            self.readiness_status = status
+            if self.on_readiness is not None:
+                self.on_readiness(status)
+        return status
+
+    def relay(
```
Trecek, line 118: [warning] tests: The new bidirectional PTY relay has no test that sends bytes through the PTY master/stdin paths or verifies terminal and signal restoration. The byte-transparency test calls only observe_output, while the PTY process test sends proof through an unrelated inherited pipe, leaving actual relay behavior uncovered.

src/autoskillit/execution/backends/_codex_session_storage.py
```diff
@@ -0,0 +1,1309 @@
...
+def _decode_mount_path(value: str) -> str:
+    return re.sub(
+        r"\\([0-7]{3})",
+        lambda match: chr(int(match.group(1), 8)),
+        value,
+    )
+
+
+def _filesystem_type(path: Path) -> str:
+    if not sys.platform.startswith("linux"):
```
Trecek, line 207: [critical] bugs: Durable Codex sessions are unconditionally rejected on every non-Linux platform, even though this diff treats Darwin as a supported POSIX canary platform and the filesystem allowlist includes APFS/HFS. Every macOS cook attempt therefore fails in storage-root validation before creating its view.

src/autoskillit/server/_lifespan.py
```diff
@@ -592,9 +608,16 @@ async def _autoskillit_lifespan(server: Any) -> Any:
             and _boot_ctx.backend is not None
             and _boot_ctx.backend.capabilities.mcp_config_capable
         ):
+            codex_backend = cast("CodexBackend", _boot_ctx.backend)
```
Trecek: [warning] arch: The server gates on generic mcp_config_capable and then casts the backend to CodexBackend to access source_codex_home. Any other MCP-config-capable backend would fail here; configuration should dispatch through a backend-owned operation instead of bypassing the backend abstraction.

src/autoskillit/workspace/session_skills.py
```diff
@@ -57,6 +64,107 @@
...
+        cls,
+        path: Path,
+        *,
+        blocking: bool,
+    ) -> _SessionLease | None:
+        path.parent.mkdir(parents=True, mode=0o700, exist_ok=True)
+        path.parent.chmod(0o700)
+        fd: int | None = None
+        try:
+            fd = os.open(path, os.O_RDWR | os.O_CREAT, 0o600)
```
Trecek: [warning] defense: The generated-home lease opens its lock path without proving that .session-leases and the lock file are real non-symlink entries and without O_NOFOLLOW. A stale or replaced symlink can redirect acquisition and invalidate the exclusive-ownership guard used before recursive cleanup.

src/autoskillit/workspace/session_skills.py
```diff
@@ -518,9 +627,171 @@ def init_session(
...
+        self._validate_session_id(session_id)
+        if (
+            session_id in self._session_roots
+            or session_id in self._session_leases
+            or session_id in self._session_skills_subdirs
+        ):
+            raise RuntimeError(f"Session is already owned by this manager: {session_id}")
+
+        persistent = backend is not None and backend.capabilities.session_dir_persistent
+        configured_root = self._codex_root if persistent else self._root
```
Trecek: [warning] arch: Every backend with session_dir_persistent is routed to _codex_root, and the persistent path later requires the hard-coded Codex rollout names sessions and archived_sessions. The nominally generic capability therefore encodes a concrete backend's storage root and layout instead of letting the backend supply them.

tests/integration/test_codex_startup_canary.py
```diff
@@ -0,0 +1,495 @@
...
+    backend = CodexBackend()
+    sequence = 0
+
+    def measure(profile_name: str, *, retain: bool) -> dict[str, object]:
+        nonlocal sequence
+        sequence += 1
+        generated_home = tmp_path / "homes" / f"{sequence:02d}-{profile_name}"
+        generated_home.parent.mkdir(exist_ok=True)
+        _prepare_home(generated_home)
+        assert _rollouts(generated_home) == []
```
Trecek: [warning] tests: The small, many-file, and large-byte profile histories are written under tmp_path/history/<profile>, but measure() launches every sample with a fresh generated home whose rollout directory is asserted empty. The selected histories are never staged or connected to Codex, so all matrix entries exercise the same empty-history startup path.

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments for the required fixes.

src/autoskillit/cli/_init_helpers.py
Trecek: Agreed — fixed in bee41c9e63facd1192e4e542d263ff2a553621ce. Initialization now calls the selected backend ensure_pre_launch contract and propagates preparation errors without a Codex cast.

src/autoskillit/cli/session/_session_cook.py
Trecek: Agreed — fixed in b1f69f521fb1d70981717d799bbabb4e1400b8b9. Cook startup tracing is now gated by the explicit cook_startup_observer_capable backend capability.

src/autoskillit/cli/session/_session_cook.py
Trecek: Agreed — fixed in 4f407d81cfc7584ffc85d663029ec45c3b1d970f. Added enabled-path regression coverage proving the cook observer is constructed and wired when tracing is active.

src/autoskillit/cli/session/_session_cook.py
Trecek: Agreed — fixed in b1f69f521fb1d70981717d799bbabb4e1400b8b9. Removed the reflective version fallback; cook startup now calls the backend version protocol directly.

src/autoskillit/cli/session/pty/_observer.py
Trecek: Agreed — fixed in 271796a5d5cea45529ab0dfa5421314c46b6f115. Readiness is latched, final buffered events are polled during teardown, and terminal attributes plus the signal handler are restored.

src/autoskillit/cli/session/pty/_observer.py
Trecek: Agreed — fixed in 271796a5d5cea45529ab0dfa5421314c46b6f115. Added deterministic bidirectional relay coverage for stdin-to-PTY and PTY-to-stdout behavior.

src/autoskillit/execution/backends/_codex_session_storage.py
Trecek: Agreed — fixed in faca0e7599c1d6a6081399921bfa44c0dabd53cd. macOS durability classification now parses bounded diskutil plist output instead of treating every Darwin filesystem as durable.

src/autoskillit/server/_lifespan.py
Trecek: Agreed — fixed in c4d4cbceaacc8f79dbad847b8f69b435177400a3. Server startup now delegates MCP prelaunch preparation through the backend contract rather than reaching into Codex internals.

src/autoskillit/workspace/session_skills.py
Trecek: Agreed — fixed in 67f65df0ee33b7dee8efa74936d8d8609369e59d. Lease acquisition now uses directory-relative O_NOFOLLOW opens, validates the descriptor type, and rejects symlink substitution.

src/autoskillit/workspace/session_skills.py
Trecek: Agreed — fixed in cdbc4e3bad72de6ca4c7e9b18fd8d177dc24e430. Persistent generated-home layout is now declared by backend conventions and consumed generically by the session manager.

tests/integration/test_codex_startup_canary.py
Trecek: Agreed — fixed in 7d284fcb9479ca6598da45ea92c97ab66a2da481. The startup canary now stages both profile histories into the generated home and asserts profile coverage plus the newly produced rollout.

src/autoskillit/cli/session/_session_picker.py
```diff
@@ -31,42 +50,49 @@ def pick_session(session_type: str, project_dir: Path, project_log_dir: Path) ->
...
-        return []
-
-    return [e for e in entries if not e.get("isSidechain")]
+def _registry_entry(
+    summary: SessionSummary,
+    registry: _Registry,
+) -> Mapping[str, object] | None:
+    if summary.launch_id is not None:
+        return registry.get(summary.launch_id)
+    if summary.backend_name != AGENT_BACKEND_CLAUDE_CODE:
```
Trecek: [warning] arch: The backend-neutral picker now branches on AGENT_BACKEND_CLAUDE_CODE and understands the Claude-specific `claude_session_id` registry field; this diff even allowlists the picker from the no-backend-name guard. Any backend with a different registry correlation scheme therefore requires editing generic CLI code. Carry the correlation identifier in SessionSummary or resolve it through a backend-owned adapter.

src/autoskillit/cli/session/_session_startup_trace.py
```diff
@@ -0,0 +1,374 @@
...
+                return
+            recorded_at = self._now()
+            record = self._record("stage", recorded_at, diagnostics=diagnostics)
+            record.update(stage=stage, attempt=attempt, view_id=view_id)
+            if stage == "spawn" and attempt in self._spawned_attempts:
+                raise RuntimeError(f"startup spawn already recorded for attempt {attempt}")
+            self._append_record(record)
+            if stage == "spawn":
+                self._spawned_attempts.add(attempt)
+                if self._spawn_at is None:
```
Trecek, line 203: [warning] bugs: The cook loop reuses one StartupTrace across reload attempts, but `_spawn_at` and `_hook_review_at` are only set when still None. Attempts after the first therefore run `require_startup_budgets()` against attempt 1's timestamps, allowing a slow or missing reload startup to pass and producing incorrect summary durations. Reset or track these anchors per attempt.

src/autoskillit/cli/session/pty/_observer.py
```diff
@@ -0,0 +1,360 @@
...
+_SUPPORTED_STATE_CONTRACTS = {
+    "codex-cli 0.145.0": _StateReadinessDef(
+        database_name="state_5.sqlite",
+        upstream_commit=_CODEX_STATE_READINESS_COMMIT,
+    )
+}
+
+
+@dataclass(frozen=True, slots=True)
+class CodexStateReadinessProbe:
```
Trecek: [warning] arch: CodexStateReadinessProbe embeds a pinned Codex version, state database filename, and SQLite schema/query contract inside the generic CLI PTY package. Codex storage compatibility changes must now modify the CLI transport layer instead of remaining under execution/backends; expose a generic readiness interface to the PTY layer and keep this adapter with the Codex backend.

src/autoskillit/cli/session/pty/_observer.py
```diff
@@ -0,0 +1,360 @@
...
+            if remaining <= 0:
+                return ObserverStatus.TIMEOUT
+            self._sleep(min(self.poll_interval_seconds, remaining))
+
+
+@dataclass(slots=True)
+class PtyObserver:
+    """Own PTY I/O observation while leaving process-group policy to its caller."""
+
+    readiness_probe: CodexStateReadinessProbe | None
```
Trecek: [warning] cohesion: The otherwise generic PtyObserver relay is concretely coupled to CodexStateReadinessProbe and ObserverStatus, combining terminal transport, output recognition, and Codex SQLite readiness in one class/module. Injecting a small generic readiness-check contract would keep PTY relay behavior cohesive and backend-independent.

src/autoskillit/execution/backends/_codex_config_lock.py
```diff
@@ -0,0 +1,180 @@
...
+        if self._fd is not None:
+            raise RuntimeError("Codex config lock instance is already acquired")
+
+        pid = self._claim_process_ownership()
+        fd: int | None = None
+        acquired = False
+        try:
+            self.lock_path.parent.mkdir(parents=True, exist_ok=True)
+            flags = os.O_RDWR | os.O_CREAT | getattr(os, "O_CLOEXEC", 0)
+            fd = os.open(self.lock_path, flags, 0o600)
```
Trecek, line 113: [warning] defense: The lock sidecar is opened without O_NOFOLLOW or a regular-file fstat check, and `_write_owner_diagnostics` subsequently truncates that descriptor. A pre-existing `.config.toml.autoskillit.lock` symlink can therefore redirect acquisition and overwrite another writable file. Open fail-closed against symlinks and validate the opened inode.

Trecek (commented):

src/autoskillit/execution/backends/claude.py
```diff
@@ -78,6 +82,16 @@
 
 log = logging.getLogger(__name__)  # noqa: TID251 — stdlib fallback: used before configure_logging(); structlog proxy would emit to stderr via import-time WriteLoggerFactory
 
+_ORDER_GREETING_PREFIXES = (
```
Trecek, line 84: [warning] cohesion: The exact `_ORDER_GREETING_PREFIXES` classification table is duplicated from `_session_picker.py`, and both locations independently classify sessions from the prompt. Adding or changing a greeting can make the locator's `session_type_hint` disagree with the picker's fallback; keep one shared classifier/source of truth.

tests/arch/test_capability_consumption.py
```diff
@@ -108,6 +100,17 @@ def test_all_capability_fields_have_production_consumers():
     )
 
 
+def test_hook_trust_policy_has_a_real_production_consumer() -> None:
+    from autoskillit.core import BackendCapabilities, paths
+
+    field_names = frozenset(field.name for field in dataclasses.fields(BackendCapabilities))
+    reads = _collect_attribute_reads(paths.pkg_root(), field_names)
```
Trecek: [warning] tests: The test claims to enforce translation at the interactive launch boundary, but it only requires any `hook_trust_policy` attribute read anywhere under `pkg_root()`. An unrelated log or diagnostic read would keep this passing after the actual launch-policy consumer was removed; assert the expected translator module/call or exercise the launch behavior.

tests/cli/test_interactive_subprocess_contracts.py
```diff
@@ -129,3 +129,146 @@ def test_interactive_subprocess_calls_wrapped_in_terminal_guard(py_file: Path) -
...
+    assert "login_tty(" not in exec_source
+    assert violations == []
+
+
+def test_terminal_and_lease_ownership_are_not_duplicated_across_pty_layers() -> None:
+    session_dir = CLI_DIR / "session"
+    process_source = (session_dir / "_session_process.py").read_text(encoding="utf-8")
+    observer_source = (session_dir / "pty" / "_observer.py").read_text(encoding="utf-8")
+
+    assert "terminal_guard" in process_source
```
Trecek: [warning] tests: This ownership guard uses raw substring presence/absence for `terminal_guard`, `tcsetpgrp`, `LOCK_UN`, and `subprocess.Popen`. Comments, unused imports, or unreachable code can satisfy these assertions without preserving terminal or lease ownership semantics; use the AST/control-flow helpers already defined in this file or a behavioral assertion.

Trecek (commented):

src/autoskillit/cli/session/_session_picker.py
Trecek: Agreed — fixed in 0138fda56. ClaudeSessionLocator now performs the Claude-specific session-ID-to-launch-ID correlation, while the backend-neutral picker resolves registry entries only through SessionSummary.launch_id. The backend-name branch and its architecture-test exemption were removed.

src/autoskillit/cli/session/_session_startup_trace.py
Trecek: Agreed — fixed in 0138fda56. StartupTrace now resets spawn and hook-review anchors at every record_attempt_anchor() call and uses a per-attempt budget start for reloads. A regression test proves attempt 2 fails closed when its own stages are missing and reports only attempt-2 durations.

src/autoskillit/cli/session/pty/_observer.py
Trecek: Agreed — fixed in 0138fda56. The version, state_5.sqlite, schema, and query contract now live in the Codex backend's CodexStateReadinessProbe. The PTY observer consumes only the public core ReadinessProbe/ObserverStatus contract through package gateways.

src/autoskillit/cli/session/pty/_observer.py
Trecek: Agreed — fixed in 0138fda56. PtyObserver is now typed against the backend-neutral ReadinessProbe protocol and no longer defines or imports Codex storage details. The concrete Codex adapter is injected by the cook launch boundary.

src/autoskillit/execution/backends/_codex_config_lock.py
Trecek: Agreed — fixed in 0138fda56. CodexConfigLock now requires O_NOFOLLOW, validates the opened inode with fstat() as a regular file before flock/truncation, and fails closed when no-follow semantics are unavailable. The regression test verifies a symlink sidecar cannot modify its target.

src/autoskillit/execution/backends/claude.py
Trecek: Agreed — fixed in 0138fda56. The picker no longer carries a second greeting-prefix table or classifies prompts itself; it consumes the backend locator's session_type_hint and defaults unclassified summaries to cook. ClaudeSessionLocator remains the single greeting classifier.

tests/arch/test_capability_consumption.py
Trecek: Agreed — fixed in 0138fda56. The guard now parses CodexBackend.build_interactive_cmd and asserts the exact _should_bypass_hook_trust(self.capabilities.hook_trust_policy, automated_session=False) translation call, rather than accepting an arbitrary attribute read elsewhere.

tests/cli/test_interactive_subprocess_contracts.py
Trecek: Agreed — fixed in 0138fda56. The ownership guard now parses both modules and structurally verifies the terminal_guard context, Popen placement, os.tcsetpgrp ownership, absence of fcntl.LOCK_UN in the process owner, and absence of Popen/tcsetpgrp calls in the observer.

src/autoskillit/cli/session/_session_cook.py
```diff
@@ -171,91 +144,243 @@ def cook(
...
+                    add_dirs=[managed_home.skills_dir],
+                    generated_home=managed_home.generated_home,
+                    initial_prompt=current_initial_prompt,
+                    resume_spec=current_resume_spec,
+                    env_extras=cook_env_extras,
+                )
+                final_cmd = built_spec.cmd
+                final_origin = built_spec.origin
+                final_env = dict(built_spec.env)
+                for reserved_key in CODEX_COOK_RESERVED_ENV_VARS:
```
Trecek: [warning] bugs: Cook forces CODEX_HOME and CODEX_SQLITE_HOME into the finalized environment for every backend. Claude cook sessions therefore inherit a bogus Codex home pointing at their Claude-generated directory, which can break Codex commands launched inside them. Gate these variables on the Codex backend or capability.

src/autoskillit/execution/backends/_codex_session_storage.py
```diff
@@ -0,0 +1,1330 @@
...
+            fcntl.flock(instance.fd, operation)
+        except BaseException:
+            os.close(instance.fd)
+            raise
+        owner = {
+            "pid": os.getpid(),
+            "host": socket.gethostname(),
+            "acquired_ns": time.time_ns(),
+        }
+        os.ftruncate(instance.fd, 0)
```
Trecek: [warning] bugs: After acquiring the flock, owner-diagnostic ftruncate/write/fsync runs outside exception cleanup. An I/O failure can leave the descriptor and lock live until traceback or garbage-collection cleanup, stalling later view, thread, or lifecycle operations. Close the descriptor on every post-acquisition failure.

src/autoskillit/execution/backends/codex.py
```diff
@@ -370,57 +457,335 @@ def read_session(self, path: Path) -> list[dict]:
...
+    if not isinstance(transport, dict):
+        transport = actual
+    errors: list[str] = []
+    if transport.get("type", "stdio") != "stdio":
+        errors.append("Codex MCP autoskillit transport is not stdio")
+    if transport.get("command") != expected.get("command"):
+        errors.append("Codex MCP autoskillit command does not match final config")
+    expected_args = expected.get("args", [])
+    actual_args = transport.get("args", [])
+    if list(actual_args or []) != list(expected_args or []):
```
Trecek: [warning] defense: External Codex JSON fields are coerced with list() and set() without validating that args and env_vars are sequences of strings. Malformed or changed probe output can raise TypeError out of validation or produce misleading comparisons instead of a structured fail-closed error. Validate field types before comparison.

Trecek (commented):

tests/execution/backends/test_codex_session_storage.py
```diff
@@ -0,0 +1,700 @@
...
+    )
+
+    context = multiprocessing.get_context("spawn")
+    ready = context.Event()
+    release = context.Event()
+    holder = context.Process(
+        target=_hold_flock,
+        args=(str(store.locks_root / "lifecycle.lock"), ready, release),
+    )
+    holder.start()
```
Trecek, line 391: [warning] tests: The non-daemon holder process and recovery thread are started without a try/finally cleanup path. Any assertion or exception before release.set() leaves the child blocked and the thread waiting on the flock, potentially hanging an xdist worker; always signal, join, and terminate in finally.

tests/execution/backends/test_codex_config_validation.py
```diff
@@ -0,0 +1,165 @@
+"""Interprocess serialization and exact-input Codex validation contracts."""
```
Trecek, line 1: [warning] tests: The diff removes the previous config-validation failure tests, but this replacement file only covers locking and snapshotting. No normal-suite test exercises the new bounded subprocess probe, inventory parser, cache, or real validate_interactive_invocation path; add success, malformed-output, nonzero, timeout, byte-limit, and cache coverage.

Trecek (commented):

src/autoskillit/cli/session/_session_cook.py
Trecek: Agreed — fixed in 2497c9869e1b534c90cb6d3e554b76dbb1824226. Generated homes are now injected only by CodexBackend; Claude cook environments remain free of Codex home variables.

src/autoskillit/execution/backends/_codex_session_storage.py
Trecek: Agreed — fixed in a49cf2fa97223c063d5222ad58bf3339a5dc95ed. _FileLease.acquire now closes its descriptor on every acquisition or diagnostic-write failure, and the regression test proves an immediate nonblocking reacquire.

src/autoskillit/execution/backends/codex.py
Trecek: Agreed — fixed in 974f477c89d70a9b9643080c157a4ad5764effc4. Native Codex inventory args and env_vars must now be JSON arrays of strings; invalid shapes return structured validation errors.

tests/execution/backends/test_codex_session_storage.py
Trecek: Agreed — fixed in 1eb7ec848f431c28ee0e926695046305b73424cf. The lifecycle race test now guarantees daemon recovery, release and join cleanup, with termination fallback in a finally path.

tests/execution/backends/test_codex_config_validation.py
Trecek: Agreed — fixed in 8263ee8bfcaeac2fb882d33124435ef090401047. The normal suite now covers bounded probe success, timeout, byte cap, malformed output, nonzero exit, caching, and the real validator path.

src/autoskillit/execution/backends/_codex_session_storage.py
```diff
@@ -0,0 +1,1331 @@
...
+    try:
+        temporary.unlink()
+    except FileNotFoundError:
+        pass
+    temporary.symlink_to(target)
+    os.replace(temporary, path)
+    _fsync_directory(path.parent)
+
+
+@dataclass(slots=True)
```
Trecek, line 244: [warning] cohesion: _FileLease independently reimplements the same lock-path validation, flock acquisition, descriptor ownership, and release lifecycle as workspace.session_skills._SessionLease and CodexConfigLock. The implementations already differ in directory-fd/inode validation and failure aggregation, so related session leases receive inconsistent safety behavior. Extract one hardened lease primitive with optional owner-diagnostic behavior and reuse it across these call sites.

src/autoskillit/execution/backends/_codex_session_storage.py
```diff
@@ -0,0 +1,1331 @@
...
+    @classmethod
+    def acquire(cls, lock_path: Path, *, nonblocking: bool = False) -> _FileLease:
+        if lock_path.suffix != ".lock":
+            raise ValueError(f"Lock path must use the .lock suffix: {lock_path}")
+        lock_path.parent.mkdir(parents=True, exist_ok=True)
+        _require_real_directory(lock_path.parent, label="lock directory")
+        if _lexists(lock_path) and lock_path.is_symlink():
+            raise RuntimeError(f"Refusing symlink lock file: {lock_path}")
+        instance = cls(path=lock_path)
+        instance.fd = os.open(
```
Trecek, line 258: [warning] defense: The storage lease boundary opens caller-predictable lock paths and immediately flocks, truncates, and writes them without validating the opened descriptor with fstat. O_NOFOLLOW rejects symlinks but still permits a pre-existing FIFO/device or other non-regular inode; require a regular file before mutating owner diagnostics.

src/autoskillit/execution/backends/codex.py
```diff
@@ -383,57 +470,349 @@ def read_session(self, path: Path) -> list[dict]:
...
+    cwd: str,
+) -> _BoundedProbeResult:
+    """Run a normal Codex config-load probe with hard time and byte bounds."""
     try:
-        result = subprocess.run(
-            ["codex", "doctor", "--json"],
-            capture_output=True,
-            text=True,
-            timeout=20,
+        process = subprocess.Popen(
```
Trecek, line 519: [warning] bugs: The bounded validation probe starts no dedicated process group, while timeout cleanup kills only the direct Popen child. If Codex spawns a descendant that outlives or ignores the parent, the probe returns a bounded timeout but leaks that subprocess; launch the probe in its own group and terminate/reap the group on every failure path.

tests/cli/test_cook_interactive.py
```diff
@@ -3,978 +3,454 @@
...
+    backend = _Backend()
+    _install_harness(monkeypatch, tmp_path)
+
+    cli.cook(backend=backend, session_id="thread-explicit")
+
+    assert backend.recover_count == 0
+    assert backend.build_calls[0]["resume_spec"] == NamedResume("thread-explicit")
+
+
+def test_cook_marks_onboarded_only_after_success(
```
Trecek, line 265: [warning] tests: The test named `marks_onboarded_only_after_success` exercises only a successful attempt. The separate nonzero-exit test uses the harness default `first_run=False`, so moving `mark_onboarded()` before the return-code check would leave every test passing. Add a first-run, nonempty-prompt, nonzero-exit case that asserts no onboarding event is recorded.

tests/cli/test_cook_interactive.py
```diff
@@ -3,978 +3,454 @@
...
-            patch("shutil.which", return_value="/usr/bin/claude"),
-            patch("builtins.input", return_value=""),
-            patch("autoskillit.workspace.DefaultSessionSkillManager", return_value=mock_mgr),
-            patch("subprocess.run", return_value=MagicMock(returncode=0)),
-        ):
-            import autoskillit.cli.session._session_cook as module
+    cli.cook()
 
-            module.cook()
+    assert requested
```
Trecek, line 317: [warning] tests: The replacement default-backend test asserts only that `get_backend` was called at least once. It would pass if cook requested the wrong configured backend or resolved it repeatedly, despite removed coverage previously checking the exact name. Assert the complete call list against the configured backend value.

tests/cli/test_cook_interactive.py
```diff
@@ -3,978 +3,454 @@
...
+    assert next(event[1] for event in events if event[0] == "registry") == launch_id
+
+
+def test_cook_does_not_treat_persistent_sessions_as_codex() -> None:
+    """Persistent storage alone must not activate Codex command behavior."""
+    import inspect
+
+    from autoskillit.cli.session._session_cook import cook
+
+    assert "session_dir_persistent" not in inspect.getsource(cook)
```
Trecek: [warning] tests: This guard proves backend neutrality only by checking that one literal token is absent from `inspect.getsource(cook)`. Codex behavior introduced through an alias or helper would pass. Exercise cook with a persistent non-Codex backend and assert that no Codex environment, readiness probe, or storage behavior is activated.

Trecek (commented):

Trecek (commented):

src/autoskillit/execution/backends/_codex_session_storage.py
Trecek: Agreed — fixed in 23b32911b2673fdae131f1661f589725dc2cc09f. `_FileLease.acquire()` now rejects any opened lock inode that is not a regular file before flocking or writing owner diagnostics, with FIFO regression coverage.

src/autoskillit/execution/backends/codex.py
Trecek: Agreed — fixed in c963d2c6186d1f8678746a84d2d92332b15ca6d3. The probe now owns a dedicated process group, kills that group on bounded failures, reaps the leader, and closes streams on unexpected post-spawn errors; regression coverage pins the group signal.

tests/cli/test_cook_interactive.py
Trecek: Investigated — this failure path is already covered by `tests/cli/test_cook_order_command.py::test_cook_mark_onboarded_not_called_on_failure`, which uses a first-run nonempty prompt and nonzero exit, then asserts `mark_onboarded` was not called. Production checks the return code before marking onboarding complete.

tests/cli/test_cook_interactive.py
Trecek: Investigated — the adjacent parametrized `test_cook_routes_through_get_backend` already asserts exactly one `get_backend` call and the exact configured backend value for both supported backends. This is the requested complete-call-list coverage.

tests/cli/test_cook_interactive.py
Trecek: Agreed — fixed in f0a84c1e5. The structural source-token guard is replaced by a behavioral cook test using a persistent non-Codex backend; it canaries Codex readiness/storage entry points and asserts no Codex environment reaches the child spec.

src/autoskillit/execution/backends/_codex_session_storage.py
Trecek: Valid observation — flagged for an architecture decision. `_FileLease`, workspace `_SessionLease`, and `CodexConfigLock` share OS-locking mechanics but intentionally differ in contention, timeout, reentrancy, directory hardening, diagnostics, and failure aggregation. Enforced import boundaries prohibit execution/workspace coupling, so consolidation requires a designed IL-0 primitive and cross-domain migration rather than a local extraction.

PR #4374, opened 2026-07-26
Memoize Skill Capability Classification by Semantic Input

Trecek:
## Summary

Introduce a process-local, thread-safe, weighted LRU around capability-evidence classification in `src/autoskillit/workspace/skill_capabilities.py`. The cache normalizes the effective logical skill name before lookup and uses the exact `(canonical_content, effective_logical_name)` pair as its semantic key, preserving classification behavior while reusing immutable evidence tuples for repeated inputs.

## Requirements

- Cache exact semantic inputs while ensuring content or effective-name changes classify independently.
- Bound retained state by entry count, aggregate accounted bytes, and a per-input admission ceiling.
- Keep full-document scans outside the cache lock and coordinate concurrent cold callers with generation-scoped single-flight state.
- Preserve existing evidence fields, ordering, source spans, authenticity and capability decisions, projection behavior, and resume authority.
- Keep the implementation dependency-free within the IL-1 workspace layer.

### Verification

- `task test-check` — PASS (`15057 passed, 1491 skipped, 27 xfailed`)
- Exact regression manifest — PASS (`293 passed`)
- Cache-focused tests — PASS (`26 passed`)
- `pre-commit run --all-files` — all hooks passed
- Representative benchmark — 2,708 classifier calls reduced to 141 underlying scans across 141 unique semantic keys, with an unchanged findings fingerprint
- General paired benchmark — unchanged 23,753 JUnit test cases; median elapsed time reduced from 306.70s to 199.52s

## Implementation Plan

Plan file: `.autoskillit/temp/make-plan/memoize_skill_capability_classification_plan_2026-07-25_125953.md`

### DIFF
--- a/src/autoskillit/workspace/skill_capabilities.py
+++ b/src/autoskillit/workspace/skill_capabilities.py
@@ -192,7 +192,9 @@ class _SkillCapabilityEvidenceCache:
         with self._lock:
             self._inflight_waiters -= 1
             if state.error is not None:
-                raise state.error
+                raise RuntimeError(
+                    "Capability evidence build failed in another thread"
+                ) from state.error
             result = state.result
             if result is None:
                 raise RuntimeError("Capability evidence build completed without a result")
--- a/src/autoskillit/workspace/skill_capabilities.py
+++ b/src/autoskillit/workspace/skill_capabilities.py
@@ -25,6 +25,9 @@ CapabilityDirection = Literal["outbound", "inbound", "descriptive"]
 CapabilitySourceClassification = Literal["executable", "artifact"]
 _SkillCapabilityEvidenceKey = tuple[str, str]
 
+# Accounted resident payload includes exact key strings, evidence source strings,
+# and a stable policy charge per immutable evidence record. Entry count bounds
+# the remaining fixed per-entry overhead.
 _SKILL_CAPABILITY_EVIDENCE_RECORD_WEIGHT_BYTES = 192
 _SKILL_CAPABILITY_EVIDENCE_CACHE_MAX_ENTRIES = 256
 _SKILL_CAPABILITY_EVIDENCE_CACHE_MAX_BYTES = 16 * 1024 * 1024
--- a/src/autoskillit/workspace/skill_capabilities.py
+++ b/src/autoskillit/workspace/skill_capabilities.py
@@ -7,8 +7,10 @@ whether the source is executable instruction or merely documentary artifact.
 
 from __future__ import annotations
 
-from dataclasses import dataclass
+from collections import OrderedDict
+from dataclasses import dataclass, field
 from functools import cache
+from threading import Event, RLock
 from typing import TYPE_CHECKING, Literal
 
 import regex as re
@@ -21,6 +23,12 @@ if TYPE_CHECKING:
 CapabilityActor = Literal["self", "parent", "external"]
 CapabilityDirection = Literal["outbound", "inbound", "descriptive"]
 CapabilitySourceClassification = Literal["executable", "artifact"]
+_SkillCapabilityEvidenceKey = tuple[str, str]
+
+_SKILL_CAPABILITY_EVIDENCE_RECORD_WEIGHT_BYTES = 192
+_SKILL_CAPABILITY_EVIDENCE_CACHE_MAX_ENTRIES = 256
+_SKILL_CAPABILITY_EVIDENCE_CACHE_MAX_BYTES = 16 * 1024 * 1024
+_SKILL_CAPABILITY_EVIDENCE_CACHE_MAX_INPUT_BYTES = 512 * 1024
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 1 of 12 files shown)


src/autoskillit/workspace/skill_capabilities.py
```diff
@@ -67,6 +78,187 @@ def valid(self) -> bool:
...
+            state.event.wait()
+        except BaseException:
+            with self._lock:
+                self._inflight_waiters -= 1
+            raise
+
+        with self._lock:
+            self._inflight_waiters -= 1
+            if state.error is not None:
+                raise state.error
```
Trecek: [warning] defense: Concurrent waiters re-raise the same mutable exception instance stored by the builder. Each raise rewrites that shared object's traceback, so simultaneous waiters can receive race-dependent tracebacks containing frames from other threads and lose reliable failure context. Raise a per-waiter wrapper chained from the builder failure, or otherwise preserve immutable exception details instead of sharing one exception object.

tests/recipe/test_rules_cmd.py
```diff
@@ -509,21 +513,103 @@ def test_bare_rebase_does_not_match_rebase_abort():
     assert all(f.rule != "run-cmd-bare-rebase-without-conflict-routing" for f in findings)
 
 
-def test_bundled_recipes_have_no_bare_rebase_findings():
+def test_bundled_recipes_have_no_bare_rebase_findings(monkeypatch, record_property):
```
Trecek: [warning] cohesion: This existing bare-rebase semantic-rule test now also instruments the workspace capability cache, verifies scan counts and byte accounting, and emits cache telemetry/fingerprints. Those concerns have independent failure causes and obscure the test's named contract. Split the cache workload assertions into a dedicated recipe/workspace integration test, leaving this test focused on bare-rebase findings.

tests/workspace/test_skill_capabilities_cache.py
```diff
@@ -0,0 +1,710 @@
...
+
+    counts = Counter(name for _, name in scan_calls)
+    assert counts == Counter({"beta": 2, "alpha": 1, "gamma": 1})
+    assert cache.info().entry_count == 2
+    assert cache.info().weight_bytes <= cache.info().max_bytes
+
+
+def test_exact_aggregate_byte_boundary_and_one_byte_overflow(monkeypatch, scan_calls) -> None:
+    content = _document("boundary", "No recognized capability.")
+    key_weight = capabilities._skill_capability_evidence_input_weight_bytes(content, "boundary")
```
Trecek, line 253: [warning] tests: The byte-boundary expectations are derived from the production weight helpers themselves, so a regression that counts characters instead of UTF-8 bytes would still pass. Add an independently calculated non-ASCII case whose character length fits but encoded byte length exceeds max_input_bytes, and assert bypass/no retention.

tests/workspace/test_skill_capabilities_cache.py
```diff
@@ -0,0 +1,710 @@
...
+
+    def fail_after_insertion(self):
+        raise failure
+
+    monkeypatch.setattr(
+        capabilities._SkillCapabilityEvidenceCache,
+        "_evict_if_needed_locked",
+        fail_after_insertion,
+    )
+    with pytest.raises(RuntimeError) as raised:
```
Trecek: [warning] tests: The bookkeeping-failure test exercises only the builder. Scanner failures test waiter release through _publish_failure, but no test covers waiters when _complete_build fails after insertion; omitting its error publication/event signal could deadlock callers while this suite passes. Synchronize a waiter before forcing eviction bookkeeping to raise and assert both futures finish with the failure and inflight counters return to zero.

Trecek (commented):

Trecek (commented):

src/autoskillit/workspace/skill_capabilities.py
Trecek: Agreed — fixed in 90476be3c. Waiters now raise fresh wrappers chained to the builder failure, and the regression test verifies distinct waiter exception identities and preserved causes.

tests/recipe/test_rules_cmd.py
Trecek: Agreed — fixed in a8b596a31 with path-registration follow-up 09d82ebb0. The bare-rebase test now checks only its semantic-rule contract; capability-cache workload assertions and telemetry live in a dedicated recipe/workspace integration test.

tests/workspace/test_skill_capabilities_cache.py
Trecek: Agreed — fixed in 6776dcbaf. The new non-ASCII case independently sets a character-count limit that fits while UTF-8 bytes exceed it, then verifies two uncached scans and zero retained entries/bytes.

tests/workspace/test_skill_capabilities_cache.py
Trecek: Agreed — fixed in 46dd26f84. The bookkeeping-failure regression now synchronizes a waiter before forced eviction failure and verifies both futures complete, the waiter preserves the builder failure as its cause, inflight counters return to zero, and retry succeeds.

src/autoskillit/workspace/skill_capabilities.py
```diff
@@ -67,6 +78,187 @@ def valid(self) -> bool:
...
+                self._entries.move_to_end(key)
+                return entry.evidence, None, False
+
+            state = self._inflight.get(key)
+            if state is not None:
+                self._inflight_waiters += 1
+                return None, state, False
+
+            state = self._new_build_state()
+            self._inflight[key] = state
```
Trecek, line 177: [warning] defense: Every distinct cold key is retained in `_inflight` without an entry or byte limit. Concurrent callers can therefore retain arbitrarily many inputs of up to 512 KiB each, defeating the cache's resource bounds; add bounded in-flight admission or bypass single-flight after a limit.

tests/workspace/test_skill_capabilities_cache.py
```diff
@@ -0,0 +1,710 @@
...
+    monkeypatch.setattr(capabilities, "_SKILL_CAPABILITY_EVIDENCE_CACHE", overflow)
+    capabilities.classify_skill_capability_evidence(content, "boundary")
+    capabilities.classify_skill_capability_evidence(content, "boundary")
+
+    assert overflow.info().entry_count == 0
+    assert overflow.info().weight_bytes == 0
+    assert len(scan_calls) == 3
+
+
+def test_record_charge_and_multi_entry_byte_eviction(monkeypatch, scan_calls) -> None:
```
Trecek, line 303: [warning] cohesion: `test_record_charge_and_multi_entry_byte_eviction` combines two independent contracts: exact evidence-record weight calculation and multi-entry LRU byte eviction. The first setup/result is unused by the eviction half, so unrelated failures share one test; split them into focused accounting and eviction tests.

tests/workspace/test_skill_capabilities_cache.py
```diff
@@ -0,0 +1,710 @@
...
+        charged_content, "charged"
+    )
+    charged_key_weight = capabilities._skill_capability_evidence_input_weight_bytes(
+        charged_content, "charged"
+    )
+    charged_weight = capabilities._skill_capability_evidence_entry_weight_bytes(
+        charged_key_weight,
+        charged_result,
+    )
+    assert charged_weight == (
```
Trecek, line 318: [warning] tests: The retained-weight expectation is calculated with the production `_retained_string_weight_bytes` helper, so an undercounting bug in that helper would affect both implementation and expectation and leave the cache's byte bound untested. Compute expected UTF-8/surrogatepass byte lengths independently in the test.

tests/workspace/test_skill_capabilities_cache.py
```diff
@@ -0,0 +1,710 @@
...
+    assert cache.info().entry_count == 1
+    assert cache.info().weight_bytes == large_weight
+    assert capabilities.classify_skill_capability_evidence(large) is large_result
+    capabilities.classify_skill_capability_evidence(small_a)
+    assert Counter(name for _, name in scan_calls)["a"] == 2
+    assert cache.info().entry_count <= cache.info().max_entries
+    assert cache.info().weight_bytes <= cache.info().max_bytes
+
+
+def test_character_preflight_bypasses_exact_accounting_and_retention(
```
Trecek, line 347: [warning] tests: This covers only the character-count preflight bypass with oversized ASCII. No test drives the separate branch where character count passes but encoded UTF-8 input weight exceeds `max_input_bytes`; a regression could retain multibyte input above the byte budget. Add a multibyte case and assert repeated scans with zero retention.

tests/workspace/test_skill_capabilities_cache.py
```diff
@@ -0,0 +1,710 @@
...
+
+    assert max_active == 2
+    assert evidence_cache.info().entry_count == 2
+
+
+def test_synchronized_warm_callers_reuse_resident_identity(evidence_cache, scan_calls) -> None:
+    content = _document("warm", 'git commit -m "warm"')
+    resident = capabilities.classify_skill_capability_evidence(content)
+
+    with ThreadPoolExecutor(max_workers=8) as executor:
```
Trecek, line 505: [warning] tests: The purported synchronized warm-call test submits workers without a barrier or other overlap control, so the executor may run every lookup serially and the test can pass without exercising concurrent resident-entry access. Synchronize worker entry before releasing the lookups.

Trecek (commented):

Trecek (commented):

src/autoskillit/workspace/skill_capabilities.py
Trecek: Agreed — fixed in 042207452733d86aa6c3f006ed4b902a6e62be06. `_lookup_or_register()` now accounts each distinct admitted key against both the entry and byte budgets; saturated distinct keys scan uncached, while same-key waiters still share the admitted build. The regression test covers entry and byte saturation and verifies in-flight accounting returns to zero.

tests/workspace/test_skill_capabilities_cache.py
Trecek: Agreed — fixed in 0402f1e44b013eb0554cdebf4821f481c0708c9f. The evidence-record weight calculation and multi-entry byte-eviction scenarios are now separate focused tests, so their setup and failure signals are independent.

tests/workspace/test_skill_capabilities_cache.py
Trecek: Agreed — fixed in 8b338b20c82f284d9faea05cd5b337bc0f9e8f8f. The test now uses a non-ASCII evidence source and independently calculates each source byte length with UTF-8 plus surrogatepass, without calling the production retained-weight helper.

tests/workspace/test_skill_capabilities_cache.py
Trecek: Investigated — this is already covered by 6776dcbaf43be8ccf7d32e605f6914a419819dee. `test_non_ascii_input_over_byte_limit_bypasses_cache()` constructs content whose character weight fits while its UTF-8 byte weight exceeds `max_input_bytes`, invokes classification twice, and asserts two scans with zero entries and zero retained bytes. No duplicate test was added.

tests/workspace/test_skill_capabilities_cache.py
Trecek: Agreed — fixed in 4599e7132deae266d7e1f3078b0e3895cb8bccbd. All eight warm-call workers now rendezvous on a barrier with the main thread before lookup release, guaranteeing overlap before resident identity and single-scan assertions run.

Issue #4351, opened 2026-07-25
CODEX_INTAKE_DISCIPLINE_DIGEST contradicts upstream Codex base instructions on reading instruction files


Trecek:
## Problem

`CODEX_INTAKE_DISCIPLINE_DIGEST` instructs Codex sessions to never read a file end-to-end. OpenAI's own base instructions — injected into the same sessions by the provider — instruct the opposite, and specifically forbid partially reading an instruction file. The repo policy is the more recent and more specific text, so the model resolves the conflict in its favour and reads SKILL.md files in fragments.

This is a correctness defect, not only a cost one.

## The conflict, verbatim

**Upstream** (`session_meta.payload.base_instructions.text`, 17,730 chars, present in every Codex session):

> - "If a read is truncated or paginated, **continue until EOF**."
> - "Progressive disclosure applies to selecting relevant resources, **not partially reading a selected instruction file**."
> - "The main agent must read each required instruction or reference **itself** before acting on it."
> - "Do not delegate reading, summarizing, or interpreting skill instructions to a subagent."

**Repo policy** (`src/autoskillit/core/types/_type_constants.py:171-172`):

> - "**Never** read a file end-to-end. Use `rg -n` with context flags, or `sed -n` ranges of at most 250 lines."

They collide hardest on the case upstream names explicitly: **instruction files**.

## Injection is confirmed and near-universal

`codex_discipline_suffix()` (`src/autoskillit/execution/backends/_claude_prompt.py:190-195`) concatenates `OUTPUT_DISCIPLINE_DIGEST` + `CODEX_INTAKE_DISCIPLINE_DIGEST` + the recipe-delivery contract, gated by `include_intake_discipline`. Hardcoded `True` at `codex.py:800-801, 914-915`; hardcoded `False` at `claude.py:579-580, 670-671`. No feature flag.

Measured presence of the clause across 2026-07-20 → 07-24:

| Scope | Presence |
|---|---|
| Full window | 747/775 (96%) |
| 07-22 / 07-23 / 07-24 (98% of volume) | 99% / 99% / 100% |
| Pre-digest control, 07-10 → 07-16 (added 07-17, `38e08ad35`) | **0/766 (0%)** |

The 0% control confirms the constant is the source.

## Observed harm

Session `rollout-2026-07-24T14-31-29-019f960a-5c24-7fb1-ba2b-3379f96bd6a8.jsonl`:

[...]

Issue #4375, opened 2026-07-26
No mechanism can observe a locally-authored instruction contradicting a provider-authored one (split from #4351)


Trecek:
## Problem

Split out of #4351 during plan review to keep that ticket scoped to its own fix.

#4351 fixes one instance of a locally-authored instruction contradicting OpenAI Codex's shipped `base_instructions`. Its fix makes every injected intake rule declare a machine-checkable `basis`. One permitted basis value is `"upstream-aligned"` — but that is checked only for *citation shape* (the evidence string must look like `openai/codex#27044`). **Nothing verifies that upstream actually says what we claim it says, or that it still says it.**

That is the generalised form of #4351. The defect was never "someone wrote a bad sentence." It was that a locally-authored instruction and a provider-authored instruction can disagree, and no mechanism in this repository can observe the disagreement. Rewording our sentence does not change that; neither does the rule registry on its own.

## The upstream text is obtainable offline and exactly

It is the `session_meta` record's `payload.base_instructions.text` in any Codex rollout JSONL — **17,766 bytes at CLI 0.145.0**, and it already reproduces all four sentences #4351 quotes.

`tests/execution/backends/_conformance_assertions.py:225-249` **already parses that field** — but only one-way, to assert our digest landed inside it. The proposed change turns that same parse into a two-way check.

## Proposed direction

A provider-claim registry: verbatim sentences from the Codex CLI's shipped `base_instructions`, each pinned to the CLI version it was verified against, each bound by subject to the local rules that speak to it. Three layers:

| Layer | Runs | Catches |
|---|---|---|
| Offline contract (`tests/contracts/`) | every `task test-check` | a new or edited rule contradicting a claimed provider position, or claiming alignment with a claim that does not exist |
| Live drift (conformance probe) | when Codex conformance probes run | upstream changing its text under us |
| Version pin (doctor check) | every `autoskillit doctor` | installed CLI moving past the version the claims were verified against |

A drafted plan already exists from the #4351 rectify session and should be treated as input, not gospel — it was written before this was split out:

`.autoskillit/temp/rectify/rectify_codex_intake_discipline_digest_2026-07-26_154500_part_b.md` (in the #4351 run clone)

## Sequencing — read before starting

**This must be sequenced after #4369.** The plan registers a new doctor check (`codex_provider_claims_verified`), which forces:

- `tests/docs/test_doc_counts.py::test_doctor_check_count_is_49` → 50
[...]

Issue #4373, opened 2026-07-26
AGENTS.md "Never commit unless told to do so" contradicts implement-worktree "Commit per phase if possible"


Trecek:
## Problem

Two instruction surfaces give a session opposite commit instructions, and neither states which wins.

**Verified, both present at the cited lines on `develop @ 0abc3d237`:**

- `AGENTS.md:43` — "**Never commit unless told to do so**"
- `src/autoskillit/skills_extended/implement-worktree/SKILL.md:153` — "4. Commit per phase if possible"

A session running `implement-worktree` receives both. `AGENTS.md` is loaded as project instructions on every session; the SKILL.md is loaded when the skill is invoked. The AGENTS.md rule is phrased as an unconditional prohibition with emphasis, which is exactly the shape a literal reader obeys hardest.

## Why it matters

The resolution is presumably "the skill invocation *is* being told to commit" — but that inference is not written anywhere. A model that resolves it the other way silently produces a single squashed commit, or no commits, in a worktree whose plan assumed per-phase commits. The multi-part plan green-gate invariant (`AGENTS.md:37`) depends on parts being independently testable and merged in sequence, which per-phase commits support.

This is the same class the report identifies as its central finding: an unconditional imperative that is correct in its original context and wrong in a context nobody re-scoped it for.

## Proposed direction

Give the AGENTS.md rule an observable precondition and an explicit else-branch, per the shape recommended for #4265:

> observable condition → action → explicit else-branch

Roughly: the prohibition is about *unrequested* commits in interactive sessions; skill-directed commits inside an implementation worktree are the requested case. Whatever the resolution, it must be stated on the surface that currently reads as unconditional — not left as an inference.

Check whether other skills issue commit instructions that collide with the same rule before settling the wording.

## Acceptance

- A session running `implement-worktree` has an unambiguous, written answer for whether to commit per phase.
- The resolution is stated on the `AGENTS.md` side, not only in the skill.
- No remaining instruction surface states an unconditional commit prohibition that a skill then contradicts.

## Source

Codex context-consumption report, 2026-07-26 — §"Unconditional imperatives worth re-scoping". Listed there with no tracking ticket. Related in shape to #4265 (unqualified test rule) and #4351 (digest contradicts upstream).

Issue #4369, opened 2026-07-26
Codex limits pin is stale (0.144.1 vs installed 0.145.0) and its doctor check is warning; one governed constant is already non-functional


Trecek:
## Problem

`CODEX_LIMITS_LAST_VERIFIED_VERSION` is pinned at `0.144.1`. The installed Codex CLI is `0.145.0`. The repo already ships a doctor check whose entire purpose is to warn in this state, and it is warning now.

```
$ grep -n 'CODEX_LIMITS_LAST_VERIFIED_VERSION: tuple' src/autoskillit/execution/backends/_codex_config.py
160:CODEX_LIMITS_LAST_VERIFIED_VERSION: tuple[int, int, int] = (0, 144, 1)
$ codex --version
codex-cli 0.145.0
```

`_check_codex_limits_verified` (`src/autoskillit/cli/doctor/_doctor_runtime.py`) returns `Severity.WARNING` when `ver.parsed > CODEX_LIMITS_LAST_VERIFIED_VERSION`, with the message:

> re-verify the recipe outer-result parser/protected evidence contract, `CODEX_HISTORY_RETENTION_TOKEN_LIMIT` (later history only), and `CODEX_AUTO_COMPACT_LIMIT` against upstream behavior, then bump `CODEX_LIMITS_LAST_VERIFIED_VERSION`

## Why this is not cosmetic

At least one constant the pin governs is already demonstrably not behaving as configured. `CODEX_AUTO_COMPACT_LIMIT` is set to `999_999_999` to disable auto-compaction, yet **131 of the 2,083 July rollout files contain compaction events** (66 top-level sessions, 65 subagent threads):

```
python3 - <<'EOF'
import glob,os
t=s=p=0
for f in glob.glob(os.path.expanduser('~/.local/share/autoskillit/logs/codex-sessions/2026/07/**/rollout-*.jsonl'),recursive=True):
    h=None; c=False
    for l in open(f,errors='ignore'):
        if '"compacted"' in l: c=True
        if h is None and '"session_meta"' in l: h=('thread_spawn' in l) or ('"subagent"' in l)
    if c: t+=1; s+= bool(h); p+= not h
print('compacted',t,'subagent',s,'top-level',p)
EOF
```

[...]

PR #4378, opened 2026-07-27
[FIX] Rectify: Evidence-Bound Injected Instruction Policy

Trecek:
## Summary

**This PR implements PART A ONLY.** Part B was split into issue #4375, which must be sequenced after #4369.

`CODEX_INTAKE_DISCIPLINE_DIGEST` is a 790-byte, six-rule free-prose string at
`src/autoskillit/core/types/_type_constants.py:170-197`, injected into every Codex session through
five delivery surfaces and into no Claude session. Two of its six rules have a cited basis; four do
not. Rule 2 — *"Never read a file end-to-end"* — contradicts both OpenAI Codex's shipped
`base_instructions` (verified verbatim from a first-party rollout) and the sibling
`OUTPUT_DISCIPLINE_DIGEST` concatenated into the *same* injected string two bytes away.

Rewording rule 2 is a one-line diff, and the report's own evidence says it will not hold: between
one in four and one in two closed issues here is later contradicted by a recurrence, and the
dominant cause is aiming at the wrong layer. The wrong layer here is *prose*. The right layer is
**the absence of any structure that could represent, and therefore check, a relationship between
two instruction texts**.

Part A replaces the free-prose constant with an **evidence-bound rule registry** — each rule
carries a machine-checkable `basis`, `evidence`, `evidence_anchor`, `subject`, `exception` and
`path_classes` — and renders the wire text from it. Six guards then make the bug class fail loudly:

| Guard | Bug class it makes impossible |
|---|---|
| G1 — no absolute imperative without a declared exception | *"An unconditional imperative correct in its original context and wrong in a context nobody re-scoped it for"* — the report's central finding, and the generative cause of #4351, #4265 and #4373 |
| G2 — every rule's `basis` resolves to a live capability value, an existing ADR, or a cited issue/doc containing the rule's own anchor | Unmeasured prose (R1's `2`, R4's `2`) frozen into a contract by a substring test |
| G3 — no rule may name a path class the `recipe_read_guard` denies | An unexecutable carve-out — the exact trap the investigation flagged for recipe files |
| G4 — the *composed* suffix must round-trip through `tomllib` | The real gap at `codex.py:847-849`, which inspects only the agent body and never the appended suffix |
| G5 — declared byte budget on the digest and on the composed suffix | Unbudgeted growth in an always-on injection replicated 12× per session |
| G6 — co-injected policy texts must claim disjoint subjects | The `OUTPUT_DISCIPLINE_DIGEST` ↔ intake-digest contradiction, with no reference to upstream needed |

Behavioural verification needs **zero new plumbing**: the digest header carries its own version at
the wire (`Context Intake Discipline v2:`), and `sessions.jsonl` already stores `codex_log`, a
direct path to each Codex session's rollout. Bumping the version to 2 *is* the cohort key. Part A
ships a reproducible measurement script so re-measuring the 54.6% repeat-read rate is a command,
not archaeology.

## Scope Note

This PR is **Part A only** of a two-part rectification. Part B — the remainder of the
rectification — was split out into issue #4375, which must be sequenced **after** #4369.

Closes #4351

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/fix-4351-intake-digest-20260726-144540-308354/.autoskillit/temp/rectify/rectify_codex_intake_discipline_digest_2026-07-26_154500_part_a.md`

### DIFF
--- a/scripts/measure_codex_read_repetition.py
+++ b/scripts/measure_codex_read_repetition.py
@@ -1,0 +1,259 @@
+#!/usr/bin/env python3
+"""Measure the Codex intake-discipline repeat-read rate across rollout sessions (#4351).
+
+Three repeat-read definitions have been published for this signal; this tool
+implements the REPORT method, and states here how the other two differ:
+
+  - report method (implemented here): counts, per session, only bounded-read
+    command shapes whose *leading* verb (before any pipe) is `sed -n`,
+    `head`/`tail` applied directly to a path, `nl -ba | sed -n`, or `rg -n`.
+    A trailing `| head -c N` output-safety wrapper is NOT itself a
+    bounded-read signal — nearly every command in this harness carries one,
+    so counting it naively would classify the vast majority of all commands
+    as "reads" and destroy the signal. A repeat is the 2nd+ bounded read of
+    the same resolved path within one session (rollout file).
+  - leading-command method (2026-07-24, 60-rollout sample): the same
+    leading-shape classifier described above, applied to a smaller, earlier
+    sample. This tool reproduces that method exactly at full corpus size, so
+    its output supersedes that one-off run.
+  - #4351 issue-body method (22,282 exec calls, 07-22 -> 07-24): counted
+    every exec_command call as a "read" with no bounded-shape filter and no
+    repeat-of-same-file grouping. It is not reproduced here because it does
+    not isolate repeat reads of a file already resident in context, which is
+    the friction signal this tool exists to measure.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import re
+from collections import defaultdict
+from pathlib import Path
+from typing import Any
+
+DEFAULT_LOG_ROOT = Path("~/.local/share/autoskillit/logs/codex-sessions").expanduser()
+DEFAULT_SESSIONS_INDEX = Path("~/.local/share/autoskillit/logs/sessions.jsonl").expanduser()
+DEFAULT_OUT = Path(".autoskillit/temp/codex_read_repetition_report.json")
+
+# Leading-command bounded-read shapes. Anchored at ^ so a trailing
+# `| head -c N` safety wrapper elsewhere in the command does not qualify it.
+_LEADING_SED_N_PAT = re.compile(r"^\s*(?:\{\s*)?sed\s+-n\s+'[^']*'\s+\S+")
+_LEADING_HEAD_PAT = re.compile(r"^\s*(?:\{\s*)?head\s+-[cn]\s*\d+\s+\S+")
+_LEADING_TAIL_PAT = re.compile(r"^\s*(?:\{\s*)?tail\s+-[cn]\s*\d+\s+\S+")
+_LEADING_NL_BA_PAT = re.compile(r"^\s*(?:\{\s*)?nl\s+-ba\s+\S+\s*\|\s*sed\s+-n\b")
+_LEADING_RG_N_PAT = re.compile(r"^\s*(?:\{\s*)?rg\s+(?:-\S+\s+)*(-n\b|--line-number\b)")
+
+_BOUNDED_PATTERNS = (
+    _LEADING_SED_N_PAT,
+    _LEADING_HEAD_PAT,
+    _LEADING_TAIL_PAT,
+    _LEADING_NL_BA_PAT,
+    _LEADING_RG_N_PAT,
+)
+
+_PATH_TOKEN_PAT = re.compile(r"(?:[./~][\w./\-]+|[\w\-]+/[\w./\-]+)")
+
+# The rarer custom_tool_call/exec shape embeds the command in a JS-template
+# string: tools.exec_command({cmd:"..."}).
+_CUSTOM_TOOL_CALL_CMD_PAT = re.compile(
+    r"tools\.exec_command\(\{\s*cmd\s*:\s*[\"']((?:[^\"'\\]|\\.)*)[\"']"
+)
+
+_VERSION_V2_MARKER = "Context Intake Discipline v2:"
+_VERSION_V1_MARKER = "Context Intake Discipline v1:"
+
+_ROLLOUT_DATE_PAT = re.compile(r"(\d{4})/(\d{2})/(\d{2})/rollout-")
+
+
+def is_bounded_read(cmd: str) -> bool:
+    """Return True when cmd's *leading* command (before any pipe) is a bounded file read."""
+    return any(pattern.search(cmd) for pattern in _BOUNDED_PATTERNS)
+
+
+def extract_target_path(cmd: str) -> str | None:
+    """Best-effort extraction of the file path a bounded-read command targets."""
+    m = re.search(r"sed\s+-n\s+'[^']*'\s+([^\s|;]+)", cmd)
+    if m:
+        return m.group(1)
+    m = re.search(r"\brg\s+[^|;]*?(-n\b|--line-number\b)[^|;]*", cmd)
+    if m:
+        # Strip a trailing `2>&1` (or `2>/dev/null`, `>&2`, ...) redirect token
+        # before tokenizing -- an unstripped redirect was mis-attributed as a
+        # path in 8.5% of matches during calibration.
+        segment = re.sub(r"\d*[<>]&?\d*(/\S+)?\s*$", "", m.group(0)).strip()
+        tokens = [t for t in segment.split() if not t.startswith("-")]
+        candidates = [t for t in tokens if t != "rg" and not re.match(r"^\d*[<>]", t)]
+        if candidates:
+            return candidates[-1].strip("'\"")
+    m = re.search(r"\b(?:head|tail)\s+-[cn]\s*\d+\s+([^\s|;]+)", cmd)
+    if m:
+        return m.group(1)
+    m = re.search(r"\bnl\s+-ba\s+([^\s|;]+)", cmd)
+    if m:
+        return m.group(1)
+    tokens = [t for t in _PATH_TOKEN_PAT.findall(cmd) if len(t) > 3]
+    return max(tokens, key=len) if tokens else None
+
+
+def classify_rollout_records(rollout_path: Path) -> tuple[list[tuple[int, str]], int]:
+    """Return (exec_commands, unclassified_count) for one rollout JSONL file.
+
+    Malformed JSONL lines are skipped, never fatal. A record that is
+    exec-shaped (function_call/exec_command or custom_tool_call/exec) but
+    whose command could not be extracted increments the unclassified count
+    instead of silently vanishing.
+    """
+    commands: list[tuple[int, str]] = []
+    unclassified = 0
+    with rollout_path.open("r", encoding="utf-8") as f:
+        for idx, line in enumerate(f):
+            line = line.strip()
+            if not line:
+                continue
+            try:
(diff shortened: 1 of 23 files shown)


Trecek (commented):

scripts/measure_codex_read_repetition.py
```diff
@@ -0,0 +1,259 @@
...
+    worst_overall.sort(key=lambda row: -row[2])
+    return {
+        "cohorts": cohorts,
+        "worst_offenders": worst_overall[:20],
+        "unclassified_record_count": total_unclassified,
+    }
+
+
+def _find_rollouts(log_root: Path, since: str | None, until: str | None) -> list[Path]:
+    paths = sorted(log_root.glob("*/*/*/rollout-*.jsonl"))
```
Trecek: [critical] bugs: _find_rollouts globs log_root.glob("*/*/*/rollout-*.jsonl") (three wildcard directory levels), but the real on-disk Codex rollout layout is only two levels deep: codex-sessions/YYYY/MM/rollout-*.jsonl (confirmed by tests/execution/backends/test_codex_session_storage.py, e.g. Path("2026/07/rollout-resume.jsonl"), and by _codex_session_storage.py's promotion logic). Against a real log directory this glob matches zero files, so the tool silently reports rollouts_scanned=0 and measures nothing, with no error surfaced.

scripts/measure_codex_read_repetition.py
```diff
@@ -0,0 +1,259 @@
...
+# The rarer custom_tool_call/exec shape embeds the command in a JS-template
+# string: tools.exec_command({cmd:"..."}).
+_CUSTOM_TOOL_CALL_CMD_PAT = re.compile(
+    r"tools\.exec_command\(\{\s*cmd\s*:\s*[\"']((?:[^\"'\\]|\\.)*)[\"']"
+)
+
+_VERSION_V2_MARKER = "Context Intake Discipline v2:"
+_VERSION_V1_MARKER = "Context Intake Discipline v1:"
+
+_ROLLOUT_DATE_PAT = re.compile(r"(\d{4})/(\d{2})/(\d{2})/rollout-")
```
Trecek: [critical] bugs: _ROLLOUT_DATE_PAT assumes a YYYY/MM/DD/rollout- three-component date path, but real rollout paths are YYYY/MM/rollout-*.jsonl (two components, no day). Even after fixing the glob depth at L222, this pattern will never match, so --since/--until date filtering silently falls through to "include everything" instead of actually filtering.

scripts/measure_codex_read_repetition.py
```diff
@@ -0,0 +1,259 @@
...
+    """Return True when cmd's *leading* command (before any pipe) is a bounded file read."""
+    return any(pattern.search(cmd) for pattern in _BOUNDED_PATTERNS)
+
+
+def extract_target_path(cmd: str) -> str | None:
+    """Best-effort extraction of the file path a bounded-read command targets."""
+    m = re.search(r"sed\s+-n\s+'[^']*'\s+([^\s|;]+)", cmd)
+    if m:
+        return m.group(1)
+    m = re.search(r"\brg\s+[^|;]*?(-n\b|--line-number\b)[^|;]*", cmd)
```
Trecek: [critical] bugs: The rg path-extraction regex r"\brg\s+[^|;]*?(-n\b|--line-number\b)[^|;]*" stops at the first literal | character, including one inside a quoted ripgrep pattern. This repo's own AGENTS.md documents `rg -n 'foo|bar'` (POSIX ERE alternation) as the idiomatic search invocation; for that command shape, extract_target_path returns a bogus token extracted from inside the pattern (e.g. "foo") instead of the actual file path, silently corrupting the repeat-read grouping key.

scripts/measure_codex_read_repetition.py
```diff
@@ -0,0 +1,259 @@
...
+    """Return (exec_commands, unclassified_count) for one rollout JSONL file.
+
+    Malformed JSONL lines are skipped, never fatal. A record that is
+    exec-shaped (function_call/exec_command or custom_tool_call/exec) but
+    whose command could not be extracted increments the unclassified count
+    instead of silently vanishing.
+    """
+    commands: list[tuple[int, str]] = []
+    unclassified = 0
+    with rollout_path.open("r", encoding="utf-8") as f:
```
Trecek, line 120: [warning] defense: In classify_rollout_records, a json.JSONDecodeError on a raw JSONL line is swallowed with a bare `continue` and no counter increment, unlike the exec-shaped-but-unextractable branch below it (which increments `unclassified`). A corrupted line that would have been an exec_command record vanishes silently instead of being surfaced via the unclassified count.

scripts/measure_codex_read_repetition.py
```diff
@@ -0,0 +1,259 @@
...
+    try:
+        parsed = json.loads(raw_args)
+    except (json.JSONDecodeError, TypeError):
+        return None
+    return parsed.get("cmd") if isinstance(parsed, dict) else None
+
+
+def classify_policy_cohort(rollout_path: Path) -> str:
+    """Return 'v2', 'v1', or 'none' from the intake-discipline header seen on the wire."""
+    text = rollout_path.read_text(encoding="utf-8", errors="ignore")
```
Trecek: [warning] defense: classify_policy_cohort() calls rollout_path.read_text(...) with no exception handling, invoked once per rollout file in a loop over potentially thousands of files. A single unreadable file raises an uncaught OSError/UnicodeError that aborts the entire batch measurement with no indication of which file failed, unlike the per-line JSONDecodeError handling just above it.

scripts/measure_codex_read_repetition.py
```diff
@@ -0,0 +1,275 @@
...
+def _find_rollouts(log_root: Path, since: str | None, until: str | None) -> list[Path]:
+    paths = sorted(log_root.glob("*/*/rollout-*.jsonl"))
+    if since is None and until is None:
+        return paths
+    filtered = []
+    for p in paths:
+        m = _ROLLOUT_DATE_PAT.search(p.as_posix())
+        date = "-".join(m.groups()) if m else None
+        if date is not None:
+            if since is not None and date < since:
```
Trecek: [critical] bugs: `_find_rollouts` compares the month-precision date string extracted by `_ROLLOUT_DATE_PAT` (`"YYYY-MM"`, since the real on-disk layout has no day component) against `--since`/`--until` via plain Python string comparison — but the argparse help text for both flags documents day-precision examples (`"2026-07-18"`). For any day-precision `--since` value in the target month, `"YYYY-MM" < "YYYY-MM-DD"` is always `True` (a string always sorts less than a longer string it prefixes), so `--since 2026-07-18` silently excludes the *entire* month of July, including rollouts from July 19-31 that should pass the filter. Symmetrically, `"YYYY-MM" > "YYYY-MM-DD"` is always `False`, so `--until 2026-07-18` silently *includes* the full month rather than bounding it at the 18th. This is a real, demonstrable regression introduced by this exact fix commit (verified directly: `"2026-07" < "2026-07-18"` → `True` in Python) — before this diff, the date pattern never matched real 2-level paths at all, so filtering was a no-op; now it matches but at the wrong precision for the documented CLI usage. `test_find_rollouts_date_filter_actually_filters` only exercises month-precision since/until values (`"2026-07"`), which masks this exact defect. Suggested fix: normalize `since`/`until` to `YYYY-MM` precision before comparing (e.g. `since[:7]`), or validate/reject day-precision input and correct the help text to require month-only input.

Trecek (commented):
AutoSkillit PR Review (round 2) — Verdict: changes_requested

Trecek (commented):
AutoSkillit review (round 3) — Verdict: approved (posted as COMMENT, not APPROVE — GitHub disallows self-approval on this self-authored PR). Verified the round-2 day-precision date-filter fix (commit 7d59e278e) is correct and complete: manually traced `_date_within_bound` against all boundary cases and bisected the rewritten test against pre-fix logic to confirm it genuinely pins the regression, corroborated by two independent audit subagents. Round-1 fixes (97d8e8b84) remain intact; the swallowed-JSONDecodeError item remains the deliberate DISCUSS deferral, correctly left unaddressed. No critical/warning findings — only two info-level test-coverage suggestions (non-blocking). task test-check: 30702 passed, 0 failures (an initial run hit an unrelated /dev/shm xdist-contention flake, traced and confirmed not caused by this PR; clean re-run passed). pre-commit run --all-files: clean. This PR implements Part A of #4351 only; Part B is tracked separately as #4375.

scripts/measure_codex_read_repetition.py
Trecek: Fixed in `a44ae9173` (branch commit `97d8e8b84`, squashed on merge). The glob is now `log_root.glob("*/*/rollout-*.jsonl")` at `scripts/measure_codex_read_repetition.py:252` — two wildcard directory levels (`YYYY/MM/`), matching the real on-disk layout rather than three.

scripts/measure_codex_read_repetition.py
Trecek: Fixed in `a44ae9173` (branch commit `7d59e278e`, squashed on merge). `_ROLLOUT_DATE_PAT` no longer parses the directory path. It is now `re.compile(r"rollout-(\d{4}-\d{2}-\d{2})T")` at line 75 and is matched against `p.name`, so the date is derived at day precision from the filename and does not depend on directory depth at all.

scripts/measure_codex_read_repetition.py
Trecek: Fixed in `a44ae9173` (branch commit `97d8e8b84`, squashed on merge). The extraction regex is now built from `_QUOTED_OR_UNPIPED` (line 89), which is alternation-safe and no longer truncates at the first literal `|`. A leading-command guard `_LEADING_RG_N_PAT` (line 45) was added alongside it.

scripts/measure_codex_read_repetition.py
Trecek: Adjudicated **DISCUSS — deliberately not changed**, and recorded as such in the programme handoff rather than silently dropped.

Counting the swallowed `json.JSONDecodeError` into `unclassified` would flip `test_unparseable_records_are_skipped_not_fatal`, which asserts `unclassified == 0` on purpose. It would also conflate two different things: a *record I could not parse* and a *read I could not classify*. `unclassified` means the latter.

If this visibility is wanted later, the correct shape is a distinct `skipped_unparseable` counter reported alongside `unclassified`, not a merge of the two. Judged not worth its own ticket at the time; the reasoning is preserved in the handoff so it is not re-litigated from scratch.

scripts/measure_codex_read_repetition.py
Trecek: Fixed in `a44ae9173`. `classify_policy_cohort()` now wraps the read:

```python
try:
    text = rollout_path.read_text(encoding="utf-8", errors="ignore")
except OSError:
    return "none"
```

with a docstring stating the intent explicitly — one unreadable file in a batch of thousands must not abort the whole measurement run.

scripts/measure_codex_read_repetition.py
Trecek: Fixed in `a44ae9173` (branch commit `7d59e278e`, squashed on merge) by the same change that resolved the `_ROLLOUT_DATE_PAT` thread. The comparison is no longer month-precision: the date is extracted at day precision from the filename, so `_date_within_bound` compares like with like.

PR #4384, opened 2026-07-27
Rectify: Codex CLI Limits Pin — From "Someone Looked" to "Here Is What They Found"

Trecek:
## Summary

`CODEX_LIMITS_LAST_VERIFIED_VERSION = (0, 144, 1)` is a bare tuple with a prose comment. The installed CLI is `codex-cli 0.145.0`. Re-verification against upstream `rust-v0.145.0` (peeled commit `25af12f7e61572b0bc18ddb1008be543b91519b0`) is complete and produced one positive, one negative, and one unverifiable result. The deliverable is **verification, a pin bump, and durable recording of what was checked — including the negative findings**.

Closes #4369

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/impl-4369-20260726-193308-115864/.autoskillit/temp/rectify/rectify_codex_limits_pin_verification_record_2026-07-26_211500.md`

### DIFF
--- a/docs/decisions/0004-recipe-redelivery.md
+++ b/docs/decisions/0004-recipe-redelivery.md
@@ -77,5 +77,19 @@ page carrying `next_part`. They must not guess or repair a malformed continuatio
 - Future work relaxing `CODEX_AUTO_COMPACT_LIMIT` must add integration tests verifying
   `load_recipe` / `open_kitchen` / `get_recipe_section` re-delivery restores full
   pipeline execution capability, including envelope re-delivery plus per-step pulls.
-- `test_copied_config_has_auto_compact_limit` validates the primary defense path:
+- `test_snapshotted_config_has_auto_compact_limit` validates the primary defense path:
   `setup_session_dir` preserves the override in the copied `config.toml`.
+
+## Re-verification at codex-cli 0.145.0 (2026-07-26)
+
+Issue #4369 re-verified the primary defense against upstream `rust-v0.145.0`
+(commit `25af12f7e61572b0bc18ddb1008be543b91519b0`). `ModelInfo::auto_compact_token_limit()`
+clamps the configured `model_auto_compact_token_limit` to 90% of the resolved context
+window; for gpt-5.6-sol (`resolved_context_window = 272_000`) the clamped, effective
+threshold is **244,800** — not the configured `999_999_999`. **`CODEX_AUTO_COMPACT_LIMIT`
+is measurably neutralized upstream at this CLI version**: 34 compaction events were
+observed under 0.145.0. The forward obligation recorded above — testing the
+`load_recipe` / `open_kitchen` / `get_recipe_section` end-to-end recovery path — is
+therefore live, not hypothetical, and is owned by #4271. The finding is recorded
+machine-readably in `CODEX_LIMIT_VERIFICATION_REGISTRY["CODEX_AUTO_COMPACT_LIMIT"]`
+(`execution/backends/_codex_config.py`).
--- a/docs/decisions/0004-recipe-redelivery.md
+++ b/docs/decisions/0004-recipe-redelivery.md
@@ -15,7 +15,8 @@ Two defenses exist:
 1. **Primary — disable compaction**: `ensure_codex_mcp_registered` writes
    `model_auto_compact_token_limit = 999_999_999` (an unreachable threshold) to
    `~/.codex/config.toml`. `setup_session_dir` copies this config into each session
-   directory, ensuring every headless session inherits the setting.
+   directory, ensuring every headless session inherits the setting. *Re-verified at
+   codex-cli 0.145.0 and found neutralized upstream — see "Re-verification" below.*
 
 2. **Guard — block raw recipe reads**: The `recipe_read_guard.py` PreToolUse hook
    blocks `run_cmd`/`Bash` from reading recipe YAML, SKILL.md, or agent definition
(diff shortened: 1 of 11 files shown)


Trecek (commented):

src/autoskillit/execution/backends/_codex_config.py
```diff
@@ -150,14 +150,192 @@ def codex_recipe_delivery_calling_contract(*, mcp_prefix: str = "") -> str:
...
+                f"{key}: status upstream_neutralized requires configured_value and "
+                "upstream_effective_value to both be set and differ"
+            )
+    elif entry.status == "locally_unreachable":
+        unreachable = (
+            entry.upstream_effective_value is None
+            and entry.configured_value is None
+            and entry.codex_config_key is None
+        )
+        if not unreachable:
```
Trecek, line 222: [warning] defense: The if/elif chain over entry.status (upstream_honored / upstream_neutralized / locally_unreachable) has no else/final branch that rejects an unrecognized status value. CodexLimitVerificationStatus is only enforced as a Literal at static-analysis time; at runtime a NamedTuple performs no validation, so a value outside the three known strings (e.g. via entry._replace(status=...) with a typo, or a record from external/future data) would silently pass validate_codex_limit_verification with none of the status-specific consistency checks applied. Since this function exists to catch status/data contradictions, an unrecognized status should raise ValueError instead of passing through unconstrained.

docs/decisions/0004-recipe-redelivery.md
```diff
@@ -77,5 +77,19 @@ page carrying `next_part`. They must not guess or repair a malformed continuatio
...
+- `test_snapshotted_config_has_auto_compact_limit` validates the primary defense path:
   `setup_session_dir` preserves the override in the copied `config.toml`.
+
+## Re-verification at codex-cli 0.145.0 (2026-07-26)
+
+Issue #4369 re-verified the primary defense against upstream `rust-v0.145.0`
+(commit `25af12f7e61572b0bc18ddb1008be543b91519b0`). `ModelInfo::auto_compact_token_limit()`
+clamps the configured `model_auto_compact_token_limit` to 90% of the resolved context
+window; for gpt-5.6-sol (`resolved_context_window = 272_000`) the clamped, effective
+threshold is **244,800** — not the configured `999_999_999`. **`CODEX_AUTO_COMPACT_LIMIT`
```
Trecek, line 90: [warning] cohesion: The new "Re-verification at codex-cli 0.145.0" section states CODEX_AUTO_COMPACT_LIMIT is "measurably neutralized upstream" (effective threshold 244,800, 34 compaction events observed), but the ADR's own unmodified Context section still calls it "Primary — disable compaction" and "an unreachable threshold." The appended section contradicts the body of the same document without updating the superseded framing.

docs/decisions/0005-output-budget-protocol.md
```diff
@@ -198,7 +198,13 @@ insertion control, not a measurement of remaining context.
...
   (`CODEX_AUTO_COMPACT_LIMIT`) against the upstream registry AND observed session
   windows, then bump `CODEX_LIMITS_LAST_VERIFIED_VERSION`. Doctor Check 39
-  (`codex_limits_verified`) mechanizes the reminder. Issue #4280's investigation
+  (`codex_limits_verified`) mechanizes the reminder. `CODEX_LIMIT_VERIFICATION_REGISTRY`
+  is the durable, machine-readable record of what was checked and found;
+  `CODEX_LIMITS_LAST_VERIFIED_VERSION` is derived from it as the minimum
+  `checked_at_cli_version` across its entries. At codex-cli 0.145.0,
+  `CODEX_AUTO_COMPACT_LIMIT` was found neutralized upstream:
+  `ModelInfo::auto_compact_token_limit()` clamps it to 90% of the resolved context
+  window, an effective threshold of 244,800 for gpt-5.6-sol. Issue #4280's investigation
```
Trecek, line 218: [warning] cohesion: This diff adds a note that CODEX_AUTO_COMPACT_LIMIT was found neutralized upstream (effective threshold 244,800) and that CODEX_HISTORY_RETENTION_TOKEN_LIMIT's scope was mischaracterized ("the earlier (later history only) qualifier was factually wrong and is removed" — per the registry finding text). But the untouched "Numeric Limits and Rationale" table further down this same file still reads CODEX_HISTORY_RETENTION_TOKEN_LIMIT as "controls later stored history and does not select the current outer result" and CODEX_AUTO_COMPACT_LIMIT as "Retain the unreachable sentinel" — both superseded claims persist uncorrected in the same document.

src/autoskillit/execution/backends/_codex_config.py
Trecek: Agreed — fixed in 9242b56aad4c206afb19ddf80078ebc6db7b9d0d. Added `else: raise ValueError(f"{key}: unrecognized status {entry.status!r}")` after the elif chain in `validate_codex_limit_verification`, closing the gap where a status value outside the three known Literal members would silently bypass all consistency checks.

docs/decisions/0004-recipe-redelivery.md
Trecek: Agreed — fixed in caa297608e3d76b0ff66231682a97525ed33d532. Appended an in-place forward-pointing annotation to the Context section ("Re-verified at codex-cli 0.145.0 and found neutralized upstream — see \"Re-verification\" below") rather than rewriting the historical framing, following the same precedent ADR-0005 already uses for its own corrections.

docs/decisions/0005-output-budget-protocol.md
Trecek: Agreed — fixed in caa297608e3d76b0ff66231682a97525ed33d532. Added a paragraph to the existing "Corrections of Record" section calling out that both superseded table claims (CODEX_HISTORY_RETENTION_TOKEN_LIMIT's scope framing and CODEX_AUTO_COMPACT_LIMIT's unreachable-sentinel framing) are contradicted by the codex-cli 0.145.0 re-verification, following the section's own established pattern rather than rewriting the table in place.

Issue #4385, opened 2026-07-27
Test conftest feature resolution is fail-open: an unloadable config silently skips feature-gated test classes and reports green


Trecek:
## Problem

`_resolve_test_config()` in `tests/conftest.py` is **fail-open**: it swallows an exception raised while loading the project config and falls back to each feature's `FEATURE_REGISTRY` default. When a feature defaults to `False`, every test class marked `@pytest.mark.feature("<name>")` is then **silently skipped** rather than run.

The consequence is that a test suite can report a large green number locally while whole feature-gated classes never executed, and the same tests run — and fail — in CI, where the config resolves differently.

## Observed

While landing #4303 (PR #4383):

- `task test-check` passed locally on the exact tree, twice, reporting `30781 passed, 0 failed`.
- CI on the same SHA failed with three `AttributeError` failures in `tests/cli/test_doctor_fleet_checks.py`.
- Running that file directly showed **all 30 tests skipped**, not run. The class is `@pytest.mark.feature("fleet")`.
- Forcing the feature on with `AUTOSKILLIT_TEST_FEATURES=fleet` reproduced the CI failure exactly.

The regression itself was real (a refactor moved `load_config` out of `autoskillit.cli.doctor`, breaking three pre-existing `monkeypatch.setattr` targets) and is fixed in `bf330713f`. **This issue is about the masking**, not that regression: the local suite could not see it, and the only signal was a CI failure on a PR.

## Why this matters

The green count is the pipeline gate. `test_check` is documented as a gate, not a diagnostic, and orchestrators route on its boolean. A fail-open feature resolver makes that boolean unsound in exactly the case where it matters most — a developer or headless session whose config differs from CI. There is no warning emitted, no skip reason naming the config failure, and no count of features that fell back.

It also interacts badly with #4303's own thesis: config problems are structurally invisible to this suite. Here a config problem does not merely go untested, it **suppresses unrelated tests** and reports success.

## Root cause

`_resolve_test_config()` catches broadly and returns registry defaults. A config that fails to load is treated as identical to a config that loaded and disabled the feature. Those are different states and only one of them is safe to proceed from silently.

## Proposed remediation

1. Distinguish the two states. On config-load failure, do not fall back silently — emit a loud warning naming the exception and the features whose resolution was affected.
2. Make the skip reason honest. A test skipped because the config could not be read should say so, not report the feature as disabled.
[...]

PR #4389, opened 2026-07-27
Rectify: Test Conftest Feature Resolution — Fail-Open Silent Skip Immunity

Trecek:
## Summary

The test feature gate in `tests/conftest.py` silently skips feature-gated tests when config resolution fails or when `experimental_enabled` resolves differently than expected. Two cooperating root causes drive this behavior: local-vs-CI feature scope delta and fail-open config resolution.

The architectural fix explicitly sets the experimental feature environment variable for `test-check`, makes config resolution fail closed, and emits a feature-scope summary through the terminal writer so diagnostics remain visible even with warnings disabled.

Closes #4385

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/fix-4385-20260727-070042-106829/.autoskillit/temp/rectify/rectify_conftest_feature_resolution_2026-07-27_071500.md`

### DIFF
--- a/src/autoskillit/config/_config_dataclasses.py
+++ b/src/autoskillit/config/_config_dataclasses.py
@@ -154,6 +154,15 @@ class CoreRunConfig:
     def __post_init__(self) -> None:
         if not self.default_model:
             raise ValueError("CoreRunConfig.default_model must not be empty")
+        # Coerce None recipe_overrides entries to empty dicts. YAML sections with
+        # all children commented out (or otherwise empty) parse to None; treating
+        # them as empty matches user intent and avoids spurious validation errors
+        # during collection when a user-level ~/.autoskillit/config.yaml contains
+        # such a section.
+        self.recipe_overrides = {
+            recipe: (overrides if overrides is not None else {})
+            for recipe, overrides in self.recipe_overrides.items()
+        }
         for step, model_val in self.step_overrides.items():
             if not isinstance(model_val, str):
                 raise ValueError(
--- a/src/autoskillit/config/_config_dataclasses.py
+++ b/src/autoskillit/config/_config_dataclasses.py
@@ -547,6 +547,15 @@ class ProvidersConfig:
                         f"profiles[{name!r}][{k!r}] must be a string or null, "
                         f"got {type(v).__name__!r}"
                     )
+        # Coerce None recipe_overrides entries to empty dicts. YAML sections with
+        # all children commented out (or otherwise empty) parse to None; treating
+        # them as empty matches user intent and avoids spurious validation errors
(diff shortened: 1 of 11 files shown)


src/autoskillit/config/_config_dataclasses.py
```diff
@@ -154,6 +154,15 @@ class CoreRunConfig:
     def __post_init__(self) -> None:
         if not self.default_model:
             raise ValueError("CoreRunConfig.default_model must not be empty")
+        # Coerce None recipe_overrides entries to empty dicts. YAML sections with
+        # all children commented out (or otherwise empty) parse to None; treating
+        # them as empty matches user intent and avoids spurious validation errors
+        # during collection when a user-level ~/.autoskillit/config.yaml contains
+        # such a section.
+        self.recipe_overrides = {
```
Trecek, line 162: [warning] defense: The None→{} coercion here is applied only to CoreRunConfig.recipe_overrides, but ProvidersConfig.recipe_overrides (~L550) and AgentBackendConfig.recipe_overrides (~L660) have the identical field shape and validation (`raise ValueError` if `not isinstance(overrides, dict)`). A commented-out `providers.recipe_overrides.<recipe>:` or `agent_backend.recipe_overrides.<recipe>:` YAML section also parses to `None` and will still raise ValueError during config load — the same spurious-validation-error class this fix targets, just via the two sibling dataclasses instead of CoreRunConfig.

Trecek, line 162: [warning] tests: No test covers this new None-coercion logic (mapping None `recipe_overrides` entries to `{}`). `tests/config/test_config.py` has coverage for non-dict and inner-type validation but nothing asserting `CoreRunConfig(recipe_overrides={'impl': None})` normalizes to `{'impl': {}}`.

Trecek (commented):

src/autoskillit/config/_config_dataclasses.py
Trecek: Agreed — fixed in 2daeea9924f17952a310054a2421cfa6b26c17d0. Extended the None->{} recipe_overrides coercion to ProvidersConfig and AgentBackendConfig, which share the identical field shape and validation as CoreRunConfig.recipe_overrides.

Trecek: Agreed — fixed in d6b8ac56914bac433cc22c8a2f718b7b98b9948e. Added None-entry coercion coverage for CoreRunConfig, and matching coverage for ProvidersConfig/AgentBackendConfig since the coercion fix (finding above) extended to those two dataclasses as well.

Issue #4377, opened 2026-07-27
Sessions that completed successfully return failure envelopes — recurrence of closed #477 and #747


Trecek:
## Problem

Sessions that complete their work successfully return failure or retry envelopes. Two independent instances occurred inside a single orchestration run on 2026-07-26. Both are recurrences of previously-closed issues.

This is the exact pattern the Codex context-consumption report names as this repository's dominant failure mode: fixes land, are declared complete, and then fail to hold.

## Instance 1 — `investigate` returns `contract_recovery` after completing

```
step_name: investigate    model: claude-opus-5[1m]
subtype: adjudicated_failure      cli_subtype: success
exit_code: 0                      kill_reason: natural_exit
needs_retry: true                 retry_reason: contract_recovery
has_progress_evidence: true       has_implementation_progress: true
write_call_count: 6               fs_writes_detected: true
session_id: 35f8235a-3219-4696-9d3a-83db3f674828
```

The session wrote a complete 45,433-byte investigation report and independently reproduced two corpus measurements to the digit. Nothing was incomplete. Only the structured-output marker was malformed.

**Predecessors, both CLOSED:**
- **#477** — "adjudicated_failure false positive: make-plan session killed after successful plan write (exit_code=0)"
- **#747** — "Contract recovery nudge: resume session to emit missing structured output token"

## Instance 2 — `prepare_issue` reports failure after creating and labelling the issue

```
success: False
status:  failed
error:   result block contained invalid JSON
subtype: success        exit_code: 0
partial_issue_url:    https://github.com/TalonT-Org/AutoSkillit/issues/4372
partial_issue_number: 4372
session_id: 483e5e91-be89-4348-9b94-e43adf64eb06
```

Issue #4372 was created **and** correctly labelled `enhancement, recipe:implementation`. The triage work completed in full; only the result block failed to serialise. The caller is told the operation failed.

No existing issue was found for this one.

## Why this matters more than a cosmetic status bug

**For `prepare_issue` specifically, the failure mode is duplicate issues.** The documented contract for a failed call is to retry. A retry here files a second copy of an issue that already exists — which directly defeats the dedup step `prepare_issue` exists to perform. The `partial_issue_number` field is the only thing preventing that, and only if the caller reads it.

[...]

Issue #384, opened 2026-03-15
prepare_issue tool fails with 'no result block found' intermittently


Trecek:
## Problem

The `prepare_issue` MCP tool fails with `error: no result block found` when attempting to create and triage issues. The failure is persistent — 3 consecutive calls with the same input all failed.

### Observed Behavior

```
success: False
status: failed
error: no result block found
```

The first call to `prepare_issue` (for issue #382) succeeded, but the next 2 calls (for a different issue) both failed with the same error. The tool launches a headless session via `/autoskillit:prepare-issue` — the session appears to complete but does not produce a parseable result block.

### Expected Behavior

`prepare_issue` should reliably create issues or return a meaningful error.

## Acceptance Criteria

- `prepare_issue` should not fail silently with "no result block found"
- If the headless session fails to produce output, the error should include diagnostic context (session_id, stderr)

PR #4390, opened 2026-07-27
[FIX] Rectify: Sessions That Completed Successfully Return Failure Envelopes

Trecek:
## Summary

The session adjudication pipeline has two independent structural gaps, each allowing completed work to be reported as failed through a different failure mode:

1. **CONTRACT_RECOVERY gate uses a hardcoded string predicate** (`sr.subtype == "adjudicated_failure"` at `_headless_result.py:859`) that structurally excludes UNPARSEABLE sessions from evidence-based recovery, even when `returncode == 0` and write evidence proves the work was done. This produces a hard failure (`needs_retry=False`) instead of a retriable failure at the headless level.
2. **Tool-level parsers independently override adjudication** — when the headless level returns `result.success=True` but the result block's JSON is malformed or absent, `_parse_prepare_result` / `_parse_enrich_result` (`tools_issue_headless.py:209-217`, `:239-247`) return an error dict, and the handler code at lines 360-366 / 488-494 returns a failure envelope via `_build_headless_error_response`, overriding the headless-level success determination.

These are independent failure modes on independent control-flow branches. Root Cause A affects the headless adjudication layer (all session types). Root Cause B affects the tool-level result handling layer (only `prepare_issue` / `enrich_issues`, only when headless adjudication already succeeded).

The immunity approach: (a) replace the single-string gate — at both the sr-mutating site (line 856) AND the parallel audit-log site (line 707) — with a set-based predicate guarded by `returncode == 0` for non-adjudicated_failure subtypes, making the gate self-documenting and trivially extensible; (b) make tool-level result handling produce a degraded-success envelope (not a failure envelope) when `result.success is True` but block parsing fails (covering both "invalid JSON" and "no result block found" sentinels in `_BLOCK_PARSE_ERRORS`).

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/fix-4377-20260727-070043-055029/.autoskillit/temp/rectify/rectify_successful_sessions_returning_failure_envelopes_2026-07-27_070500.md`

Closes #4377

### DIFF
--- a/src/autoskillit/server/tools/tools_issue_headless.py
+++ b/src/autoskillit/server/tools/tools_issue_headless.py
@@ -360,9 +360,16 @@ async def prepare_issue(
             if parsed.get("error") in _BLOCK_PARSE_ERRORS:
                 extra = _extract_partial_issue_data(result.result)
                 return json.dumps(
-                    _build_headless_error_response(
-                        result, error=parsed["error"], extra_fields=extra
-                    )
+                    {
+                        "success": True,
+                        "status": "degraded",
+                        "warning": parsed["error"],
+                        "session_id": result.session_id,
+                        "stderr": result.stderr or "",
+                        "subtype": result.subtype or "",
+                        "exit_code": result.exit_code if result.exit_code is not None else -1,
+                        **extra,
+                    }
                 )
 
             # Block parsed successfully. result.success=True is the authoritative signal —
@@ -488,9 +495,16 @@ async def enrich_issues(
             if parsed.get("error") in _BLOCK_PARSE_ERRORS:
                 extra = _extract_partial_enrich_data(result.result)
                 return json.dumps(
-                    _build_headless_error_response(
-                        result, error=parsed["error"], extra_fields=extra
-                    )
+                    {
+                        "success": True,
+                        "status": "degraded",
+                        "warning": parsed["error"],
+                        "session_id": result.session_id,
+                        "stderr": result.stderr or "",
+                        "subtype": result.subtype or "",
+                        "exit_code": result.exit_code if result.exit_code is not None else -1,
+                        **extra,
+                    }
                 )
 
             return json.dumps(
--- a/src/autoskillit/server/tools/tools_issue_headless.py
+++ b/src/autoskillit/server/tools/tools_issue_headless.py
@@ -38,10 +38,10 @@ _BLOCK_PARSE_ERRORS: frozenset[str] = frozenset(
     {"no result block found", "result block contained invalid JSON"}
 )
 
-# Canonical 7 fields of the headless error envelope. extra_fields cannot overwrite
+# Canonical fields of the headless response envelope. extra_fields cannot overwrite
 # any of these — see _build_headless_error_response for the contract.
 _CANONICAL_KEYS: frozenset[str] = frozenset(
-    {"success", "status", "error", "session_id", "stderr", "subtype", "exit_code"}
+    {"success", "status", "error", "warning", "session_id", "stderr", "subtype", "exit_code"}
 )
 
 # Regex for partial-issue-data extraction: a GitHub issue URL anywhere in the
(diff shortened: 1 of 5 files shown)


src/autoskillit/server/tools/tools_issue_headless.py
```diff
@@ -360,9 +360,16 @@ async def prepare_issue(
             if parsed.get("error") in _BLOCK_PARSE_ERRORS:
                 extra = _extract_partial_issue_data(result.result)
                 return json.dumps(
-                    _build_headless_error_response(
-                        result, error=parsed["error"], extra_fields=extra
-                    )
+                    {
```
Trecek: [warning] arch: Hand-rolled degraded-success response dict duplicates the canonical envelope-construction logic that `_build_headless_error_response()` centralizes. That helper's docstring states: 'Every failure path that derives a response from a SkillResult MUST use this builder. Do not hand-roll error dicts — that pattern caused silent omission of diagnostic fields (issue #384).' This new degraded-success branch reintroduces that exact anti-pattern (here in prepare_issue and identically in enrich_issues at L498) instead of extending the builder (e.g., a status='degraded' mode) to cover the new success-but-degraded case.

src/autoskillit/server/tools/tools_issue_headless.py
```diff
@@ -360,9 +360,16 @@ async def prepare_issue(
...
+                        "success": True,
+                        "status": "degraded",
+                        "warning": parsed["error"],
+                        "session_id": result.session_id,
+                        "stderr": result.stderr or "",
+                        "subtype": result.subtype or "",
+                        "exit_code": result.exit_code if result.exit_code is not None else -1,
+                        **extra,
```
Trecek: [warning] defense: The degraded-success envelope spreads `**extra` (from `_extract_partial_issue_data`) last in a raw dict literal without filtering against `_CANONICAL_KEYS`, unlike `_build_headless_error_response` which explicitly does `resp.update({k: v for k, v in extra_fields.items() if k not in _CANONICAL_KEYS})` to prevent extra_fields from clobbering diagnostic keys. Currently `_extract_partial_issue_data` only emits `partial_issue_url`/`partial_issue_number`, so there's no live collision today, but this path bypasses the established safeguard and relies on an unenforced, untested invariant. The same gap recurs in enrich_issues (L506).

src/autoskillit/server/tools/tools_issue_headless.py
```diff
@@ -488,9 +495,16 @@ async def enrich_issues(
             if parsed.get("error") in _BLOCK_PARSE_ERRORS:
                 extra = _extract_partial_enrich_data(result.result)
                 return json.dumps(
-                    _build_headless_error_response(
-                        result, error=parsed["error"], extra_fields=extra
-                    )
+                    {
```
Trecek: [warning] arch: Duplicate of the hand-rolled degraded-success dict at L363 (prepare_issue) — enrich_issues repeats the identical field literal instead of sharing one helper, the exact anti-pattern `_build_headless_error_response` (and its issue #384 history) was introduced to prevent.

src/autoskillit/server/tools/tools_issue_headless.py
```diff
@@ -488,9 +495,16 @@ async def enrich_issues(
...
+                        "success": True,
+                        "status": "degraded",
+                        "warning": parsed["error"],
+                        "session_id": result.session_id,
+                        "stderr": result.stderr or "",
+                        "subtype": result.subtype or "",
+                        "exit_code": result.exit_code if result.exit_code is not None else -1,
+                        **extra,
```
Trecek: [warning] defense: Same unguarded `**extra` spread as prepare_issue's degraded-success branch (L371) — `_extract_partial_enrich_data`'s output can silently clobber canonical envelope keys since there is no `_CANONICAL_KEYS`-style filter here, unlike `_build_headless_error_response`.

Trecek (commented):

src/autoskillit/server/tools/tools_issue_headless.py
Trecek: Agreed — fixed in 32ff31d38870f41c928342d8f7d1b76290ef117a. Extended `_build_headless_error_response()` with `success`/`warning` keyword params (defaults preserve all existing failure-path call sites) instead of hand-rolling a duplicate dict; both `prepare_issue` and `enrich_issues` degraded-success paths now route through the single builder.

src/autoskillit/server/tools/tools_issue_headless.py
Trecek: Agreed — fixed in 32ff31d38870f41c928342d8f7d1b76290ef117a. Added `warning` to `_CANONICAL_KEYS` and route the degraded-success dict through `_build_headless_error_response()`'s existing `extra_fields` filter (`resp.update({k: v for k, v in extra_fields.items() if k not in _CANONICAL_KEYS})`), so `**extra` can no longer clobber canonical fields in the degraded-success path — matching the guarantee `test_build_headless_error_response_extra_fields_cannot_override_canonical` already enforces for the failure path.

src/autoskillit/server/tools/tools_issue_headless.py
Trecek: Agreed — fixed in 32ff31d38870f41c928342d8f7d1b76290ef117a. Same root cause as the L363 finding — this duplicate hand-rolled dict in `enrich_issues` is now replaced by the same `_build_headless_error_response(..., success=True, status=\"degraded\", warning=..., extra_fields=extra)` call, eliminating the duplication.

src/autoskillit/server/tools/tools_issue_headless.py
Trecek: Agreed — fixed in 32ff31d38870f41c928342d8f7d1b76290ef117a. Same fix as the L371 finding, mirrored at this call site: extra_fields from _extract_partial_enrich_data now passes through the builder's _CANONICAL_KEYS filter (including the newly-protected warning key) instead of an unfiltered **extra spread.

Issue #4392, opened 2026-07-27
bridge_investigation completeness gate proxies on a hard-coded '## Recommendations' heading — halts the pipeline for 15 of 30 marked issues


Trecek:
## Problem

The completeness gate shipped in #4381 (PR #4388, merged to `develop` 2026-07-27) treats
the presence of a `## Recommendations` heading as proof that an `## Investigation` section
is complete. Sections that are complete but do not use that heading are rejected:

```
ValueError: extracted '## Investigation' section lacks '## Recommendations' heading
            — investigation report is truncated
```

Because #4381 also (correctly) changed `bridge_investigation`'s `on_failure` from
`rectify` to `release_issue_failure`, this is no longer a warning — **it halts the
pipeline.**

## Measured, on the merged code

Replayed `smoke_utils.extract_investigation` against the 120 most recent issues, over the
30 that carry `<!-- investigation_complete: true -->` **and** have a real
`## Investigation` section:

```
issues with a real ## Investigation section : 30
OLD awk  — retained <90% of section          : 25/30
NEW code — retained <90% of section          : 15/30
```

The improvement is real: 25 → 15. But **all 15 remaining cases are the same failure**,
and none of them is a truncation — every one raises the `ValueError` above and produces
no output at all.

Affected: #4361, #4356, #4343, #4273, #4271, #4267, #4262, #4261, #4260, #4258, #4257,
#4256, #4237, #4235, #4233.

So #4381's stated acceptance criterion — *"replaying the extraction across the issue
corpus retains ≥90% for every issue carrying the marker; the 35-of-49 figure goes to
zero"* — **is not met**. It was not measured before merge; this run measured it after.

## Why this is not a request to revert

The direction is right and matches #4381's own "make truncation loud" requirement. Silent
truncation was the worse failure. The defect is narrower: **the completeness predicate is
a single hard-coded heading name**, which is a proxy for completeness rather than a
measure of it — the same shape of error as the `test -s` guard it replaced (#3324), one
layer up.

The practical consequence is operational and immediate: roughly **half the
`investigation_complete` backlog can no longer enter the remediation pipeline at all**.
A pipeline that stops is much better than one that plans from a fragment, but this will
be hit on the next such issue processed.

## Proposed direction

1. **Measure completeness, don't proxy it.** #4381's own item 2 asked for "extracted bytes
   versus bytes from the heading to end-of-body, failing below a threshold." That was not
[...]

PR #4388, opened 2026-07-27
[FIX] Rectify: bridge_investigation Silently Truncates Investigation Content

Trecek:
## Summary

The `bridge_investigation` step in `remediation.yaml` uses an inline awk command to extract investigation content from GitHub issue bodies. The awk terminates at the first `## ` heading inside the section being extracted, discarding 96-100% of the investigation report. This is the third manifestation of defects on the same awk line (#3195, #3324, #4381). The `test -s` guard and `type: path` capture validation catch only 0-byte files, not truncation. All four existing tests are structural (check string presence and routing) — none execute the command against realistic data.

The architectural immunity replaces the untestable inline awk with a `run_python` callable in `smoke_utils/`, following the project's established pattern for complex recipe step logic. This makes the extraction unit-testable, structurally validatable, and heading-level-correct by construction.

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/fix-4381-20260727-070041-148642/.autoskillit/temp/rectify/rectify_bridge_investigation_truncation_2026-07-27_072500.md`

Closes #4381

🤖 Generated with [Claude Code](https://claude.com/claude-code) via AutoSkillit

Issue #4387, opened 2026-07-27
audit_impl preflight rejects the plan path the recipe supplies — AUDIT_CYCLE_INVENTORY allowed root is the orchestrator temp dir, never the clone


Trecek:
## Problem

The `audit_impl` step of the `remediation` recipe cannot be invoked at all in a normal
clone-based pipeline. Its `AUDIT_CYCLE_INVENTORY` preflight rejects the plan path the
recipe itself supplies, crashing the session before it starts.

```
AuditCycleVerificationError: artifact containment/read failed: Path escapes allowed root
```

The session never launches: `subtype=crashed`, `exit_code=-1`, `lifespan_started=false`,
`write_call_count=0`. There is no verdict, no audit cycle, and no artifact.

## Mechanism

`src/autoskillit/server/_recipe_execution.py:371`

```python
return factory(snapshot=snapshot, allowed_root=tool_ctx.temp_dir)
```

The audit-cycle verifier's allowed root is the **orchestrator's** `tool_ctx.temp_dir`.
The plan paths it is asked to verify come from `context.all_plan_paths`, which the
recipe populates from `rectify`, whose `output_dir` is `.autoskillit/temp/rectify`
**inside the clone** (`context.work_dir`).

The clone is created by `bootstrap_clone` under `autoskillit-runs/`, which is never
inside the orchestrator's temp dir. So the containment check in
`src/autoskillit/core/path_containment.py:43` can never pass for a cloned pipeline —
the two roots are structurally disjoint.

## Observed

Reproduced on #4377 (remediation, clone-based, `open_pr=true`) on 2026-07-27, with three
different spellings of the same input — all crash identically:

| Attempt | `all_plan_paths` value | Result |
|---|---|---|
| 1 | absolute path in clone `.autoskillit/temp/rectify/` | `Path escapes allowed root` |
| 2 | path relative to `cwd` (the clone) | `Path escapes allowed root` |
| 3 | absolute path, file physically copied under the primary checkout's `.autoskillit/temp/` | `Path escapes allowed root` |

Attempt 3 is the informative one: staging the artifact under what appeared to be the
allowed root still failed, so `tool_ctx.temp_dir` is neither the clone temp nor the
primary checkout's `.autoskillit/temp` — it is some third location that a recipe-declared
plan path can never land in. Verified there are no symlink components on any tested path
and the plan file is a regular file with `st_nlink == 1`.

## Why this matters

[...]

Issue #4386, opened 2026-07-27
investigation_complete marker is substring-matched over the whole body, including fenced code blocks — skips investigate and reaches rectify with an empty investigation


Trecek:
## Problem

`investigation_complete` is resolved by testing whether the marker string appears
**anywhere** in the issue body, with no structural qualification:

`src/autoskillit/server/tools/tools_issue_labels.py:93`

```python
investigation_complete = INVESTIGATION_COMPLETE_MARKER in (result.get("body") or "")
```

A body that merely *mentions* the marker — inside a fenced code block, a quoted
example, or a report describing the marker mechanism — is classified as carrying a
completed investigation. The `investigate` step is then skipped for an issue that has
no investigation at all.

## Observed live, on the issue that documents the sibling bug

While processing **#4381** on 2026-07-27, `claim_and_resolve_issue` returned
`investigation_complete: true`. #4381 has **no `## Investigation` section**. The marker
was matched inside a ```bash fenced block that reproduces the sibling measurement:

```python
MARK = "<!-- investigation_complete: true -->"
```

That single false positive produced the following chain, end to end, with no failure
surfaced to the orchestrator:

| Step | What happened |
|---|---|
| `claim_and_resolve` | `investigation_complete=true` — **false positive** |
| `investigate` | **skipped** — the issue is treated as already investigated |
| `bridge_investigation` | no `## Investigation` heading → extracts **0 bytes** |
| `test -s` guard | fails, so the step reports failure |
| routing | `on_failure: rectify` — **rectify runs anyway**, with an empty investigation path |

So the plan-producing step would have been reached with *nothing*, on an issue whose own
subject is silent corruption of that very input. Verified by hand:

```
$ gh issue view 4381 --json body -q .body | grep -c '^## Investigation'
0
$ wc -c .autoskillit/temp/investigate/investigation_from_issue.md
0
```

## Why this is not #4381

#4381 is the `awk` extraction range in `remediation.yaml` — it fixes *how much* of an
existing `## Investigation` section survives. This defect is upstream of that and lives
in a different file: the **decision to skip `investigate` at all**, made in
`tools_issue_labels.py` before any extraction is attempted. Fixing #4381's regex does not
make a body without an Investigation section stop being classified as investigated, and
the two changes do not touch the same surface.

They compound: this bug decides the investigation is present, #4381 decides how much of
it to pass on, and the shared `on_failure: rectify` route means neither can halt the
pipeline.

## Lineage

[...]

PR #4393, opened 2026-07-27
Fix: completeness gate proxied on a heading name — restore 16 of 34 blocked investigations (#4392)

Trecek:
Closes #4392.

Follow-up to #4381 (PR #4388), found by running that issue's own acceptance test
immediately after it merged.

## What was wrong

#4381 shipped a completeness gate requiring a literal `## Recommendations` heading.
Replayed against the live issue corpus, it rejected **16 of 34** real investigations.
Because #4381 also — correctly — changed `bridge_investigation`'s `on_failure` from
`rectify` to `release_issue_failure`, that is not a warning: those issues can no longer
enter the remediation pipeline at all.

## Two findings that changed the fix

**1. The heading test was never guarding extraction.** Extraction already runs from
`## Investigation` to end-of-body, so the section *cannot* be truncated. The check was
demanding one report vocabulary, not measuring completeness. This also rules out the
byte-ratio check #4381 proposed as remediation — with end-of-body extraction, extracted
always equals heading-to-end-of-body, so the ratio is always 100% and the check is vacuous.

**2. The remaining failures are not truncated reports.** The issues that fail even a
generic structural check use `## Investigation` as an *attestation* — "prior investigation
completed interactively … included **above**" — with the analysis written above the
heading. Extracting downward from the heading gets a three-line note; the content is
upward. Rejecting them was wrong, and so was extracting from the heading down.

## What this does

- Completeness is now **"the section contains at least one `## ` subsection"** — agnostic
  to heading vocabulary, and still rejecting the original defect shape (marker plus a
  one-line note, no subsections).
- An attestation-style section **falls back to handing rectify the whole issue body**,
  which demonstrably contains the analysis. This is exactly what the programme's machinery
  notes already instructed operators to do by hand.
- A hard failure is kept only when neither the section **nor the text above it** carries
  any structure.

The fallback deliberately tests the text *preceding* the heading. Testing the whole body
matches the `## Investigation` heading itself and always passes — caught by a failing test
during development.

## Verification

Acceptance re-measured on the live corpus, which is what #4381 asked for and what this PR
is answerable to:

```
marked issues with an ## Investigation heading : 34
failing / empty  — before this change          : 16
failing / empty  — after  this change          :  0
```

- 11/11 `extract_investigation` unit tests pass, including three new ones: the attestation
  fallback, a structured report with no `## Recommendations` heading, and a `### ` decoy.
- Full suite: **33,024 passed**, 625 skipped, 26 xfailed.
- `pre-commit run --all-files` clean.

## Scope flags

None. Single callable plus its tests; no sibling-ticket surfaces touched. Specifically
does **not** touch `server/_recipe_delivery.py` or `server/tools/tools_recipe.py` (#4362,
actively in progress), `tools_issue_labels.py` (#4386), or the audit-cycle verifier (#4387).

Does **not** revert #4381 — halting loudly is strictly better than the silent truncation it
replaced, and reverting would restore the worse behaviour.

### DIFF
--- a/src/autoskillit/smoke_utils/_investigation.py
+++ b/src/autoskillit/smoke_utils/_investigation.py
@@ -9,9 +9,23 @@ import regex as re
 from autoskillit.core import atomic_write, run_gh
 
 _INVESTIGATION_HEADING = "## Investigation"
-_RECOMMENDATIONS_HEADING = "## Recommendations"
 _INVESTIGATION_HEADING_RE = re.compile(r"^" + re.escape(_INVESTIGATION_HEADING), re.MULTILINE)
-_RECOMMENDATIONS_HEADING_RE = re.compile(r"^" + re.escape(_RECOMMENDATIONS_HEADING), re.MULTILINE)
+# A structured report carries at least one ``## `` subsection. This is a property of the
+# content, not of any particular heading vocabulary — see _is_structured.
+_SUBSECTION_RE = re.compile(r"^## ", re.MULTILINE)
+
+
+def _is_structured(text: str) -> bool:
+    """Whether ``text`` looks like a structured report rather than a bare stub.
+
+    The defect this guards (#4381) produced a preamble of a few lines — a marker plus a
+    one-line note — with no subsections at all. Requiring at least one ``## `` subsection
+    rejects exactly that shape while remaining agnostic about which headings a report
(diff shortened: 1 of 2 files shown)


Trecek (commented):
AutoSkillit review passed. No blocking issues found. (Posted as COMMENT, not APPROVE — GitHub does not allow self-approval on own PR.)

Issue #4321, opened 2026-07-22
mtime-only shell-capture cleanup can unlink an active quiet artifact


Trecek:
## Problem

Shell-capture cleanup uses pathname `mtime` as its only writer-liveness signal. A
harness-managed writer can remain open but quiet for longer than the TTL, so a
sweep can unlink its pathname while the producer still holds the file open.

On POSIX the open inode survives until the descriptor closes, but the pathname
needed by the harness's post-execution measurement and replay disappears. Later
writes are no longer reachable through that path, and an oversized artifact
cannot subsequently provide a stable retrieval path. This contradicts the
cleanup helper's stated liveness assumption and is broken existing behavior, so
it is a remediation.

## Evidence

- `src/autoskillit/hooks/shell_capture_hook.py:102-108` opens the artifact as the
  managed command's stdout/stderr before execution; a running command may remain
  quiet while holding that descriptor.
- `src/autoskillit/hooks/_capture_cleanup.py:9-15` states that the one-hour age
  threshold is the liveness guard and claims an in-flight harness will never be
  swept.
- `src/autoskillit/hooks/_capture_cleanup.py:43-62,83-97` performs filename and
  containment checks, but its only writer-liveness decision is `st_mtime`; it has
  no independent evidence that the harness has finished.
- The registered duplicate sweep in
  `src/autoskillit/hooks/session_start_hook.py:83-98` makes the same mtime-only
  liveness decision.
- After the managed command exits, `shell_capture_hook.py:109-117` reopens the
  pathname for sizing, hashing, and replay. A concurrent unlink can therefore
  break the current wrapper postlude as well as later retrieval.
- `tests/hooks/test_shell_capture_conformance.py:290-305,347-365` backdates
  completed files and invokes the helper directly; it does not cover a live
  writer or the registered production sweep.
- A bounded helper reproduction opened a matching capture file, backdated its
  mtime, and swept it. The helper reported one deletion, the pathname disappeared,
  and the still-open descriptor accepted five more bytes. This proves the
  algorithmic flaw and POSIX behavior; reproducing the production race also
  requires its lifecycle trigger to run concurrently in the same project.

## Expected Behavior

Cleanup must use authoritative liveness evidence independent of `mtime` and must
not remove the path of a harness-managed writer that still holds the capture
open. Time may remain the retention signal after liveness/finalization is proven.
Abandoned captures must still become eligible through a defined, bounded recovery
rule.

## Dependencies

[...]

PR #4405, opened 2026-07-28
Rectify: Codex Shell-Capture Cleanup Lifecycle

Trecek:
## Summary

Adds a durable lifecycle owner for Codex shell-capture artifacts, including writer liveness, crash recovery, bounded reclamation, quarantine deletion, runner-tail cleanup, and a cleanup-only SessionStart hook.

## Verification

- pre-commit run --all-files
- task test-all: 33,368 passed, 628 skipped, 26 xfailed
- task test-check: 31,896 passed, 559 skipped, 26 xfailed (PASS)
- task coverage-audit

Closes #4320
Closes #4321


Trecek (commented):

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,958 @@
...
+    ) -> None:
+        frame = _encode_frame(record, generation)
+        if size + len(frame) > _COMPACTION_THRESHOLD_BYTES:
+            latest = dict(records)
+            latest[record.capture_id] = record
+            self._compact_locked(latest, generation + 1)
+            return
+        fd = self._open_ledger()
+        try:
+            os.write(fd, frame)
```
Trecek: [critical] bugs: The append path and compaction loop (line 448) ignore os.write's returned byte count. A short regular-file write can therefore be treated as a durable commit; recovery may discard the transition, while compaction can replace the ledger with malformed frames and make lifecycle loading fail. Use a write-all loop that handles zero and short writes.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,958 @@
...
+                os.close(public.fd)
+            if quarantine is not None:
+                os.close(quarantine.fd)
+
+    def _retry(self, record: CaptureLifecycleRecord) -> CaptureLifecycleRecord:
+        retry_count = record.retry_count + 1
+        delay = min(float(2 ** min(retry_count, 12)), _MAX_RETRY_SECONDS)
+        return replace(
+            record,
+            state=CaptureState.RETRY,
```
Trecek: [critical] bugs: RETRY discards the cleanup phase that failed. If abandoned-state normalization fails while only the staging entry exists, the next sweep skips normalization, sees no public or quarantine entry, records DELETED, and strands the staging file. A failure after moving public to quarantine similarly gets a new nonce and strands the old quarantine. Preserve and resume the phase-specific state.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,958 @@
...
+                    record,
+                    state=CaptureState.ABANDONED,
+                    artifact_identity=identity,
+                    retention_at=record.created_at,
+                    next_attempt_at=self._wall_clock(),
+                    size=lease_target.size,
+                ),
+                lease_target,
+            )
+        finally:
```
Trecek, line 816: [warning] bugs: When _try_artifact_lease raises _WriterLive, lease_target is excluded from this finally block and is never returned to the caller, so each live-writer sweep leaks the observed file descriptor. Close lease_target on exceptional exits and transfer ownership only on success.

src/autoskillit/hook_registry.py
```diff
@@ -465,9 +517,122 @@ def __post_init__(self) -> None:
...
+
+
+def validate_lifecycle_contracts(
+    registry: Sequence[HookDef],
+    lifecycle_contracts: Sequence[LifecycleContractDef],
+    *,
+    backend: Literal["claude_code", "codex"],
+) -> None:
+    """Fail closed when a deployed producer loses a required cleanup owner."""
+    contract_keys = {
```
Trecek, line 588: [warning] arch: Lifecycle coverage is keyed only by resource and producer_script, ignoring backend. A producer reachable on multiple backends can satisfy completeness with a contract for one backend; validation for another backend then has no applicable contract and skips its owner checks. Require coverage for every producer/resource/backend tuple.

src/autoskillit/hook_registry.py
```diff
@@ -40,6 +49,25 @@ def __post_init__(self) -> None:
...
+class LifecycleContractDef:
+    """Static ownership contract for a hook-produced persistent resource."""
+
+    resource: str
+    producer_script: str
+    backend: Literal["claude_code", "codex"]
+    session_scope: Literal["any", "headless_only", "interactive_only"]
+    required_owner_roles: frozenset[Literal["same_runner", "session_start"]]
+
+    def __post_init__(self) -> None:
```
Trecek, line 72: [warning] defense: LifecycleContractDef validates only non-emptiness. Runtime-invalid backend, session_scope, or owner-role values can still be constructed despite Literal annotations; an unknown backend is silently excluded by validate_lifecycle_contracts and bypasses ownership checks. Validate every enum-like field and collection element here.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,958 @@
...
+        }:
+            raise CaptureLifecycleError("invalid terminal lifecycle transition")
+        now = self._wall_clock()
+        return self._commit(
+            replace(
+                record,
+                state=CaptureState.FAILED if failed else CaptureState.FINALIZED,
+                retention_at=now,
+                next_attempt_at=now + _RETENTION_SECONDS,
+                size=size,
```
Trecek, line 662: [warning] defense: finalize_capture commits caller-supplied size and sha256 without validating non-negative integer and digest constraints. A negative size is written successfully, but the next ledger load rejects that record and blocks later lifecycle operations. Validate these values before replace and _commit.

src/autoskillit/hooks/capture_lifecycle_hook.py
```diff
@@ -0,0 +1,65 @@
...
+    CaptureLifecycleError,
+)
+
+_MAX_INPUT_BYTES = 64 * 1024
+_MAX_DIAGNOSTIC_BYTES = 512
+
+
+def _bounded_stderr(message: str) -> None:
+    try:
+        sys.stderr.write(message[:_MAX_DIAGNOSTIC_BYTES])
```
Trecek: [warning] defense: The 512-byte diagnostic limit is enforced by slicing Unicode characters, so a multibyte exception message can exceed the advertised failure-path byte bound. Encode and truncate by bytes before writing.

tests/hooks/test_capture_lifecycle.py
```diff
@@ -0,0 +1,271 @@
...
+def _finalized_capture(project: Path, clock: _Clock, capture_id: str = _CAPTURE_ID):
+    anchor, root, store = _open_store(project, clock)
+    artifact = create_capture_artifact(root, capture_id, store)
+    os.write(artifact.fd, b"captured")
+    os.fsync(artifact.fd)
+    store.finalize_capture(capture_id, size=8, sha256=_DIGEST, failed=False)
+    return anchor, root, store, artifact
+
+
+def test_managed_artifact_is_published_only_after_durable_identity(tmp_path: Path) -> None:
```
Trecek, line 119: [warning] tests: The lifecycle safety tests observe only completed happy paths, so they do not prove that identity is durable before publication or that identity replacement and interrupted staging/quarantine cleanup preserve replacement files and retryable state. Add fault-injection tests at the staged commit, link/fsync, published commit, and quarantine boundaries.

tests/hooks/test_capture_lifecycle.py
```diff
@@ -0,0 +1,271 @@
...
+        for index in range(5):
+            capture_id = f"{index + 1:016x}"
+            artifact = create_capture_artifact(root, capture_id, store)
+            store.finalize_capture(capture_id, size=index + 1, sha256=_DIGEST, failed=False)
+            artifact.close()
+            artifact.release_lease()
+            artifacts.append(artifact)
+
+        clock.advance(3601)
+        first = store.sweep(max_items=2, max_duration_seconds=1)
```
Trecek, line 1183: [warning] tests: This bounded-sweep test exercises only max_items. Its monotonic clock never advances during sweeping, so it cannot cover the duration cutoff that keeps SessionStart cleanup within its timeout. Advance the clock during processing and assert the sweep stops with due work remaining.

tests/hooks/test_capture_lifecycle.py
```diff
@@ -0,0 +1,271 @@
...
+    project = tmp_path / "project"
+    clock = _Clock()
+    anchor, root, store = _open_store(project, clock)
+    try:
+        store.reserve_capture(_CAPTURE_ID)
+        ledger = _capture_dir(project) / capture_lifecycle.LEDGER_NAME
+        with ledger.open("ab") as stream:
+            stream.write(capture_lifecycle.FRAME_MAGIC + b"\x00\x00")
+        assert store.get_record(_CAPTURE_ID).state is CaptureState.RESERVED
+        assert ledger.stat().st_size < capture_lifecycle.MAX_LEDGER_BYTES
```
Trecek: [warning] tests: The recovery assertion is ineffective: this small ledger is below MAX_LEDGER_BYTES before and after loading, so the test passes even if the incomplete tail is never truncated. Record the valid size before appending the partial frame and assert the post-load size returns exactly to it.

Trecek (commented):
AutoSkillit review: blocking issues detected. See inline comments; changes are required. GitHub does not allow the PR author to submit a REQUEST_CHANGES review, so this verdict is recorded as a comment.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 227780790. `_write_all` now loops through short and interrupted writes, fails closed on zero progress, and is used by both ledger append and compaction. Regression tests cover short writes in both paths and zero-byte writes.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 56b226f30. Retry scheduling now preserves the failed lifecycle phase instead of replacing it with `RETRY`, so staging normalization and the committed quarantine nonce resume safely. Regression tests cover interrupted staging and quarantine recovery.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in a41b0bfe2. `_normalize_abandoned` now closes every observation descriptor on exceptional exits and transfers the lease-target descriptor only after a successful return. The regression test verifies the live-writer observation FD is closed.

src/autoskillit/hook_registry.py
Trecek: Agreed — fixed in 8b8dfe404. Lifecycle contract coverage now keys resource, producer script, and backend, while requiring a contract only when that producer is reachable on the backend being validated. A multi-backend regression test proves a one-backend contract cannot satisfy the other backend.

src/autoskillit/hook_registry.py
Trecek: Agreed — fixed in 135654400. `LifecycleContractDef.__post_init__` now validates string fields, backend, session scope, the frozen role collection, and every owner-role element. Parameterized tests cover each runtime-invalid value.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 51f54b307. `finalize_capture` now rejects negative, boolean, and non-integer sizes plus malformed or missing required digests before any commit; ledger decoding also rejects malformed non-empty digests. The regression test confirms invalid metadata leaves the prior record readable.

src/autoskillit/hooks/capture_lifecycle_hook.py
Trecek: Agreed — fixed in 7955d26ad. Diagnostics are now encoded as UTF-8, truncated to the byte ceiling, and decoded without an incomplete trailing code point before text output. A multibyte regression test enforces the 512-byte bound.

tests/hooks/test_capture_lifecycle.py
Trecek: Agreed — fixed in 2f1d614b8, with phase-resume coverage in 56b226f30. Fault-injection tests now cover the durable staged identity, interrupted link/root-fsync publication, failed published commit, staged identity replacement, interrupted quarantine cleanup, and quarantine replacement. They assert replacement files survive and interrupted phases remain recoverable.

tests/hooks/test_capture_lifecycle.py
Trecek: Agreed — fixed in b28cbe31c. The new duration-bound regression advances the monotonic clock during each item, proves processing stops after two items at the elapsed-time ceiling, and asserts three due records remain.

tests/hooks/test_capture_lifecycle.py
Trecek: Agreed — fixed in a4d1be9fe. The recovery test now records the exact valid ledger size, proves the injected partial tail enlarged it, and asserts loading truncates it exactly back to the valid boundary.

src/autoskillit/hook_registry.py
```diff
@@ -30,8 +30,17 @@ class HookDef:
...
-    mechanism: Literal["deny", "additionalContext", "output-rewrite", "input-rewrite"] = "deny"
+    mechanism: Literal[
+        "deny",
+        "additionalContext",
+        "output-rewrite",
+        "input-rewrite",
+        "side-effect",
+    ] = "deny"
     enforcement_strength: dict[str, str] = field(default_factory=dict)
+    produces_resources: frozenset[str] = field(default_factory=frozenset)
```
Trecek, line 41: [warning] defense: HookDef does not construction-validate the new resource collections. Mutable/non-frozenset values are accepted, and string reclaim metadata can satisfy membership checks, allowing malformed ownership metadata through lifecycle validation. Require frozenset[str] with non-empty elements.

src/autoskillit/hook_registry.py
```diff
@@ -465,9 +523,132 @@ def __post_init__(self) -> None:
...
 
 
+def hook_applies_to_backend(
+    hook_def: HookDef,
+    *,
+    backend: Literal["claude_code", "codex"],
+    session_scope: Literal["headless", "interactive"],
+) -> bool:
+    """Return whether a hook is reachable for one deployed backend/session scope."""
+    match backend:
```
Trecek, line 547: [warning] defense: hook_applies_to_backend has no fallback for runtime-invalid backend values, so it returns None despite its bool annotation; validate_lifecycle_contracts can then silently skip producer coverage. Reject unsupported backend and session_scope values at this public boundary.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -182,6 +172,11 @@ def close(self) -> None:
             os.close(self.fd)
```
Trecek, line 136: [warning] cohesion: `CaptureArtifact` owns both `fd` and `lease_fd`, but `close()` closes only `fd`; callers must know to invoke the separately named `release_lease()`. This asymmetric ownership contract makes conventional `finally: artifact.close()` silently retain a live writer lock. Split the lease into its own scoped handle or make the partial-close naming explicit.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,986 @@
...
+    except (KeyError, TypeError, ValueError) as exc:
+        raise CaptureLedgerError("invalid lifecycle record") from exc
+    if (
+        not isinstance(capture_id, str)
+        or not _CAPTURE_ID_RE.fullmatch(capture_id)
+        or not isinstance(staging_name, str)
+        or not _STAGING_NAME_RE.fullmatch(staging_name)
+        or not isinstance(public_name, str)
+        or not _PUBLIC_NAME_RE.fullmatch(public_name)
+        or not isinstance(created_at, (int, float))
```
Trecek: [warning] defense: Ledger timestamp validation accepts NaN and infinity. A NaN next_attempt_at never compares as due, silently retaining the record and artifact indefinitely. Require finite created_at, next_attempt_at, and retention_at values.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,986 @@
...
+        except BaseException:
+            os.close(fd)
+            try:
+                os.unlink(temp_name, dir_fd=self._root_fd)
+            except OSError:
+                pass
+            raise
+        else:
+            os.close(fd)
+        os.replace(
```
Trecek: [warning] bugs: The compaction temporary file is cleaned only while writing it. If `os.replace` fails, the randomly named `.capture-lifecycle-compact-*` file remains permanently, and repeated compaction failures accumulate orphan control files. Unlink the temporary entry on replacement failure as well.

Trecek (commented):

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,986 @@
...
+                >= MAX_ACTIVE_RECORDS
+            ):
+                raise CaptureLedgerError("active lifecycle record bound reached")
+            self._append_locked(record, records, generation, size)
+        return record
+
+    def mark_staged(
+        self,
+        capture_id: str,
+        artifact_identity: tuple[int, int],
```
Trecek, line 522: [warning] defense: mark_staged commits caller-supplied artifact_identity without runtime validation. Malformed tuple contents serialize successfully but are rejected on the next ledger load, wedging subsequent lifecycle operations. Validate two non-bool, nonnegative integers before commit.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,986 @@
...
+            return fd, lease_fd, record.public_name, identity
+        except (CaptureLifecycleError, OSError):
+            try:
+                self.finalize_capture(
+                    capture_id,
+                    size=os.fstat(fd).st_size if fd >= 0 else 0,
+                    sha256="",
+                    failed=True,
+                )
+            except (CaptureLifecycleError, OSError):
```
Trecek: [warning] defense: Recovery-state commit failures are discarded here and similarly in _capture_artifacts.py at line 841. Callers receive only the primary failure, losing the fact that durable lifecycle recovery also failed and may have stranded an artifact. Preserve or attach the secondary exception context.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,986 @@
...
+        except OSError as exc:
+            if exc.errno in (errno.EAGAIN, errno.EWOULDBLOCK):
+                raise _WriterLive from exc
+            raise CaptureLifecycleError("artifact lease capability failure") from exc
+
+    def _normalize_abandoned(
+        self,
+        record: CaptureLifecycleRecord,
+    ) -> tuple[CaptureLifecycleRecord, _ObservedArtifact | None]:
+        staging = self._observe(
```
Trecek: [warning] bugs: Observation descriptors can leak on error paths because `_normalize_abandoned` opens `staging` before observing `public`, and `_quarantine_delete` opens `public` before observing `quarantine`, but their cleanup `try/finally` begins only after both calls return. If the second `_observe` raises for a tampered entry or I/O failure, the first fd is never closed. Start the protected region before the first observation; `_observe` should also close its fd if `os.fstat` raises.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,986 @@
...
+        started = self._monotonic()
+        examined = deleted = deleted_bytes = writer_live = 0
+        not_due = tampered = errors = retry_count = 0
+        for capture_id in self._due_ids(self._wall_clock())[:max_items]:
+            if self._monotonic() - started >= max_duration_seconds:
+                break
+            examined += 1
+            result, logical_bytes, retries = self._sweep_one(capture_id)
+            deleted_bytes += logical_bytes
+            retry_count += retries
```
Trecek: [warning] cohesion: `CaptureCleanupOutcome` otherwise contains per-sweep event counts, but `retry_count` sums each failed record's lifetime retry depth returned by `_sweep_one`; a single third failure contributes 3 while `errors` contributes 1. Choose whether this field represents retries performed in this sweep or cumulative retry depth, then align its name and aggregation.

src/autoskillit/hooks/capture_lifecycle_hook.py
```diff
@@ -0,0 +1,69 @@
...
+import json
+import os
+import sys
+from pathlib import Path
+
+_HOOKS_DIR = str(Path(__file__).resolve().parent)
+if _HOOKS_DIR not in sys.path:
+    sys.path.insert(0, _HOOKS_DIR)
+
+from _capture_artifacts import (  # type: ignore[import-not-found]  # noqa: E402
```
Trecek: [warning] arch: The cleanup-only SessionStart owner imports the entire `_capture_artifacts` producer/runner module solely to obtain lifecycle authority. This couples reclamation to unrelated command execution, policy, replay, and subprocess dependencies, so a producer-side import regression disables the supposedly independent cleanup path and leaves retained artifacts unreclaimed. Extract the project/root authority and `open_capture_lifecycle` factory into a small stdlib-only shared module used by both hooks.

tests/docs/test_output_containment_decision.py
```diff
@@ -0,0 +1,130 @@
...
+    [
+        "RESERVED",
+        "STAGED",
+        "PUBLISHED_WRITING",
+        "FINALIZED",
+        "FAILED",
+        "ABANDONED",
+        "DELETING",
+        "DELETED",
+        "RETRY",
```
Trecek: [warning] cohesion: The documentation ratchet requires `RETRY` as a durable lifecycle state, but `CaptureState` defines no such state and `_retry()` preserves the current state while updating counters. This locks the docs/tests to a lifecycle the implementation never emits; decide whether `RETRY` should be implemented or removed from both ADRs and this ratchet.

tests/hooks/test_capture_lifecycle.py
```diff
@@ -0,0 +1,761 @@
...
+        assert outcome.tampered == 1
+        assert store.get_record(_CAPTURE_ID).state is CaptureState.TAMPERED
+        assert staging.read_bytes() == b"replacement"
+        assert public.read_bytes() == b"replacement"
+    finally:
+        root.close()
+        anchor.close()
+
+
+def test_quarantine_replacement_is_preserved_as_tampered(
```
Trecek, line 422: [warning] tests: `test_quarantine_replacement_is_preserved_as_tampered` never replaces the quarantine inode; it creates a new public file instead. The diff also removes the prior symlink/FIFO/hardlink/world-writable tests without equivalent lifecycle coverage, leaving the destructive identity-revalidation boundary largely untested. Exercise actual quarantine replacement and the documented unsafe survivor classes, asserting substitutes survive and records become TAMPERED.

tests/hooks/test_capture_lifecycle.py
```diff
@@ -0,0 +1,761 @@
...
+        second = store.sweep(max_items=8, max_duration_seconds=1)
+        assert second.deleted == 1
+        assert store.get_record(_CAPTURE_ID).state is CaptureState.DELETED
+        assert not (_capture_dir(project) / retry.quarantine_name).exists()
+    finally:
+        root.close()
+        anchor.close()
+
+
+def test_sweep_is_bounded_and_repeated_calls_make_progress(tmp_path: Path) -> None:
```
Trecek, line 1175: [warning] tests: The bounded-progress test covers only successful deletions. It does not inject an error or tampered first record and prove a later due record is still processed, despite the new contract that one failed row cannot starve the backlog. Add a mixed failure/success sweep case and verify the failed row is durably retried while later work completes.

tests/hooks/test_capture_lifecycle.py
```diff
@@ -0,0 +1,761 @@
...
+            stream.write(capture_lifecycle.FRAME_MAGIC + b"\x00\x00")
+        assert ledger.stat().st_size > valid_size
+        assert store.get_record(_CAPTURE_ID).state is CaptureState.RESERVED
+        assert ledger.stat().st_size == valid_size
+    finally:
+        root.close()
+        anchor.close()
+
+
+@pytest.mark.parametrize("force_compaction", [False, True])
```
Trecek, line 1456: [warning] tests: Ledger tests cover short/zero writes and checksum recovery but never exercise the new active-record, ledger-size, or compaction-size bounds. Regressions in these fail-closed limits would therefore pass. Lower the constants via monkeypatch and assert each boundary rejects excess state without replacing or corrupting the valid ledger.

Trecek (commented):

src/autoskillit/hook_registry.py
Trecek: Agreed — fixed in 24281b7657a545b769546edf98d33c8d6ec650df. HookDef now construction-validates and freezes resource metadata; coverage rejects mutable and string-backed collections.

src/autoskillit/hook_registry.py
Trecek: Agreed — fixed in 7b1fc4c2dc19a80f04357a30803880c047e52795. hook_applies_to_backend now rejects unsupported runtime values instead of returning a non-boolean result, with regression coverage.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in 691272ce13f1c0caa11fcf5550c665da8ed8e2d0. CaptureArtifact now exposes partial-close state explicitly while preserving deliberate lease ownership, with coverage for both descriptors.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 2dceaab924f13e1adc1f475ba14bd6abac7e5530. Lifecycle timestamp validation now rejects NaN and infinity so records cannot become permanently non-due.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 1050555911b56a60fd4d6846d84f7d28312df9b6. Failed compaction replacement now cleans the temporary artifact, including the os.replace failure path.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 08aeed5f1f24d64cc992882ee407ab8c241ecb8e. mark_staged now validates artifact identity shape and integer values before committing the ledger record.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 72725e3eb0e3089d598047f78a7be45edb012934. Secondary recovery failures now remain attached to the primary lifecycle failure so callers retain both contexts.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 5fb4957e7156b8c25aa5e5207acc4db19df7676f. Observation descriptors are now closed across normalization and quarantine error paths, with leak regression coverage.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 914b767603ee5b7ecc8a5680fc2de0af48673f3f. retry_count now reports retry events from the current sweep instead of summing lifetime retry depth.

src/autoskillit/hooks/capture_lifecycle_hook.py
Trecek: Agreed — fixed through aab8a510a1fe7f26d03c6dd1f69acf33775c029c. The cleanup hook now imports a private stdlib-only descriptor authority and no longer loads the producer/runner module.

tests/docs/test_output_containment_decision.py
Trecek: Agreed — fixed in a9922f2eda97ad17c3d15c70947ebbbface4a8ef. Documentation and its ratchet now describe the implemented durable states rather than requiring a nonexistent RETRY state.

tests/hooks/test_capture_lifecycle.py
Trecek: Agreed — fixed in 382d3583263bec30b607a079e93f53149d09a03d. The test now replaces the quarantine inode and restores coverage for unsafe surviving artifacts.

tests/hooks/test_capture_lifecycle.py
Trecek: Agreed — fixed in 3e4c66c750d6e4a93c82caddedfb2d87db3f9def. Mixed failure/success coverage now proves an errored due record does not starve later cleanup work.

tests/hooks/test_capture_lifecycle.py
Trecek: Agreed — fixed in 6758f3baad7dc5d3f949a2e4fceb3b1040a84d23. Tests now exercise active-record, ledger-size, and compaction-size bounds and their failure behavior.

src/autoskillit/hooks/_capture/_authority.py
```diff
@@ -0,0 +1,250 @@
...
+from typing import TYPE_CHECKING
+
+_HOOKS_DIR = str(Path(__file__).resolve().parent.parent)
+if _HOOKS_DIR not in sys.path:
+    sys.path.insert(0, _HOOKS_DIR)
+
+if TYPE_CHECKING:
+    from autoskillit.hooks._capture_lifecycle import CaptureLifecycleStore
+else:
+    from _capture_lifecycle import CaptureLifecycleStore
```
Trecek: [warning] arch: When this module is imported package-qualified, the runtime branch still loads top-level _capture_lifecycle, while other package code loads autoskillit.hooks._capture_lifecycle. One process then has distinct CaptureLifecycleStore, exception, and enum class identities, so package callers cannot reliably catch or compare the package-qualified types. Select the import from __package__ or centralize the bootstrap around one canonical module identity.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -967,5 +742,50 @@ def _main(argv: list[str] | None = None) -> int:
         return 1
 
 
+def _emit_cleanup_failure(detail: str) -> None:
+    safe_detail = " ".join(detail.split()).replace("]", "\\u005d")[:240]
```
Trecek: [warning] defense: _emit_cleanup_failure truncates to 240 Unicode characters, so multibyte exception text can exceed the intended bounded failure-path output. Encode and truncate by bytes before writing, as capture_lifecycle_hook._bounded_stderr already does.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,998 @@
...
+    def _locked(self) -> Iterator[None]:
+        self._validate_root()
+        try:
+            fd = os.open(LOCK_NAME, _CONTROL_FLAGS, 0o600, dir_fd=self._root_fd)
+        except OSError as exc:
+            raise CaptureLifecycleError("cannot open lifecycle lock") from exc
+        try:
+            _validate_control_file(fd, LOCK_NAME)
+            try:
+                fcntl.flock(fd, fcntl.LOCK_EX)
```
Trecek: [warning] bugs: flock(LOCK_EX) can block without bound, but sweeps 50 ms budget is checked only after _due_ids acquires this lock, and each item plus the final count locks again. Under contention, SessionStart and synchronous runner-tail cleanup can exceed the advertised bound indefinitely. Use nonblocking or bounded acquisition and defer cleanup on contention.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,998 @@
...
+                os.close(fd)
+            raise
+
+    def finalize_capture(
+        self,
+        capture_id: str,
+        *,
+        size: int,
+        sha256: str,
+        failed: bool,
```
Trecek, line 630: [warning] defense: finalize_capture validates size and digest but trusts the failed annotation at runtime. A truthy non-bool commits FAILED and a falsy non-bool commits FINALIZED, silently selecting a terminal state from malformed boundary input. Require failed to be an actual bool before evaluating it.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,998 @@
...
+            raise CaptureLifecycleError("cannot inspect managed capture") from exc
+        if (
+            not stat.S_ISREG(observed.st_mode)
+            or observed.st_uid != os.geteuid()
+            or observed.st_mode & _UNTRUSTED_WRITE_BITS
+            or observed.st_nlink not in (1, 2)
+        ):
+            raise _Tampered
+        try:
+            fd = os.open(name, _OBSERVE_FLAGS, dir_fd=self._root_fd)
```
Trecek, line 710: [warning] bugs: _observe maps every os.open failure to _Tampered. Transient operational failures such as EMFILE, ENFILE, or EIO therefore make _sweep_one durably mark the record TAMPERED and exclude it from cleanup forever instead of retrying. Treat only unsafe-entry evidence as tampering and route operational errors through the retry/backoff path.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,998 @@
...
+        return replace(
+            record,
+            retry_count=retry_count,
+            next_attempt_at=self._wall_clock() + delay,
+        )
+
+    def _sweep_one(self, capture_id: str) -> tuple[str, int, int]:
+        with self._locked():
+            records, generation, size = self._load_locked()
+            record = records.get(capture_id)
```
Trecek, line 915: [critical] bugs: Each record persists project_identity and root_identity, but _sweep_one never compares them with the stores current authorities before normalization or deletion. Moving an intact capture-root directory between projects preserves the root and artifact inodes, so a destination sweep accepts the foreign ledger and deletes its artifacts despite the recorded project boundary. Reject mismatched records before any destructive step.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,998 @@
...
+                return ("deleted", deleted_bytes, 0)
+            except _WriterLive:
+                live = replace(record, next_attempt_at=now + 30.0)
+                self._append_locked(live, records, generation, size)
+                return ("writer_live", 0, 0)
+            except _Tampered:
+                tampered = replace(record, state=CaptureState.TAMPERED)
+                self._append_locked(tampered, records, generation, size)
+                return ("tampered", 0, 0)
+            except (CaptureLifecycleError, OSError):
```
Trecek, line 970: [warning] defense: _sweep_one discards each CaptureLifecycleError or OSError and returns only an error count; both new production callers ignore CaptureCleanupOutcome. Persistent cleanup failures therefore become silent after retry state is recorded. Preserve bounded error context or have callers emit a bounded diagnostic when outcome.errors is nonzero.

src/autoskillit/hooks/capture_lifecycle_hook.py
```diff
@@ -0,0 +1,69 @@
...
+            errors="ignore",
+        )
+        sys.stderr.write(bounded)
+    except (OSError, RuntimeError, UnicodeError):
+        pass
+
+
+def main() -> int:
+    try:
+        raw = sys.stdin.read(_MAX_INPUT_BYTES + 1)
```
Trecek: [warning] defense: _MAX_INPUT_BYTES is enforced with text-mode read and len(raw), which count Unicode characters rather than encoded bytes. A multibyte hook payload can exceed the advertised 64 KiB boundary severalfold and still be parsed. Read bounded bytes from sys.stdin.buffer and decode afterward.

src/autoskillit/hooks/capture_lifecycle_hook.py
```diff
@@ -0,0 +1,69 @@
...
+            not isinstance(payload_cwd, str)
+            or not payload_cwd
+            or not os.path.isabs(payload_cwd)
+            or "\x00" in payload_cwd
+        ):
+            return 0
+        try:
+            with open_capture_lifecycle(payload_cwd, create=False) as lifecycle:
+                lifecycle.sweep(max_items=32, max_duration_seconds=0.05)
+        except CaptureSetupError:
```
Trecek: [warning] defense: CaptureSetupError represents both the expected absent-store case and unsafe ownership or mode, unavailable filesystem capabilities, and an unopenable project anchor. Suppressing every instance silently disables cleanup on safety failures. Distinguish absence from unsafe/setup failures and report the latter through bounded stderr.

tests/hooks/test_capture_artifacts.py
```diff
@@ -608,217 +640,154 @@ def fail_identity(_fd, _expected):
...
-    assert len(calls) == 2
-    assert {path.name for path in stale_paths} == set(calls)
-    assert all(path.exists() for path in stale_paths)
+    assert capture_artifacts._main(["run", "encoded", "/abs/project", _CAPTURE_ID]) == 37
+    assert events == ["dispatch", "open", "sweep", "close"]
 
 
-def test_setup_failure_prevents_user_command_and_emits_failure_marker(
-    tmp_path: Path,
+def test_runner_tail_cleanup_failure_does_not_replace_user_result(
```
Trecek, line 734: [warning] tests: This cleanup-failure test asserts only an ASCII substring and never verifies a byte bound for the new runner-tail diagnostic. Inject a multibyte error and assert len(stderr.encode(utf-8)) stays within the containment budget.

tests/hooks/test_capture_lifecycle.py
```diff
@@ -0,0 +1,1226 @@
...
+        assert failed.state is CaptureState.FAILED
+        assert failed.artifact_identity == staged_identity
+        assert (_capture_dir(project) / failed.staging_name).exists()
+        assert not (_capture_dir(project) / failed.public_name).exists()
+    finally:
+        root.close()
+        anchor.close()
+
+
+def test_creation_failure_preserves_failed_state_recovery_error(
```
Trecek, line 221: [warning] tests: This test makes both mark_staged and failed-state recovery fail, so it never covers mark_staged failing while recovery successfully records FAILED with artifact_identity=None. A later normalization then observes the staging name without an expected identity and can adopt/delete a replacement file. Add that case, replace or collide with the staged entry, and prove the unverified entry survives.

tests/hooks/test_capture_lifecycle_hook.py
```diff
@@ -0,0 +1,197 @@
...
+    "payload",
+    [
+        [],
+        {},
+        {"cwd": ""},
+        {"cwd": "relative"},
+        {"cwd": 1},
+    ],
+)
+def test_cleanup_hook_rejects_invalid_payload_without_output(
```
Trecek, line 161: [warning] tests: The invalid-payload cases are all tiny and do not exercise the advertised 64 KiB input-byte limit. Add a multibyte JSON payload whose UTF-8 encoding exceeds the limit while its character count does not, and assert cleanup is not attempted.

Trecek (commented):

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments — verdict: changes_requested. GitHub does not allow the PR author to submit a REQUEST_CHANGES event on their own pull request.

src/autoskillit/hook_registry.py
```diff
@@ -465,9 +533,136 @@ def __post_init__(self) -> None:
...
+    session_scope: Literal["headless", "interactive"],
+) -> bool:
+    """Return whether a hook is reachable for one deployed backend/session scope."""
+    if backend not in ("claude_code", "codex"):
+        raise ValueError(f"unsupported hook backend: {backend!r}")
+    if session_scope not in ("headless", "interactive"):
+        raise ValueError(f"unsupported hook session scope: {session_scope!r}")
+    match backend:
+        case "codex":
+            return hook_def.session_scope != "interactive_only" and hook_def.codex_status not in {
```
Trecek: [warning] bugs: The Codex branch ignores the requested session_scope: a headless_only hook is reported reachable for interactive sessions, while interactive_only is never reachable even when requested. Lifecycle validation can therefore accept a headless-only cleanup owner as satisfying an interactive contract.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,998 @@
...
+        if (
+            not isinstance(sha256, str)
+            or bool(sha256)
+            and not _SHA256_RE.fullmatch(sha256)
+            or not failed
+            and not sha256
+        ):
+            raise CaptureLifecycleError("invalid terminal capture digest")
+        record = self.get_record(capture_id)
+        if record is None or record.state not in {
```
Trecek, line 645: [warning] defense: Successful finalization is accepted from RESERVED or STAGED and does not require artifact_identity. This can create a FINALIZED record whose cleanup later binds by filename alone. Require PUBLISHED_WRITING with a persisted artifact identity when failed is false; reserve earlier-state recovery for failed finalization.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -0,0 +1,998 @@
...
+            self._try_artifact_lease(lease_target)
+            identity = lease_target.identity
+            if staging is not None and public is not None:
+                if staging.identity != public.identity or staging.nlink != 2 or public.nlink != 2:
+                    raise _Tampered
+                os.unlink(record.staging_name, dir_fd=self._root_fd)
+                os.fsync(self._root_fd)
+            elif staging is not None:
+                try:
+                    os.link(
```
Trecek: [warning] bugs: Both recovery hard-link paths (also L840) create a new link and then reject nlink != 2 without removing the link they just created. If an unexpected hardlink already exists, the sweep marks the record TAMPERED but leaves a new public or quarantine entry behind permanently.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek, line 915: [warning] defense: Loaded records are processed without checking record.project_identity and record.root_identity against the store authorities. A copied but checksum-valid ledger can therefore authorize cleanup under a different project/root. Validate both persisted bindings before observing or deleting artifacts.

tests/docs/test_output_containment_decision.py
```diff
@@ -0,0 +1,129 @@
...
+DECISION = REPO_ROOT / "docs/decisions/0006-output-containment.md"
+OUTPUT_BUDGET_DECISION = REPO_ROOT / "docs/decisions/0005-output-budget-protocol.md"
+SAFETY_GUIDE = REPO_ROOT / "docs/safety/hooks.md"
+
+pytestmark = [pytest.mark.layer("docs"), pytest.mark.small]
+
+
+@pytest.fixture(scope="module")
+def lifecycle_docs() -> str:
+    return "\n".join(
```
Trecek: [warning] tests: This fixture concatenates ADR-0006, ADR-0005, and the safety guide before every assertion. Removing a required guarantee from ADR-0006 can still pass when the phrase remains in a companion document, so the tests do not ratchet the named decision. Assert ADR-specific requirements against DECISION separately.

tests/hooks/test_capture_lifecycle.py
```diff
@@ -0,0 +1,1226 @@
...
+
+        assert first_fd >= 0
+        with pytest.raises(OSError):
+            os.fstat(first_fd)
+    finally:
+        root.close()
+        anchor.close()
+
+
+def test_writer_lease_is_visible_to_an_independent_process(tmp_path: Path) -> None:
```
Trecek, line 841: [warning] tests: The only independent-process test checks the artifact writer lease. The new cross-process ledger locking and process-termination recovery guarantees have no test that terminates a producer and reopens/sweeps the store from another process, leaving the central durability boundary unverified.

tests/hooks/test_capture_lifecycle_hook.py
```diff
@@ -0,0 +1,197 @@
...
+    assert completed.returncode == 0
+    assert completed.stdout == ""
+    assert completed.stderr == ""
+    assert not (project / ".autoskillit").exists()
+
+
+def test_cleanup_hook_fails_open_with_bounded_stderr(tmp_path: Path) -> None:
+    project = tmp_path / "project"
+    project.mkdir()
+    _seed_due_capture(project)
```
Trecek: [warning] tests: The fail-open corruption test discards the seeded artifact path and never verifies that cleanup preserved it. A cleanup implementation could delete the managed artifact, emit the expected bounded error, and still pass. Retain the path and assert its contents remain unchanged.

Trecek (commented):

Trecek (commented):

src/autoskillit/hooks/_capture/_authority.py
Trecek: Agreed — fixed in a82d38655. `_capture._authority` now resolves `_capture_lifecycle` from its package identity, preserving the package-qualified `CaptureLifecycleStore`/exception/enum identities while retaining standalone hook loading. A regression asserts the shared class identity.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in 301d9b441. `_emit_cleanup_failure` now normalizes, UTF-8 encodes, truncates to the 240-byte detail budget, and decodes safely before writing; the regression uses a multibyte exception and asserts the total stderr byte bound.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 219124f2b. Sweep-only ledger acquisition now uses `LOCK_NB`; contention returns a deferred outcome within the duration budget, including initial, per-item, and final due-list acquisition. A cross-process lock-holder test asserts prompt return and artifact preservation.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 6703ecf53. `finalize_capture` now requires `failed` to be an actual `bool` before digest/state evaluation, so malformed truthy or falsy boundary values cannot select a terminal state.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 66e5c34f7. `_observe` now treats `ELOOP` as unsafe-entry evidence but maps operational `os.open` failures to `CaptureLifecycleError`, which follows the durable retry/backoff path. The EMFILE regression proves the record is retried rather than marked TAMPERED.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 0bddc70fc. `_sweep_one` now compares both persisted `project_identity` and `root_identity` with the open store authorities before normalization or deletion. Foreign checksum-valid records become TAMPERED and their artifacts survive; both bindings are covered.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 138a180f0. Both the runner-tail owner and SessionStart owner now inspect `CaptureCleanupOutcome.errors` and emit a bounded deferred-cleanup diagnostic when retries were recorded, while preserving fail-open command/hook behavior.

src/autoskillit/hooks/capture_lifecycle_hook.py
Trecek: Agreed — fixed in 03b0e8538. The cleanup hook now reads at most 64 KiB + 1 from `sys.stdin.buffer`, applies the limit to raw bytes, and decodes only afterward. A UTF-8 multibyte payload regression proves oversized input cannot trigger cleanup.

src/autoskillit/hooks/capture_lifecycle_hook.py
Trecek: Agreed — fixed in 2ec79c06b. `CaptureStoreAbsentError` now represents the expected create=False missing-store case; unsafe ownership/mode, unavailable capabilities, and anchor/setup failures remain `CaptureSetupError` and produce bounded diagnostics. The absent path stays silent.

tests/hooks/test_capture_artifacts.py
Trecek: Agreed — fixed in 301d9b441. The runner-tail cleanup-failure regression now injects a long multibyte exception and asserts `len(stderr.encode("utf-8")) <= 512`, covering the actual containment boundary.

tests/hooks/test_capture_lifecycle.py
Trecek: Agreed — fixed in 390d79998. Failed-state recovery with `artifact_identity=None` no longer adopts a staging/public entry by filename: any such entry is classified TAMPERED and preserved. The regression makes `mark_staged` fail, lets FAILED recovery succeed, replaces staging, and proves survival.

tests/hooks/test_capture_lifecycle_hook.py
Trecek: Agreed — fixed in 03b0e8538. The invalid-input coverage now sends JSON whose UTF-8 encoding exceeds 64 KiB while its character count does not, then proves cleanup was not attempted and the due artifact remains unchanged.

src/autoskillit/hook_registry.py
Trecek: Agreed — fixed in e1f3fa405. Codex reachability now applies the requested `session_scope` after the Codex status gate: `headless_only` is false interactively and `interactive_only` is true only for interactive checks. The full scope matrix is covered.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 6703ecf53. Successful finalization now requires `PUBLISHED_WRITING` plus a persisted `artifact_identity`; RESERVED/STAGED recovery remains available only for failed finalization. Separate regressions cover both earlier states.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 882d7fc01. Both recovery-link paths now roll back the just-created destination when post-link identity/nlink verification fails. Regressions seed an unexpected existing hard link and prove no public/quarantine entry is stranded.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 0bddc70fc. Loaded records are now authority-checked against the store's current project and root identities before any observation, normalization, quarantine, or deletion; mismatches are terminally preserved as TAMPERED.

tests/docs/test_output_containment_decision.py
Trecek: Agreed — fixed in c2ad24fb6 (with ADR wording pinned in 1d5e89a09). The ADR-0006 ratchet now reads `DECISION` alone, so companion documents cannot mask removal of owners, states, retention, survivor classes, durability, or accepted-gap guarantees.

tests/hooks/test_capture_lifecycle.py
Trecek: Agreed — fixed in 364b8cb86. A producer subprocess now creates and writes a PUBLISHED_WRITING artifact, is terminated while holding its descriptors, and an independently reopened store advances retention, sweeps, and proves durable recovery/deletion.

tests/hooks/test_capture_lifecycle_hook.py
Trecek: Agreed — fixed in 31f68e042. The fail-open corrupt-ledger test now retains the seeded artifact path and asserts its original bytes remain after the bounded cleanup error, so accidental deletion can no longer pass.

Issue #4399, opened 2026-07-28
Claude Code full open_kitchen returns no recipe body, and close_kitchen -> open_kitchen empties the MCP tool registry


Trecek:
## Summary

Two independent defects in the Claude Code delivery/visibility path, split out of #4362.

1. **Envelope truncation.** A full `open_kitchen` returns literally `## open_kitchen ✓ v`
   — 19 characters, no recipe body. This is *not* a byte-level truncation: it is the
   correct output of the PostToolUse formatter rendering an ENVELOPE-mode payload whose
   `content`/`summary`/`diagram` were replaced by `step_flow_skeleton`, a key the
   formatter has never read.
2. **Tool registry loss.** After a `close_kitchen` → `open_kitchen` sequence, all 41
   kitchen-tagged MCP tools (including `run_skill`) become uncallable while the server
   process stays alive and un-gated tools keep answering normally.

**Split rationale.** #4362's title and its 11-item `## Acceptance criteria` are about the
*Codex* problem — a 123-step flow skeleton not fitting the intentional 10,000-byte
conservative envelope, requiring a governed compact-skeleton / pagination / recovery
protocol. The two defects above were found on **Claude Code** during sessions 4 and 5 and
were appended to that thread. `dry-walkthrough` correctly refused a plan targeting only
these two against #4362's Codex acceptance criteria. #4362 remains open and unchanged for
the Codex work; this issue owns the Claude Code defects.

## Evidence — envelope truncation

Reproduced 100% on **0.10.905**, `develop` @ `d26d38e45`, recipe `remediation`, with zero
ingredient overrides and with overrides. The persisted payload is written **in full** every
time — only the returned envelope is degenerate. Recovery in use today: read
`recipe_execution.execution_id` and `invocation_template_digests` out of
`.autoskillit/temp/recipe-delivery/<kitchen_id>/open_kitchen/<recipe>/descriptor-1-schema-1/sha256~*/payload.json`.
That tax is paid on **every** orchestration.

## Evidence — registry loss, and the controlled test that isolated it

| # | session | version | preceded by `close_kitchen`? | registry after |
|---|---|---|---|---|
| 1 | 4 | 0.10.904 | no | intact |
| 2 | 4 | 0.10.904 | **yes** | **emptied** |
| 3 | 4 | 0.10.904 | **yes** | **emptied** |
| 4 | 5 | 0.10.905 | no | intact |
| 5 | 5 | 0.10.905 | no | intact |
[...]

PR #4407, opened 2026-07-28
Rectify: open_kitchen Registry Loss — Client-Cache Coherence

Trecek:
## Summary

After `close_kitchen()` → `open_kitchen()`, all 41 kitchen-tagged MCP tools vanish from the client's tool list. The server-side state is correct — the global `mcp.enable()` calls added in PR #4401 work and the server reports tools as enabled — but the client is never notified to re-query `tools/list`, so it keeps serving a stale cache from the close transition.

The architectural weakness is a **notification asymmetry**: `close_kitchen` unconditionally sends `ToolListChangedNotification` (via `ctx.reset_visibility()`), but `open_kitchen` deliberately suppresses notification when `_skip_notify=True` (all production backends). The `_skip_notify` flag correctly prevents notification at boot (no client connected yet) but is incorrectly reused mid-session (client IS connected and synced to the disabled state).

The fix is to send an explicit `ToolListChangedNotification` after the global re-enables in `open_kitchen`, and to structurally prevent this class of bug from recurring via an architectural guard and a session-level round-trip regression test.

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260728-095428-498581/.autoskillit/temp/rectify/rectify_open_kitchen_registry_loss_residual_2026-07-28_101500.md`

Closes #4399

### DIFF
--- a/src/autoskillit/server/tools/tools_kitchen.py
+++ b/src/autoskillit/server/tools/tools_kitchen.py
@@ -15,6 +15,7 @@ if TYPE_CHECKING:
 
 from fastmcp import Context
 from fastmcp.dependencies import CurrentContext
+from mcp.types import ToolListChangedNotification
 
 from autoskillit import __version__
 from autoskillit.config import (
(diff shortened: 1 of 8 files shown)


Trecek (commented):

(Verdict: approved. Self-authored PR — GitHub does not permit a self-APPROVE review, so this is posted as a COMMENT review instead.)

Issue #4406, opened 2026-07-28
audit-impl writes authority.json pretty-printed but the verifier demands strict canonical JSON — audit_impl crashes on the artifact it just wrote, after a completed audit


Trecek:
## Summary

`audit-impl` writes `authority.json` **pretty-printed** (`indent=2`, trailing newline), but the
server-side verifier that reads it back demands **strict canonical versioned JSON**. Producer
and consumer disagree on serialization form, so `audit_impl` crashes on the authority file it
just wrote — *after* having completed a full, correct audit.

## Why this only surfaced now

`audit_impl` could not launch at all in any clone-based pipeline until **#4387** landed
(containment root anchored to the orchestrator temp dir). The 2026-07-28 run on #4399 is the
first time it has actually executed since. It launched, ran the audit, reached a verdict, wrote
its artifacts — and then hit this. **#4387's fix works; this is the next barrier behind it.**

## Observed 2026-07-28 (clone `remediation-20260728-095428-498581`, #4399 pipeline)

`audit_impl` produced complete, well-formed output before the crash:

- `remediation_open_kitchen_registry_loss_residual_*.md` — a **NO GO** with one MISSING finding
- `cycles/gen1/.../round-1/authority.json` (3,318 bytes)
- `cycles/gen1/.../round-1/inventory.json`

The NO GO itself was correct and actionable (a docstring the plan named explicitly was not
renamed), and was remediated in one commit. So the audit's *reasoning* is sound — only the
artifact handshake is broken.

The authority file is emitted human-readable:

```json
{
  "schema_version": 1,
  ...
}
```

The verifier rejects it with `"authority is not strict canonical versioned JSON"`
(`core/audit_cycle_verifier.py:432`; sibling messages at 452 for the disposition report and 580
for the inventory, so the same mismatch likely affects all three artifact kinds).
Consumer side: `decode_versioned_json_bytes` (`core/io.py:343`).

## Impact

An infrastructure crash **after** a completed audit. The verdict, the findings and the
remediation plan all exist on disk but never reach the pipeline, so the orchestrator sees a
tool failure rather than a NO GO. Combined with **#4404** (in `remediation.yaml` a preflight
crash is still routed as `register_clone_failure`, indistinguishable from a NO GO), this can
tear down a complete, green implementation on the basis of a serialization mismatch.

## Proposed direction

1. Make producer and consumer share **one** serializer. The canonical writer already exists —
[...]

PR #4408, opened 2026-07-29
Rectify: Marker Substring Match False Positives — Centralized Code-Region-Aware Detection

Trecek:
## Summary

`INVESTIGATION_COMPLETE_MARKER` and `REVIEW_APPROACH_MARKER` are detected via naive `str.__contains__` against raw, unsanitized GitHub issue bodies at two duplicated call sites. This produces false positives when the marker text appears inside markdown fenced code blocks or inline code spans. A production-grade `strip_markdown_code_regions` utility does not exist in `src/` — only a test-only version in `tests/_helpers.py`. Additionally, the `build-execution-map` SKILL.md propagates the same vulnerability in prose by instructing the LLM to perform a "simple substring check."

The architectural weakness is twofold: (1) no centralized marker-detection predicate exists — each call site reimplements raw `in` checks independently, and (2) there is no shared code-region-stripping utility in production code. The fix promotes the stripping utility into IL-0 `core/`, creates a centralized `detect_body_marker()` predicate that strips before matching, replaces both duplicated call sites, fixes the SKILL.md prose, and adds false-positive rejection tests for both markers.

## Implementation Plan

Plan file: `/home/talon/projects/autoskillit-runs/remediation-20260728-142843-376918/.autoskillit/temp/rectify/rectify_marker_substring_match_2026-07-28_223749.md`

Closes #4386

---

## Audit gate bypassed — #4406

`audit_impl` did **not** gate this PR. It crashed with
`AuditCycleVerificationError: authority is not strict canonical versioned JSON` — that is
#4406: `audit-impl` writes `authority.json` pretty-printed, while its own verifier requires
the strict canonical profile (`sort_keys=True`, `separators=(",", ":")`). The crash is in the
cycle-artifact machinery *after* the audit itself completed, so it is not a quality finding
against this change.

The bypass was explicitly authorized by the repository owner. Evidence standing in for the gate:

- `task test-check` — 31848 passed, 557 skipped, 26 xfailed
- `pre-commit run --all-files` — all hooks passed
- `dry-walkthrough` — plan verified and stamped after three auditor passes
- no recorded deviations from the plan

### DIFF
--- a/src/autoskillit/core/types/_type_helpers.py
+++ b/src/autoskillit/core/types/_type_helpers.py
@@ -30,11 +30,13 @@ __all__ = [
     "extract_path_arg",
     "extract_positional_args",
     "extract_skill_name",
+    "detect_body_marker",
     "fleet_error",
     "render_target_skill_command",
     "resolve_skill_name",
     "resolve_target_skill",
     "session_type",
+    "strip_markdown_code_regions",
     "truncate_text",
 ]
 
@@ -426,6 +428,22 @@ def truncate_text(text: str, max_len: int = 5000) -> str:
     if len(text) <= max_len:
         return text
     return f"...[truncated {len(text) - max_len} chars]...\n" + text[-max_len:]
+
+
+_CODE_BLOCK_RE = re.compile(r"(```|~~~).*?\1", re.DOTALL)
+_INLINE_CODE_RE = re.compile(r"`[^`]*`")
+
+
+def strip_markdown_code_regions(text: str) -> str:
+    """Remove fenced code blocks and inline code spans from markdown text."""
+    text = _CODE_BLOCK_RE.sub("", text)
+    text = _INLINE_CODE_RE.sub("", text)
+    return text
+
+
+def detect_body_marker(body: str, marker: str) -> bool:
+    """Check whether *marker* appears in *body* outside markdown code regions."""
+    return marker in strip_markdown_code_regions(body)
 
 
 def fleet_error(
--- a/src/autoskillit/core/types/_type_helpers.py
+++ b/src/autoskillit/core/types/_type_helpers.py
@@ -436,9 +436,12 @@ _INLINE_CODE_RE = re.compile(r"`[^`]*`")
 
 def strip_markdown_code_regions(text: str) -> str:
     """Remove fenced code blocks and inline code spans from markdown text."""
-    text = _CODE_BLOCK_RE.sub("", text)
-    text = _INLINE_CODE_RE.sub("", text)
-    return text
+    while True:
+        stripped = _CODE_BLOCK_RE.sub("", text)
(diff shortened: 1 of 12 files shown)


src/autoskillit/core/types/_type_helpers.py
```diff
@@ -428,6 +430,25 @@ def truncate_text(text: str, max_len: int = 5000) -> str:
     return f"...[truncated {len(text) - max_len} chars]...\n" + text[-max_len:]
 
 
+_CODE_BLOCK_RE = re.compile(r"(```|~~~).*?\1", re.DOTALL)
+_INLINE_CODE_RE = re.compile(r"`[^`\n]*`")
+
+
+def strip_markdown_code_regions(text: str) -> str:
```
Trecek, line 437: [warning] bugs: strip_markdown_code_regions requires a matched opening+closing fence pair via the (```|~~~).*?\1 pattern. An unclosed/odd-count fence (e.g. an issue body opens a ``` block and never closes it) leaves everything after the dangling opening delimiter completely unstripped, including any marker text that follows it. On GitHub, an unterminated fence causes all subsequent content to render as code (until EOF), so a marker appearing there is visually in code and should not be treated as live — but detect_body_marker(body, marker) still returns True in this case, reproducing the class of false-positive this PR is meant to fix. Repro: strip_markdown_code_regions("```\ncode\n\nmore, unclosed") leaves the marker present in the stripped text. Consider treating an unmatched fence delimiter as extending to EOF (matching GitHub's rendering) so trailing content is also stripped.

Trecek (commented):

src/autoskillit/core/types/_type_helpers.py
Trecek: Valid observation — confirmed the repro (`strip_markdown_code_regions("```\ncode\n<!-- investigation_complete: true -->\nmore, unclosed")` leaves the marker present). A literal fix (extend the unmatched fence to EOF, e.g. `(```|~~~).*?(?:\1|\Z)`) works for this repro, but `strip_markdown_code_regions` is a single shared implementation — `tests/_helpers.py` re-exports the *same* function for `tests/docs/test_banned_phrases.py`, `tests/docs/test_glossary_spelling.py`, `tests/skills/test_investigate_design_intent_contracts.py`, and `tests/workspace/test_skills.py`, which strip code fences from SKILL.md/docs prose before banned-phrase/mention-guard scanning. `tests/test_helpers_strip.py::test_unclosed_fence_is_fail_open` explicitly locks the opposite (fail-open) behavior as intentional for that use case — flipping the shared regex to fail-closed would silently drop trailing prose in doc scans, inverting the safety direction those consumers rely on. Flagging for a design decision: fork a stricter fail-closed helper scoped to `detect_body_marker`, or accept the shared behavior change and update the fail-open test/consumers.

PR #4410, opened 2026-07-29
[FIX] Restore branch-targeted CI policy for #4405

Trecek:
## Summary

- Restore branch-targeted CI runner and test-filter policy through one fail-closed event resolver.
- Give develop-targeted pull requests and merge groups the immutable event base SHA for conservative filtering.
- Remove PR #4405's redundant lifecycle-exception import suppression and add behavioral, manifest, and documentation ratchets.

## Target policy

| Target | Runners | Filter |
|---|---|---|
| `develop` | `ubuntu-latest` | `conservative` |
| `main` | `ubuntu-latest` | `none` |
| `stable` | `ubuntu-latest`, `macos-15` | `none` |

## Stacked PR

This correction is based on and targets PR #4405's head branch: `impl-rectify_codex_shell_capture_cleanup_lifecycle_2026-07-27_233858-20260728-080540`.

## Verification

- Focused CI, manifest, hook, and Pyright suppression gate passed.
- `task test-check`: 33,799 passed.
- `pre-commit run --all-files` passed.
- Hook registry hash and `.autoskillit/test-source-map.json` remained unchanged.

### DIFF
--- a/scripts/ci_target_policy.py
+++ b/scripts/ci_target_policy.py
@@ -107,7 +107,11 @@ def resolve_ci_profile(
 
     if target not in allowed_targets:
         raise ValueError(f"target {target!r} is not allowed for event {event_name!r}")
-    policy = CI_TARGET_POLICIES[target]
+    policy = CI_TARGET_POLICIES.get(target)
+    if policy is None:
+        raise ValueError(
+            f"target {target!r} is allowed for event {event_name!r} but has no registered policy"
+        )
 
     base_revision = ""
     if policy.filter_mode == "conservative":
--- a/scripts/ci_target_policy.py
+++ b/scripts/ci_target_policy.py
@@ -1,0 +1,152 @@
+#!/usr/bin/env python3
+"""Resolve the runner and test-filter policy for a GitHub Actions event."""
+
+from __future__ import annotations
+
+import json
+import os
+import re
+import sys
+from collections.abc import Mapping
+from dataclasses import dataclass
+from pathlib import Path
+from types import MappingProxyType
+
+_HEAD_REF_PREFIX = "refs/heads/"
+_COMMIT_SHA_PATTERN = re.compile(r"[0-9a-f]{40}")
+
+
+@dataclass(frozen=True, slots=True)
+class CiTargetPolicyDef:
+    """Static CI behavior registered for one target branch."""
+
+    os_runners: tuple[str, ...]
+    filter_mode: str
+
+
+CI_TARGET_POLICIES: Mapping[str, CiTargetPolicyDef] = MappingProxyType(
+    {
+        "develop": CiTargetPolicyDef(("ubuntu-latest",), "conservative"),
+        "main": CiTargetPolicyDef(("ubuntu-latest",), "none"),
+        "stable": CiTargetPolicyDef(("ubuntu-latest", "macos-15"), "none"),
+    }
+)
+
+ALLOWED_TARGETS_BY_EVENT: Mapping[str, frozenset[str]] = MappingProxyType(
+    {
+        "pull_request": frozenset({"develop", "main", "stable"}),
+        "merge_group": frozenset({"develop", "main", "stable"}),
+        "push": frozenset({"main", "stable"}),
+    }
+)
+
+
+def _require_mapping(value: object, field: str) -> Mapping[str, object]:
+    if not isinstance(value, Mapping):
+        raise ValueError(f"{field} must be an object")
+    return value
+
+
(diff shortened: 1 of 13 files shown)


tests/hooks/test_capture_lifecycle_hook.py
```diff
@@ -34,10 +38,43 @@ def test_cleanup_hook_imports_minimal_shared_authority() -> None:
     source = SCRIPT.read_text(encoding="utf-8")
     assert "from _capture._authority import" in source
     assert "from _capture_artifacts import" not in source
+    assert "from _capture_lifecycle import" not in source
+    for lifecycle_name in (
```
Trecek: [warning] tests: These assertions search the entire source text, so a lifecycle name used in an exception handler, comment, or unrelated statement would satisfy the test even if it were omitted from the `_capture._authority` import. Inspect the authority import block specifically so this architectural guard cannot false-pass.

tests/infra/test_ci_workflow.py
```diff
@@ -1,27 +1,465 @@
...
+    raw_payload: str | None,
+) -> None:
+    completed = _run_cli(
+        tmp_path,
+        event_name=event_name,
+        payload=payload,
+        raw_payload=raw_payload,
+    )
+
+    assert completed.returncode != 0
```
Trecek: [warning] tests: The malformed-input CLI tests accept any nonzero exit and ignore stderr, so an unexpected traceback or unrelated crash passes just like the intended handled rejection. Assert the defined exit code and a concise diagnostic (including absence of a traceback); the environment-requirement case below has the same weakness.

tests/infra/test_ci_dev_config.py
```diff
@@ -249,13 +249,14 @@ def test_ci_pull_request_trigger_includes_develop(self) -> None:
         assert "develop" in pr_branches, "CI must trigger on PRs targeting develop branch"
 
     def test_ci_preflight_outputs_os_matrix(self) -> None:
```
Trecek: [warning] cohesion: CI-target-policy assertions are split across this file and `tests/infra/test_ci_workflow.py`, where the new workflow-authority test repeats the same preflight outputs, resolver invocation, and test-step environment wiring. Consolidate the duplicated coverage in the policy test module to keep this feature local and avoid two maintenance points.

scripts/ci_target_policy.py
```diff
@@ -0,0 +1,152 @@
...
+        base_sha = merge_group.get("base_sha")
+    else:
+        target = _target_from_head_ref(
+            _require_string(event_payload, "ref"),
+            "ref",
+        )
+
+    if target not in allowed_targets:
+        raise ValueError(f"target {target!r} is not allowed for event {event_name!r}")
+    policy = CI_TARGET_POLICIES[target]
```
Trecek: [warning] defense: The target is validated against `ALLOWED_TARGETS_BY_EVENT` and then indexed in the independently maintained `CI_TARGET_POLICIES` registry without checking that the registries are consistent. An allowlist-only change therefore raises an uncaught `KeyError` rather than the contextual `ValueError` handled by `main`. Validate registry consistency or use a guarded lookup with explicit context.

Trecek (commented):

Trecek (commented):

tests/hooks/test_capture_lifecycle_hook.py
Trecek: Agreed — fixed in aeaa2fced. The guard now parses the hook AST, selects the single _capture._authority import, and checks its exact imported-name set, so incidental runtime uses cannot satisfy the import contract.

tests/infra/test_ci_workflow.py
Trecek: Agreed — fixed in abaf3d588. Both malformed-input and missing-environment cases now require exit code 2, a single ci_target_policy diagnostic containing the expected cause, empty stdout, and no traceback.

tests/infra/test_ci_dev_config.py
Trecek: Agreed — fixed in 2848f683c. The four duplicated preflight/output-wiring tests were removed from test_ci_dev_config.py; the stricter combined policy-authority coverage remains in test_ci_workflow.py.

scripts/ci_target_policy.py
Trecek: Agreed — fixed in ea1fc00ea. The resolver now uses a guarded policy lookup and raises a contextual ValueError when an allowed target lacks a registered policy; a regression test covers the inconsistent-registry case.

Issue #4322, opened 2026-07-22
shell-capture complete=true metadata can become stale after descendant writes


Trecek:
## Problem

The oversized shell-capture preview computes its byte count, optional SHA-256,
head slice, marker, and tail slice through separate pathname operations after the
immediate subshell exits. A descendant outside that shell's active child/job set
can retain the inherited output descriptor and write between or after those
operations.

The resulting preview can combine multiple artifact states. Its metadata may not
describe the emitted slices or the artifact later retrieved at the advertised
path, while the marker unconditionally says `complete=true`. Detached-child
writes are an accepted ADR-0006 gap, but these existing metadata semantics are
undefined and misleading, so this is a remediation.

## Evidence

- `src/autoskillit/hooks/shell_capture_hook.py:102-109` redirects the subshell to
  the artifact and uses bare `wait` in its EXIT trap. That does not establish
  quiescence for arbitrary descriptor-holding descendants.
- `docs/decisions/0006-output-containment.md:64-68,75-76` explicitly places
  detached/job-table-escaped descendants outside the wait guarantee.
- The oversized path is sequential: size at
  `shell_capture_hook.py:110`, optional digest at lines 91-95 and 114, head at
  line 115, unconditional `complete=true` marker at lines 96-100 and 116, then
  tail at line 117.
- `sha256sum` absence produces `sha256=unavailable` rather than a verifiable
  digest (`shell_capture_hook.py:91-95`).
- Existing tests cover safe marker interpolation and ordinary final-artifact
  parity, but not count/digest/slice consistency under concurrent descendant
  writes (`tests/hooks/test_shell_capture_hook.py:118-139` and
  `tests/hooks/test_shell_capture_conformance.py:211-224`).
- A bounded reproduction emitted a marker reporting 13,000 bytes and
  `complete=true`; afterward the artifact was 13,004 bytes and ended in `LATE`.
  This proves the count described an earlier state and that `complete=true`
  cannot truthfully mean an immutable final artifact.

## Expected Behavior

The marker and emitted slices must describe one coherent, documented state.
Unless ADR-0006 is deliberately changed, completion should be scoped to managed
producer/postlude or snapshot completion rather than immutable filesystem
completion. Later retrieval must have an explicit verification contract for
detecting whether the advertised path still represents that snapshot.

## Dependencies

- Depends on #4319 for stable, race-safe artifact identity and reads.
- Depends on #4321's selected writer-lifecycle/finalization protocol. This issue
[...]

PR #4418, opened 2026-07-29
Rectify: Shell-Capture Snapshot Authority

Trecek:
## Summary

- Makes a descriptor-verified, factory-only snapshot the sole authority for a FINAL shell capture.
- Separates capture, snapshot, reference, delivery, and retention state, with every transition routed through one locked, revision-checked lifecycle core.
- Makes reference publication and stdout delivery transactional while retaining the producer lease through delivery.
- Hardens framed-ledger migration, restart reconciliation, descriptor rebinding defenses, and cleanup/reader handoff.
- Adds the planned fault, race, migration, import-order, durability, integration, architecture, and documentation coverage.

Closes #4322

## Validation

- `task test-check` — PASS (`34,009 passed`, `630 skipped`, `27 xfailed`)
- `pre-commit run --all-files` — PASS
- Exact-commit implementation audit — GO (`22/22` requirements covered; no missing, odd, or conflicting findings)
- Dry-walkthrough — verified against the rectification plan

## Implementation

The work is organized as the four planned implementation packages—verified snapshot
authority, reader leases, transactional delivery, and documentation/integration
ratchets—followed by one audit-remediation commit that closes every finding from the
first audit pass.


src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -621,68 +653,219 @@ def create_artifact(self, capture_id: str) -> tuple[int, int, str, tuple[int, in
...
+                else:
+                    if (
+                        current is not None
+                        and replace(
+                            current,
+                            compaction_epoch=candidate.compaction_epoch,
+                        )
+                        == candidate
+                    ):
+                        return finalized
```
Trecek: [critical] bugs: This reconciliation treats a frame that is merely readable as durably committed. If _append_locked() wrote the frame and os.fsync() then failed, get_record() can read dirty page-cache bytes and this returns FINAL authority anyway. That allows a success/reference from an uncertain FINAL. Reconcile only faults known to occur after successful sync; fsync failures must fail closed.

Trecek (commented):

src/autoskillit/hooks/_capture/_delivery.py
```diff
@@ -0,0 +1,657 @@
...
+    runtime_errors: RuntimeErrors,
+) -> None:
+    try:
+        store.transition_delivery(
+            value,
+            expected=expected,
+            target=target,
+        )
+    except runtime_errors:
+        if delivery_status(store, value) == target:
```
Trecek, line 401: [critical] bugs: After any transition exception, observing the target state through get_record() is accepted as success. An os.fsync() failure can leave the new frame readable from page cache without making it durable, so ATTEMPTING or DELIVERED may be reported as durably finished and output can proceed despite an uncertain transition. Only reconcile errors proven to occur after successful fsync.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -555,37 +557,81 @@ def _resolve_bash() -> str:
...
-            return None
+        return lifecycle.publish_reference(finalized)
     except _CAPTURE_RUNTIME_ERRORS:
-        return None
+        reconciled = _reference_result_after_transition(
+            lifecycle,
+            finalized,
+            unavailable_reason="PUBLICATION_FAILED",
+        )
+        if type(reconciled) is PublishedCaptureReference:
```
Trecek, line 613: [critical] bugs: Publication reconciliation returns a bearer reference whenever the just-written PUBLISHED state is readable, including when publish_reference() failed at ledger fsync. The hook can therefore emit a token whose publication transition was never durably committed. Restrict reconciliation to post-sync faults; an fsync error must not authorize publication.

src/autoskillit/hooks/_capture/_delivery.py
```diff
@@ -0,0 +1,657 @@
...
+
+from __future__ import annotations
+
+import os
+import sys
+from collections.abc import Callable
+from dataclasses import replace
+from typing import Any, Protocol, TypeAlias
+
+from . import _ledger, _snapshot, _sweep
```
Trecek: [warning] arch: Delivery orchestration imports the cleanup/recovery _sweep module solely to call its private _same_record helper at L609. This couples delivery transitions to a sweep implementation detail. Move the comparison into _ledger or a shared lifecycle/store primitive used by both modules.

src/autoskillit/hooks/_capture/_delivery.py
```diff
@@ -0,0 +1,657 @@
...
+        raise RuntimeError("conflicting shell-capture delivery module identity")
+
+DeliveryValue: TypeAlias = (
+    _snapshot.FinalizedCapture
+    | _snapshot.PublishedCaptureReference
+    | _snapshot.UnavailableCaptureReference
+)
+RuntimeErrors: TypeAlias = tuple[type[Exception], ...]
+
+_REFERENCE_TRANSITIONS = {
```
Trecek: [warning] cohesion: _REFERENCE_TRANSITIONS and _DELIVERY_TRANSITIONS restate the successor graphs already defined by _ledger._REFERENCE_SUCCESSORS and _ledger._DELIVERY_SUCCESSORS, differing mainly by omitted self/terminal entries. Transition legality now has two sources of truth that can drift as statuses evolve; derive action transitions from one shared policy.

src/autoskillit/hooks/_capture/_snapshot.py
```diff
@@ -0,0 +1,994 @@
...
+]
+
+SNAPSHOT_SCHEMA_VERSION = 2
+SNAPSHOT_PRODUCER = "codex_shell_capture"
+MANAGED_STREAM_DOMAIN = "combined_stdout_stderr_pipe_eof"
+MAX_MANIFEST_BYTES = 8 * 1024
+MAX_REFERENCE_TOKEN_BYTES = 192
+
+_AUTHORITY_FACTORY_TOKEN = object()
+_CAPTURE_ID_RE = re.compile(r"^[0-9a-f]{16}$")
```
Trecek: [warning] cohesion: Capture identity syntax is scattered across independent regex copies: snapshot defines capture/incarnation/digest/reference patterns here, ledger repeats capture/incarnation, the transport contract repeats capture/digest/reference, and lifecycle repeats capture/incarnation. A format change can make manifests, persisted records, and emitted markers disagree; centralize the shared validators.

src/autoskillit/hooks/_capture/_resolver.py
```diff
@@ -0,0 +1,272 @@
...
+            if (
+                current is None
+                or current.revision != revision
+                or current.manifest_bytes != manifest_bytes
+                or rechecked_manifest != current_manifest
+            ):
+                raise lifecycle_error("producer capture changed during verification")
+        return _reader._make_verified_reader(fd, current_manifest, revision)
+    except BaseException:
+        os.close(fd)
```
Trecek: [warning] defense: The cleanup close runs unguarded inside an active BaseException handler. If os.close(fd) fails, its OSError replaces the original authority or lifecycle exception and destroys the diagnostic context. Preserve the primary exception while handling cleanup failure separately.

src/autoskillit/hooks/_capture_contract.py
```diff
@@ -1,8 +1,479 @@
...
+        "status",
+    }
+)
+
+
+class CaptureContractError(ValueError):
+    """Raised when a V2 capture transport value is invalid or noncanonical."""
+
+
+@dataclass(frozen=True, slots=True)
```
Trecek, line 104: [warning] defense: The exported CaptureV2Fields and CaptureFailureV2 boundary dataclasses accept invalid values at construction and validate only during rendering. Invalid transport objects can propagate and fail far from their source; enforce their invariants in __post_init__.

src/autoskillit/hooks/_capture/_ledger.py
```diff
@@ -0,0 +1,966 @@
...
+
+
+_CAPTURE_ID_RE = re.compile(r"^[0-9a-f]{16}$")
+_INCARNATION_RE = re.compile(r"^[0-9a-f]{32}$")
+_PUBLIC_NAME_RE = re.compile(r"^shell_[0-9a-f]{16}\.log$")
+_STAGING_NAME_RE = re.compile(r"^\.capture-staging-[0-9a-f]{16}-[0-9a-f]{16}$")
+_QUARANTINE_NAME_RE = re.compile(r"^\.capture-quarantine-[0-9a-f]{16}-[0-9a-f]{16}$")
+
+
+@dataclass(frozen=True, slots=True)
```
Trecek, line 101: [warning] defense: CaptureLifecycleRecord is the exported durable-state model but performs no construction-time validation; malformed identities, revisions, and incompatible state axes are accepted until a later append or serialization. Validate invariants in __post_init__ so invalid lifecycle authority cannot exist in memory.

tests/execution/backends/test_codex_capture_lifecycle_integration.py
```diff
@@ -39,12 +55,21 @@ def _seed_due_capture(project: Path) -> Path:
...
+        carrier_identity=(artifact.identity.device, artifact.identity.inode),
+        measurement=CaptureMeasurement.from_bytes(b"due", inline_bytes=3),
+        command_outcome=CommandOutcome.exited(0),
+        expected_revision=artifact.authority.expected_revision,
+        finalized_at=old,
+        retention_deadline=old + 3600.0,
     )
+    lifecycle.commit_verified_snapshot(verified, issue_reference=False)
     artifact_path = project.joinpath(*CAPTURE_PATH_COMPONENTS, artifact.name)
     artifact.close_artifact_fd()
```
Trecek, line 74: [warning] tests: The updated due-capture helper closes only the artifact data descriptor before closing the root, leaving the separate exclusive writer-lease descriptor open. The same omission appears in tests/hooks/test_capture_lifecycle_hook.py at L170. These helpers leak a flock/file descriptor into the xdist worker and do not model producer release before cleanup; call artifact.release_lease().

tests/hooks/test_capture_lifecycle.py
```diff
@@ -90,15 +138,487 @@ def _capture_dir(project: Path) -> Path:
...
+
+
+def _race_calls(*calls: Callable[[], object]) -> list[object]:
+    barrier = threading.Barrier(len(calls) + 1)
+
+    def invoke(call: Callable[[], object]) -> object:
+        barrier.wait(timeout=5)
+        try:
+            return call()
+        except (CaptureLifecycleError, OSError) as exc:
```
Trecek: [warning] tests: _race_calls converts every OSError into an ordinary loser result, while the race tests only count Exception versus success. An unrelated I/O or descriptor failure can satisfy the expected one-loser outcome and leave the tests green; assert the precise allowed loser exception and relevant message or errno.

tests/hooks/test_capture_verified_reader.py
```diff
@@ -0,0 +1,818 @@
...
+    carrier.unlink()
+    os.mkfifo(carrier, 0o600)
+    code = (
+        "import sys;"
+        "from autoskillit.hooks._capture_artifacts import open_capture_lifecycle;"
+        "from autoskillit.hooks._capture_lifecycle import CaptureLifecycleError;"
+        "\ntry:\n"
+        "  with open_capture_lifecycle(sys.argv[1],create=False) as lifecycle:\n"
+        "    lifecycle.open_verified_capture(sys.argv[2])\n"
+        "except (CaptureLifecycleError,OSError):\n"
```
Trecek: [warning] tests: The FIFO-substitution subprocess treats any CaptureLifecycleError or OSError as success, so unrelated ledger, project-open, or resolver failures satisfy the test without proving the FIFO was rejected for unsafe carrier metadata. Have the child validate the expected exception/message or emit a reason that the parent asserts.

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments; reconciliation must not treat readable-but-unsynced ledger frames as durable authority.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in `aec494e07`. Lifecycle appends now distinguish the narrow, proven post-`fsync` descriptor-close case with `CaptureTransitionCommittedError`; an ordinary transition exception is no longer reconciled from a merely readable record. Regression tests cover reported sync failures and unclassified post-append faults.

src/autoskillit/hooks/_capture/_delivery.py
Trecek: Agreed — fixed in `aec494e07`. Delivery reconciliation now accepts only the explicit durable-commit signal, so generic exceptions cannot promote readable `ATTEMPTING` or `DELIVERED` state to success. The new tests exercise classified and unclassified transition failures.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in `aec494e07`. Publication now returns a bearer reference after an exception only when the ledger raised the explicit post-durability commit signal; an unclassified publication failure remains fail-closed. Regression coverage verifies that distinction.

src/autoskillit/hooks/_capture/_delivery.py
Trecek: Agreed — fixed in `76be67361`. Lifecycle-record equality now lives in the ledger authority module as `same_record`, and sweep/delivery consumers share that implementation instead of reaching into `_sweep` internals.

src/autoskillit/hooks/_capture/_delivery.py
Trecek: Agreed — fixed in `b2691815c` and finalized in `f60795f6e`. The lifecycle enums and all state/reference/delivery/retention successor graphs now have one canonical home in `_capture/_lifecycle_policy.py`; ledger validation and delivery transitions consume the same policy helpers.

src/autoskillit/hooks/_capture/_snapshot.py
Trecek: Agreed — fixed in `5ed469817` with the import-locality follow-up in `f60795f6e`. Capture ID, incarnation, digest, and reference syntax are centralized in `_capture/_syntax.py`, and contract, snapshot, ledger, and lifecycle boundaries reuse those canonical patterns.

src/autoskillit/hooks/_capture/_resolver.py
Trecek: Agreed — fixed in `77d84c9a7`. Resolver cleanup no longer replaces an active authority/validation exception: the primary exception propagates, and a simultaneous descriptor-close failure is attached as a note. A regression test exercises both failures together.

src/autoskillit/hooks/_capture_contract.py
Trecek: Agreed — fixed in `cb0f78ecf`. `CaptureV2Fields` and `CaptureFailureV2` now validate in `__post_init__`, so invalid transport values cannot exist between construction and rendering. Tests assert rejection at construction time.

src/autoskillit/hooks/_capture/_ledger.py
Trecek: Agreed — fixed in `c0f747717`. `CaptureLifecycleRecord.__post_init__` now runs the canonical record validator, preventing invalid state/status combinations from existing before append or serialization; regression tests exercise direct construction and `replace`.

tests/execution/backends/test_codex_capture_lifecycle_integration.py
Trecek: Agreed — fixed in `80f29e4e8`. Both capture helpers now release the artifact lease before closing the capture root/project authority, and their cleanup tests preserve that ownership order.

tests/hooks/test_capture_lifecycle.py
Trecek: Agreed — fixed in `9df8272ec`. The race helper now catches only `CaptureLifecycleError`, and each race test asserts the precise allowed loser message plus the expected winner, so unrelated `OSError` failures can no longer count as success.

tests/hooks/test_capture_verified_reader.py
Trecek: Agreed — fixed in `219baf518`. The FIFO subprocess now reports unexpected exceptions distinctly, the test asserts the exact `CaptureAuthorityError` reason (`capture carrier metadata changed`), and the test clock is current so expiry cannot produce a false positive.

src/autoskillit/hooks/_capture/_sweep.py
```diff
@@ -58,6 +85,517 @@ def create_verified_recovery_link(
...
+            compaction_epoch=compaction_epoch,
+            ledger_size=size,
+            authority=store._authority_for(expected),
+            allowed_states={expected.state},
+            transform=transform,
+        )
+
+
+def sweep_one(
+    store: Any,
```
Trecek: [warning] arch: The new lifecycle implementation shards have no stable store boundary: sweep_one accepts store as Any and drives its private locking, loading, transition, authority, cleanup, and identity internals; normalize_interrupted_deliveries does the same, while resolver's protocol exposes private store fields. This disperses atomic lifecycle ownership outside CaptureLifecycleStore and prevents type checking from catching internal-contract drift. Keep transaction choreography in the store or introduce a typed internal store port.

src/autoskillit/hooks/_capture/_delivery.py
```diff
@@ -0,0 +1,621 @@
...
+            lifecycle_error=lifecycle_error,
+        )
+
+    store._transition_current(
+        manifest.capture_id,
+        manifest.incarnation,
+        allowed_states={_ledger.CaptureState.FINALIZED},
+        transform=transform,
+    )
+    return _snapshot._make_unavailable_reference(snapshot, reason_code)
```
Trecek: [critical] defense: reason_code is validated only when constructing the return value, after _transition_current has durably changed reference_status to UNAVAILABLE. An invalid public argument therefore commits the transition and then raises, leaving durable state mutated without returning the corresponding authority object. Validate the reason code before the transition.

src/autoskillit/hooks/_capture/_snapshot.py
```diff
@@ -0,0 +1,990 @@
...
+                    not isinstance(self.settlement_returncode, int)
+                    or isinstance(self.settlement_returncode, bool)
+                )
+            )
+        ):
+            raise CaptureAuthorityError("invalid capture failure evidence")
+
+
+@dataclass(frozen=True, slots=True)
+class CaptureManifestWire:
```
Trecek, line 234: [warning] defense: CaptureManifestWire is exported but has no construction-time validation; arbitrary schema versions, identities, statuses, lengths, and timestamps can exist until a later private restoration call happens to invoke _validate_manifest. Enforce the wire invariants in __post_init__ so the public typed boundary cannot create malformed transport state.

src/autoskillit/hooks/_capture/_reader.py
```diff
@@ -0,0 +1,145 @@
...
+        ):
+            raise CaptureAuthorityError("invalid verified capture reader")
+
+    def __enter__(self) -> VerifiedCaptureReader:
+        if self._descriptor < 0:
+            raise CaptureAuthorityError("verified capture reader is closed")
+        return self
+
+    def __exit__(self, *_exc: object) -> None:
+        self.close()
```
Trecek: [warning] defense: __exit__ calls close() without preserving an active exception. If os.close fails while the with-body is already raising, the cleanup OSError replaces the authority/read failure and destroys its diagnostic context.

src/autoskillit/hooks/_capture/_resolver.py
```diff
@@ -0,0 +1,278 @@
...
+) -> int:
+    try:
+        lease_fd = os.dup(artifact_fd)
+    except OSError as exc:
+        raise lifecycle_error("cannot duplicate writer lease descriptor") from exc
+    try:
+        fcntl.flock(lease_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        return lease_fd
+    except OSError as exc:
+        os.close(lease_fd)
```
Trecek: [warning] defense: The lease descriptor close runs unguarded while handling flock failure. A close OSError replaces the original contention or capability error, unlike the context-preserving cleanup already used later in this module.

src/autoskillit/hooks/_capture/_resolver.py
```diff
@@ -0,0 +1,278 @@
...
+                or current.manifest_bytes != manifest_bytes
+                or current_manifest != manifest
+            ):
+                raise lifecycle_error("verified capture changed during resolution")
+        reader = _reader._make_verified_reader(fd, manifest, revision)
+        fd = -1
+        return reader
+    finally:
+        if fd >= 0:
+            os.close(fd)
```
Trecek: [warning] defense: The finally close can replace any token parsing, ledger, lease, or descriptor-verification exception from open_verified_capture. Preserve the primary exception and attach cleanup failure as context.

src/autoskillit/hooks/_capture/_delivery.py
```diff
@@ -0,0 +1,621 @@
...
+                        _ledger.CaptureReferenceStatus.PUBLISHED,
+                    }
+                )
+            )
+        ]
+    for expected in sorted(candidates, key=lambda record: record.capture_id):
+        lease = None
+        try:
+            lease = store._acquire_cleanup_lease(expected)
+        except (lease_live, tampered, lifecycle_error, OSError):
```
Trecek: [warning] defense: Restart normalization silently treats every lifecycle_error and OSError from cleanup-lease acquisition like ordinary lease contention. Unexpected descriptor, ledger, or authority failures therefore leave ATTEMPTING/PUBLISHED state stale while store construction reports success, with no context or outcome exposing the failed normalization.

src/autoskillit/hooks/_capture/_delivery.py
```diff
@@ -0,0 +1,621 @@
...
+                    records=records,
+                    compaction_epoch=compaction_epoch,
+                    ledger_size=size,
+                    authority=store._authority_for(current),
+                    allowed_states={_ledger.CaptureState.FINALIZED},
+                    transform=lambda _record: candidate,
+                )
+        finally:
+            if lease is not None:
+                os.close(lease.fd)
```
Trecek, line 558: [warning] defense: The cleanup-lease close is unguarded in a finally block, so a close failure can replace the transition error being propagated or turn a successfully normalized candidate into an apparent failure. Preserve the primary result/error and record cleanup failure separately.

tests/hooks/test_capture_lifecycle.py
```diff
@@ -90,15 +160,538 @@ def _capture_dir(project: Path) -> Path:
...
+        "staging_name": f".capture-staging-{capture_id}-{'1' * 16}",
+        "state": state.value,
+    }
+
+
+def _race_calls(*calls: Callable[[], object]) -> list[object]:
+    barrier = threading.Barrier(len(calls) + 1)
+
+    def invoke(call: Callable[[], object]) -> object:
+        barrier.wait(timeout=5)
```
Trecek, line 222: [warning] tests: The race harness only synchronizes threads immediately before invoking each operation. It does not force both operations to read the same predecessor before either commits, so the accepted one-success/one-stale-loser result is also produced by fully serial execution. Locking or lost-update regressions can therefore remain green; add a deterministic interleaving barrier inside the transition path.

tests/hooks/test_capture_lifecycle.py
```diff
@@ -90,15 +160,538 @@ def _capture_dir(project: Path) -> Path:
...
+    )
+
+
 def _finalized_capture(project: Path, clock: _Clock, capture_id: str = _CAPTURE_ID):
     anchor, root, store = _open_store(project, clock)
     artifact = create_capture_artifact(root, capture_id, store)
     os.write(artifact.fd, b"captured")
-    os.fsync(artifact.fd)
-    store.finalize_capture(capture_id, size=8, sha256=_DIGEST, failed=False)
+    _commit_verified(store, artifact, b"captured", clock)
```
Trecek: [warning] tests: _finalized_capture acquires an anchor, root, artifact descriptor, and exclusive lease, then calls _commit_verified before establishing any cleanup guard. If verification or commit raises—the failure these tests are meant to expose—the helper exits without closing those resources, contaminating the xdist worker. Clean up partially constructed resources before re-raising.

tests/hooks/test_capture_verified_reader.py
```diff
@@ -0,0 +1,862 @@
...
+
+
+def _finalize(
+    project: Path,
+    clock: _Clock,
+    data: bytes,
+    *,
+    issue_reference: bool = True,
+):
+    anchor, root, store = _open_store(project, clock)
```
Trecek: [warning] tests: _finalize and the _publish wrapper acquire anchor/root descriptors plus an artifact descriptor and flock lease without exception cleanup. Any create, verify, commit, or publication regression leaks all acquired resources into the continuing xdist worker. Use an ExitStack or an exception path that closes the artifact, releases its lease, and closes root/anchor before re-raising.

tests/hooks/test_capture_verified_reader.py
```diff
@@ -0,0 +1,862 @@
...
+) -> None:
+    clock = _Clock()
+    anchor, root, store, artifact, _finalized, _published = _publish(
+        tmp_path / "project",
+        clock,
+        b"authority",
+    )
+    _close_artifact(artifact)
+    try:
+        with pytest.raises((CaptureAuthorityError, CaptureLifecycleError)):
```
Trecek: [warning] tests: This test accepts either CaptureAuthorityError or CaptureLifecycleError for syntactically invalid, non-capability inputs. If reference parsing regresses and accepts one of these values, the later missing-record lookup raises CaptureLifecycleError and the test still passes. Assert CaptureAuthorityError (and its parse-validation message) so rejection is proven at the intended boundary.

tests/hooks/test_capture_verified_reader.py
```diff
@@ -0,0 +1,862 @@
...
+    clock = _Clock()
+    anchor, root, store, artifact, _finalized, published = _publish(
+        tmp_path / "project-a",
+        clock,
+        b"bound",
+    )
+    _close_artifact(artifact)
+    other_anchor, other_root, other_store = _open_store(tmp_path / "project-b", clock)
+    try:
+        with pytest.raises(CaptureLifecycleError):
```
Trecek: [warning] tests: The physical-project binding test opens the token against an empty second project. It therefore passes at the earlier 'record unavailable' check and never exercises the project/root identity comparison; removing physical-binding validation would leave it green. Seed project B with an otherwise valid published record and assert the specific physical-binding mismatch.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -619,76 +675,170 @@ def run_capture(command: str, cwd: str, capture_id: str) -> int:
...
+                    payload = _render_inline_capture(delivery_value)
+                elif isinstance(
+                    delivery_value,
+                    (PublishedCaptureReference, UnavailableCaptureReference),
+                ):
+                    payload = _render_oversized_capture(delivery_value)
+                else:
+                    raise CaptureSetupError("capture delivery value is unavailable")
+                failure_stage = "capture stdout write and flush"
+                _write_and_flush_hook_stdout(payload)
```
Trecek: [warning] bugs: The progress flag is set only after the combined write-and-flush helper returns. If a short write succeeds and a later write or flush raises, stdout already contains capture bytes but delivery_bytes_flushed remains false, so settlement records FAILED instead of UNKNOWN. Track partial write progress so an ambiguous delivery is never represented as a definite failure.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -163,32 +234,32 @@ def create_capture_artifact(
...
         identity = FileIdentity(device=raw_identity[0], inode=raw_identity[1])
         return CaptureArtifact(
             fd=fd,
             name=public_name,
             identity=identity,
             lease_fd=lease_fd,
+            authority=authority,
             _factory_token=_ARTIFACT_FACTORY_TOKEN,
         )
     except (CaptureLifecycleError, OSError) as exc:
```
Trecek: [warning] bugs: CaptureTransitionCommittedError directly subclasses RuntimeError, but both artifact-creation exception handlers omit it. A post-fsync descriptor-close fault from mark_staged, mark_published, or commit_capture_failure therefore bypasses the fd/lease cleanup below this block, leaking the exclusive carrier lease and leaving a durably transitioned record without reconciliation.

src/autoskillit/hooks/_capture/_sweep.py
```diff
@@ -58,6 +85,517 @@ def create_verified_recovery_link(
...
+        ):
+            raise Tampered
+        return ObservedArtifact(
+            fd=fd,
+            identity=identity,
+            nlink=value.st_nlink,
+            size=value.st_size,
+        )
+    except BaseException:
+        os.close(fd)
```
Trecek: [warning] bugs: The cleanup close is unguarded while handling another exception. If validation raises Tampered and os.close(fd) also fails, the OSError replaces the tampering exception; sweep_one then takes its retry/error path instead of durably marking the record TAMPERED. Preserve the primary exception and attach cleanup failure as secondary context.

src/autoskillit/hooks/_capture/_descriptor.py
```diff
@@ -0,0 +1,97 @@
...
+
+import hashlib
+import hmac
+import json
+import os
+import stat
+import sys
+from typing import Protocol
+
+_THIS_MODULE = sys.modules[__name__]
```
Trecek: [warning] cohesion: Dual-import identity registration is copy-pasted across 14 changed modules, each restating its own alias tuple and conflict check. Adding or renaming one module can omit one spelling and recreate the split class/enum identities this machinery prevents. Centralize a derived alias-registration bootstrap.

src/autoskillit/hooks/_capture/_snapshot.py
```diff
@@ -0,0 +1,990 @@
...
+        raise CaptureAuthorityError(f"invalid {field}")
+    return float(value)
+
+
+class CommandOutcomeKind(StrEnum):
+    EXITED = "exited"
+    SIGNALED = "signaled"
+
+
+class CaptureStatus(StrEnum):
```
Trecek: [warning] cohesion: The finalized-snapshot module owns lifecycle-only concepts: CaptureStatus includes PENDING, FAILED, and LEGACY_CLEANUP_ONLY, and the same module defines CaptureFailureEvidence and LegacyCleanupOnly. _ledger must consequently import non-snapshot lifecycle policy from _snapshot. Move outcome/evidence types beside the other lifecycle axes or split manifest status from lifecycle outcome.

Trecek (commented):
AutoSkillit PR Review — Verdict: changes_requested (self-authored PR; GitHub does not permit REQUEST_CHANGES)

src/autoskillit/hooks/_capture/_resolver.py
```diff
@@ -0,0 +1,278 @@
...
+
+
+def adopt_verified_capture(
```
Trecek, line 246: [warning] cohesion: adopt_verified_capture independently duplicates open_verified_capture's security-sensitive load/authorize, descriptor-inspect, out-of-lock digest verification, reload, and revision/manifest comparison sequence. The two paths already differ in acquisition and predicates, so later hardening can land in only one. Consider a shared linearized verification primitive parameterized by authority and descriptor acquisition.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -619,76 +675,170 @@ def run_capture(command: str, cwd: str, capture_id: str) -> int:
...
+                    _failure_transport(
+                        stage="artifact_write",
+                        detail="capture artifact write failed",
+                        shell_returncode=command_returncode,
+                        settlement=None,
+                    )
+                )
             failure_stage = "capture artifact integrity verification"
-            _verify_capture_artifact(artifact, result)
+            finalized_at = time.time()
```
Trecek: [warning] cohesion: Successful finalization uses time.time() and a separate literal 3600-second deadline here, while failed/abandoned lifecycle paths use CaptureLifecycleStore._wall_clock and _RETENTION_SECONDS. The same retention policy therefore has two clock and duration authorities; injected clocks or a future retention change can make successful and failed captures age differently. Have the lifecycle policy supply both timestamps.

src/autoskillit/hooks/_capture/_replay.py
```diff
@@ -0,0 +1,223 @@
...
+    write_all_stream(sys.stdout.buffer, payload, boundary="hook stdout")
+    sys.stdout.buffer.flush()
+
+
+def _write_and_flush_hook_stderr(payload: bytes) -> None:
+    write_all_stream(sys.stderr.buffer, payload, boundary="hook stderr")
+    sys.stderr.buffer.flush()
+
+
+def _failure_stage(value: str) -> str:
```
Trecek, line 120: [warning] cohesion: Failure metadata policy is restated across three modules: replay normalizes/truncates stages and details to 64/240, CaptureFailureEvidence independently enforces those byte limits, and CaptureFailureV2 independently defines the transport regex and 240-byte limit. Changing one boundary can make persisted evidence and emitted failure transport incompatible. Share the limits and conversion/normalization contract.

src/autoskillit/hooks/_capture/_snapshot.py
```diff
@@ -0,0 +1,990 @@
...
+    PENDING = "pending"
+    COMPLETE = "complete"
+    PARTIAL = "partial"
```
Trecek: [warning] overengineering: CaptureStatus advertises PARTIAL even though the lifecycle makes that value impossible. Remove the enum member and the explicit PARTIAL rejection; persisted "partial" input will still fail enum decoding, preserving every reachable behavior while avoiding a fictitious state.

Trecek (commented):
AutoSkillit review: blocking findings detected. See inline comments; changes are requested. GitHub does not permit REQUEST_CHANGES on a self-authored PR.

src/autoskillit/hooks/_capture/_sweep.py
Trecek: Valid observation — flagged for design decision. The store currently owns the transaction boundary, but replacing that with a typed store port is a broader architectural choice that should not be folded into this review fix.

src/autoskillit/hooks/_capture/_delivery.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. The unavailability reason is now validated before the lifecycle transition, so invalid input cannot mutate state first.

src/autoskillit/hooks/_capture/_snapshot.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. `CaptureManifestWire` now validates its cross-field invariants at construction time.

src/autoskillit/hooks/_capture/_reader.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Reader cleanup now preserves an active body exception and attaches cleanup failure as secondary evidence.

src/autoskillit/hooks/_capture/_resolver.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Writer-lease cleanup now preserves the primary exception and records close failure without masking it.

src/autoskillit/hooks/_capture/_resolver.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Verified-open cleanup now preserves the primary verification error while retaining descriptor-close evidence.

src/autoskillit/hooks/_capture/_delivery.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Interrupted-delivery normalization now limits recovery to the intended lease/tamper cases and surfaces unexpected lifecycle or OS failures.

src/autoskillit/hooks/_capture/_delivery.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Normalization cleanup now preserves the primary recovery error if descriptor close also fails.

tests/hooks/test_capture_lifecycle.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. The race harness now performs its mutation inside the transition boundary, making the interleaving deterministic.

tests/hooks/test_capture_lifecycle.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. The lifecycle test helper now reliably closes the store it creates.

tests/hooks/test_capture_verified_reader.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. The verified-reader helper now closes both the reader and its lifecycle store during teardown.

tests/hooks/test_capture_verified_reader.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. The parser regression test now asserts the exact authority exception instead of accepting an overly broad class.

tests/hooks/test_capture_verified_reader.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. The physical-binding test now supplies the copied ledger state needed to reach and verify the intended root-binding guard.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Positive stdout write progress is recorded before return-count validation, so ambiguous partial delivery transitions to `UNKNOWN`.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Artifact creation and recovery now treat committed-transition failures as cleanup-requiring outcomes.

src/autoskillit/hooks/_capture/_sweep.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Sweep cleanup now keeps the tamper error primary and attaches cleanup failure as secondary evidence.

src/autoskillit/hooks/_capture/_descriptor.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Supported shell-capture import aliases are now registered through one shared module-identity bootstrap.

src/autoskillit/hooks/_capture/_snapshot.py
Trecek: Valid observation — flagged for design decision. The lifecycle concepts in the snapshot module are real coupling, but moving those contracts across module boundaries is a broader architecture migration than this review-fix pass.

src/autoskillit/hooks/_capture/_resolver.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Open and adopt now share one linearized verification primitive while retaining path-specific cleanup context.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Finalization and retention timestamps now come from the lifecycle store's clock authority.

src/autoskillit/hooks/_capture/_replay.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. Capture failure-field policy is now centralized in the capture subpackage and shared by both producers.

src/autoskillit/hooks/_capture/_snapshot.py
Trecek: Agreed — fixed in 985470467e22f366fc3e21de74887a3444f67212. The unreachable `PARTIAL` manifest status was removed from the model and its dependent logic.

src/autoskillit/hooks/_capture/_sweep.py
Trecek: Follow-up after investigation: implemented in a17b79f13d2cf37ec961b5bd64c7c95c38da499b. The cleanup choreography remains in the intentional `_sweep` shard, but `_store_port.py` now defines canonical interface-segregated read, transition, delivery-normalization, resolver, and sweep capabilities. `_sweep`, `_delivery`, and `_resolver` no longer use `store: Any` or duplicate local store protocols, so mypy now detects capability drift without changing lock/CAS ordering. Added an architecture regression guard against noncanonical shard store annotations.

src/autoskillit/hooks/_capture/_snapshot.py
Trecek: Follow-up after investigation: implemented in a17b79f13d2cf37ec961b5bd64c7c95c38da499b. `CaptureStatus` now lives with the canonical lifecycle axes in `_lifecycle_policy.py`; `CaptureFailureEvidence` and `LegacyCleanupOnly` now live with lifecycle outcomes/internal signals in `_types.py`. `_ledger` and orchestration import the canonical owners directly, while `_snapshot` retains only FINAL manifest/verification/reference authority. Wire values and the manifest `COMPLETE` invariant are unchanged, with packaged/standalone identity coverage added.

src/autoskillit/hooks/_capture/_delivery.py
```diff
@@ -0,0 +1,589 @@
...
+        lease = None
+        try:
+            lease = store._acquire_cleanup_lease(expected)
+        except (lease_live, tampered):
+            continue
+        try:
+            with store._locked():
+                records, compaction_epoch, size = store._load_locked()
+                current = records.get(expected.capture_id)
+                if current is None or not _ledger.same_record(expected, current):
```
Trecek, line 533: [warning] bugs: A stale-record race leaks the cleanup-lease descriptor. After _acquire_cleanup_lease() succeeds, this continue skips the try-else block containing the only normal-path os.close(lease.fd). Close the lease in a finally block so every non-exception exit releases it.

src/autoskillit/hooks/_capture_lifecycle.py
```diff
@@ -1,70 +1,86 @@
...
 _COMPACTION_THRESHOLD_BYTES = 3 * 1024 * 1024
 _MAX_COMPACTION_BYTES = 4 * 1024 * 1024
 _MAX_TOMBSTONES = 256
-_FRAME_HEADER_FORMAT = ">4sI"
-_FRAME_HEADER_BYTES = 8
-_CHECKSUM_BYTES = hashlib.sha256().digest_size
-_CAPTURE_ID_RE = re.compile(r"^[0-9a-f]{16}$")
-_SHA256_RE = re.compile(r"^[0-9a-f]{64}$")
+_CAPTURE_ID_RE = _capture_syntax.CAPTURE_ID_RE
 _PUBLIC_NAME_RE = re.compile(r"^shell_[0-9a-f]{16}\.log$")
```
Trecek: [warning] cohesion: Artifact-name syntax now has multiple authorities: these patterns are duplicated in _capture/_ledger.py, and the carrier pattern is duplicated again in _capture/_snapshot.py, while _capture/_syntax.py centralizes the related ID and digest syntax. Centralize the filename patterns too so ledger validation, snapshot verification, and cleanup cannot drift.

src/autoskillit/hooks/_capture_artifacts.py
```diff
@@ -69,14 +84,46 @@
...
+        CaptureTransitionCommittedError,
+    )
     from _hook_settings import (
         HOOK_CONFIG_FILENAME,
         HOOK_CONFIG_OVERLAY_FILENAME,
         merge_hook_configs,
     )
-    from _policy_event import PolicyEvent, render_capture_marker
+
+_RunnerSettlementEvidence = _capture_replay.RunnerSettlementEvidence
```
Trecek: [warning] cohesion: This twelve-name replay alias façade leaves two ownership surfaces: run_capture and the changed tests continue through aliases on capture_artifacts instead of the new replay module. Call a stable replay interface directly and move the monkeypatch targets with it so future replay changes have one authoritative seam.

Trecek (commented):

Trecek (commented):

src/autoskillit/hooks/_capture/_delivery.py
Trecek: Agreed — fixed in f262fbd46d300d3798d150ae4809051aa8488ea1. The stale-record control-transfer path now closes every successfully acquired cleanup-lease descriptor from a finally block, and the regression test verifies the descriptor is closed when the record changes after lease acquisition.

src/autoskillit/hooks/_capture_lifecycle.py
Trecek: Agreed — fixed in 43135380a07bbeaffb6c5d96b577bd6a99f8df68. Public, staging, and quarantine filename regexes now have one authority in _capture/_syntax.py; ledger validation, snapshot verification, lifecycle cleanup, and the structural contract consume that canonical syntax.

src/autoskillit/hooks/_capture_artifacts.py
Trecek: Agreed — fixed in 3c4bb80f00f8afc86e35dbbacfc77e1b936b53b9. _capture_artifacts.py now calls the replay module directly, the twelve-name alias façade is removed, and replay-focused tests monkeypatch and exercise that same authoritative module seam.

PR #4423, opened 2026-07-30
Rectify: Contracted Update, Install, and Plugin Publication Transaction

### LINKED ISSUE #4409: Codex-scoped update silently skips Claude plugin refresh and breaks Cook
## Problem

Running `autoskillit update` with the effective backend forced to Codex can upgrade the
AutoSkillit executable while silently skipping the required Claude plugin refresh. The
update still reports success. A later Claude Cook launch then fails before spawning:

```text
PluginArtifactValidationError: installed plugin root is invalid:
/home/talon/.claude/plugins/cache/autoskillit-local/autoskillit/0.10.910
```

This leaves a valid-looking but unusable cross-backend installation. A user should never
have to diagnose or manually repair this split transaction.

## Impact

- Claude Cook cannot launch.
- Codex Cook continues to work, masking the installation skew.
- `autoskillit update` can report success and restart despite not synchronizing the
  Claude plugin artifact.
- Existing diagnostics do not directly identify the absent current-version artifact.
- Recovery currently requires the user to rerun `autoskillit install` without the Codex
  backend override.

## Investigation

> Prior investigation completed interactively. The root cause, affected components,
> test gaps, history, and recommended remediation are documented below. Whoever picks
> this up should proceed from these findings rather than repeat root-cause investigation.

### Summary

The failure is caused by a split installation transaction. The running AutoSkillit
package is `0.10.910`, while Claude's registry, marketplace projection, and only cache
directory are all still `0.10.909`. The latest matching shell-history entry ran
`autoskillit update` with `AUTOSKILLIT_AGENT_BACKEND__BACKEND=codex`. The updater
inherited that override into its required post-upgrade `autoskillit install`;
installation declined because Codex is not plugin-install capable, but the CLI
represented that decline as exit status zero. The updater therefore accepted the binary
version advance as success and restarted without publishing the matching Claude
artifact. Claude Cook subsequently synthesized the required `0.10.910` cache path and
failed closed because it did not exist. Codex works because it uses generated-home mode
and never consumes the installed Claude plugin artifact. **[SUPPORTED]**

### Root Cause

The initiating contract failure spans three boundaries:

1. `run_update_command()` copies the entire caller environment into both subprocesses,
   including the Codex backend override, then invokes `autoskillit install`
   (`src/autoskillit/cli/update/_update.py:34-60`).
2. `install()` resolves that effective Codex backend and returns `False` before mutation
   because `plugin_install_capable` is false
   (`src/autoskillit/cli/_marketplace.py:348-360`). The public CLI wrapper does not
   translate `False` into a non-zero exit (`src/autoskillit/cli/app.py:245-256`).
3. The updater checks only the subprocess return code and later verifies only that
   Python package metadata advanced (`src/autoskillit/cli/update/_update.py:61-84`;
   `src/autoskillit/cli/update/_update_checks.py:80-104`). It never establishes that the
   matching Claude plugin transaction was published.

This exactly permits the observed state:

- executable/package and repository: `0.10.910`;
- `installed_plugins.json`, marketplace manifests, and cache: `0.10.909`;
- `~/.claude/plugins/cache/autoskillit-local/autoskillit/0.10.910`: absent;
- no persisted Claude metadata refers to `0.10.910`.

The immediate launch exception is intentional fail-closed behavior, not the originating
defect. Claude sees the installed plugin key, selects `IMPLICIT_INSTALLED`, and
constructs the root from the executing `__version__`
(`src/autoskillit/cli/_plugin_artifact.py:68-80,119-149`). Acquisition requires that
exact root to resolve strictly
(`src/autoskillit/cli/_plugin_artifact.py:250-267,459-469`), producing the reported
`FileNotFoundError` before Claude starts.

### Affected Components

- `src/autoskillit/cli/update/_update.py:34-84`: inherits the backend override, invokes
  install, and treats zero exit plus package advancement as update success.
  **[SUPPORTED]**
- `src/autoskillit/cli/_marketplace.py:321-371`: declines plugin installation for a
  non-capable backend by returning `False`. **[SUPPORTED]**
- `src/autoskillit/cli/app.py:245-256`: drops the declined-install result at the process
  boundary. **[SUPPORTED]**
- `src/autoskillit/cli/_plugin_artifact.py:68-149,243-302,459-469`: selects and validates
  the exact current-version installed artifact. **[SUPPORTED]**
- `src/autoskillit/cli/session/_session_cook.py:197-205,264-276`: selects and acquires
  the binding for Cook. **[SUPPORTED]**
- `src/autoskillit/workspace/_install_state.py:236-265` and
  `src/autoskillit/cli/doctor/_doctor_mcp.py:386-425`: diagnostics do not reliably
  identify an absent current-version cache artifact. **[SUPPORTED]**

### Data Flow

```text
AUTOSKILLIT_AGENT_BACKEND__BACKEND=codex autoskillit update
  → binary advances to 0.10.910
  → inherited Codex override reaches autoskillit install
  → installer returns False without refreshing Claude state
  → CLI exits zero
  → updater verifies only package version and reports success
  → Claude registry key still selects implicit-installed mode
  → authority synthesizes .../0.10.910
  → strict root validation raises
```

For Codex, `plugin_install_capable=False` causes `interactive_plugin_authority()` to
return generated-home mode instead
(`src/autoskillit/execution/backends/codex.py:1025-1034`;
`src/autoskillit/cli/_plugin_artifact.py:141-149`), so the skew is not on its launch
path.

### Test Gap Analysis

- Update tests confirm that `autoskillit install` is invoked and that the caller
  environment is forwarded, but do not compose that behavior with a Codex override and
  the real installer (`tests/cli/test_update_command.py:119-158`).
- The install tests explicitly require Codex-backed installation to return `False`
  (`tests/cli/test_install.py:303-326`), while app-wrapper tests accept that `False`
  without asserting a non-zero process result.
- The update failure test covers only a subprocess return code of one
  (`tests/cli/test_update_command.py:256-284`); the production decline returns zero at
  the CLI boundary.
- Cook tests replace `interactive_plugin_authority()` wholesale
  (`tests/cli/test_cook_interactive.py:52-75`), and the named dangling-path Cook test
  directly exercises projected authority rather than Cook's current selector
  (`tests/cli/test_plugin_source_authority.py:76-101`).
- Installed-authority tests publish an existing exact root or a wrong identity; no
  launch-level case combines an old valid install with a missing current-version root
  (`tests/cli/test_plugin_source_authority.py:267-347,423-442`).
- `_derived_version_findings()` skips a missing current-version plugin file
  (`src/autoskillit/workspace/_install_state.py:246-255`), so the diagnostic matrix does
  not fail on this state.

### Similar Patterns

Projected plugin authority may rematerialize an absent projection from the live package
(`src/autoskillit/workspace/_projected_artifact/authority.py:393-428`). Installed
authority deliberately cannot do that because Claude owns installation and publication
occurs only after successful `claude plugin install`
(`src/autoskillit/cli/_marketplace.py:440-468`). Reusing the old `0.10.909` artifact
would violate the exact semantic identity and artifact-lifetime design; the appropriate
correction is to make update/install transaction success truthful and verifiable.

### Design Intent Findings

- Commit `d7727e043f67` (2026-07-28), **Rectify: Issue #4382 — Live Plugin Artifact
  Lifetime**, introduced exact installed-artifact identities, shared launch leases,
  identity-validated mutation, and deferred retirement. Its strict validation is
  designed to prevent mixed-version or retired content from entering a live child
  process.
- Commit `9ccb19d4dd26` (2026-07-24) stopped using Claude-owned
  `installed_plugins.json.installPath` as a projection source after stale and dangling
  paths caused crashes and mixed-version execution.
- Commit `81280b5b0b20` (2026-05-16) introduced the backend guard that returns `False`
  for non-Claude installation.
- Commit `6ad19340a8d0` (2026-04-24) attempted to eliminate silent post-update install
  failures, but detects only non-zero subprocess exits. The `False`-to-zero path
  remained outside that fix's model.

### Historical Context

No prior `$investigate` invocation was found in the current Claude project logs. Git
history does show repeated plugin-cache and update/install correctness fixes
(`0ff269a6c229`, `ffc24b89ccdb`, `6ad19340a8d0`, `9ccb19d4dd26`, and `d7727e043f67`).
This incident is a recurrence of the stale install-state/Cook-crash family through a new
boundary: a declined install is reported as successful during a backend-scoped update.

### External Research

Claude's official documentation confirms that installed marketplace plugin versions are
copied into distinct versioned cache directories and that older versions are
garbage-collected. That supports exact versioned roots as expected behavior, but it is
not the cause here: local metadata and timestamps show `0.10.910` was never published to
Claude's cache.

Reference:
https://code.claude.com/docs/en/plugins-reference#plugin-caching-and-file-resolution

### Scope Boundary

**Investigated:** live package, registry, marketplace and cache state; the exact Cook
launch path; Codex divergence; update/install status propagation; diagnostics; tests;
design intent; relevant history; and official cache semantics.

**Not captured:** the full terminal output of the historical update command was not
persisted, so the shell-history command plus code and filesystem chronology establish
the initiating path without a captured transcript of its printed messages.

### Recommended Remediation

1. Make the update-to-install handoff independent of the session backend override, or
   explicitly synchronize an already-registered Claude plugin.
2. Give declined/deferred install a non-success process outcome and require update to
   verify the exact current installed artifact before reporting success or restarting.
3. Add an end-to-end regression using the real selector and subprocess contract:
   Codex-scoped update with an existing older Claude install, matching current install,
   missing current root, dangling registry path, and missing/corrupt identity across
   Claude and Codex launches.
4. Make doctor/install-state verification report a missing current-version artifact
   directly.

## Deduplication

No relevant open duplicate was found. Closed historical issues #150 and #4382 cover
related update/install consistency and artifact-lifetime concerns, but neither covers
the Codex-backend override causing a semantic install refusal to cross the CLI boundary
as exit status zero.

---

## Verification 2026-07-29 — symptom cleared, defect intact. Do not close.

**Current state is healthy.** Installed package is `0.10.912` and
`~/.claude/plugins/cache/autoskillit-local/autoskillit/` contains `0.10.909`, `0.10.910` and
`0.10.912` (the last written 2026-07-28 21:57). So the artifact for the running version exists and
Claude Cook launches. Someone ran `autoskillit install` without the Codex override — that repaired
the *instance*.

**The mechanism is unchanged.** `git log --since=2026-07-28` on `cli/update/_update.py` and
`cli/_marketplace.py` shows no fix; the only touch is `d7727e043` (#4382 plugin artifact lifetime),
which is unrelated. All three original links still hold:

1. `_update.py:35-39` still builds `_skip_env = {**os.environ, ...}` and passes it to the
   post-upgrade `autoskillit install`, so a Codex backend override is still inherited.
2. `_marketplace.py:355-360` still `return False` when `plugin_install_capable` is false.
3. `app.py:254-256` still discards that boolean:

```python
completed = _install(scope=scope)
if completed:
    _print_next_steps(context="install")
```

No `sys.exit(1)`, no raise — the command returns normally, so **the decline exits 0**.

### The guard that looks like protection but cannot fire

`_update.py:61-68` *does* now check the child's status:

```python
if install_result.returncode != 0:
    print("\nautoskillit install exited with an error. "
          "Hooks and plugin cache may be stale. ...")
```

This is dead code for this failure mode. The decline path exits **0**, so the branch never
executes and the warning never prints. Anyone auditing `_update.py` alone would reasonably conclude
the case is handled. **It is not** — the check is downstream of a boolean that was already thrown
away one process earlier.

That makes the smallest correct fix a one-liner at the CLI boundary:

```python
completed = _install(scope=scope)
if not completed:
    raise SystemExit(1)
_print_next_steps(context="install")
```

With that, the existing `returncode != 0` guard starts working as intended and the rest of this
issue's remediation becomes about *quality* of the message rather than presence of one.

### What "broken" means, concretely

Not degraded — **hard launch failure**. Claude Cook synthesizes the cache path for the currently
running version and fails closed when it is absent:

```
PluginArtifactValidationError: installed plugin root is invalid:
/home/talon/.claude/plugins/cache/autoskillit-local/autoskillit/<version>
```

No order/cook session can start at all. Codex Cook continues working because it uses generated-home
mode and never consumes the installed Claude artifact, which is what masks the skew.

### Why it will recur

Multiple long-lived `autoskillit cook` processes on this machine carry
`AUTOSKILLIT_AGENT_BACKEND__BACKEND=codex` in their environment. Any `autoskillit update` invoked
from such a shell reproduces the split transaction exactly. Recovery is one command
(`autoskillit install` with no override), but the failure is silent until the next Claude launch,
so the user learns about it at the worst moment.

### Suggested acceptance addition

A regression test that runs the update path with a non-plugin-capable backend forced in the
environment and asserts the command exits non-zero — i.e. that the decline is *representable* as
failure, not merely printed.

## Resolution — 2026-08-27

Resolved on `develop` by **#4423**, merged as `295c591` (“Rectify: Contracted Update,
Install, and Plugin Publication Transaction”) on 2026-07-31. That work landed after the
2026-07-29 verification above and replaces the split update/install path with a contracted
transaction:

- update snapshots an existing Claude plugin registration and makes matching publication an
  explicit maintenance obligation even when the caller's ambient backend is Codex;
- the maintenance child environment is sealed and excludes Codex backend overrides, Claude
  session state, and unrelated ambient authority;
- maintenance installation is distinct from a direct Codex install decline;
- declined, deferred, failed, recovery-required, and indeterminate install outcomes cross the
  CLI and updater process boundaries as distinct non-success statuses;
- update reports success and restarts only after verifying the exact expected-version plugin
  generation, semantic identity, and digest; and
- doctor/install-state verification reports a registered plugin whose current generation is
  missing.

Focused and integration tests cover the historical Codex-scoped update with an older Claude
registration, missing/wrong/corrupt current generations, typed process-status propagation, and
success-effect suppression.

The `in-progress` label was applied on 2026-07-30 at 02:36:18 UTC for the historical #4423
implementation run. #4423 merged the following day. There is no open PR, attached worktree,
current claim, or active implementation session for this issue; the surviving local implementation
branch has had no activity since 2026-07-31. The label is stale, and this issue is closed as fixed
by #4423.

Trecek:
## Summary

Issue #4409 is a split-transaction defect, not a Cook artifact-selection defect. An update can advance the AutoSkillit package, decline the required Claude plugin refresh, lose that semantic decline at the CLI boundary, verify only package metadata, and then report success and restart. Claude later selects installed-plugin mode and correctly fails closed because the exact current-version artifact does not exist. Codex continues through generated-home mode and masks the skew.

Related issue: [#4409 — Codex-scoped update silently skips Claude plugin refresh and breaks Cook](https://github.com/TalonT-Org/AutoSkillit/issues/4409)

Closes #4409

## Implementation Plan

Plan file: `/home/talon/projects/generic_automation_mcp/.autoskillit/temp/rectify/rectify_update_install_publication_transaction_2026-07-29_192008.md`


src/autoskillit/cli/_install_contract.py
```diff
@@ -0,0 +1,183 @@
...
+    DECLINED = 10
+    DEFERRED = 11
+    FAILED_PREFLIGHT = 20
+    FAILED_CHILD = 21
+    FAILED_POSTCONDITION = 22
+    RECOVERY_REQUIRED = 23
+    INDETERMINATE = 24
+
+
+@dataclass(frozen=True, slots=True)
```
Trecek, line 72: [warning] defense: [L65] defines the cross-boundary InstallRequest without construction validation. Runtime strings such as mode='maintenance-update' survive construction but fail every identity check against InstallMode, causing install() to use the direct ambient context while skipping both maintenance requirements and direct backend/workspace checks.

src/autoskillit/cli/_install_snapshot/_snapshot.py
```diff
@@ -0,0 +1,418 @@
...
+
+def _installed_plugin_root(version: str | None = None) -> Path:
+    if version is None:
+        from autoskillit.cli._plugin_artifact import current_installed_plugin_root
+
+        return current_installed_plugin_root()
+    return _plugin_cache_dir() / version
+
+
+def _installed_plugins_json_path() -> Path:
```
Trecek, line 43: [warning] cohesion: [L32] adds another owner for the `installed_plugins.json` path. The same diff adds a home-aware path helper in `core/_plugin_ids.py` [L30-L33] and spells the path inline in `update/_transaction.py` [L272-L274]. Consolidate this load-bearing registry-path policy behind one shared helper/repository.

src/autoskillit/cli/_install_snapshot/_snapshot.py
```diff
@@ -0,0 +1,418 @@
...
+        self._remove(self._stage_dir)
+        self._entries.clear()
+        self._workspace_temp_shape = None
+        self._staged = False
+
+    def commit(self) -> None:
+        """Discard staged state only after every required postcondition passes."""
+        if not self._staged:
+            raise RuntimeError("install snapshot was not staged")
+        shutil.rmtree(self._stage_dir)
```
Trecek: [critical] defense: [L414] deletes the staged rollback tree before marking the snapshot committed. If rmtree partially removes a directory backup and then raises, install() treats the cleanup error as a transaction failure and rolls back from the truncated backup; _restore_entry can then accept that truncated directory as authoritative and report compensation completed after silently dropping original files.

src/autoskillit/cli/_marketplace.py
```diff
@@ -200,304 +218,503 @@ def _ensure_marketplace() -> Path:
...
-    — a dangling pointer roughly two hours later, once the sweeper ran. Every
-    failure path now restores this snapshot instead.
-    """
 
-    def __init__(self) -> None:
-        from autoskillit.cli._plugin_artifact import installed_artifact_manifest_path
+def _settings_path(scope: str, cwd: Path) -> Path:
+    if scope == "user":
+        return Path.home() / ".claude" / "settings.json"
+    return cwd / ".claude" / "settings.json"
```
Trecek: [warning] bugs: `_settings_path` maps both `project` and `local` installs to `.claude/settings.json`. The replaced `_claude_settings_path(scope)` distinguished Claude's local settings file, so `install --scope local` now snapshots, evicts, and verifies the project-scoped file while stale AutoSkillit hooks in `.claude/settings.local.json` survive, causing duplicate hook registration after plugin activation.

src/autoskillit/cli/_marketplace.py
```diff
@@ -200,304 +218,503 @@ def _ensure_marketplace() -> Path:
...
 
-def _marketplace_manifest_path() -> Path:
-    return Path.home() / ".autoskillit" / "marketplace" / ".claude-plugin" / "marketplace.json"
+def _validate_transaction_target(target: Path, expected_version: str) -> None:
+    """Validate an explicit-version target without consulting cached package state."""
+    expected_parent = _plugin_cache_dir()
+    if (
+        not target.is_absolute()
+        or target != expected_parent / expected_version
+        or target.is_symlink()
```
Trecek: [warning] tests: This changed guard rejects a symlinked install target, but the diff removes the prior install-level symlink regression test. The replacement symlink test exercises `_InstallSnapshot` directly and therefore bypasses `_validate_transaction_target`. Add tests proving symlink, non-directory, and escaping targets are rejected before mutation and that an external sentinel survives.

src/autoskillit/cli/_marketplace.py
```diff
@@ -200,304 +218,503 @@ def _ensure_marketplace() -> Path:
...
+        InstallFailureKind.POSTCONDITION,
+        f"Fetch cache remains after invalidation: {fetch_cache_path}",
+    )
 
-    cfg = load_config(Path.cwd())
 
-    from autoskillit.execution import get_backend
+def install(
+    *,
+    request: InstallRequest | None = None,
```
Trecek: [warning] abstraction_surface: The new typed request is layered over the old independent scope input. The only production caller, cli.app.install(), always constructs InstallRequest and then redundantly passes request.scope again; when both inputs are supplied, this implementation silently ignores scope in favor of request.scope. The request=None/scope-only variant is exercised only by direct unit calls to this private module. Retain one authority—preferably a required InstallRequest—and remove the conflicting fallback surface.

src/autoskillit/cli/_marketplace.py
```diff
@@ -200,304 +218,503 @@ def _ensure_marketplace() -> Path:
...
+                        )
+                    verified_identity = verification.identity.incarnation_id
+                    snapshot.commit()
+                except _InstallFailed as exc:
+                    return _compensated_result(
+                        snapshot,
+                        exc,
+                        owned_lease_fd=target_writer.fileno(),
+                    )
+                except Exception as exc:
```
Trecek: [critical] defense: [L689] catches only Exception after snapshot.stage(), so KeyboardInterrupt or SystemExit during the subsequent destructive install steps bypasses _compensated_result() entirely and leaves already-mutated shared install surfaces without rollback.

src/autoskillit/cli/update/_transaction.py
```diff
@@ -0,0 +1,479 @@
...
+    "UPDATE_TRANSACTION_PHASES",
+    "UpdateProcessStatus",
+    "UpdateTransactionOutcome",
+    "UpdateTransactionPhase",
+    "UpdateTransactionResult",
+    "process_status_for_update_outcome",
+    "run_update_transaction",
+]
+
+_PLUGIN_REF = "autoskillit@autoskillit-local"
```
Trecek: [warning] cohesion: [L50] introduces a local `_PLUGIN_REF` literal for update registration and verification, while the changed install-state authority uses the shared core `_AUTOSKILLIT_PLUGIN_KEY` at `_install_state.py` [L105] and [L215]. These parallel identity-policy sources can drift; the transaction should consume the shared authority.

src/autoskillit/cli/update/_transaction.py
```diff
@@ -0,0 +1,479 @@
...
+            logger.warning("update_artifact_verification_failed", exc_info=True)
+            return progress.finish(
+                UpdateTransactionOutcome.FAILED_POSTCONDITION,
+                expected_version=expected_version,
+                install_result=install_result,
+                findings=(f"Installed plugin verification failed: {exc}",),
+            )
+
+        verified_identity = (
+            verification.identity.semantic_key if verification.identity is not None else None
```
Trecek, line 455: [warning] cohesion: [L447] stores `verification.identity.semantic_key` in `UpdateTransactionResult.verified_identity`, while the install producer stores `verification.identity.incarnation_id` in the same-named field at `_marketplace.py` [L681], and `_map_install_result` forwards that representation at [L235]. The field therefore changes meaning by completion path; use explicit semantic-key/incarnation fields or normalize the representation.

src/autoskillit/workspace/_install_state.py
```diff
@@ -199,23 +201,47 @@ def verify_install_state() -> tuple[InstallStateFinding, ...]:
...
+    if not separator or plugin_ref != _AUTOSKILLIT_PLUGIN_KEY:
+        return False
     try:
-        identity = read_installed_plugin_artifact_identity(
-            record.managed_path,
-            expected_semantic_key=record.semantic_key,
-            manifest_path=record.manifest_path,
-        )
-    except PluginArtifactValidationError:
+        home = record.managed_path.parents[5]
```
Trecek: [warning] cohesion: [L218] reverse-engineers `home` from `managed_path.parents[5]`; `_plugin_artifact.py` independently repeats the identical positional rule at [L269]. This duplicates the managed-root layout outside `InstallStateSpec`, which otherwise owns that layout. Centralize the inverse mapping or pass trusted home explicitly.

src/autoskillit/workspace/_installed_artifact.py
```diff
@@ -0,0 +1,318 @@
...
+    home: Path
+    plugin_ref: str
+    expected_version: str
+    require_registered_plugin: bool
+    require_shared_lease: bool
+    supplied_lease: ArtifactLease | None = None
+
+    def __post_init__(self) -> None:
+        home = Path(self.home)
+        if not home.is_absolute():
```
Trecek, line 77: [warning] tests: The new `InstallStateSpec` path-safety boundary rejects relative homes, unsafe plugin/version path components, and non-boolean flags, but the added tests construct only valid specs. Add parameterized rejection coverage for every validation branch; otherwise a regression could permit managed-path escape or weaken trusted-input enforcement unnoticed.

src/autoskillit/workspace/_installed_artifact.py
```diff
@@ -0,0 +1,318 @@
...
+            / "plugins"
+            / "cache"
+            / DIRECT_INSTALL_CACHE_SUBDIR
+            / plugin_name
+            / self.expected_version
+        )
+
+    @property
+    def semantic_key(self) -> str:
+        """Return the exact plugin/version transaction identity."""
```
Trecek, line 138: [warning] abstraction_surface: The persisted artifact identity format now has two authorities: publication and current launch authority call cli._plugin_artifact.installed_plugin_semantic_key(), while verify_installed_plugin_artifact() consumes this separately implemented f"{plugin_ref}:{expected_version}" property; launch and retirement additionally parse the format with raw rpartition(":"). All production consumers use the same autoskillit identity across versions, so no supported variant needs a separate codec. Centralize the formatter/parser in a shared lower-layer primitive to prevent publication, verification, and retirement compatibility from drifting.

tests/workspace/test_installed_artifact.py
```diff
@@ -0,0 +1,358 @@
...
+            assert result.lease is None
+            assert not state.lease_path.exists()
+        elif kind is PluginArtifactStateKind.VALID_CURRENT:
+            assert result.identity == state.identity
+            assert result.findings == ()
+            assert result.lease is not None
+        else:
+            assert kind in INVALID_PLUGIN_ARTIFACT_STATE_KINDS
+            assert result.identity is None
+            assert result.findings
```
Trecek: [warning] tests: The complete invalid-artifact matrix only requires `result.findings` to be non-empty; subsequent assertions check generic severity and message fragments. All distinct corrupt states could collapse to one incorrect generic finding and still pass, and the contract/doctor matrix tests repeat the same weakness. Define expected check-code sets and identity/lease outcomes per `PluginArtifactStateKind` and assert them exactly.

Trecek (commented):

Trecek (commented):
AutoSkillit review: blocking findings detected. GitHub does not allow the PR author to request changes on their own PR; see the 13 inline comments for required fixes.

src/autoskillit/cli/_install_contract.py
Trecek: Agreed — fixed in d665ecd9df8e674c7874fd22e37d472e384fc341. `InstallRequest.__post_init__` now rejects raw-string modes and other untyped cross-boundary values, with parameterized coverage for every field.

src/autoskillit/cli/_install_snapshot/_snapshot.py
Trecek: Agreed — fixed in 6fb2884318f29e92fe927199fe9b72802595631c. Snapshot and update code now consume the home-aware core `_installed_plugins_path()` authority instead of spelling the registry path independently.

src/autoskillit/cli/_install_snapshot/_snapshot.py
Trecek: Agreed — fixed in 1d4c8a36d5fd9129f42c0ad4f8ac01f43324f60a. Snapshot commit state is finalized before best-effort staging cleanup, so a partial `rmtree` failure cannot make truncated rollback evidence authoritative. The regression test injects partial deletion and verifies rollback is disabled while installed bytes remain current.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 062ee29fc447b2b584d82e3484628476819485c8. Local scope now maps to `.claude/settings.local.json`; user, project, and local path distinctions are covered directly.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 6d70472b92795c82f99513120cbb778b63a97b6c. Install-level tests now cover symlink, non-directory, and escaping targets, assert rejection occurs before any child mutation, and verify an external sentinel remains untouched.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in 553736c7d02e119d2dab6b01337a44c99b4f0360. The private install transaction now requires one `InstallRequest`; the redundant `scope` fallback was removed and production/tests use the typed request as the sole authority.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in e0a1f17a029755d067e79315a02bbfa80fbba020. Post-stage failures now catch `BaseException`, compensate first, and re-raise control-flow exceptions. Regression coverage proves both `KeyboardInterrupt` and `SystemExit` restore mutated state before propagation.

src/autoskillit/cli/update/_transaction.py
Trecek: Agreed — fixed in 6ac42939e29522e26c407b769744a80726de1d52. The update transaction now consumes core `_AUTOSKILLIT_PLUGIN_KEY`; its parallel `_PLUGIN_REF` policy literal was removed.

src/autoskillit/cli/update/_transaction.py
Trecek: Agreed — fixed in a60d5d68f2da451cc1a225f801bb1545ac45702d. Install and update now both place the verified semantic key in `verified_identity`; the install success path asserts the normalized representation.

src/autoskillit/workspace/_install_state.py
Trecek: Agreed — fixed in c1eddb82c607594b8e55fedaf9ba620991b4c56b. `InstallStateSpec.from_managed_root()` now owns the inverse layout mapping and exact-root validation; workspace retirement and CLI launch authority both use it instead of duplicating `parents[5]`.

src/autoskillit/workspace/_installed_artifact.py
Trecek: Agreed — fixed in ee81e81640d55a45318a07678cf8ab634d20e4ee. Parameterized tests now cover relative home, empty/unsafe plugin and version components, and both non-boolean trust flags.

src/autoskillit/workspace/_installed_artifact.py
Trecek: Agreed — fixed in 49898b92597a619d5238e63a13386dd577ecf435. Formatting and parsing now live in one stdlib-only core semantic-key codec, consumed by publication, workspace verification, launch, and retirement without crossing import layers.

tests/workspace/test_installed_artifact.py
Trecek: Agreed — fixed in 780edeb658beb690065ebaf3ddef42274ff42a4a. The shared production-shaped fixture now declares exact check-code, identity, and lease expectations for every `PluginArtifactStateKind`, and the workspace, contract, and doctor matrices assert them exactly.

src/autoskillit/workspace/_installed_artifact.py
```diff
@@ -0,0 +1,349 @@
...
+        )
+        if spec.managed_root != root:
+            raise ValueError(f"installed plugin root is invalid: {root}")
+        return spec
+
+    @property
+    def managed_root(self) -> Path:
+        """Return the sole managed root authorized by the trusted inputs."""
+        plugin_name = self.plugin_ref.partition("@")[0]
+        return (
```
Trecek: [warning] cohesion: InstallStateSpec.managed_root re-derives the managed plugin-cache layout independently of the install-side owners added in the same diff: _install_snapshot/_snapshot.py _plugin_cache_dir()/_installed_plugin_root(version), which _marketplace._validate_transaction_target uses to authorize the very target this spec later verifies. The same duplication repeats one level down: manifest_path and lease_path re-implement _plugin_artifact.installed_artifact_manifest_path()/installed_artifact_lock_path(). Publication and verification therefore compute the same three paths by two independent rules and can silently disagree. Give the managed-root/manifest/lease layout a single owner and have both sides consume it.

src/autoskillit/workspace/_installed_artifact.py
```diff
@@ -0,0 +1,349 @@
...
+            f"contains {observed!r}"
+        )
+
+
+def _validate_supplied_lease(
+    spec: InstallStateSpec,
+) -> tuple[ArtifactLease | None, InstallStateFinding | None]:
+    lease = spec.supplied_lease
+    if lease is None:
+        if not spec.require_shared_lease:
```
Trecek: [warning] cohesion: require_shared_lease is a mode selector wearing a requirement's name. At L221 not spec.require_shared_lease is treated as 'the caller must have handed us an open exclusive publication lease'. A reader of _marketplace.py L663 (require_shared_lease=False) naturally reads 'a shared lease is not required' when the field actually mandates a supplied exclusive lease. Rename it to an explicit two-valued mode (e.g. lease_mode) so the call sites state the policy they mean.

Trecek (commented):

Five warning-level findings posted inline below. Five info-level observations (no-op callback, dead params, asymmetric result builder, self-alias re-export, dead nested install_result field) are summarised here:

- **L224** [info/cohesion] src/autoskillit/cli/update/_update_checks.py: [info] cohesion: _run_update_sequence keeps info and current in its signature only to discard them immediately via `del info, current`, and 
- **L710** [info/cohesion] src/autoskillit/cli/_marketplace.py: [info] cohesion: Every other terminal in install() builds its result through _typed_result(...), which is also the single place findings are
- **L35** [info/cohesion] src/autoskillit/cli/_marketplace.py: [info] cohesion: from autoskillit.cli._install_snapshot import _installed_plugins_json_path as _installed_plugins_json_path is a pure re-exp
- **L141** [info/abstraction-surface] src/autoskillit/cli/update/_transaction.py: [info] abstraction-surface: UpdateTransactionResult.install_result (line 141) carries the nested InstallResult through every progress.finish
- **L284** [info/abstraction-surface] src/autoskillit/cli/_install_snapshot/_snapshot.py: [info] abstraction-surface: _InstallSnapshot.track_retirement (line 284) is a no-op method - body is a single docstring, _record_id paramete

src/autoskillit/cli/_marketplace.py
```diff
@@ -200,304 +218,502 @@ def _ensure_marketplace() -> Path:
...
+                        )
+                    except PluginArtifactPublicationError as exc:
+                        raise _InstallFailed(
+                            InstallFailureKind.POSTCONDITION,
+                            f"Failed to publish installed plugin identity: {exc}",
+                        ) from exc
+
+                    if evict_direct_mcp_entry(_user_claude_json_path()):
+                        print("Removed stale direct MCP entry from ~/.claude.json")
+                    _hooks_mod._evict_stale_autoskillit_hooks(settings_path)
```
Trecek, line 636: [warning] cohesion: Hook eviction is still owned by _hooks_mod, but the scope->settings-file mapping it depends on has been forked into _marketplace._settings_path and is passed in from the outside here and at L513. The hooks module retains its own _claude_settings_path(scope) (the function this call replaced), so the same three-way scope policy now lives in two modules that must be changed together. Keep one owner - export the mapping from the hooks module and have the install transaction consume it.

src/autoskillit/cli/update/_transaction.py
```diff
@@ -0,0 +1,478 @@
...
+    DECLINED = "declined"
+    DEFERRED = "deferred"
+    RECOVERY_REQUIRED = "recovery-required"
+    INDETERMINATE = "indeterminate"
+
+
+class UpdateProcessStatus(IntEnum):
+    """Stable public statuses for the explicit update process boundary."""
+
+    SUCCESS = int(InstallProcessStatus.SUCCESS)
```
Trecek, line 103: [warning] abstraction-surface: UpdateProcessStatus (lines 96-106) is a parallel enum that re-aliases every value as int(InstallProcessStatus.X) - eight of eight members map 1:1 to the public InstallProcessStatus enum defined in _install_contract.py:52-62, and the only production consumer (_update.py:36) compares the result against UpdateProcessStatus.SUCCESS only. The _PROCESS_STATUS_BY_OUTCOME table (lines 109-124) reproduces the same mapping that process_status_for_result(result.install_result) would already express: UpdateTransactionOutcome values are derived inside _map_install_result from InstallOutcome (lines 218-231) and then re-projected onto the identical InstallProcessStatus codes. The two enums and the lookup table form a duplicated policy layer with no observable behavior change - run_update_command could call process_status_for_result(result.install_result) directly and UpdateProcessStatus would vanish.

src/autoskillit/cli/update/_transaction.py
```diff
@@ -0,0 +1,478 @@
...
+    outcome: UpdateTransactionOutcome
+    expected_version: str | None = None
+    install_result: InstallResult | None = None
+    verified_identity: str | None = None
+    findings: tuple[str, ...] = ()
+    phase_history: tuple[UpdateTransactionPhase, ...] = ()
+    irreversible_pivot_crossed: bool = False
+
+
+class _TransactionProgress:
```
Trecek, line 152: [warning] abstraction-surface: _TransactionProgress (lines 148-189) maintains a strict phase-prepend state machine with two surface fields - phase_history (line 144) and irreversible_pivot_crossed (line 145) - that have zero production readers: grep across src/ for .phase_history and .irreversible_pivot_crossed returns no matches. Every enforcement call (enter(...) lines 257-477) and every record in UpdateTransactionResult carries this state, but the only readers are tests/cli/test_update_transaction.py asserting result.phase_history == UPDATE_TRANSACTION_PHASES. The IRREVERSIBLE_PIVOT_PHASE constant (line 80) likewise has only test-side readers. This is generic state machinery whose invariants cannot be violated by any caller because no caller observes the result - either expose the state to a real consumer (e.g., logging, retry routing) or remove the surface fields, the phase enum, and the progress wrapper in favor of plain function calls.

src/autoskillit/workspace/_installed_artifact.py
Trecek: [warning] cohesion: InstallStateSpec.managed_root re-derives the managed plugin-cache layout independently of the install-side owners added in the same diff: _install_snapshot/_snapshot.py _plugin_cache_dir()/_installed_plugin_root(version), which _marketplace._validate_transaction_target uses to authorize the very target this spec later verifies. The same duplication repeats one level down: manifest_path and lease_path re-implement _plugin_artifact.installed_artifact_manifest_path()/installed_artifact_lock_path(). Publication and verification therefore compute the same three paths by two independent rules and can silently disagree. Give the managed-root/manifest/lease layout a single owner and have both sides consume it.

src/autoskillit/workspace/_installed_artifact.py
Trecek: [warning] cohesion: require_shared_lease is a mode selector wearing a requirement's name. At L221 not spec.require_shared_lease is treated as 'the caller must have handed us an open exclusive publication lease'. A reader of _marketplace.py L663 (require_shared_lease=False) naturally reads 'a shared lease is not required' when the field actually mandates a supplied exclusive lease. Rename it to an explicit two-valued mode (e.g. lease_mode) so the call sites state the policy they mean.

src/autoskillit/cli/_marketplace.py
Trecek, line 636: [warning] cohesion: Hook eviction is still owned by _hooks_mod, but the scope->settings-file mapping it depends on has been forked into _marketplace._settings_path and is passed in from the outside here and at L513. The hooks module retains its own _claude_settings_path(scope) (the function this call replaced), so the same three-way scope policy now lives in two modules that must be changed together. Keep one owner - export the mapping from the hooks module and have the install transaction consume it.

src/autoskillit/cli/update/_update_checks.py
```diff
@@ -258,38 +219,21 @@ def _run_update_sequence(
...
-    latest: str = (
-        _fetch_latest_version(target_branch, home) or current
-        if target_branch is not None
-        else current
-    )
-    install_result: subprocess.CompletedProcess[bytes] = subprocess.CompletedProcess(
-        args=["autoskillit", "install"], returncode=0
-    )
+    """Present the automatic caller around the shared update transaction."""
+    del info, current
```
Trecek: [info] cohesion: _run_update_sequence keeps info and current in its signature only to discard them immediately via `del info, current`, and the call site still computes and passes them. The shared transaction now derives both internally. Drop the two parameters and the arguments rather than carrying dead inputs through the presentation adapter.

src/autoskillit/cli/_marketplace.py
```diff
@@ -200,304 +218,502 @@ def _ensure_marketplace() -> Path:
...
+                    return compensated
+    except (OSError, RuntimeError, ValueError) as exc:
+        return _typed_result(
+            InstallOutcome.FAILED,
+            failure_kind=InstallFailureKind.PREFLIGHT,
+            findings=(f"install lock failure: {exc}",),
+        )
+
+    print(f"Plugin installed: {plugin_ref} (scope: {effective_scope})")
+    return InstallResult(
```
Trecek: [info] cohesion: Every other terminal in install() builds its result through _typed_result(...), which is also the single place findings are printed; the success path alone constructs InstallResult(...) directly and prints its message inline at L709. The asymmetry means any future change to result construction or finding presentation silently skips the completed-install case. Route the success return through _typed_result like the other outcomes.

src/autoskillit/cli/_marketplace.py
```diff
@@ -20,8 +20,22 @@
...
+    InstallOutcome,
+    InstallRequest,
+    InstallResult,
+)
+from autoskillit.cli._install_snapshot import (
+    _installed_plugin_root,
+    _InstallSnapshot,
+    _plugin_cache_dir,
+)
+from autoskillit.cli._install_snapshot import (
```
Trecek: [info] cohesion: from autoskillit.cli._install_snapshot import _installed_plugins_json_path as _installed_plugins_json_path is a pure re-export: the symbol has no remaining use in this module after the snapshot extraction, and the self-alias exists only to keep the old import path importable. This preserves a second address for a helper that now has a home, contrary to the project's no-backward-compatibility-shim rule. Delete the re-export and have consumers import from _install_snapshot.

src/autoskillit/cli/update/_transaction.py
Trecek, line 103: [warning] abstraction-surface: UpdateProcessStatus (lines 96-106) is a parallel enum that re-aliases every value as int(InstallProcessStatus.X) - eight of eight members map 1:1 to the public InstallProcessStatus enum defined in _install_contract.py:52-62, and the only production consumer (_update.py:36) compares the result against UpdateProcessStatus.SUCCESS only. The _PROCESS_STATUS_BY_OUTCOME table (lines 109-124) reproduces the same mapping that process_status_for_result(result.install_result) would already express: UpdateTransactionOutcome values are derived inside _map_install_result from InstallOutcome (lines 218-231) and then re-projected onto the identical InstallProcessStatus codes. The two enums and the lookup table form a duplicated policy layer with no observable behavior change - run_update_command could call process_status_for_result(result.install_result) directly and UpdateProcessStatus would vanish.

src/autoskillit/cli/update/_transaction.py
```diff
@@ -0,0 +1,478 @@
...
+    return _PROCESS_STATUS_BY_OUTCOME[outcome]
+
+
+@dataclass(frozen=True, slots=True)
+class UpdateTransactionResult:
+    """Immutable update result and the evidence needed to present it."""
+
+    outcome: UpdateTransactionOutcome
+    expected_version: str | None = None
+    install_result: InstallResult | None = None
```
Trecek, line 145: [info] abstraction-surface: UpdateTransactionResult.install_result (line 141) carries the nested InstallResult through every progress.finish(...) call site, but no production consumer reads it back: grepping src/ for result.install_result or .install_result returns zero hits outside of _transaction.py itself. Tests/cli/test_update_transaction.py inspects it in 12 places, but the field is published state that is never observed. The runtime already aggregates install_result.findings/verified_identity onto the top-level UpdateTransactionResult, so the nested object is pure data duplication of fields the outer record already exposes - either the field should be removed or its consumers should be promoted to read nested state instead of duplicating it on the outer record.

src/autoskillit/cli/update/_transaction.py
Trecek, line 152: [warning] abstraction-surface: _TransactionProgress (lines 148-189) maintains a strict phase-prepend state machine with two surface fields - phase_history (line 144) and irreversible_pivot_crossed (line 145) - that have zero production readers: grep across src/ for .phase_history and .irreversible_pivot_crossed returns no matches. Every enforcement call (enter(...) lines 257-477) and every record in UpdateTransactionResult carries this state, but the only readers are tests/cli/test_update_transaction.py asserting result.phase_history == UPDATE_TRANSACTION_PHASES. The IRREVERSIBLE_PIVOT_PHASE constant (line 80) likewise has only test-side readers. This is generic state machinery whose invariants cannot be violated by any caller because no caller observes the result - either expose the state to a real consumer (e.g., logging, retry routing) or remove the surface fields, the phase enum, and the progress wrapper in favor of plain function calls.

src/autoskillit/cli/_install_snapshot/_snapshot.py
```diff
@@ -0,0 +1,429 @@
...
+                "install_snapshot_stage_failed",
+                stage_dir=str(self._stage_dir),
+                exc_info=True,
+            )
+            shutil.rmtree(self._stage_dir, ignore_errors=True)
+            self._entries.clear()
+            self._workspace_temp_shape = None
+            raise
+
+    def track_retirement(self, _record_id: str) -> None:
```
Trecek: [info] abstraction-surface: _InstallSnapshot.track_retirement (line 284) is a no-op method - body is a single docstring, _record_id parameter is intentionally unused (leading underscore), and it is wired in only to satisfy the Callable[[str], None] callback type passed to _clear_plugin_cache(on_retirement_created=snapshot.track_retirement, ...) at _marketplace.py:565. The callback signature is required by InstalledPluginArtifactRetirementOwner.enqueue_retirement for notification of new retirement records, but the snapshot has no retirement state to track (the docstring admits as much: 'Retirement state is covered by the exact staged cache file'). The method, its underscore-parameter, and the wrapper indirection in _marketplace.py:565 add an interface without adding behavior - either pass None and adapt the called API to accept that, or drop the parameter entirely.

Trecek (commented):
AutoSkillit PR Review — Follow-up: posting the 5 info-level findings inline as requested. Verdict remains approved_with_comments.

src/autoskillit/workspace/_installed_artifact.py
Trecek: Agreed — fixed in 1a0bc5c9a. Canonical IL-0 helpers now own the managed root, manifest, and lease paths, and workspace/CLI consumers share them.

Trecek: Agreed — fixed in 1a0bc5c9a. Canonical IL-0 helpers now own the managed root, manifest, and lease paths, and workspace/CLI consumers share them.

src/autoskillit/workspace/_installed_artifact.py
Trecek: Agreed — fixed in c07c7051d. Install verification now uses explicit SHARED/EXCLUSIVE lease modes, with coverage for shared auto-acquire, supplied shared, supplied exclusive, and missing exclusive ownership.

Trecek: Agreed — fixed in c07c7051d. Install verification now uses explicit SHARED/EXCLUSIVE lease modes, with coverage for shared auto-acquire, supplied shared, supplied exclusive, and missing exclusive ownership.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in bc8e84e2a. The canonical hook settings resolver now handles user/project/local scopes from an explicit cwd, and marketplace consumes that owner.

Trecek: Agreed — fixed in bc8e84e2a. The canonical hook settings resolver now handles user/project/local scopes from an explicit cwd, and marketplace consumes that owner.

src/autoskillit/cli/update/_update_checks.py
Trecek: Agreed — fixed in acb54c3fa. The source change in 0768de701 removes the dead inputs and acb54c3fa updates the remaining callers.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in db18a0993. Marketplace install success now flows through `_typed_result` with its typed finding and identity.

src/autoskillit/cli/_marketplace.py
Trecek: Agreed — fixed in d41bbc336. The marketplace compatibility re-export is removed and tests import from the canonical `_install_snapshot` facade.

src/autoskillit/cli/_install_snapshot/_snapshot.py
Trecek: Agreed — fixed in b8650515e. The no-op retirement callback is removed and the optional callback is omitted.

src/autoskillit/cli/update/_transaction.py
Trecek: Investigated — this is intentional. The nested InstallResult preserves lossless failure_kind evidence that the outer update outcome intentionally collapses; tests cover preflight/child distinctions and later verification failures.

src/autoskillit/cli/update/_transaction.py
Trecek: This needs a design decision. Direct canonical install-status mapping cannot represent pre-install or upgrade failures where no install_result exists; changing the public boundary requires an explicit contract decision.

Trecek: This needs a design decision. Direct canonical install-status mapping cannot represent pre-install or upgrade failures where no install_result exists; changing the public boundary requires an explicit contract decision.

src/autoskillit/cli/update/_transaction.py
Trecek: This needs a design decision. Transaction history and pivot currently have no production reader, while progress.enter actively enforces phase and finalization ordering; surfacing or removing that audit state changes the contract.

Trecek: This needs a design decision. Transaction history and pivot currently have no production reader, while progress.enter actively enforces phase and finalization ordering; surfacing or removing that audit state changes the contract.

Trecek (commented):
FALLBACK: I was unable to post inline comments (GitHub secondary rate limit blocked the batch POST for 24+ minutes). Posting summary as review body instead. This is a DEGRADED review.

## AutoSkillit Review Findings

**Verdict:** changes_requested (2 critical blocking, 16 warning, 11 info, 1 decision; 30 total after suppressing 3 prior-resolved matches)

### src/autoskillit/cli/_install_contract.py
- **L1** [info/slop]: Module docstring at L1-L5 ('dependency leaf: performs no I/O and imports only from stdlib') is boilerplate that restates
- **L107** [info/defense]: InstallResult.__post_init__ allowlist duplicates _FAILED_STATUS_BY_KIND.keys() at [L107]. Iterate the registry to keep o
- **L116** [info/slop]: _FAILED_STATUS_BY_KIND and _OUTCOME_BY_STATUS are two parallel lookup tables; could be unified into a single enum-to-enu
- **L116** [warning/abstraction_surface]: Three parallel lookup tables encode install-outcome <-> process-status: _FAILED_STATUS_BY_KIND (L116-120), _OUTCOME_BY_S

### src/autoskillit/cli/_install_snapshot/_snapshot.py
- **L31** [info/arch]: Two version authorities inside one install transaction. `_installed_plugin_root(version=None)` falls back to a lazy `fro
- **L51** [warning/arch]: New cross-subpackage reach into `cli/update/` private internals. `_InstallSnapshot.__init__` lazily imports `autoskillit
- **L180** [info/slop]: Multi-line comment at L180-L182 in _restore_owned_lease_entry explains the obvious rationale for keeping an inode-locked
- **L352** [warning/reachability]: The `if lease_entry is None:` branch is unreachable. self._lease_path is unconditionally added to self._paths in __init_
- **L423** [warning/defense]: commit() catches Exception when removing staged directory at [L792]. Broad clause masks ValueError/TypeError/AttributeEr

### src/autoskillit/cli/_marketplace.py
- **L271** [info/reachability]: _run_claude_admin raises ValueError when argv is empty, but both production call sites pass hard-coded non-empty tuples.
- **L281** [info/slop]: _typed_result helper (L281-L299) prints every finding to stdout as a side-effect; tests use capsys to assert findings - 
- **L362** [warning/arch]: `_verify_cleanup` re-implements the nested settings.json hook-traversal that `_hooks._evict_stale_autoskillit_hooks` own
- **L414** [info/slop]: Old multi-paragraph prose about a dangling-pointer incident was deleted from _InstallSnapshot docstring at L243-L247 but
- **L509** [critical/bugs]: The new install transaction removed the `with _InstallLock():` wrapper that previously serialized concurrent installs. _
- **L721** [critical/defense]: _compensated_result assigns failure_kind=primary.kind. If primary.kind is InstallFailureKind.ROLLBACK, the resulting Ins
- **L721** [warning/defense]: install() catches BaseException at [L1676] but _compensated_result can itself raise (e.g., from _restore_owned_lease_ent

### src/autoskillit/cli/update/_transaction.py
- **L56** [info/slop]: _MAINTENANCE_EXTRAS module-level dict at L56-L59 holds AUTOSKILLIT_SKIP_* keys consumed only inside build_maintenance_en
- **L113** [warning/slop]: _PROCESS_STATUS_BY_OUTCOME table reproduces the InstallProcessStatus mapping already expressible through process_status_
- **L165** [warning/bugs]: _TransactionProgress.enter(RESULT_FINALIZATION) only guards against double-finalization; it does NOT enforce that all pr
- **L196** [info/slop]: _process_finding and _upgrade_failure are tiny shims used once each in the same file - inlining both would shrink the fi
- **L482** [warning/defense]: UpdateTransactionResult.verified_identity is populated from two sources depending on completion path: install_result.ver

### src/autoskillit/core/_claude_env.py
- **L168** [warning/defense]: _is_protected_maintenance_extra at [L3140-L3151] protects only AUTOSKILLIT_{BACKEND,SESSION,CAMPAIGN,ORDER}. An attacker

### src/autoskillit/workspace/_install_state.py
- **L232** [warning/arch]: Duplicated contract expressed as a bare string literal across a module boundary. `_record_matches_current_installed_arti

### src/autoskillit/workspace/_installed_artifact.py
- **L110** [warning/slop]: from_managed_root reconstructs home via positional parents[5] rule - duplicates the managed-root layout that _install_sn
- **L180** [info/arch]: IL-1 workspace diagnostics embed CLI operator instructions (e.g., 'Run `autoskillit install` to reinstall the plugin.') 
- **L312** [warning/reachability]: Defensive post-read re-validation `if identity.semantic_key != spec.semantic_key or identity.managed_path != spec.manage
- **L355** [warning/defense]: The no-create fast path at [L3761-L3766] uses is_symlink() OR exists() - a dangling symlink returns True from is_symlink
- **L355** [warning/defense]: _validate_supplied_lease catches (OSError, RuntimeError, ValueError) at [L3736] and reduces them to a single string find
- **L355** [warning/defense]: verify_installed_plugin_artifact catches BaseException to release the lease. While it re-raises at [L3853], pairing a Ba

### tests/cli/test_doctor.py
- **L24** [info/slop]: _filesystem_snapshot helper at L24-L40 is a once-used fixture-style helper for a single parametrized test; same logic re

Trecek (commented):
Resolved the findings from review 4824384286. GitHub placed that review in the top-level body after the inline batch hit a secondary rate limit, so this is the corresponding consolidated response.

Addressed 7 findings:

- Centralized the unreadable installed-artifact check name (79f771d7c2).
- Moved fetch-cache path policy to the shared install-snapshot layer (5b617f76ca).
- Shared hook cleanup traversal between eviction and verification (74edd47d14).
- Derived failed-kind validation and reverse failed-status entries from one mapping, addressing two findings (16e0a7bf13).
- Removed redundant installed-artifact identity revalidation (3f678346f8).
- Preserved KeyboardInterrupt/SystemExit when compensation itself fails, with regression tests (257e16cfac).
- Updated the existing line-based JSON-write allowlist after the helper move (80535c0bc7).

I also investigated all 23 remaining findings and did not change code where the claim did not hold:

<details>
<summary>Rejected findings and evidence</summary>

1. Workspace installation guidance is the intentional domain diagnostic surfaced unchanged by multiple callers.
2. The install transaction always derives its target from explicit expected_version; there is no second transaction authority.
3. RESULT_FINALIZATION is intentionally valid after every operational prefix, as enforced by transaction-progress tests.
4. install() still holds _InstallLock across staging, leases, child commands, publication, and verification; _lock_owned only prevents nested acquisition.
5. Update-saga outcomes such as FAILED_UPGRADE are not representable by the install-result mapping.
6. The rollback comment documents the non-obvious inode-descriptor versus lexical-path distinction.
7. The dependency-leaf docstring records a useful enforced architecture boundary.
8. Moving managed-root inversion into the IL-3 CLI snapshot module would violate the import-layer boundary; InstallStateSpec.from_managed_root is already the validated inverse.
9. The alleged relocated incident narrative is absent from the current marketplace file.
10. _typed_result is the CLI presentation boundary; printing findings is required and tested behavior.
11. The doctor and install-transaction filesystem helpers have different snapshot contracts and test domains.
12. Core owns the allowed-key taxonomy; _MAINTENANCE_EXTRAS owns this caller's choice to set both guards.
13. _upgrade_failure has multiple callers, and _process_finding serves both child processes.
14. The staged-lease fallback converts future invariant corruption into recovery diagnostics instead of a secondary unpacking failure.
15. Rejecting empty argv is a valid subprocess boundary invariant.
16. Only the two install-contract mapping directions overlap; the update mapping belongs to a separate domain.
17. No production _compensated_result caller supplies ROLLBACK as the primary failure kind.
18. The dangling-symlink branch was read in reverse: lexical evidence prevents the no-create return and causes lease acquisition.
19. Expected lease failures intentionally normalize into structured findings; no consumer branches on the original exception type.
20. Verifier-owned leases must close for control-flow exceptions too, and close_preserving retains the primary outcome.
21. Broad staged-directory cleanup suppression is deliberate post-commit finality with regression coverage.
22. Failure paths preserve available child identity evidence, while successful registered-plugin paths require fresh verification; those sources are intentionally lifecycle-specific.
23. Unsupported AUTOSKILLIT extras, including RUNTIME and DEBUG, are rejected by the independent two-key allowlist.

</details>

Validation is green at head 80535c0bc774fa7070888a526df740b93593e5d9: 32,853 passed, 570 skipped, 27 expected failures; pre-commit run --all-files also passed. The four older unresolved inline discussion threads already contain prior autoskillit resolution replies and were left untouched by the idempotency guard.

Issue #4417, opened 2026-07-29
Orchestrator asks the user instead of retrying when MCP is still connecting — #4398 deleted the silent-retry instruction, and the guard test passes on a bare 'do not' substring


Trecek:
## Summary

`autoskillit order` now asks the user what to do when the AutoSkillit MCP tools are not yet
connected, instead of retrying. It used to retry silently. The behaviour regressed in
**`3f10896f1` (#4398, 2026-07-28)**, which rewrote `_MCP_RETRY_INSTRUCTION`.

The MCP server can take up to ~30 seconds to finish connecting after session launch. This is
expected and unavoidable (see #1330 — the race is structural in Claude Code's interactive path).
Retrying is the entire mitigation, and it is gone.

## Observed 2026-07-29

```
● I'll start by opening the kitchen for the remediation recipe.
  ☐ MCP missing
  I need to call open_kitchen first, but the AutoSkillit MCP tools don't appear to be
  available in this session. How would you like to proceed?
    1. Retry — tools should be there
    2. Abort session
```

The user selected "Retry" and it worked immediately. The session had everything it needed to
recover on its own and asked instead.

## What changed

Before `3f10896f1`:

```
If calling open_kitchen produces ANY error — including "No such tool available", a
tool_use_error, or any is_error response — retry the call once immediately. Do not explain
the error or suggest troubleshooting steps to the user. Just retry silently.
If the retry also fails, output "AutoSkillit MCP server did not start — ending session." and end.
```

After: a five-case decision procedure headed **"RECIPE STARTUP RESULT ORDER — evaluate these cases
in order"**. Every branch classifies *the result of a call that was made*.

Two things were lost:

1. **The unconditional trigger.** `ANY error` became an enumeration of five specific result
   shapes. Classification replaced reflex.
2. **The prohibition on talking to the user.** *"Do not explain the error or suggest
   troubleshooting steps to the user"* was deleted outright. #1330 added that sentence
   specifically because a session had given troubleshooting advice instead of retrying — the exact
   behaviour that just recurred.

## Why the regression guard didn't fire — this is the important part

#1330 shipped three contract tests to prevent exactly this. They all still pass.
`test_mcp_retry_instruction_covers_silent_retry`
(`tests/cli/test_orchestrator_prompt_contract.py:182`) asserts:

```python
assert "silently" in instr or "do not explain" in instr or "do not" in instr
```

[...]

PR #4425, opened 2026-07-30
Rectify: MCP Startup Readiness and Replay-Safe Kitchen Opening

Trecek:
## Summary

Issue #4417 exposes two missing protocols: interactive startup lacked an owner for ensuring `open_kitchen` was addressable before dispatch, and kitchen opening lacked an explicit re-entry/reconciliation state machine after partial effects.

This change adds a timeout-bounded client-readiness barrier and semantic retry policy, a typed replay-safe kitchen transition with durable effect receipts, and generalized fleet effect provenance for timeout, cancellation, crash, cleanup, and resume paths.

## Verification

- `pre-commit run --all-files`
- `AUTOSKILLIT_TEST_FILTER=conservative AUTOSKILLIT_TEST_BASE_REF=develop task test-check`

## Implementation Plan

Plan file: `.autoskillit/temp/rectify/rectify_mcp_startup_readiness_and_retry_immunity_2026-07-29_134135.md`


Trecek (commented):

src/autoskillit/server/tools/tools_fleet_dispatch.py
```diff
@@ -538,27 +703,66 @@ async def dispatch_food_truck(
...
+                effect_provenance=snapshot,
+            ).to_envelope()
 
         if campaign_state_path_str and isinstance(result, DispatchResult):
-            _write_dispatch_to_campaign_state(
+            provenance.start(
+                DispatchEffectName.CAMPAIGN_STATE_WRITE,
+                identities={"campaign_state_path": campaign_state_path_str},
+            )
+            campaign_write_confirmed = _write_dispatch_to_campaign_state(
```
Trecek, line 788: [critical] bugs: The campaign record is written with the pre-write provenance snapshot, then CAMPAIGN_STATE_WRITE is confirmed only in memory. The durable record therefore permanently omits the confirmation (or records the effect as STARTED), so restart/recovery consumers can see reconcile_required or otherwise stale provenance while the returned envelope reports a confirmed write. Persist the post-confirmation snapshot, using a replay-safe receipt mechanism.

tests/cli/test_session_launch.py
```diff
@@ -94,11 +95,22 @@ def cook_session_context(
 
 
 def _capture_subprocess(monkeypatch: pytest.MonkeyPatch) -> dict:
-    """Replace subprocess.run with a capturing stub. Stubs shutil.which to /usr/bin/claude."""
+    """Replace subprocess.run with a capturing stub and preserve real binary lookup."""
     captured: dict = {}
-    monkeypatch.setattr(shutil, "which", lambda _: "/usr/bin/claude")
+    real_which = shutil.which
+    monkeypatch.setattr(shutil, "which", lambda name, **kwargs: real_which(name, **kwargs))
```
Trecek: [warning] tests: This helper now delegates executable resolution to the host's real PATH. The launch unit tests therefore pass or fail depending on whether matching Claude/Codex binaries happen to be installed. Create executable fixtures under tmp_path and resolve those explicitly so these tests remain hermetic.

tests/execution/backends/test_cli_conformance_probes.py
```diff
@@ -1418,3 +1437,349 @@ def echo(x: str) -> str:
...
+            + "\n"
+        )
+    return result
+
+
+@_skip_unless_claude_startup_smoke
+@pytest.mark.timeout(120)
+@pytest.mark.parametrize(
+    ("delay_ms", "connect_timeout_ms"),
+    [(0, 5_000), (1_000, 5_000), (2_500, 1_000)],
```
Trecek, line 1684: [warning] tests: The fresh-over-budget-retry case asserts only eventual readiness and output bounds; it never checks the trace for a failed first attempt followed by the intended retry. A single delayed/blocking attempt can satisfy every assertion while retry behavior is broken. Assert the attempt count and ordering recorded in the trace.

src/autoskillit/execution/process/_process_kill.py
```diff
@@ -36,31 +36,52 @@ def kill_process_tree(pid: int, timeout: float = 2.0) -> None:
 
     # Include the parent in the kill list
     all_procs = children + [parent]
+    identities: list[tuple[int, float]] = []
+    for proc in all_procs:
+        try:
+            identities.append((proc.pid, proc.create_time()))
```
Trecek, line 43: [warning] defense: Process.create_time() can raise psutil.AccessDenied, but this identity-capture loop catches only NoSuchProcess. One inaccessible descendant therefore aborts the entire cleanup before SIGTERM, instead of recording access-denied evidence and continuing with the remaining processes.

src/autoskillit/fleet/_api.py
```diff
@@ -326,6 +357,52 @@ def _reject(error_code: FleetErrorCode, message: str, **kwargs: Any) -> Dispatch
...
+                            name=effective_name,
+                            status=failure_status,
+                            dispatch_id=identities.get("dispatch_id", ""),
+                            dispatched_session_id=identities.get("dispatched_session_id", ""),
+                            reason=str(FleetErrorCode.FLEET_L3_STARTUP_OR_CRASH),
+                            diagnostic_message=diagnostic_message,
+                            effect_provenance=snapshot.to_dict(),
+                        ),
+                    )
+                except Exception:
```
Trecek, line 389: [warning] defense: If crash-state persistence fails, the exception is only logged and the returned DispatchResult still advertises the intended non-null per_dispatch_state_path. The campaign writer can then treat a missing or stale file as authoritative. Clear the path or carry explicit persistence-failure provenance in the result.

src/autoskillit/core/types/_type_backend.py
```diff
@@ -41,6 +45,119 @@
...
+        """Resolve and seal the exact executable selected by the effective env."""
+        explicit = environment.get(explicit_path_env, "") if explicit_path_env else ""
+        if explicit:
+            candidate = Path(explicit).expanduser()
+            if not candidate.is_absolute():
+                raise ValueError(f"{explicit_path_env} must be an absolute path")
+        else:
+            try:
+                resolved = shutil.which(binary_name, path=environment.get("PATH"))
+            except TypeError:
```
Trecek: [warning] defense: This broad TypeError fallback exists for legacy test doubles, but it drops the effective environment's PATH and retries against the host PATH. It can therefore bind a different executable than the sealed launch environment specifies. Update the test doubles and remove the compatibility fallback (or detect capability without swallowing lookup failures).

src/autoskillit/server/_recipe_delivery.py
```diff
@@ -117,6 +121,7 @@ class FinalizedRecipeResponse:
     initialization_activating: bool = False
     initialization_id: str | None = None
     initialization_requirements: tuple[RecipeInitializationRequirement, ...] = ()
+    kitchen_transition_token: KitchenTransitionToken | None = None
```
Trecek, line 127: [warning] abstraction_surface: kitchen_transition_token is a write-only ownership-proof carrier: checkout-wide references only define, construct, and store it; complete_finalized_recipe_response and all consumers never read it. The token therefore cannot enforce operation/effect ownership despite adding concurrency-sensitive surface. Consume it at finalization or remove the inert abstraction.

src/autoskillit/cli/_prompts.py
```diff
@@ -26,37 +28,278 @@
...
+            _POST_RECEIPT_SCOPE: rendered[post_start:],
+        }
+        return tuple(
+            clause.clause_id
+            for clause in self.clauses
+            if rendered.count(f"[{clause.clause_id}]") != 1
+            or sections[clause.scope].count(clause.render()) != 1
+        )
+
+    def reduce_pre_dispatch_failure(self, attempt: int) -> McpStartupRecoveryEvent:
```
Trecek: [warning] reachability: This startup-recovery reducer is unreachable from production control flow. Production only renders the spec into prompt text; reduce_pre_dispatch_failure, reduce_received_result, validate_rendered, and the event model are called only by tests. Those tests exercise a parallel model rather than the orchestrator behavior it appears to govern. Integrate the reducer with the real path or remove the test-only state machine.

src/autoskillit/fleet/state_types.py
```diff
@@ -53,6 +55,273 @@ class ErrorCodeCategory(StrEnum):
...
+        ):
+            return DispatchAggregatePhase.COMMITTED
+        if any(effect.phase == DispatchEffectPhase.CONFIRMED for effect in relevant):
+            return DispatchAggregatePhase.STARTED
+        return DispatchAggregatePhase.NOT_STARTED
+
+    @property
+    def retry_disposition(self) -> DispatchRetryDisposition:
+        aggregate = self.aggregate_phase
+        if self.fresh_retry_authorized or aggregate == DispatchAggregatePhase.NOT_STARTED:
```
Trecek: [warning] abstraction_surface: The fresh-retry-after-compensation branch has no reachable producer: COMPENSATION has no production consumer and authorize_fresh_retry_after_compensation has no callers, so fresh_retry_authorized can never become true through the shipped request path. Remove this unsupported policy branch or implement and validate the compensation transition end to end.

Trecek (commented):
AutoSkillit review: blocking issue detected. See the inline comments; the automated verdict is changes_requested. GitHub does not permit an author to request changes on their own PR, so this review was submitted as COMMENT.

src/autoskillit/server/tools/tools_fleet_dispatch.py
Trecek: Agreed — fixed in 0862cea93a0a27fde6c9bd019dcfe4a5ab318ac4 with the boundary-preserving follow-up 2435963158524e1fbeba89bf678afceb5762cad3. The campaign writer now confirms CAMPAIGN_STATE_WRITE and atomically persists that post-confirmation provenance onto the existing campaign record; persistence failure marks the effect ambiguous.

tests/cli/test_session_launch.py
Trecek: Agreed — fixed in 4dde4330fbac4056dbaa1599db4b5baae520850e. Session-launch tests now provision executable fixtures under tmp_path and resolve them through the patched PATH instead of depending on host-installed claude or codex binaries.

tests/execution/backends/test_cli_conformance_probes.py
Trecek: Agreed — fixed in 20688f77d1100911812b1929e97396d547cdf310. The over-budget conformance case now records per-process server identities and asserts that the first delayed attempt never reaches exec, a distinct retry does, and the successful open_kitchen result occurs afterward.

src/autoskillit/execution/process/_process_kill.py
Trecek: Agreed — fixed in a2e35a81312200b4ccc90cc4ad73ca82477ae370. AccessDenied while reading one process create_time is now recorded as degraded identity evidence without aborting cleanup of the remaining process tree; a regression test verifies signaling continues.

src/autoskillit/fleet/_api.py
Trecek: Agreed — fixed in 4ee840d12218845174dbebe106893ebe4a9f94a6. If crash-record persistence fails, execute_dispatch now clears per_dispatch_state_path before returning so downstream campaign handling cannot trust an unpersisted state file; the failure envelope remains intact.

src/autoskillit/core/types/_type_backend.py
Trecek: Agreed — fixed in 49254a74a3364ab1a4ccdfbc92ef856a5a41719e. Executable binding now always calls shutil.which with the effective environment PATH, and test doubles accept that keyword explicitly; the host-PATH TypeError fallback was removed.

src/autoskillit/server/_recipe_delivery.py
Trecek: Agreed — fixed in 71ff78e81e2c61a97c44d912e4d39965f0c496c8. Finalized response completion now consumes the KitchenTransitionToken by validating the current operation_id and recipe-serving effect_id under the transition lock before committing lifecycle or receipt state.

src/autoskillit/cli/_prompts.py
Trecek: Agreed — fixed in 7a7b0167c21db7330464d9b84fe6f577b833c10b. The production-unreachable startup event reducer, validator, and synthetic harness were removed; direct tests retain the rendered prompt clauses, phase boundary, attempt cap, and canonical instruction contract.

src/autoskillit/fleet/state_types.py
Trecek: Agreed — fixed in a81b2e7c16bb2f4f95febf39a5bfece530c420cc. The unproducible compensation effect and fresh-retry override were removed from the provenance model, serializer, tracker, and retry-disposition branch so decisions remain fail-closed.

src/autoskillit/core/types/_type_backend.py
```diff
@@ -41,6 +45,114 @@
...
+            ]
+        ).encode("utf-8")
+        return cls(
+            path=canonical,
+            device=stat_result.st_dev,
+            inode=stat_result.st_ino,
+            size=stat_result.st_size,
+            mtime_ns=stat_result.st_mtime_ns,
+            file_sha256=_sha256_file(canonical),
+            environment_fingerprint=hashlib.sha256(fingerprint_payload).hexdigest(),
```
Trecek: [warning] reachability: `environment_fingerprint` is computed but never consumed, so it cannot protect executable/environment ownership. Launch code checks and uses the sealed `launch_environment` mapping directly; checkout-wide consumers only assert that this hash exists. Remove the inert parallel identity or verify/emit it at a real boundary.

src/autoskillit/execution/backends/claude.py
```diff
@@ -891,8 +908,54 @@ def validate_interactive_invocation(self, spec: CmdSpec) -> list[str]:
...
+                timeout=5,
+                env=dict(environment),
+                cwd=str(executable.cwd),
+            )
+        except subprocess.TimeoutExpired:
+            return ["Claude Code capability probe timed out"]
+        except OSError as exc:
+            return [f"Claude Code capability probe failed: {exc}"]
+        if result.returncode != 0:
+            return [f"Claude Code capability probe failed with exit code {result.returncode}"]
```
Trecek: [warning] defense: The capability probe captures stdout and stderr but discards both whenever `--version` exits nonzero, leaving only an exit code at the pre-launch boundary. Include bounded stderr/stdout so installation, loader, and policy failures retain their actionable diagnostic context.

src/autoskillit/fleet/_api.py
```diff
@@ -810,14 +1000,38 @@ def _on_spawn(pid: int, ticks: int) -> None:
         from autoskillit.core import read_boot_id
 
         _dispatched_pid.append(pid)
+        provenance.confirm(
```
Trecek: [critical] defense: `PROCESS_SPAWN` is confirmed before a dispatched session ID exists. A timeout after spawn but before session-ID capture therefore reports `resume_by_identity` with an empty `dispatched_session_id`, violating the new contract that resume requires the dispatch ID, session ID, and downstream identities. Keep the retry state reconciliation-required until the authoritative session identity is recorded.

src/autoskillit/fleet/state_types.py
```diff
@@ -53,6 +55,264 @@ class ErrorCodeCategory(StrEnum):
...
+    def start(
+        self,
+        name: DispatchEffectName,
+        *,
+        retry_relevant: bool = True,
+        identities: dict[str, object] | None = None,
+    ) -> None:
+        with self._lock:
+            existing = self._effects.get(name)
+            if existing is not None and existing.phase != DispatchEffectPhase.NOT_STARTED:
```
Trecek: [warning] reachability: The `existing.phase == NOT_STARTED` replacement path is unreachable through `DispatchProvenanceTracker`: its private map starts empty, `start()` creates STARTED, `confirm()` creates CONFIRMED, and `mark_ambiguous()` creates STARTED; no loader or setter inserts NOT_STARTED. Treat any existing record as already journaled, or add a real producer if this phase is intended.

src/autoskillit/fleet/state_types.py
```diff
@@ -419,20 +695,29 @@ class DispatchCompleted:
...
             "l3_parse_source": self.l3_parse_source,
             "lifespan_started": self.lifespan_started,
             "stderr": self.stderr,
             "elapsed_seconds": self.elapsed_seconds,
         }
+        if not self.success:
+            try:
+                error_code = FleetErrorCode(reason_text).value
+            except ValueError:
+                error_code = FleetErrorCode.FLEET_L3_STARTUP_OR_CRASH.value
```
Trecek: [warning] bugs: Every failed `DispatchCompleted` whose free-form reason is not exactly a `FleetErrorCode` is relabeled `FLEET_L3_STARTUP_OR_CRASH`. Clean L3 results may return ordinary domain failure reasons, so this changes their error category and can trigger incorrect retry routing. Preserve or explicitly classify the actual failure reason.

src/autoskillit/server/tools/tools_fleet_dispatch.py
```diff
@@ -367,18 +562,32 @@ async def dispatch_food_truck(
...
+                        DispatchEffectName.CAMPAIGN_STATE_WRITE,
+                        identities={"campaign_state_path": campaign_state_path_str},
+                    )
                     upsert_dispatch_record_by_name(
                         Path(campaign_state_path_str),
                         DispatchRecord(
                             name=effective_name,
                             status=DispatchStatus.SKIPPED,
                             reason="skip_when condition evaluated to true",
+                            effect_provenance=provenance.snapshot().to_dict(),
```
Trecek, line 575: [warning] bugs: The `skip_when` path persists the snapshot while `CAMPAIGN_STATE_WRITE` is still STARTED, then confirms only the in-memory tracker. After restart, the campaign record reports UNKNOWN/reconcile-required while the returned envelope reports a confirmed write. Persist the post-confirmation receipt as the normal dispatch path does.

src/autoskillit/server/tools/tools_fleet_dispatch.py
```diff
@@ -538,27 +748,66 @@ async def dispatch_food_truck(
...
             )
+            snapshot = provenance.snapshot()
+            identities = _dispatch_effect_identities(snapshot)
+            if snapshot.aggregate_phase == DispatchAggregatePhase.NOT_STARTED:
+                return DispatchRejected(
+                    error_code=FleetErrorCode.FLEET_L3_TIMEOUT,
+                    message=timeout_message,
+                    effect_provenance=snapshot,
+                ).to_envelope()
+            return DispatchCompleted(
```
Trecek, line 773: [warning] defense: The started-effect timeout path computes `timeout_message` but does not pass it as `diagnostic_message`. `DispatchCompleted.to_envelope()` therefore falls back to the enum text and drops the configured timeout duration and context that the previous `fleet_error` response exposed.

src/autoskillit/server/tools/tools_kitchen.py
```diff
@@ -111,6 +139,283 @@
 _PR_CREATE_RECIPES: frozenset[str] = frozenset(
     {"merge-prs", "implementation", "implementation-groups", "remediation"}
 )
+_OPEN_KITCHEN_REQUEST_CTX: ContextVar[ToolContext] = ContextVar("open_kitchen_request_context")
+
+
+def _ensure_kitchen_transition(tool_ctx: ToolContext) -> None:
+    """Create infrastructure identity once, before request arguments are bound."""
+    state = getattr(tool_ctx, "kitchen_open_state", None)
+    if not isinstance(state, KitchenOpenState):
```
Trecek: [warning] reachability: This invalid-context repair branch is unreachable from the production `open_kitchen` path. `_get_ctx()` supplies a typed `ToolContext`, its factories always construct `KitchenOpenState` and `RLock`, every assignment preserves those types, and no persistence lifecycle can load an older context object. Remove this duplicate state-minting path instead of repairing impossible process-local state.

tests/arch/test_kitchen_id_assignment.py
```diff
@@ -43,13 +44,21 @@ def test_kitchen_id_only_assigned_via_resolve_kitchen_id():
                         f"ctx.kitchen_id assignment uses "
                         f"{rhs}, not resolve_kitchen_id()"
                     )
-                else:
-                    pytest.fail(
-                        f"{file_path}:{node.lineno}: "
-                        f"ctx.kitchen_id assignment is not via resolve_kitchen_id()"
-                    )
+                if isinstance(node.value, ast.Attribute) and node.value.attr == "kitchen_id":
```
Trecek: [warning] arch: The ownership guard accepts any right-hand attribute named `kitchen_id`, so `ctx.kitchen_id = unrelated.kitchen_id` would pass even though only transition state may supply it. Match the qualified transition-state expression rather than only the terminal attribute name.

tests/arch/test_subpackage_isolation.py
```diff
@@ -91,6 +91,9 @@ def _get_call_func_name(node: ast.Call) -> str | None:
         "_sessions",  # cli/_sessions.py: sessions_app = App(name="sessions", ...)
         "_validate",  # cli/_validate.py: validate_app = App(name="validate", ...)
         "_type_backend",  # core/types/_type_backend.py: CLAUDE_CODE_CAPABILITIES constant
+        "_prompts",  # cli/_prompts.py: immutable startup recovery spec and rendering
+        "tools_kitchen",  # request-scoped open-kitchen transition ContextVar
```
Trecek: [warning] arch: Adding `tools_kitchen` to `SINGLETON_ALLOWED_MODULES` disables singleton-locality checks for the whole module, not only the request-scoped ContextVar. Any unrelated future module-level constructor in this 2,000+ line module will now bypass the guard; narrow the exemption to the intended assignment.

tests/arch/test_subpackage_isolation.py
```diff
@@ -1046,12 +1049,15 @@ def test_data_directories_are_not_python_packages() -> None:
         "filter cascade",
     ),
     "_api.py": (
-        1200,
+        1550,
```
Trecek, line 1052: [warning] arch: This fleet-specific line-limit increase is keyed by basename `_api.py`, and the guard falls back to `py_file.name`. It therefore raises the ceiling for every package's `_api.py`; key the exemption as `fleet/_api.py` so unrelated facades remain constrained.

tests/cli/test_session_launch.py
```diff
@@ -218,7 +244,7 @@ def test_run_interactive_session_holds_binding_through_reap_and_passes_descripto
     authority.acquire_launch_binding = lambda **_kwargs: binding  # type: ignore[method-assign]
     events: list[str] = []
 
-    monkeypatch.setattr(shutil, "which", lambda _: "/usr/bin/claude")
+    monkeypatch.setattr(shutil, "which", lambda _, **_kwargs: "/usr/bin/claude")
```
Trecek: [warning] tests: These per-test overrides replace the new hermetic executable fixture with nonexistent `/usr/bin/claude` or `/usr/bin/codex` paths. On the current PR HEAD, nine `test_session_launch.py` cases fail in CI when strict executable binding resolves those paths. Reuse the fixture binaries instead of overriding `shutil.which` with host-dependent paths.

Trecek (commented):

Trecek (commented):
AutoSkillit review: blocking issue detected. See the inline comments; the automated verdict is changes_requested. GitHub does not permit an author to request changes on their own PR, so this summary was submitted as COMMENT.

src/autoskillit/core/types/_type_backend.py
Trecek: Agreed — fixed in `15b304d2e`. Removed the inert `environment_fingerprint` capability field and its test-only assertions; the sealed launch-environment mapping remains the single executable/environment ownership source.

src/autoskillit/execution/backends/claude.py
Trecek: Agreed — fixed in `31b9a56bc`. Nonzero Claude capability probes now retain bounded stderr/stdout diagnostics in the readiness failure instead of reporting only the exit code.

src/autoskillit/fleet/_api.py
Trecek: Agreed — fixed in `55189a60e` with cancellation-state persistence completed in `d2b81e67f`. `PROCESS_SPAWN` starts when the PID is observed, confirms only after `_on_session_id` records an authoritative session ID, and interrupted state now persists the resulting retry provenance.

src/autoskillit/fleet/state_types.py
Trecek: Agreed — fixed in `c6e2a09db`. `DispatchProvenanceTracker.start()` now treats every existing effect as already journaled and no longer carries the unreachable `NOT_STARTED` replacement branch.

src/autoskillit/fleet/state_types.py
Trecek: Agreed — fixed in `fad033109`. Failed `DispatchCompleted` envelopes preserve ordinary domain failure reasons instead of relabeling every non-enum reason as `FLEET_L3_STARTUP_OR_CRASH`.

src/autoskillit/server/tools/tools_fleet_dispatch.py
Trecek: Agreed — fixed in `f94747e82`. The `skip_when` path now persists campaign state again after confirming `CAMPAIGN_STATE_WRITE`, so restart state and the returned envelope carry the same confirmed receipt.

src/autoskillit/server/tools/tools_fleet_dispatch.py
Trecek: Agreed — fixed in `fde8b3c83`. The started-effect timeout path now passes its detailed `timeout_message` as `diagnostic_message`, preserving the configured duration and dispatch context in the envelope.

src/autoskillit/server/tools/tools_kitchen.py
Trecek: Agreed — fixed in `82971e2ab`, with test contexts aligned in `aeb323a52`. Removed the invalid-context repair branch and its replacement lock/state minting; `_ensure_kitchen_transition` now consumes the typed `ToolContext` state guaranteed by the factory.

tests/arch/test_kitchen_id_assignment.py
Trecek: Agreed — fixed in `2dfe7b067`. The AST guard now accepts only the qualified transition expression `state.kitchen_id`; an unrelated terminal `.kitchen_id` attribute no longer passes.

tests/arch/test_subpackage_isolation.py
Trecek: Agreed — fixed in `20adc836b`. Removed the module-wide `tools_kitchen` singleton exemption and allowlisted only `_OPEN_KITCHEN_REQUEST_CTX` through the assignment-scoped registry.

tests/arch/test_subpackage_isolation.py
Trecek: Agreed — fixed in `5aa6fbcb5`. The line-limit exemption is now keyed as `fleet/_api.py`, so unrelated modules sharing the `_api.py` basename retain the default ceiling.

tests/cli/test_session_launch.py
Trecek: Agreed — fixed in `62112ceea`. Removed the per-test `/usr/bin/claude` and `/usr/bin/codex` overrides; the affected cases now reuse the executable files created by the autouse hermetic fixture.

src/autoskillit/cli/_prompts.py
```diff
@@ -26,37 +27,219 @@
...
-    "5. Only a structured, nonrecoverable application error is terminal. Print its "
-    "user_visible_message verbatim (or the raw application result if absent), do "
-    "not call AskUserQuestion, and do not diagnose it as MCP startup failure."
+_PRE_DISPATCH_SCOPE = "pre_dispatch"
+_POST_RECEIPT_SCOPE = "post_receipt"
+
+
+@dataclass(frozen=True, slots=True)
+class _McpStartupRecoveryClause:
+    clause_id: str
```
Trecek, line 36: [info] abstraction_surface: _McpStartupRecoveryClause declares five fields (actor, modality, polarity, action, object) that are only validated by __post_init__ and never read by render(). They impose typed constraints but the actual rendering only uses clause_id, scope, and text. Either wire them into the rendered text (e.g., compose a structured sentence from actor+modality+action+object) or drop them as unneeded structure.

src/autoskillit/cli/_prompts.py
```diff
@@ -26,37 +27,219 @@
...
+    scope: str
+    actor: str
```
Trecek: [warning] slop: _McpStartupRecoveryClause fields actor, modality, polarity, action, object are validated in _McpStartupRecoverySpec.__post_init__ (L67-L76) but never read anywhere else in the diff — render() on both clause and spec only uses clause_id, scope, and text. actor and modality each enforce a single constant value ("orchestrator" / "must"); polarity, action, object are stored per-clause but never consumed. Either drop the unused fields/validation or have render() incorporate them. Looks like AI-added validation scaffolding without a real consumer.

src/autoskillit/execution/backends/codex.py
```diff
@@ -1715,7 +1719,13 @@ def version(self) -> str:
...
         return []
 
-    def ensure_pre_launch(self, *, session_dir: Path | None = None) -> list[str]:
+    def ensure_pre_launch(
+        self,
+        *,
+        session_dir: Path | None = None,
+        executable: ExecutableLaunchBinding | None = None,
+    ) -> list[str]:
+        del executable
```
Trecek, line 1740: [warning] cohesion: Asymmetric capability-probe behaviour: Codex's `ensure_pre_launch` accepts the `executable: ExecutableLaunchBinding | None` parameter that the protocol was extended to carry (line 1722), but explicitly discards it via `del executable` (line 1728). Claude's `ensure_pre_launch` (claude.py line 911) uses the same parameter to enforce binding validity, sealed MCP env values, and minimum version. The shared interface contract forces asymmetric behaviour — readers cannot tell from the signature that one backend validates and the other silently ignores the binding. Either drop the parameter from Codex or document the asymmetry at the protocol boundary.

Trecek (commented):

This is a self-PR review. The following inline comments are advisory findings. 3 critical-severity issues (test hermeticity regressions) and 19 warnings must be addressed before merge. 12 info-severity findings are suggestions.

src/autoskillit/fleet/_api.py
```diff
@@ -326,6 +357,53 @@ def _reject(error_code: FleetErrorCode, message: str, **kwargs: Any) -> Dispatch
             dispatch_name=effective_name,
             exc_info=True,
         )
+        snapshot = provenance.snapshot()
```
Trecek, line 360: [info] arch: execute_dispatch now writes per-dispatch state on crash (L360-L406): when an exception escapes the inner dispatch loop and `aggregate_phase != NOT_STARTED`, the function persists a DispatchRecord and returns `DispatchCompleted(success=False, dispatch_status=FAILURE, ...)` instead of `DispatchRejected`. This is consistent with the effect-provenance model (a crash after partial dispatch is a completed-but-failed outcome, not a clean rejection), but is a behavioral change for callers that previously expected only DispatchRejected on exception.

src/autoskillit/fleet/_api.py
```diff
@@ -623,33 +764,69 @@ async def _run_dispatch(
...
-                    error_code=error_code,
+                    status=dispatch_status,
+                    reason=str(error_code),
                     diagnostic_message=message,
                     dispatch_id=dispatch_id,
+                    dispatched_session_id=dispatched_session_id,
+                    effect_provenance=provenance.snapshot().to_dict(),
                 ),
             )
         except Exception:
```
Trecek, line 827: [warning] defense: _complete_failure_with_state catches Exception around append_dispatch_record and only emits logger.warning(exc_info=True), but still returns DispatchResult(completed, per_dispatch_state_path=state_path) with the full state_path as if persistence succeeded. The caller (dispatch_food_truck) and downstream reconciliation cannot distinguish a successfully-persisted failure record from one that failed silently. On a critical-state path (post-allocation failure record), this leaves the in-memory DispatchCompleted.effect_provenance saying 'committed' while the durable campaign state has no record — defeating the very provenance-driven retry-disposition contract the PR introduces. Should either raise, return a sentinel, or change the return value to indicate persistence failure.

src/autoskillit/fleet/_api.py
```diff
@@ -849,6 +1056,16 @@ def _on_session_id(session_id: str) -> None:
         mark_dispatch_session_identity(
             state_path, effective_name, dispatched_session_id=session_id
         )
+        provenance.confirm(
```
Trecek, line 1068: [warning] defense: _on_session_id callback unconditionally calls provenance.confirm(DispatchEffectName.PROCESS_SPAWN, ...) even when the corresponding provenance.start(PROCESS_SPAWN) at L1081-1085 has not yet executed. PROCESS_SPAWN is started inside the 'try:' block AFTER the callback closure is defined and AFTER _do_dispatch returns control. Today DispatchProvenanceTracker.confirm auto-creates an effect in CONFIRMED state when existing is None, so if the executor fires _on_session_id synchronously before the start runs, PROCESS_SPAWN is recorded as CONFIRMED-without-STARTED (the same invariant gap noted above) and the aggregate_phase logic loses the ability to detect an in-progress unconfirmed spawn. The start() call must precede the closure exposure to the executor.

src/autoskillit/fleet/state_types.py
```diff
@@ -53,6 +55,264 @@ class ErrorCodeCategory(StrEnum):
...
 
+class DispatchEffectName(StrEnum):
+    """Stable vocabulary for retry-relevant fleet dispatch effects."""
+
+    CAMPAIGN_PATH_CAPTURE = "campaign_path_capture"
+    CALLER_IDENTITY = "caller_identity"
+    DISPATCH_ALLOCATION = "dispatch_allocation"
+    PRIOR_DISPATCH_BINDING = "prior_dispatch_binding"
+    REQUESTED_RESUME_BINDING = "requested_resume_binding"
+    EFFECTIVE_RESUME_BINDING = "effective_resume_binding"
```
Trecek, line 66: [info] abstraction_surface: DispatchEffectName.CALLER_IDENTITY and DispatchEffectName.CAMPAIGN_PATH_CAPTURE are declared retry_relevant=False in every call site, so they are journaled but never counted by aggregate_phase / retry_disposition. They exist purely to surface in the to_dict() envelope. Either mark them as observability-only (separate journal) or fold them into a single context-capture effect to reduce the enum surface.

src/autoskillit/fleet/state_types.py
```diff
@@ -53,6 +55,264 @@ class ErrorCodeCategory(StrEnum):
...
+class DispatchAggregatePhase(StrEnum):
+    """Conservative aggregate of all retry-relevant dispatch effects."""
+
+    NOT_STARTED = "not_started"
+    STARTED = "started"
+    COMMITTED = "committed"
+    UNKNOWN = "unknown"
+
+
+class DispatchRetryDisposition(StrEnum):
```
Trecek, line 93: [info] cohesion: Retry-disposition naming divergence between parallel layers: `DispatchRetryDisposition` (line 93) uses `FRESH_DISPATCH_ALLOWED / RESUME_BY_IDENTITY / RECONCILE_REQUIRED` while `KitchenRetryDisposition` (kitchen_transition.py line 64) uses `RETRY_SAFE / IN_PROGRESS / RECONCILE_REQUIRED / FINGERPRINT_CONFLICT / COMMITTED_REPLAY`. Both layers expose caller-facing retry semantics with no shared vocabulary, so a downstream caller that sees both kinds of provenance cannot easily apply consistent policy. A unified disposition vocabulary (or an explicit shared abstraction) would prevent silent naming drift.

src/autoskillit/fleet/state_types.py
```diff
@@ -53,6 +55,264 @@ class ErrorCodeCategory(StrEnum):
...
+
+
+@dataclass(frozen=True, slots=True)
+class DispatchEffectProvenance:
+    """Immutable request snapshot used by domain, wire, and formatter layers."""
+
+    operation_id: str
+    effects: tuple[DispatchEffectRecord, ...] = ()
+    cancel_requested: bool = False
+    local_cleanup: ProcessCleanupResult | None = None
```
Trecek, line 132: [warning] abstraction_surface: DispatchEffectProvenance carries three write-only evidence fields (local_cleanup, state_cleanup_confirmed, labels_cleanup_confirmed). They are set by record_local_cleanup / record_state_cleanup / record_labels_cleanup but no production logic ever reads them — aggregate_phase and retry_disposition compute from effects[] only. cancel_requested (L131) is identical: set by request_cancel() but only serialized via to_dict() and read by tests. The fields exist purely as opaque JSON diagnostic metadata. Either drop them or wire them into aggregate_phase / retry_disposition so consumers can actually rely on them.

src/autoskillit/fleet/state_types.py
```diff
@@ -53,6 +55,264 @@ class ErrorCodeCategory(StrEnum):
...
+                return
+            self._effects[name] = DispatchEffectRecord(
+                name=name,
+                phase=DispatchEffectPhase.STARTED,
+                effect_id=self._effect_id(name),
+                retry_relevant=retry_relevant,
+                known_downstream_identities=self._identities(identities),
+            )
+
+    def confirm(
```
Trecek, line 217: [warning] defense: DispatchProvenanceTracker.confirm does not validate that the effect was first started. When existing is None, the method writes a new DispatchEffectRecord with phase=CONFIRMED directly, bypassing STARTED. The aggregate_phase/retry_disposition logic assumes a STARTED effect exists before a CONFIRMED one — and the CRITICAL state-machine contract is that confirm is only authoritative after start. There is no invariant or raise that enforces this. Today every caller happens to start first, but the typed state machinery permits an invalid transition (CONFIRMED-without-STARTED), which silently bypasses the STARTED-ambiguity detection in aggregate_phase. Compare with KitchenOpenState which uses _LEGAL_PHASE_TRANSITIONS for the same protection; the new tracker should mirror that.

src/autoskillit/fleet/state_types.py
```diff
@@ -53,6 +55,264 @@ class ErrorCodeCategory(StrEnum):
...
+                    for name in sorted(self._effects, key=lambda item: item.value)
+                ),
+                cancel_requested=self._cancel_requested,
+                local_cleanup=self._local_cleanup,
+                state_cleanup_confirmed=self._state_cleanup_confirmed,
+                labels_cleanup_confirmed=self._labels_cleanup_confirmed,
+            )
+
+
+def _untracked_dispatch_provenance() -> DispatchEffectProvenance:
```
Trecek: [warning] slop: Backward-compat shim: `_untracked_dispatch_provenance()` is a sentinel-default factory (operation_id='untracked', hardcoded 'untracked:dispatch_allocation' effect_id, ambiguity='effect provenance was not supplied by the producer') wired in via default_factory for both `DispatchRejected.effect_provenance` ([L653-L655]) and `DispatchCompleted.effect_provenance` ([L683-L685]). The docstring explicitly says 'Fail-closed provenance for legacy constructors outside the request path' — transitional/backward-compat language the cleanup was supposed to remove. Any new DispatchRejected/DispatchCompleted constructed outside the migrated request path silently inherits an UNKNOWN-aggregated provenance that masks whether the dispatch was retry-safe. Either propagate real provenance through every caller or make `effect_provenance` a required field rather than papering over the gap with a 'legacy' sentinel.

src/autoskillit/pipeline/kitchen_transition.py
```diff
@@ -0,0 +1,523 @@
...
+    "abort_kitchen_effect",
+    "bind_kitchen_intent",
+    "canonical_kitchen_intent_fingerprint",
+    "claim_kitchen_request",
+    "closed_kitchen_open_state",
+    "commit_kitchen_response",
+    "confirm_kitchen_effect",
+    "kitchen_state_payload",
+    "mark_kitchen_effect_ambiguous",
+    "mark_kitchen_effect_degraded",
```
Trecek, line 36: [info] abstraction_surface: mark_kitchen_effect_degraded is exported in __all__ and re-exported via pipeline/__init__.py but the only production caller is _transition_degraded in the same module. Public re-export is over-eager — the function is internal-only and should be _mark_kitchen_effect_degraded (private).

src/autoskillit/pipeline/kitchen_transition.py
```diff
@@ -0,0 +1,523 @@
...
+    RECONCILE_REQUIRED = "reconcile_required"
+    FINGERPRINT_CONFLICT = "fingerprint_conflict"
+    COMMITTED_REPLAY = "committed_replay"
+
+
+@dataclass(frozen=True, slots=True)
+class KitchenEffectRecord:
+    """Immutable receipt for one effect owned by an open operation."""
+
+    name: str
```
Trecek, line 87: [warning] cohesion: Type asymmetry between parallel effect records: `KitchenEffectRecord.name` is typed as plain `str` (line 78), but the parallel `DispatchEffectRecord.name` in `fleet/state_types.py` (line 105) is typed as `DispatchEffectName` (StrEnum). The two new feature areas expose the same conceptual surface — a journal of named effects — but use different type-safety strategies. Kitchen transition should adopt a typed StrEnum (e.g., `KitchenEffectName`) so callers receive the same compile-time vocabulary that dispatch provenance provides.

src/autoskillit/pipeline/kitchen_transition.py
```diff
@@ -0,0 +1,523 @@
...
+    effect_id: str
+    phase: KitchenEffectPhase = KitchenEffectPhase.PREPARED
+    receipt: str | None = None
+    downstream_identity: str | None = None
```
Trecek, line 91: [warning] abstraction_surface: KitchenEffectRecord.downstream_identity is set by 5 callers (bind_kitchen_intent, _transition_confirm gates) and serialized by kitchen_state_payload, but no production consumer ever reads effect.downstream_identity. It is a write-only carrier on the per-effect journal. Either drop the field or expose it through a lookup helper that the open_kitchen handler actually uses (e.g., to attribute receipts back to the operation that produced them).

src/autoskillit/pipeline/kitchen_transition.py
```diff
@@ -0,0 +1,523 @@
...
+@dataclass(frozen=True, slots=True)
+class KitchenOpenState:
+    """Immutable snapshot of one process-local ``open_kitchen`` operation."""
+
+    phase: KitchenOpenPhase
+    kitchen_id: str
+    operation_id: str
+    context_id: str
+    intent_fingerprint: str | None = None
+    mode: str | None = None
```
Trecek: [info] abstraction_surface: KitchenOpenState.mode is set by bind_kitchen_intent but never read anywhere in src/, and is not even serialized by kitchen_state_payload. The field is a dead write.

src/autoskillit/pipeline/kitchen_transition.py
```diff
@@ -0,0 +1,523 @@
...
+    effects: tuple[KitchenEffectRecord, ...] = ()
+    cached_response: str | None = None
+    initialization_id: str | None = None
```
Trecek, line 115: [info] abstraction_surface: KitchenOpenState.initialization_id is set by prepare_kitchen_response / commit_kitchen_response and serialized by kitchen_state_payload, but no production consumer reads it back. The formatter (_fmt_recipe.py:235-253) only reads kitchen_id/operation_id/intent_fingerprint/received_intent_fingerprint/phase/effects/degraded_evidence/ambiguity/retry_disposition. Drop the field or thread it into the pretty output.

src/autoskillit/server/_lifespan.py
```diff
@@ -370,7 +379,23 @@ async def _pre_reveal_kitchen(ctx: Any) -> None:
     from autoskillit.server._misc import _prime_quota_cache  # circular-break
     from autoskillit.server.tools.tools_kitchen import _write_hook_config  # circular-break
 
-    ctx.kitchen_id = resolve_kitchen_id()
+    state = getattr(ctx, "kitchen_open_state", None)
+    if not isinstance(state, KitchenOpenState):
```
Trecek: [info] abstraction_surface: The defensive guard `if not isinstance(state, KitchenOpenState): ctx.kitchen_transition_lock = threading.RLock(); ctx.kitchen_open_state = closed_kitchen_open_state()` duplicates the ToolContext dataclass defaults (default_factory=threading.RLock for kitchen_transition_lock, default_factory=closed_kitchen_open_state for kitchen_open_state in context.py:207-211). The dataclass already initializes both, so the guard can only be reachable for a ctx that doesn't have these fields — i.e., a ctx that is not a ToolContext. Tighten the type check (isinstance(ctx, ToolContext)) or drop the guard.

src/autoskillit/server/_notify.py
```diff
@@ -234,6 +253,21 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:
                 result = complete_section_response(finalized_section, result)
             elif finalized_initialization is not None:
                 result = complete_initialization_response(finalized_initialization, result)
+            if kitchen_response_success and ctx is not None and isinstance(result, str):
```
Trecek, line 262: [warning] bugs: Response-enforcement confirmation is gated on parsing the enforced result as a JSON object with `success is True`. `open_kitchen` can legitimately return a rendered non-JSON response (for example the normal text/recipe response after formatting), so the `response_enforcement` effect remains STARTED despite successful enforcement. This leaves the replay journal in an in-progress/ambiguous state and prevents safe replay.

src/autoskillit/server/tools/tools_fleet_dispatch.py
```diff
@@ -70,6 +79,88 @@
...
+_BOUND_DISPATCH_PROVENANCE: ContextVar[DispatchProvenanceTracker | None] = ContextVar(
+    "bound_dispatch_provenance",
+    default=None,
+)
+_ACTIVE_DISPATCH_PROVENANCE: ContextVar[DispatchProvenanceTracker] = ContextVar(
+    "active_dispatch_provenance"
+)
+
+
+def _attach_dispatch_provenance(
```
Trecek, line 91: [warning] cohesion: Asymmetric envelope-metadata attachment: `_attach_dispatch_provenance` (line 91) operates on a raw JSON string and writes a single `effect_provenance` field, while the parallel `_attach_transition_fields` in `tools_kitchen.py` (line 194) operates on a dict and writes many transition fields (`kitchen_id`, `operation_id`, `phase`, `effects`, etc.). Both serve the same purpose — attach immutable operation metadata to an envelope — but use different signatures, return types, and naming conventions. A single shared envelope attachment helper would prevent the two new feature areas from drifting further.

src/autoskillit/server/tools/tools_kitchen.py
```diff
@@ -111,6 +137,266 @@
...
+            name=name if isinstance(name, str) else None,
+            overrides=overrides if isinstance(overrides, Mapping) else None,
+            ingredients_only=ingredients_only,
+            delivery_request=(delivery_request if isinstance(delivery_request, Mapping) else None),
+            context_id=tool_ctx.kitchen_open_state.context_id,
+        )
+        mode = "ingredients_only" if ingredients_only else ("recipe" if name else "anonymous")
+        with tool_ctx.kitchen_transition_lock:
+            active = tool_ctx.kitchen_open_state
+            committed_postconditions_hold = not (
```
Trecek, line 293: [info] arch: Re-binding flow in _bind_open_kitchen_transition: when the current state is COMMITTED and a new request arrives with a different recipe name (`getattr(tool_ctx, "recipe_name", "") != name`), the binder silently mints a new `KitchenOpenState` with the SAME `kitchen_id` but a NEW `operation_id` (L294-L304). This semantic path is not covered by tests/server/test_open_kitchen_transition.py, which only exercises the in-progress conflict and pre-dispatch retry-safe paths. The COMMITTED re-bind path could leave a session with a mismatched `kitchen_id` vs `operation_id` invariant.

src/autoskillit/server/tools/tools_kitchen.py
```diff
@@ -111,6 +137,266 @@
...
+                active.phase is KitchenOpenPhase.COMMITTED
+                and mode == "recipe"
+                and getattr(tool_ctx, "recipe_name", "") != name
+            )
+            if (
+                active.phase is KitchenOpenPhase.COMMITTED
+                and active.intent_fingerprint is not None
+                and (active.intent_fingerprint != fingerprint or not committed_postconditions_hold)
+            ):
+                tool_ctx.kitchen_open_state = new_kitchen_open_state(
```
Trecek, line 303: [warning] bugs: When a committed operation's postconditions no longer hold, this code replaces the committed state with a fresh `new_kitchen_open_state` while retaining the old `kitchen_id`. That discards the cached response and all recorded effect provenance, allowing a subsequent request to redispatch effects from an operation that was already committed. A replay-safe transition should enter reconciliation or explicitly close/reset the operation only after compensating or validating the prior effects, not silently erase the journal.

src/autoskillit/server/tools/tools_kitchen.py
```diff
@@ -130,15 +416,22 @@ def _kitchen_failure_envelope(
...
+        "error": f"{type(exc).__name__}: {exc}",
+        "stage": stage,
+    }
+    try:
+        from autoskillit.server._state import _get_ctx_or_none  # circular-break
+
+        tool_ctx = _get_ctx_or_none()
+        if tool_ctx is not None:
+            payload.update(_transition_fields(tool_ctx))
+    except Exception:
```
Trecek, line 433: [info] defense: _kitchen_failure_envelope wraps the entire _get_ctx_or_none() + _transition_fields(tool_ctx) enrichment in a broad 'except Exception' that logs 'open_kitchen_transition_failure_envelope_failed' and silently returns the basic envelope without the kitchen transition fields. During a critical-state open_kitchen failure path (the envelope is the only client-visible signal of what went wrong), losing the transition payload means the caller cannot distinguish a closed kitchen from a kitchen that failed mid-transition. A narrower except (specific exceptions from _transition_fields, e.g. AttributeError if state is missing) would preserve the diagnostic when the machinery is wired but the envelope is still returned when it isn't.

src/autoskillit/server/tools/tools_kitchen.py
```diff
@@ -1066,6 +1456,7 @@ async def open_kitchen(
                     )
             if _is_deferred_recall:
                 try:
+                    _transition_start(tool_ctx, "recipe_serving")
```
Trecek: [warning] bugs: The deferred-recall path starts the `recipe_serving` effect but never confirms it before returning the rendered response through `finalize_recipe_delivery`; the normal path similarly starts the effect and relies on the response wrapper to infer success. If `finalize_recipe_delivery` or response enforcement returns a non-JSON/string result, the effect remains STARTED and the operation becomes permanently in-progress/reconciliation-required on replay even though recipe serving may have completed. The effect needs an authoritative confirmation or an explicit ambiguity transition on every return path.

src/autoskillit/server/tools/tools_kitchen.py
```diff
@@ -1419,6 +1823,7 @@ async def open_kitchen(
 
             return render_served_response(result)
 
+        _transition_start(tool_ctx, "anonymous_response")
```
Trecek, line 1827: [warning] bugs: `anonymous_response` is started, but the result is passed through `render_served_response` and returned without confirming that effect. The outer transition binder only confirms effects that are still STARTED when it parses a JSON success envelope; the rendered response may be plain text, so this effect can remain STARTED after a successful anonymous `open_kitchen`, incorrectly blocking future calls as in-progress/reconciliation-required.

tests/cli/test_doctor_backend_guards.py
```diff
@@ -581,6 +581,28 @@ def test_version_above_minimum_returns_ok(self, monkeypatch: pytest.MonkeyPatch)
         assert result.severity == Severity.OK
         assert "0.131.0" in result.message
 
+    def test_claude_uses_its_own_floor_and_display_name(
```
Trecek, line 584: [warning] tests: test_claude_uses_its_own_floor_and_display_name calls `_check_codex_version(backend=ClaudeCodeBackend())`. The function name `_check_codex_version` suggests it's codex-specific, but the test exercises it with a Claude backend. Either the function needs renaming (e.g., `_check_backend_version`) or the test is testing implementation detail that should be split into separate per-backend functions. The test is logically valid (it does verify floor and display name), but the cross-purposing of a codex-named function is a smell.

tests/cli/test_fleet_provenance_prompt_contract.py
```diff
@@ -0,0 +1,63 @@
...
+)
+def test_clause_deletion_fails_its_own_invariant(
+    module: object,
+    clause_id: str,
+    required_text: str,
+) -> None:
+    source = inspect.getsource(module)
+    section = infrastructure_recovery_section(source)
+    mutant = source.replace(section, section.replace(required_text, "", 1), 1)
+
```
Trecek, line 37: [warning] tests: test_clause_deletion_fails_its_own_invariant mutates the source by `source.replace(section, section.replace(required_text, "", 1), 1)` and checks that `assert_provenance_contract` raises. This is a meta-test that mutates the source text in-memory and re-runs a structural assertion. If `_fleet_provenance_prompt_contract.assert_provenance_contract` is broken in a way that ALWAYS raises (e.g., bad import), the test would pass for the wrong reason. The test should additionally assert the contract succeeds on the unmutated source first (or use a positive control).

tests/cli/test_session_launch.py
```diff
@@ -969,9 +997,15 @@ def test_multi_backend_no_cross_flag_contamination(monkeypatch: pytest.MonkeyPat
     from autoskillit.execution.backends import BACKEND_REGISTRY
 
     captured: dict = {}
-    monkeypatch.setattr(shutil, "which", lambda _: "/usr/bin/fake-agent")
+    real_which = shutil.which
```
Trecek, line 1000: [critical] tests: HERMETIC EXECUTABLE FINDING: test_multi_backend_no_cross_flag_contamination uses `real_which = shutil.which` (the real shutil.which) and does NOT monkeypatch it. The test asserts `cmd[0] == Path(real_which(backend.binary_name()) or "").resolve()`, which depends on the host machine having actual `claude`/`codex` binaries installed at the same path the test runtime expects. The previous version monkeypatched `shutil.which` to return a hardcoded `/usr/bin/...` path. The PR added tmp_path-baked binaries in `_stub_artifact_authorities` and an `_stub_interactive_prelaunch` fixture for OTHER tests, but this test was migrated to use the unmocked real `shutil.which` without tmp_path. The test will fail (or be skipped silently) on hosts without these binaries; the cmd-construction logic is supposed to be unit-tested, not host-binaries-dependent.

tests/cli/test_session_launch.py
```diff
@@ -1007,9 +1040,15 @@ def test_real_backend_no_foreign_flags(monkeypatch: pytest.MonkeyPatch, backend_
     from autoskillit.execution.backends import BACKEND_REGISTRY
 
     captured: dict = {}
-    monkeypatch.setattr(shutil, "which", lambda binary: f"/usr/bin/{binary}")
+    real_which = shutil.which
```
Trecek, line 1043: [critical] tests: HERMETIC EXECUTABLE FINDING: test_real_backend_no_foreign_flags uses `real_which = shutil.which` (line 1043) and the test asserts `cmd[0] == Path(real_which(backend.binary_name()) or "").resolve()` (line 1063). Test signature `def test_real_backend_no_foreign_flags(monkeypatch: pytest.MonkeyPatch, backend_...)` takes no tmp_path. The mock_run handles --version, but the binary path resolution depends on host PATH. The previous version monkeypatched shutil.which to a fixed path. With the new `_stub_interactive_prelaunch` fixture in conftest, this test should either use the fixture or have tmp_path + a write_text executable. Note: BACKEND_REGISTRY iterates real backends (claude/codex) — host dependency is intentional but must be opt-in (skipif) and not implicit.

tests/cli/test_session_launch.py
```diff
@@ -1047,9 +1085,15 @@ def test_cross_validation_contract_all_flags_known(
...
```
Trecek, line 1088: [critical] tests: HERMETIC EXECUTABLE FINDING: test_cross_validation_contract_all_flags_known (line 1088) uses `real_which = shutil.which` and asserts `cmd[0] == Path(real_which(backend.binary_name()) or "").resolve()` (line 1108). Same regression as the other two tests: removed monkeypatch.setattr(shutil, "which", ...) and replaced with a call to the unmocked real shutil.which. Test will fail on hosts without `claude` or `codex` binaries.

tests/execution/backends/test_claude_startup_readiness.py
```diff
@@ -0,0 +1,138 @@
...
+        environment={
+            "PATH": str(tmp_path),
+            "MCP_CONNECTION_NONBLOCKING": CLAUDE_MCP_CONNECTION_NONBLOCKING,
+            CLAUDE_MCP_CONNECT_TIMEOUT_ENV_VAR: str(CLAUDE_MCP_CONNECT_TIMEOUT_MS),
+        },
+        cwd=tmp_path,
+    )
+
+
+def test_interactive_cmd_uses_bound_path_and_sealed_readiness_values(tmp_path: Path) -> None:
```
Trecek, line 37: [info] tests: test_interactive_cmd_uses_bound_path_and_sealed_readiness_values writes a fake executable to tmp_path (line 38) but the test depends on `backend.build_interactive_cmd(env_extras=extras)` succeeding without explicit `executable=`. The ExecutableLaunchBinding.resolve call at line 49 uses `candidate.env` which is the env_extras itself, not the candidate's env after sealing. The test is somewhat entangled with the contract that env_extras already includes the sealed readiness values. The test would be clearer if the test exercised only the seal-and-bind path explicitly.

tests/execution/backends/test_cli_conformance_probes.py
```diff
@@ -1418,3 +1437,376 @@ def echo(x: str) -> str:
...
+    ready: bool
+    tool_list_observed: bool
+    open_kitchen_result_observed: bool
+    question_detected: bool
+    output_bytes: int
+    output_sha256: str
+    trace_path: Path
+
+
+def _write_delayed_startup_plugin(
```
Trecek, line 1452: [warning] tests: _write_delayed_startup_plugin builds a large inline Python script (~100 lines) as a string then writes it to a tmp_path shim file. The script uses `os.fork`-style subprocess.Popen with stdin/stdout pipes and threads. This is a real subprocess test that runs `claude` interactively. The script is large enough that any test failure will be hard to diagnose. The skipif guard at line 7187 requires both CLAUDE_STARTUP_READINESS_SMOKE=1 AND host-installed `claude` AND auth tokens — this is a highly env-gated test and will run rarely. The test is valuable but the inline-script approach is a maintenance hazard.

tests/pipeline/test_kitchen_transition.py
```diff
@@ -0,0 +1,170 @@
...
+    assert "connection lost" in ambiguous.ambiguity[0]
+
+
+def test_started_effect_is_not_restarted() -> None:
+    state = bind_kitchen_intent(_state(), fingerprint=_fingerprint(), mode="recipe")
+    started = start_kitchen_effect(state, "recipe_serving")
+
+    replayed = start_kitchen_effect(started, "recipe_serving")
+
+    assert replayed == started
```
Trecek, line 109: [warning] tests: test_started_effect_is_not_restarted asserts `replayed == started` (structural equality on the entire state). This is a strong assertion that relies on exact representation — any future addition of a field (e.g., a timestamp, monotonic counter) to the state will break the test, even if the re-start-not-restarted semantics are preserved. The test should assert specifically on the effects list and the relevant phase, not on state identity.

tests/server/test_dispatch_provenance_boundary.py
```diff
@@ -0,0 +1,123 @@
...
+    _bound_dispatch_provenance,
+    _confirm_campaign_state_write,
+    _dispatch_cancellation_response,
+)
+
+pytestmark = [pytest.mark.layer("server"), pytest.mark.small]
+
+
+@pytest.mark.asyncio
+async def test_argument_binder_attaches_unique_request_provenance() -> None:
```
Trecek, line 25: [info] tests: test_argument_binder_attaches_unique_request_provenance defines an inner `@_bind_dispatch_provenance @_cancellation_shield(...)` decorated handler. This is a fine test, but each test parameter set (operation_id, response_factory, state_factory) is hardcoded; if a regression test is needed for e.g. cross-task state pollution, the test would need to be duplicated. Consider parameterizing over composite scenarios.

tests/server/test_open_kitchen_transition.py
```diff
@@ -0,0 +1,123 @@
...
+    await first
+
+    assert conflict["success"] is False
+    assert conflict["error"] == "open_kitchen_intent_fingerprint_conflict"
+    assert conflict["retry_disposition"] == "fingerprint_conflict"
+    assert tool_ctx.kitchen_open_state.request_active is False
+
+
+@pytest.mark.anyio
+async def test_proven_predispatch_failure_remains_retry_safe(
```
Trecek, line 101: [warning] tests: test_proven_predispatch_failure_remains_retry_safe calls `_transition_start(tool_ctx, "recipe_serving")` and `_transition_abort(tool_ctx, "recipe_serving")` directly inside the implementation to simulate a pre-dispatch failure. This couples the test to the internal helper functions `_transition_start`/`_transition_abort`. If those helpers are renamed or replaced, the test breaks even if the externally-observable behavior is unchanged. The test should ideally drive the failure via the public surface (e.g., raise an exception in implementation that the binder handles).

